PR #637, opened 2015-04-15
Add parse format for ISO8601 with millisecond and test case

yetisno:
parse format yyyy-MM-dd'T'HH:mm:ss.SSS'Z'


gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
```diff
@@ -39,6 +39,7 @@
   private final DateFormat enUsFormat;
   private final DateFormat localFormat;
   private final DateFormat iso8601Format;
+  private final DateFormat iso8601FormatMilli;
```
inder123: why do we need two DateFormat objects? Can this be done with just one?

yetisno: yes, i'll keep one DateFormat object and reduce precision of millisecond to fit ISO8601. Close request

PR #638, opened 2015-04-16
reduce precision of millisecond to fit ISO8601

yetisno:
make format yyyy-MM-dd'T'HH:mm:ss.SSS'Z' fit yyyy-MM-dd'T'HH:mm:ss'Z'.

Some api provide millisecond precision, but it'll throw exception. This change cut millisecond part to fit ISO8601 format.

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -101,11 +101,15 @@ final class DefaultDateTypeAdapter implements JsonSerializer<Date>, JsonDeserial
         return enUsFormat.parse(json.getAsString());
       } catch (ParseException ignored) {
       }
-      try {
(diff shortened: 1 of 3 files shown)


inder123:
@swankjesse @JakeWharton What do you think about this one?

IgorMiller:
ISO 8601 specifies that you do can have fraction of seconds in your date format.
From http://en.wikipedia.org/wiki/ISO_8601 : "There is no limit on the number of decimal places for the decimal fraction."
Another quick fix would be to add another iso8601Format with milliseconds. :-1: 
Or, try using
`return DatatypeConverter.parseDate(json.getAsString()).getTime();`
instead of 
`return iso8601Format.parse(json);`
It should be able to parse all of ISO8601 formats ;) :+1:

swankjesse:
I'm not comfortable with this proposal. My recommendation is to use `UtcDateTypeAdapter`, which is smarter & faster than the built-in date formatting.

PR #643, opened 2015-05-11
Getters and setters

steaks:
Hi Gson team,

My company ([Hurdlr](hurdlr.com)) uses gson extensively.  We really enjoy the library and appreciate all your work!

This pull request includes an attempt to implement support for getters and setters.  I realize you probably have deliberately chosen to not implement this feature.  However, the other day I stumbled across a problem where I really wanted to serialize a computed property.  So along the way I implemented support for getters and setters for my company.  (A full explanation of my use case is explained below.)

After implementing this feature for my own purposes, I figured creating a pull request couldn't hurt.  And after seeing [others with similar requests](http://stackoverflow.com/questions/6203487/why-does-gson-use-fields-and-not-getters-setters) I hope this pull request will make it into the codebase or inspire ideas for getter/setter support.

Before I invest more effort into this feature, I'm curious to hear initial thoughts and reactions.  If you want to proceed with this feature, I'd be interested to learn what I can do for a next step.

**Basic usage**
```
class Foo {
  private String bar;

  @GsonGetter
  public String getBar() {
    return this.bar;
  }

  @GsonSetter
  public String setBar(String value) {
    this.bar = value;
  }
}

//JSON
{ "bar": "baz" }  //using getter and setter, not the private field

Gson gson = new GsonBuilder().allowGetterAndSetterMethods().create();
```

*More examples can be found in MethodTest.java*

**API additions**
* @GsonGetter
* @GsonSetter
* GsonBuilder.allowGettersAndSetterMethods
* Excluder.allowGettersAndSetterMethods
* MethodAndFieldNamingStrategy (new interface)
* MethodAndFieldNamingPolicy (new enum)

**Expected common use cases**
* Computed properties via getters
* Lazily evaluated properties via getters
* Safer way to expose only public API
  * Often private fields can mistakenly be serialized because we often overlook them when thinking about our exposed API
* Data formatting via getters and/or setters
* Data validation via getters and/or setters
```
public void setFoo(String value) {
  if (value < 0) { throw new IllegalArgumentException(); }
  this.foo = value;
}
```

**Positives**
* Simple and readable annotations
* Fully backwards compatible
* Opt-in
  * allowGetterAndSetterMethods is necessary on GsonBuilder
  * @GsonGetter and @GsonSetter are required for gson to consider a method
* Intuitive default naming 
  * e.g. getBar => "bar", setBar => "bar"
* Handles collisions
  * e.g. getBar/setBar take precedence over bar

**Negatives**
* Extra annotations added...more bloated API
* Extra MethodAndFieldNamingStrategy interface and MethodAndFieldNamingPolicy policy make the API more bloated and confusing.

**Implementation / Testing gaps**
* No tests for naming
* No performance testing
* No Java support testing.  I'm pretty sure it works for Java 1.6+

**Context for the use case that prompted me to implement @GsonGetter and @GsonSetter:**

We factored out functionality in class ```Foo``` into a separate helper class.  But we wanted to preserve the JSON API in order to not break backwards compatibility.

There are a few way so solve this problem.  Three include
* Shadow fooHelper.bar in NewFoo
* Create a custom serializer/deserializer
* Use a GsonGetter for a computed property.  IMO this solution is the simplest and most elegant :)

```
class Foo {
  private String bar;
}
//JSON
{  "bar": "baz" }
```

```
class NewFoo {
   private FooHelper fooHelper;
}
//Current JSON
{  "fooHelper": { "bar": "baz" } }
//Desired JSON
{ "bar": "baz" }
```
*Solution 3: Computed getter*
```
class NewFoo {
  private FooHelper fooHelper;

  @ GsonGetter
  public getBar() {
    return this.fooHelper.bar;
  }
}
```

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -123,8 +123,8 @@ public final class Gson {
   private final boolean prettyPrinting;
 
   final JsonDeserializationContext deserializationContext = new JsonDeserializationContext() {
-    @SuppressWarnings("unchecked")
-	public <T> T deserialize(JsonElement json, Type typeOfT) throws JsonParseException {
+  @SuppressWarnings("unchecked")
+  public <T> T deserialize(JsonElement json, Type typeOfT) throws JsonParseException {
       return (T) fromJson(json, typeOfT);
     }
   };
@@ -173,13 +173,13 @@ public final class Gson {
    * </ul>
    */
   public Gson() {
-    this(Excluder.DEFAULT, FieldNamingPolicy.IDENTITY,
+    this(Excluder.DEFAULT, MethodAndFieldNamingPolicy.DEFAULT,
         Collections.<Type, InstanceCreator<?>>emptyMap(), false, false, DEFAULT_JSON_NON_EXECUTABLE,
         true, false, false, LongSerializationPolicy.DEFAULT,
         Collections.<TypeAdapterFactory>emptyList());
   }
-
-  Gson(final Excluder excluder, final FieldNamingStrategy fieldNamingPolicy,
+  
+  Gson(final Excluder excluder, final FieldNamingStrategy namingPolicy,
       final Map<Type, InstanceCreator<?>> instanceCreators, boolean serializeNulls,
       boolean complexMapKeySerialization, boolean generateNonExecutableGson, boolean htmlSafe,
       boolean prettyPrinting, boolean serializeSpecialFloatingPointValues,
@@ -241,7 +241,7 @@ public final class Gson {
     factories.add(new JsonAdapterAnnotationTypeAdapterFactory(constructorConstructor));
     factories.add(TypeAdapters.ENUM_FACTORY);
     factories.add(new ReflectiveTypeAdapterFactory(
-        constructorConstructor, fieldNamingPolicy, excluder));
+        constructorConstructor, namingPolicy, excluder));
 
     this.factories = Collections.unmodifiableList(factories);
   }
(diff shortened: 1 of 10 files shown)


swankjesse:
Personally I don't want this in the core. Have you considered implementing this using a custom `TypeAdapter`? They're very powerful.

steaks:
I think a TypeAdapter would work for a one off case.  But being able to serialize/deserialize getters and setters easily in any class, not just one off cases, would be more valuable to me.  I could implement a TypeAdapterFactory.

Maybe something like:

```
class GettersAndSettersTypeAdapterFactory implements TypeAdapterFactory {
  @Override
    GsonGettersAndSetters a = type.getClass().getAnnotation(GsonGettersAndSetters.class);
    if (a != null) {
      //Basically do everything I do in my modified ReflectiveTypeAdapterFactory
      //...get bound methods, get bound fields, create Adapter
    }
    return null;
  }
}
```
```
@GsonGettersAndSetters
class Foo { 
  @GsonGetter
   public String getBaz() { }
}

@GsonGettersAndSetters
class Bar { 
 @GsonGetter
  public String getBaz() { }
}
```

But this solution doesn't seem great because I'm basically reimplementing ReflectiveTypeAdapterFactory and miss any improvements to ReflectiveTypeAdapterFactory when new commits are made.

Are there other ways I'm missing?

steaks:
Any updates here?

swankjesse:
Nothing from me. I think a custom TypeAdapter is still the best option.

inder123:
Yes, philosophically we haven't considered supporting getters and setters. Some users have asked for it, but majority seems fine without it. We wont add it to core Gson unless there was a greater need.

Appreciate your effort in writing the type adapter. I think you should publish it for use with Gson. Would be great to see if people find it useful.

PR #645, opened 2015-05-20
Convenience methods in JsonObject

canidae:
I've added several convenience methods to JsonObject. I find that whenever I use Gson for directly parsing JSON I end up creating similar convenience methods in my own project.
See second commit message for a simple example of the change.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -18,6 +18,8 @@ package com.google.gson;
 
 import com.google.gson.internal.LinkedTreeMap;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.Map;
 import java.util.Set;
 
@@ -153,13 +155,40 @@ public final class JsonObject extends JsonElement {
   }
 
   /**
+   * Returns the member with the specified name if it exists, otherwise the supplied default value
+   * is returned.
+   *
+   * @param memberName name of the member that is being requested.
+   * @param defaultValue the value to be returned if the member does not exist.
+   * @return the member matching the name. <code>defaultValue</code> if no such member exists.
+   */
+  public JsonElement get(String memberName, JsonElement defaultValue) {
+    JsonElement result = get(memberName);
+    return result != null ? result : defaultValue;
+  }
+
+  /**
    * Convenience method to get the specified member as a JsonPrimitive element.
    *
    * @param memberName name of the member being requested.
    * @return the JsonPrimitive corresponding to the specified member.
    */
   public JsonPrimitive getAsJsonPrimitive(String memberName) {
-    return (JsonPrimitive) members.get(memberName);
+    return (JsonPrimitive) get(memberName);
+  }
+
+  /**
+   * Convenience method to get the specified member as a JsonPrimitive element if it exists,
+   * otherwise the supplied default value is returned.
+   *
+   * @param memberName name of the member being requested.
+   * @param defaultValue the value to be returned if the member does not exist.
+   * @return the JsonPrimitive corresponding to the specified member. <code>defaultValue</code>
+   * if no such member exists.
+   */
+  public JsonPrimitive getAsJsonPrimitive(String memberName, JsonPrimitive defaultValue) {
+    JsonPrimitive result = getAsJsonPrimitive(memberName);
+    return result != null ? result : defaultValue;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


canidae:
Signed the CLA thingy...

inder123:
Hi. Appreciate your efforts and the detailed tests for adding this.
However, I am concerned about the code bloat. I am not sure about the value of primitive accessor methods in JsonObject.

Similarly, Not too sure about providing the getAsBooleanWrapper() either. We are not providing wrappers for anything else. So, why do it for boolean.

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -153,13 +155,40 @@ public JsonElement get(String memberName) {
...
+  /**
+   * Convenience method to get the specified member as a JsonPrimitive element if it exists,
+   * otherwise the supplied default value is returned.
+   *
+   * @param memberName name of the member being requested.
+   * @param defaultValue the value to be returned if the member does not exist.
+   * @return the JsonPrimitive corresponding to the specified member. <code>defaultValue</code>
+   * if no such member exists.
+   */
+  public JsonPrimitive getAsJsonPrimitive(String memberName, JsonPrimitive defaultValue) {
```
inder123: Not too sure about the advantages of default value methods. Too many new methods get added for a small benefit.

inder123:
I would like to close this pull request in the next day or two. If any one has a strong preference for keeping it, please speak up.

canidae:
Thank you for the feedback. I am however slightly confused by some of the points made:
1. getBooleanAsWrapper(): I don't see the need for such methods myself, I added it only because it already exists for the parent class (JsonElement).
2. Code bloat: I share the concern for code bloat. As of such, I'm really curious why there are methods such as "getAsString()" and other data types for JsonArray that throws an unique exception if there are more than 1 element in the array.

I do however not feel that adding these convenience methods are bloating the code. I would be surprised if I'm the only one who create a wrapper for Gson to avoid the extensive code you'll need whenever you're dealing with JSON input where some fields are optional.

inder123:
1. Get as BooleanWrapper is package private, not part of the public API. We can remove it if needed.
2. If we were to redo the API, these methods will probably not make the cut. They did seem like a good idea at the time we created Gson.
With the newer releases of Gson, the DOM style API of JsonElement and subclasses is not really needed. 

I am curious, what kind of wrapper are you creating for Gson.

canidae:
I did not see that getBooleanAsWrapper() was package private, that is a fair point.

The wrapper I'm creating do exactly the same as the convenience methods I added to JsonObject: Return a default value when the member I want doesn't exist in the JSON. In all the last three projects I've used Gson I've been in the situation that some data is optional (and if not present, a default value is to be used). All these projects have been work related and I don't have access to the code now (on vacation), but basically I create an utility class with methods for all getAs...() methods, like this:
    public static String getAsString(JsonObject object, String member, String defaultValue) {
      JsonPrimitive primitive = object.getAsJsonPrimitive(member);
      String value = null;
      if (primitive != null)
        value = primitive.getAsString(member);
      return value != null ? value : defaultValue;
    }

(I'm coding by head here, I could dig up the exact code in some days when I'm back at work, but I assume the above example shows well enough what the wrappers I'm creating do). In short terms: Methods just like the ones I added to JsonObject, in a utility class.
As such, the convenience methods I've added are by no means crucial, the problem can be solved by such wrapper methods. I'm however not the only developer in the project, other developers may very well overlook the utility class and create their own solutions, which in the long term will create messy code. This is less likely to happen when the library itself present these convenience methods.

I did create a very simple example in one of the commit messages, here's a slightly larger example that hopefully shows that parsing even more complex JSON structure can lead to fairly unreadable code without convenience methods:
    {
      "optional1": {
        "optional2": {
          "optional3": {
            "valueWanted": "FooBar"
          }
        }
      }
    }
With convenience methods "valueWanted":
    JsonObject defObject = new JsonObject();
    String valueWanted = json.getAsJsonObject("optional1", defObject)
      .getAsJsonObject("optional2", defObject)
      .getAsJsonObject("optional3", defObject)
      .getAsString("valueWanted", "optionalDefaultValue");

Without convenience method, something like this:
    String valueWanted = null;
    JsonObject jsonObj = json.getAsJsonObject("optional1");
    if (jsonObj != null) {
      jsonObj = jsonObj.getAsJsonObject("optional2");
      if (jsonObj != null) {
        jsonObj = jsonObj.getAsJsonObject("optional3");
        if (jsonObj != null)
          valueWanted = jsonObj.getAsJsonPrimitive("valueWanted").getAsString();
      }
    }
    // if we want a default value when member is missing:
    if (valueWanted == null)
      valueWanted = "defaultValue";

From my point of view it would be beneficial if these methods were within the library rather than creating a utility class in projects using Gson, but I do see your point about code bloating. I feel these methods would be within the gist of Gson and not actually bloat the code significantly, if you feel differently that's well understandable.

inder123:
Thanks for the details. Yes, having a default object option for getters may be convenient and is similar to the pattern followed elsewhere (such as with Android preferences). Will keep an open mind in adding these.

Overall, I would recommend that you use the binding to Java classes instead of dealing with JsonElement parse trees directly. Also, overall, Gson has moved towards TypeAdapters since they don't generate the parse tree and are more efficient in memory and performance.

PR #649, opened 2015-05-30
Added support for duplicate keys in Map while deserializing. The last…

inder123:
… value overrides any previous values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+build
+bin
+target
+.classpath
+.project
+*.prefs
+.settings
--- a/gson/.gitignore
+++ b/gson/.gitignore
@@ -1,0 +1,7 @@
+build
+bin
+target
+.classpath
+.project
+*.prefs
+.settings
--- a/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
@@ -172,10 +172,7 @@ public final class MapTypeAdapterFactory implements TypeAdapterFactory {
           in.beginArray(); // entry array
           K key = keyTypeAdapter.read(in);
           V value = valueTypeAdapter.read(in);
-          V replaced = map.put(key, value);
-          if (replaced != null) {
-            throw new JsonSyntaxException("duplicate key: " + key);
-          }
+          map.put(key, value);
           in.endArray();
         }
         in.endArray();
@@ -185,10 +182,7 @@ public final class MapTypeAdapterFactory implements TypeAdapterFactory {
           JsonReaderInternalAccess.INSTANCE.promoteNameToValue(in);
           K key = keyTypeAdapter.read(in);
           V value = valueTypeAdapter.read(in);
-          V replaced = map.put(key, value);
-          if (replaced != null) {
-            throw new JsonSyntaxException("duplicate key: " + key);
-          }
+          map.put(key, value);
         }
         in.endObject();
       }
--- a/gson/src/test/java/com/google/gson/functional/MapAsArrayTypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapAsArrayTypeAdapterTest.java
@@ -74,11 +74,8 @@ public class MapAsArrayTypeAdapterTest extends TestCase {
         .create();
 
     String s = "[[\"1.00\",\"a\"],[\"1.0\",\"b\"]]";
-    try {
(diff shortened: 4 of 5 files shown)


inder123:
I signed it

On Sat, May 30, 2015 at 4:23 PM, googlebot <notifications@github.com> wrote:

> Thanks for your pull request. It looks like this may be your first
> contribution to a Google open source project, in which case you'll need to
> sign a Contributor License Agreement (CLA).
>
> [image: :memo:] *Please visit https://cla.developers.google.com/
> <https://cla.developers.google.com/> to sign.*
>
> Once you've signed, please reply here (e.g. I signed it!) and we'll
> verify. Thanks.
> ------------------------------
>
>    - If you've already signed a CLA, it's possible we don't have your
>    GitHub username or you're using a different email address. Check your
>    existing CLA data <https://cla.developers.google.com/clas> and verify
>    that your email is set on your git commits
>    <https://help.github.com/articles/setting-your-email-in-git/>.
>    - If you signed the CLA as a corporation, please let us know the
>    company's name.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/gson/pull/649#issuecomment-107098777>.
>



-- 
http://www.singhinderjeet.com/

swankjesse:
I think this is a bad idea. It should reject invalid JSON.

inder123:
Jesse: Can you point me to the spec that says that it is invalid JSON?
Also, we do support it for objects. Why treat it differently for Maps?

inder123:
@swankjesse I think Gson should be lenient in accepting JSON. Also, I can't seem to find spec that says that duplicate keys are not allowed.

swankjesse:
From the [JSON RFC](https://tools.ietf.org/html/rfc7159):

   An object whose names are all unique is interoperable in the sense
   that all software implementations receiving that object will agree on
   the name-value mappings.  When the names within an object are not
   unique, the behavior of software that receives such an object is
   unpredictable.  Many implementations report the last name/value pair
   only.  Other implementations report an error or fail to parse the
   object, and some implementations report all of the name/value pairs,
   including duplicates.

Duplicate keys is a potential security hole, and I'm against them.

inder123:
@swankjesse The spec says "The names within an object SHOULD be unique."  It is a SHOULD not a MUST. In fact, the para you quote implies that names can be duplicate, and many implementations pick the last name/value.

Can you elaborate on the security hole specific to this case? Clients should always validate user input anyway, I don't see duplicate keys causing any extra issues.

Overall, for Gson, we are generally lenient while receiving input. In fact, we already do that for duplicate keys in JsonObject. Applying a higher level of strictness to Map sounds a little arbitrary.

swankjesse:
The security hole is that you validate the first occurrence of a value, but use the last occurrence, or vice-versa. For example, you might write code like this:

```
  public boolean userHasLimitedPermissions(String json) {
    JsonReader r = new JsonReader(...);
    r.beginObject();
    while (r.hasNext()) {
      String key = r.nextKey();
      String value = r.nextString();
      if (key.equals("userid")) return !value.equals("root");
    }
    r.endObject();
  }
```

Bugs like this happen all the time. Most recently the Android code that checks signatures on `.apk` files failed on this.

It's a bug that we permit duplicates in JsonObject, but a more defensible one: checking the alternative there is more work.

Here's a better question: who wants this? Nobody sane should be using JSON that has duplicated keys.

mingtaozhangsc:
With v2.3.1, the behavior is inconsistent:
There seems no duplicate key detection during serialization, while deserialization throw com.google.gson.JsonSyntaxException: duplicate key.

trothwell:
What is status of this commit? Will Gson have ability to control duplicated key behavior?

svetob:
@swankjesse Someone who wants this is me - for example recent versions of the RabbitMQ management API have a bug where duplicate keys are, in fact, returned: https://github.com/rabbitmq/rabbitmq-management/issues/278 This causes our deserialisation with GSON to fail. It has since been fixed, but we need to support all versions in our library.

In the real world, you can't always control what data you get, but might still need to be able to handle it somehow.

If GSON cannot support this, we'll need to switch library, which would be a shame since GSON is otherwise convenient and easy to work with.

swankjesse:
@svetob you.can install your own TypeAdapter to workaround this I'd you like. It’s a few lines of code to write the adapter and install it.

michaelklishin:
@svetob this is a terrible idea. If all JSON parsers behaved this way we would discover https://github.com/rabbitmq/rabbitmq-management/issues/278 much later.

eamonnmcmanus:
If we were going to do this it would certainly be via a configurable option, as in #1884. I think we can close this old PR in favour of that one.

PR #652, opened 2015-06-09
Fix issues if runing in an environment with a Turkish locale

schlan:
### Description 
When using `LOWER_CASE_WITH_DASHES` or `LOWER_CASE_WITH_UNDERSCORES` as FieldNamingPolicy it could happen that certain fields won't get serialized/deserialized. 
I've seen that behavior on devices with a turkish (`tr-TR`) locale. `toLowerCase()` uses the device's locale and converts an `I` into an `ı`. ([JavaDoc](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toLowerCase()))

### Example
A field with the name: 
```
...
int lowerId = 8;
....
```
will turn into:
```
{ ... 'lower-ıd':8 ... }
```

### Recreation steps
Check out the first commit (299ee89) and run the test with the following JVM flags:  `-Duser.language=tr -Duser.region=TR`
 
### References
https://code.google.com/p/google-gson/issues/detail?id=541
http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toLowerCase()

### DIFF
--- a/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
+++ b/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
@@ -17,6 +17,7 @@
 package com.google.gson;
 
 import java.lang.reflect.Field;
+import java.util.Locale;
 
 /**
  * An enumeration that defines a few standard naming conventions for JSON field names.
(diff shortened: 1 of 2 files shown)


inder123:
:+1: 
Thanks for the fix.

PR #656, opened 2015-06-17
Update ReflectiveTypeAdapterFactory.java

skeychen:
fit declares multiple JSON fields named


skeychen:
I signed it!

swankjesse:
I don't think we want this behavior change.

PR #658, opened 2015-06-18
Add Travis CI build.

JakeWharton:


.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: java
+
+jdk:
+  - oraclejdk7
+  - oraclejdk8
+
+install: mvn -f gson install -DskipTests=true
+script: mvn -f gson test
```
JakeWharton: Would love to see the repo updated to a more traditional Java repo layout and some of the other cruft removed or consolidated.

inder123: Do you mean move the sub-projects out as their own projects? What would you suggest?

JakeWharton: I looked at it more after writing this, and didn't realize that each folder was a separate Maven-based project. With that, a simple parent pom that includes each folder as a module will probably be enough to allow everything to be built at once.

inder123: True. However, other sub-projects are in states of disrepair and not maintained well. We didn't want to tie them in with Gson.

Issue #188, opened 2015-03-19
Support handlers for unknown properties, useful for error handling


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. create a JSON string containing extra attributes
2. invoke Gson.fromJson supplying a class with fewer elements
3. GSON successfully instantiates the class without protesting about the 
existence of extra attributes in the string.

What is the expected output? What do you see instead?
1. define class A containing two fields: name and surname
2. define class B containing only one field: name
3. transform an instance of class A to Json string and use the Json string 
to create an instance of class B.

GSON doesn't complain. Even if this is the intended behavior, shouldn't 
there be an option to enforce stricter parsing?

What version of the product are you using? On what operating system?
gson-1.4

Please provide any additional information below.


```

Original issue reported on code.google.com by `mperdik...@gmail.com` on 22 Jan 2010 at 10:13

Attachments:
- [Converter.java](https://storage.googleapis.com/google-code-attachments/google-gson/issue-188/comment-0/Converter.java)

mcerina:
I submitted pull request https://github.com/google/gson/pull/660 for this

avarabyeu:
Any updates on this issue? looking forward for the fix...

pratapkishorevarma:
Is there a workaround for this issue which we can do until a fix is provided?

ikus060:
I manage to workaround this issue using a class similar to the following:
```
public class ValidatorAdapterFactory implements TypeAdapterFactory {

    @Override
    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
        // If the type adapter is a reflective type adapter, we want to modify the implementation using reflection. The
        // trick is to replace the Map object used to lookup the property name. Instead of returning null if the
        // property is not found, we throw a Json exception to terminate the deserialization.
        TypeAdapter<T> delegate = gson.getDelegateAdapter(this, type);

        // Check if the type adapter is a reflective, cause this solution only work for reflection.
        if (delegate instanceof ReflectiveTypeAdapterFactory.Adapter) {

            try {
                // Get reference to the existing boundFields.
                Field f = delegate.getClass().getDeclaredField("boundFields");
                f.setAccessible(true);
                Map boundFields = (Map) f.get(delegate);

                // Then replace it with our implementation throwing exception if the value is null.
                boundFields = new LinkedHashMap(boundFields) {

                    @Override
                    public Object get(Object key) {

                        Object value = super.get(key);
                        if (value == null) {
[...]

PR #660, opened 2015-06-19
added a strategy to manage unknown fields during deserialization

mcerina:
See issue https://github.com/google/gson/issues/188
While I understand the design choice of ignoring unknown fields during deserialization by default, I definitely think that there are use cases in which throwing an exception, or at least being able to log something is preferred.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -170,16 +170,21 @@ public final class Gson {
    *   <li>By default, Gson excludes <code>transient</code> or <code>static</code> fields from
    *   consideration for serialization and deserialization. You can change this behavior through
    *   {@link GsonBuilder#excludeFieldsWithModifiers(int...)}.</li>
+   *   <li>By default, during deserialization Gson silently ignores fields that are present in the
+   *   incoming Json but not in the Java classes. You can change this policy through
+   *   {@link GsonBuilder#setUnknownFieldHandlingStrategy(UnknownFieldHandlingStrategy)}.</li>
    * </ul>
    */
   public Gson() {
     this(Excluder.DEFAULT, FieldNamingPolicy.IDENTITY,
+        UnknownFieldHandlingPolicy.IGNORE,
         Collections.<Type, InstanceCreator<?>>emptyMap(), false, false, DEFAULT_JSON_NON_EXECUTABLE,
         true, false, false, LongSerializationPolicy.DEFAULT,
         Collections.<TypeAdapterFactory>emptyList());
   }
 
   Gson(final Excluder excluder, final FieldNamingStrategy fieldNamingPolicy,
+      UnknownFieldHandlingStrategy unknownFieldHandlingPolicy,
       final Map<Type, InstanceCreator<?>> instanceCreators, boolean serializeNulls,
       boolean complexMapKeySerialization, boolean generateNonExecutableGson, boolean htmlSafe,
       boolean prettyPrinting, boolean serializeSpecialFloatingPointValues,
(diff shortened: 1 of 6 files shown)


mcerina:
I signed it!

inder123:
Appreciate you taking the time to write this. However, I think the usecase for this is too infrequent to warrant inclusion in standard Gson. Also, the approach is too narrow (for just unknown properties). We have had many requests to provide ability to intercept and validate a Gson generated instance. We would like to see a simple elegant solution to the broader issue.

itERRatOR:
It is quite frequent. And I would only reject solutions as not good enough only when there are other better alternatives presented. But it doesn't look like @inder123 is going to come up with any idea regarding this at all.

ghuh:
I would agree that this is a common problem. For instance, a client sends a JSON object with a field name misspelled.  If it is valid for that field to be omitted, you would never know that there was an issue.  It would simply get silently ignored.

horalstvo:
I would like to have a choice to decide whether GSON should ignore unknown fields or should reject such requests. Sadly, I'm switching now to Jackson to get that.

awbgithub:
@inder123 I see a comment from you from over a year ago where you mention many requests like this one to support validation ... can you point me to a discussion where you are working on a more general and elegant solution? If not, please reconsider this PR as it is a simple and elegant approach to the specific problem of receiving unexpected fields in the JSON.

mrkb80:
I agree that the user should have the choice.

vertex-github:
"I think the usecase for this is too infrequent to warrant inclusion in standard Gson"

We would like this use case - and from the looks of the comments so do many others.  Jackson has this feature and it seems fairly frequently used (do a search on tutorials and examples).

"Also, the approach is too narrow (for just unknown properties). We have had many requests to provide ability to intercept and validate a Gson generated instance. We would like to see a simple elegant solution to the broader issue."

That maybe the case, but we came here looking for this exact feature, and only this feature.  We want a way to be be able to detect/intercept additional fields in the JSON so that we can report/handle fields that we were not expecting.  We spent quite a while trying to shoe-horn this feature in, and there's no hooks available to us to provide this functionality without duplicating all of the core type adapter factory and its supporting classes - that seems a tad wasteful.

We took a quick review of the PR code, and it seems very simple, logical and well thought out.  Keep the default behavior as is, and existing users wont have to change any code.  Provide the builder option and everyone is happy/happier. :-)

+1 for taking this PR.

inder123:
Throwing Exception on encountering an unnknown fields seems weird. JSON is an extensible and flexible format: a client shouldn't be broken by adding a new property in JSON.

What about missing fields? Wouldn't you want a similar action?

A better way is to use https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/PostConstructAdapterFactory.java

awbgithub:
@inder123 As @mcerina mentions in the first comment above, the UnknownFieldHandlingStrategy is extensible and could be used to just log unknown fields if throwing an exception doesn't suit your needs.

I looked at PostConstructAdapterFactoryTest and see that your test throws an exception when the sandwich is too cheesy ... and I can see how that could also be used to take some action if a "required" field was missing (i.e. still null during post construct validation) ... but I can't figure out how to use the PostConstructAdapterFactory to catch when an unknown field was present in the original JSON. Can you provide a code example of how the PostConstructAdapterFactory could be used to solve the OP's issue about handling unknown fields?

romankulyk1:
Can you deserialize the JSON object to get your Java object, then serialize your Java object back to JSON and see if you get the same result (compare the original JSON object to the converted Java JSON)! if not...extra fields! Haven't tried, yet, but this is just a thought.

adam66:
I am looking for this feature right now. Is the reason for its absence just adherence to dogma?

rubearen:
Still no solution for this problem.

Issue #626, opened 2015-03-19
Patch for /trunk/gson/src/main/java/com/google/gson/JsonDeserializer.java


GoogleCodeExporter:
```
Remove incorrect parentheses
```

Original issue reported on code.google.com by `mfisc...@yext.com` on 3 Feb 2015 at 1:06

Attachments:
- [JsonDeserializer.java.patch](https://storage.googleapis.com/google-code-attachments/google-gson/issue-626/comment-0/JsonDeserializer.java.patch)

Marcono1234:
Seems to apply still, see #661 for the needed changes. Though there are futher issues with that code sample, see #1425.

Marcono1234:
It is not completely clear what this issue was about since the attachment does not exist anymore, but based on the linked pull request it was most likely about the erroneous `()` in the code sample of the `JsonDeserializer` documentation. That problem was fixed by #2192, therefore I will close this issue.

PR #661, opened 2015-06-24
Small error of documentation

RoxRicardo:
Fixed name of class in JsonDeserializer.html as requested by the issue #626


swankjesse:
If you're going to fix the Javadocs, you should fix the original source code, not the generated output.

RoxRicardo:
Sorry, my bad. I fixed in the class, it's a idiot fix, but only to close the issue.

swankjesse:
Also, please http://rebaseandsqua.sh

PR #663, opened 2015-07-07
fixed the error of printing new line for system that dont use '\n'

gazzdingo:
I have changed it so that it will work on more OS when you are wanting to print a new line


swankjesse:
I don't think this makes sense. JSON doesn't care what OS it's being generated on.

gazzdingo:
But on windows the pretty mode does not do pretty mode

gazzdingo:
Because it will not do a new line

swankjesse:
I think this is a bug in Windows. Please consider sending them a Pull Request!

Issue #613, opened 2015-03-19
Document that Gson instances are thread safe


GoogleCodeExporter:
```
Issue #63 [1] and the gson mailing list [2] indicate that Gson instances are 
thread safe. It would be good to document that in the Gson class, either in the 
Javadoc or with a JSR-305 @ThreadSafe annotation.

Background for this request is that I recently found quite a few places in 
multi-threaded code where Gson instances were unnecessarily created because the 
developers didn't know that these instances are thread safe.

[1] http://code.google.com/p/google-gson/issues/detail?id=63
[2] https://groups.google.com/forum/#!topic/google-gson/Vju1HuJJUIE
```

Original issue reported on code.google.com by `st.fer...@gmail.com` on 28 Nov 2014 at 1:42

inder123:
https://github.com/google/gson/pull/685

PR #664, opened 2015-07-13
Add a JSR-305 @ThreadSafe annotation to the Gson class

ferstl:
This PR solves Issue #613 that I submitted several months ago on Google Code.

Summary:
- Add an optional dependency to com.google.code.findbugs:jsr305.
  The optional scope is used in order to avoid introducing a new
  transitive dependency to the jsr305 library. This is fine because the 
  @ThreadSafe annotation has only a documentary purpose and it is not
  retained at runtime.
- Annotate the Gson class as @ThreadSafe

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -50,6 +50,12 @@
   </organization>
   <dependencies>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.0</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>3.8.2</version>
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -31,6 +31,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.concurrent.ThreadSafe;
+
 import com.google.gson.internal.ConstructorConstructor;
 import com.google.gson.internal.Excluder;
 import com.google.gson.internal.Primitives;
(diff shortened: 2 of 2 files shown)


ferstl:
I signed the CLA

inder123:
@swankjesse  I dont like this pull request. It adds an unncessary dependency to Gson. I am not sure the benefit is all that much. Can we revert this please?

ferstl:
@inder123 @swankjesse Since the dependency is declared as `optional`, it will neither be resolved nor downloaded when Gson is used as a dependency.
That was exactly my intention.

swankjesse:
I don't feel strongly either way.

swankjesse:
(Which is to say, if you want to rollback, I'll merge!)

ferstl:
If you revert this PR, please document at least in the Javadoc that the Gson class is thread-safe. Otherwise one has to assume that it is not (which makes things complicated and slow if you want to use Gson in a multi-threaded environment).

inder123:
@ferstl  @swankjesse  I am not sure entirely how optional dependencies work. I thought they work fine when you use reflection to lookup code.

Since Gson class has the ThreadSafe annotation added, the dependency will get forced, isn't it? Otherwise, we will have compilation error. Correct?

ferstl:
@inder123 @swankjesse [Here](https://maven.apache.org/pom.html#Dependencies) is a good description about optional dependencies.

You will have the new dependency when you are compiling Gson. But when you use Gson as a dependency, it will be ignored entirely. This also won't cause any errors at runtime because the `ThreadSafe` annotation's retention policy is `CLASS`.

Btw, Guava does [exactly the same](https://github.com/google/guava/blob/master/guava/pom.xml#L25).

inder123:
@ferstl  Thanks for the explanation.
Retention policy class definition from Javadoc: Annotations are to be recorded in the class file by the compiler but need not be retained by the VM at run time.

This to me indicates that any library using Gson will probably need ThreadSafe while compiling, but it is possible that the compiler ignores annotations with missing classes. I will try it out and see what this leads to.

inder123:
As we are readying for a new release, I am having second thoughts on this one.
So far Gson has avoided all external dependencies. So I would like to continue doing the same. I will update the Gson documentation to indicate Threadsafe. I will revert this change.

PR #667, opened 2015-07-19
JsonWriter#value supports opaque JSON values.

adamtanner:
Add a polymorphic version of `value(String value)` that takes an additional boolean argument (`encode`) that can be used to write the string directly to the underlying writer without modification, while maintaining the state of the `JsonWriter` so that it can continue to be written to.

The intended use case for this is when building JSON that contains a pre-serialized JSON string as a value in an object or array.

Alternative API ideas/open questions:
*  Should a separate method be used instead of a polymorphic version of `value(String)`? Maybe something like `rawValue(String)`?

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -421,6 +421,23 @@ public class JsonWriter implements Closeable, Flushable {
   }
 
   /**
+   * Writes {@code value} directly to the writer without quoting or
+   * escaping.
+   *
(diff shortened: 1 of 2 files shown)


swankjesse:
Rename to `jsonValue()` ?

swankjesse:
...and I don't think there's much value to sharing code between strings and raw JSON strings. I'd rather you copy&paste to make this separate.

adamtanner:
Renamed to `jsonValue`, decoupled from `value(String)`, and squashed into a single commit.

Let me know if there's anything else. :smile:

Thanks!

swankjesse:
Brilliant.

PR #669, opened 2015-07-20
Replace localhost lookup with static IP to fix test.

adamtanner:
Calling InetAddress.getLocalHost() will cause a lookup to occur that may
fail with a java.net.UnknownHostException if the system the test is
running on is not configured correctly.

This is often fixed by echoing "127.0.0.1 $HOSTNAME" to /etc/hosts, but
in this case it seems easier to pick a static IP string to avoid the
lookup entirely and prevent false negatives in the test.


inder123:
Ok. Won't this force a network connection and DNS lookup?

adamtanner:
Based on my understanding of the code from JDK 1.7 it looks like it won't make a network connection if the host is an IP address.

see: http://www.docjar.com/html/api/java/net/InetAddress.java.html#1081

May have missed something on my part though.

inder123:
Ok, Thanks for checking. :+1:

Issue #668, opened 2015-07-20
Maven release?


adamtanner:
Just wondering if you have any upcoming plans to do a Maven release. I'm happy to help if I can.

Thanks!

inder123:
Any specific feature/bug that you are looking forward to?

adamtanner:
Yeah, we're looking to use the `jsonValue` feature I just added in #667 for a Java client we're writing for [Martian Proxy](https://www.github.com/google/martian). I was going to wait, but it seemed like releases are few and far between so I figured I'd ask. :smile:

Thanks!

adamtanner:
Bump. Just wondering what your thoughts are on this. We want to release and I need to know if I should rework the library to work without this feature or wait a bit longer.

inder123:
I do want to cut a release. @swankjesse @JakeWharton @joel-leitch 
Shall we do a review of the proposed changes so far?

JakeWharton:
Looks like history was rewritten but the tags weren't updated?

https://github.com/google/gson/compare/gson-2.3.1...master

inder123:
Ok. Do you know what to do to update the tags? If so, go ahead. Thanks.

johnbarr:
I too would like a new release to use this same change.

JakeWharton:
Haven't gotten around to updating the tags. Here's the commit list though: https://github.com/google/gson/compare/eba6825a...master

JakeWharton:
Let's get this done before October 1st?

swankjesse:
Yes! @inder123 ready to sign off?
https://github.com/google/gson/blob/master/CHANGELOG.md

inder123:
I am comfortable with a new release. However, do we want to do a quick
review of pending pull requests and filed issues?

On Sat, Sep 26, 2015 at 4:48 PM, Jesse Wilson <notifications@github.com>
wrote:

> Yes! @inder123 <https://github.com/inder123> ready to sign off?
> https://github.com/google/gson/blob/master/CHANGELOG.md
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/gson/issues/668#issuecomment-143505867>.
>



-- 
http://www.singhinderjeet.com/

JakeWharton:
I did a few. Will try to get to more soon.

advayDev1:
[...]

PR #670, opened 2015-07-20
Change references to Google Code to Github in pom.xml.

adamtanner:
Not sure if this is entirely correct, but thought this might be blocking new releases so I figured I'd give it a shot. Let me know if this looks right.

Related: #668


inder123:
:+1:

PR #671, opened 2015-07-24
Support for adding primitives directly to JsonArray instances

ownaginatious:
Overloaded the "add(...)" method in JsonArray to support directly putting in primitives as arguments.

Now instead of inserting them in the old overly verbose way:
```java
JsonArray array = new JsonArray();
array.add(new JsonPrimitive(true));
array.add(new JsonPrimitive(12));
array.add(new JsonPrimitive("hello"));
array.add(new JsonPrimitive('c'));
```
You can do:
```java
JsonArray array = new JsonArray();
array.add(true);
array.add(12);
array.add("hello");
array.add('c');
```

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -55,10 +55,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @param bool the boolean that needs to be added to the array.
    */
   public void add(Boolean bool) {
-    if (bool == null) {
-      elements.add(JsonNull.INSTANCE);
-    }
-    elements.add(new JsonPrimitive(bool));
+    elements.add(bool == null ? JsonNull.INSTANCE : new JsonPrimitive(bool));
   }
 
   /**
@@ -67,10 +64,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @param character the character that needs to be added to the array.
    */
   public void add(Character character) {
-    if (character == null) {
-      elements.add(JsonNull.INSTANCE);
-    }
-    elements.add(new JsonPrimitive(character));
+    elements.add(character == null ? JsonNull.INSTANCE : new JsonPrimitive(character));
   }
 
   /**
@@ -79,10 +73,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @param number the number that needs to be added to the array.
    */
   public void add(Number number) {
-    if (number == null) {
-      elements.add(JsonNull.INSTANCE);
-    }
-    elements.add(new JsonPrimitive(number));
+    elements.add(number == null ? JsonNull.INSTANCE : new JsonPrimitive(number));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ownaginatious:
I signed it!

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -50,6 +50,54 @@ JsonArray deepCopy() {
...
 
   /**
+   * Adds the specified boolean to self.
+   *
+   * @param bool the boolean that needs to be added to the array.
+   */
+  public void add(Boolean bool) {
+    if (bool == null) {
+      elements.add(JsonNull.INSTANCE);
+    }
```
JakeWharton: The lack of `return` will add both. I think you want:
```java
elements.add(bool == null ? JsonNull.INSTANCE : new JsonPrimitive(bool));
```
for all these methods.

JakeWharton:
I don't know the DOM API well because it's inefficient to use, but if this was determined as something merge-worth it needs litmus tests verifying behavior of null and non-null arguments of each of these methods.

ownaginatious:
Sorry about that; I was too hasty on committing my changes and submitting a pull request. Dumb oversight on my part.

Every other Java JSON library I've used (org.json, Jackson. etc) supports inserting primitives directly into arrays, so I found it odd that this one doesn't. I find it looks a lot cleaner if you ever have to implement your own JSON serializers.

I added a JUnit test for null/non-null tests as well as tests for other things.

JakeWharton:
If you have to implement your own serializers you would use `JsonReader` and `JsonWriter`, not the DOM-based `JsonObject` / `JsonElement` things.

ownaginatious:
Interesting. I don't think I've found any examples anywhere online of anyone using `JsonReader` and `JsonWriter` rather than the DOM elements in a serializer.

Here are all the top results when I search for how to make a Gson serializer/deserializer:
  * http://www.studytrails.com/java/json/java-google-json-custom-serializer-deserializer.jsp
  * http://www.javacreed.com/gson-serialiser-example/
  * http://www.programcreek.com/java-api-examples/index.php?api=com.google.gson.JsonSerializer

I can see why `JsonReader` and `JsonWriter` would make more sense in a streaming API, but I guess for simple things where performance is not a high priority, many find it complicated or verbose to use as opposed to the DOM.

Doesn't the API for JsonDeserializer kind of push you in the DOM direction, seeing as the method signature gives you a `JsonElement` rather than a `JsonReader` instance to work with?

Anyway, would you say this is enough reason is enough to justify this pull request?

gson/src/test/java/com/google/gson/functional/JsonArrayTest.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright (C) 2008 Google Inc.
```
swankjesse: Unlikely!

swankjesse:
I defer to @inder123 .

ownaginatious:
I should also add that `JsonObject` does expose methods for adding primitives directly instead of only `JsonElement`. Oddly though, only for `Number` and `String` and not for `Boolean` or `Character`, which `JsonElement` also supports as primitives.

```java
public void add(String property, JsonElement value);
public void addProperty(String property, Number value);
public void addProperty(String property, String value);
```

Seems like it would make sense to also add these to `JsonObject` from a completeness standpoint.
```java
public void addProperty(String property, Boolean value);
public void addProperty(String property, Character value);
```

Of course, this is just my point of view.

inder123:
@ownaginatious  Thanks for the changes. I am going to accept them.
Feel free to send another pull request with addProperty changes.

inder123:
:+1:

Issue #676, opened 2015-08-04
`RuntimeTypeAdapterFactory` overrides explicit control over `serializeNulls` behavior of contained objects


lcoote:
By default `serializeNulls` behavior is off for GSON which is useful for building compact JSON. However, for specific types of objects its necessary to switch this on to make a distinction between (a) properties for which no explicit value has been assigned vs (b) properties for which a `null` value has been explicitly assigned.

One way this is achieved is to use a `TypeAdapter` implementation to temporarily override the setting of this behavior on the `JsonWriter`.

E.g.
  
    static class TypeAdapterForSomeObject extends TypeAdapter<SomeObject> {

        @Override
        public void write(JsonWriter out, SomeObject value) throws IOException {
            boolean oldSerializeNulls = out.getSerializeNulls();
            try {
                out.setSerializeNulls(true);
                // serialization impl...
            } finally {
                out.setSerializeNulls(oldSerializeNulls);
            }
        }

        // ...
    
    }

However, when such a type is part of another class which is serialized within another wrapper object which relies on a `RuntimeTypeAdapterFactory`, these explicit null override behavior can be lost. That is because, the implementation of `TypeAdapter.write` initially serializes the outer value using a temporary `JsonTreeWriter` into a `JsonElement`. This temporary writer invokes the custom TypeAdapter and thus accurately collects the explicit `null` values. However, `RuntimeTypeAdapterFactory` then serializes the `JsonElement` using the externally provided `JsonWriter` using its own `serializeNulls` configuration and therefore the explicit `null` values in the element are discarded.

A full test is included below:

    public class Tests {
        
         // define value requiring explicit nulls
        static class Inner {
            String prop;
            Inner(String prop) {
                this.prop = prop;
            }
        }

        // define adapter to switch on explicit null
        static class TypeAdapterForInner extends TypeAdapter<Inner> {
    
            @Override
            public void write(JsonWriter out, Inner value) throws IOException {
                boolean oldSerializeNulls = out.getSerializeNulls();
                try {
                    out.setSerializeNulls(true);
                    out.beginObject().name("prop").value(value.prop).endObject();
                } finally {
                    out.setSerializeNulls(oldSerializeNulls);
                }
            }
    
            @Override
            public Inner read(JsonReader in) throws IOException {
                in.beginObject();
                in.nextName();
                String value = null;
                if (in.peek() == JsonToken.STRING) {
                    value = in.nextString();
                } else {
                    in.nextNull();
                }
                in.endObject();
[...]

PR #677, opened 2015-08-04
fix Issue #676 - "explicit" nulls within RuntimeTypeAdapterFactory

lcoote:
Using custom TypeAdapters, it is possible to force null values within a
particular type to be explicitly serialized, by overriding the
setSerializeNulls setting on the supplied JsonWriter.

However, this behavior fails in the context of a
RuntimeTypeAdapterFactory because a second JsonTreeWriter is used and
settings are not propagated from one to the other.

This change fixes this by:

- propagating the original setSerializeNulls setting to the created
JsonTreeWriter
- temporarily forcing the setSerializeNulls setting to true on the
oriiginal JsonWriter to ensure any explicit nulls are preserved.

Issue #676


lcoote:
I signed it!

Issue #675, opened 2015-07-30
StackOverflowError in delegating TypeAdapterFactory


daniel-huss:
GSON Version: 2.3.1
Description: ```gson.getDelegateAdapter(this, type)``` causes inifinite recursion when used with a @JsonAdapter annotated type.

Minimal example:
```java
package test.gson;

import com.google.gson.Gson;
import com.google.gson.TypeAdapter;
import com.google.gson.TypeAdapterFactory;
import com.google.gson.annotations.JsonAdapter;
import com.google.gson.reflect.TypeToken;

/* @formatter:off */
public class GsonStackOverflowError {

    public static void main(final String[] args) {
        new Gson().fromJson("{}", A.class);
    }

    @JsonAdapter(Adapter.class)
    public static class A {}

    public static class Adapter implements TypeAdapterFactory {
        @Override
        public <T> TypeAdapter<T> create(final Gson gson, final TypeToken<T> type)
        {
            gson.getDelegateAdapter(this, type);
            throw new IllegalStateException("Not reproducible");
        }
    }
}
```

daniel-huss:
GSON 2.3 throws IllegalArgumentException instead, see http://stackoverflow.com/questions/29445165/

swankjesse:
Yup, `getDelegateAdapter()` doesn't work with `@JsonAdapter`. Needs to be fixed.

manishgo:
Any updates on this issue?

kevinmost:
I can confirm that the issue is resolved as of Gson 2.7. Should this issue be closed?

JakeWharton:
Thanks for following up

PR #682, opened 2015-08-06
[fix] StackOverflowError in delegating TypeAdapterFactory #675

AnisovAleksey:


gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -56,12 +60,24 @@ public JsonAdapterAnnotationTypeAdapterFactory(ConstructorConstructor constructo
...
+        }
+        return factory.create(gson, fieldType);
     }
 
     throw new IllegalArgumentException(
         "@JsonAdapter value must be TypeAdapter or TypeAdapterFactory reference.");
   }
+
+    @Override
+    public boolean equals(Object obj) {
```
swankjesse: This is not an equivalence relation.

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -32,6 +35,7 @@
 public final class JsonAdapterAnnotationTypeAdapterFactory implements TypeAdapterFactory {
 
   private final ConstructorConstructor constructorConstructor;
+    private static final List<TypeAdapterFactory> createdFactories = new ArrayList<TypeAdapterFactory>();
```
swankjesse: This is a memory leak.

swankjesse:
I don't think this is the right approach to fixing the problem. Can we get by with just searching for skipPast, and trying the whole list if skipPast isn't found?

threat70:
static list is issue like a static method in JsonAdapterAnnotationTypeAdapterFactory. I have no ideas (for now) how make method non-static (maybe separate annotation for field and class). equals method - like hack, it can be any one method ;)
I think owner of created adapters by annotation (JsonAdapterAnnotationTypeAdapterFactory), must known about these adapters (like list). my solution not fully completed, but it topic for discussion. 
What do you think about this?

swankjesse:
Can we get by with just searching for skipPast, and trying the whole list if skipPast isn't found?

threat70:
I not fully understand,you mean try to create all adapters from list if skipPast isn't found?

swankjesse:
Yes, exactly.

Marcono1234:
The original issue seems to be fixed (probably by #873?), see also #675 which is referenced in the commit of this PR. The test added in this PR does not throw a StackOverflowError anymore in the latest Gson version.

PR #683, opened 2015-08-09
Made JsonArray implement the List interface.

inder123:


inder123:
@swankjesse @JakeWharton Is this a good idea (to implement List instead of Iterable for JsonArray)? Bloats the API a little bit, but useful nonetheless.

swankjesse:
It isn't a binary-compatible change.

swankjesse:
Don't forget equals().

swankjesse:
It don't feel strongly either way.

If we make this change we will need to figure out the binary compatibility problem.

Also if we make this change we should change JsonObject to implement Map for symmetry.

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -30,7 +32,7 @@
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
-public final class JsonArray extends JsonElement implements Iterable<JsonElement> {
+public final class JsonArray extends JsonElement implements List<JsonElement> {
```
JakeWharton: You'll also want the `RandomAccess` marker interface.

inder123:
I agree we should keep binary compatibility. Closing the pull request.

PR #688, opened 2015-08-14
Add METHOD target for use with AutoValue's abstract property methods

### LINKED ISSUE #687: Add METHOD Target to SerializedName attribute
I'm in the process of making an AutoValue extension that will generate a Gson TypeAdapterFactory for generated data types, and I'd like to support `@SerializedName`.  Currently, the attribute only targets `ElementType.FIELD`s, meaning users can't put it on the abstract property methods that AutoValue uses.  I'd like to have `ElementType.METHOD` added to the list of supported `ElementType`s so that I don't have to include a separate annotation for AutoValue users.

Here is what the user's class would look like in this case:

```java
@AutoValue public abstract class Foo {
  @SerializedName("Baz") public abstract String baz();
}
```

MichaelEvans:
Closes #687

### DIFF
--- a/gson/src/main/java/com/google/gson/annotations/SerializedName.java
+++ b/gson/src/main/java/com/google/gson/annotations/SerializedName.java
@@ -64,7 +64,7 @@ import java.lang.annotation.Target;
  * @author Joel Leitch
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.FIELD)
+@Target({ElementType.FIELD, ElementType.METHOD})
 public @interface SerializedName {
 
   /**


gson/src/main/java/com/google/gson/annotations/SerializedName.java
```diff
@@ -64,7 +64,7 @@
  * @author Joel Leitch
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.FIELD)
+@Target({ElementType.FIELD, ElementType.METHOD})
```
inder123: @swankjesse Do we really want to do this?

If we allow SerializedName on methods, we should actually support it in full. 
We Don't and we seem to have no plans to support it either. So, why open it up?

swankjesse:
It's a pragmatic solution to the problem AutoValue users have. I don't think it obligates us to doing getters and setters, and I'm quite against doing that!

swankjesse:
Details are here: https://github.com/google/gson/issues/687

inder123:
Ok

PR #689, opened 2015-08-14
Add unit tests for constructors from util classes

vbauer:
I think that constructors from util classes should pass the following rules (best practices):

1. Each util class should have only the one constructor without parameters.
2. Constructor should be private.
3. Code should not have possibility to instantiate this class even using reflection mechanism (UnsupportedOperationException)

### DIFF
--- a/gson/src/test/java/com/google/gson/common/MoreAsserts.java
+++ b/gson/src/test/java/com/google/gson/common/MoreAsserts.java
@@ -69,4 +69,5 @@ public class MoreAsserts {
     Assert.assertFalse(a.equals(null));
     Assert.assertFalse(a.equals(new Object()));
   }
+
 }
--- a/gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
+++ b/gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
@@ -30,6 +30,10 @@ package com.google.gson.internal;
  * @author Joel Leitch
  */
 public final class $Gson$Preconditions {
+  private $Gson$Preconditions() {
+    throw new UnsupportedOperationException();
+  }
+
   public static <T> T checkNotNull(T obj) {
     if (obj == null) {
       throw new NullPointerException();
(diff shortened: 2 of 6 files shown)


gson/src/test/java/com/google/gson/ConstructorContractTest.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.gson.internal.bind.TypeAdapters;
+import junit.framework.TestCase;
+
+/**
+ * Unit test to check constructors in util classes.
+ *
+ * @author Vladislav Bauer
+ */
+
+public class ConstructorContractTest extends TestCase {
```
swankjesse: You can delete this. If these tests fail it doesn't harm anything.

gson/src/test/java/com/google/gson/common/MoreAsserts.java
```diff
@@ -69,4 +72,30 @@ public static void assertEqualsAndHashCode(Object a, Object b) {
     Assert.assertFalse(a.equals(null));
     Assert.assertFalse(a.equals(new Object()));
   }
+
+  /**
+   * Asserts that the specified util class can not be instantiated.
+   * @param utilClass util class
+   */
+  public static void assertUtilConstructor(Class<?> utilClass) {
```
swankjesse: You can delete this too!

gson/src/test/java/com/google/gson/ConstructorContractTest.java
vbauer: If this test fails, it means that design of some class has a problem (somebody can instantiate it).

swankjesse: Yes. But this test offers extremely little utility: it's unlikely to ever detect a regression. And the severity of that bug is also extremely low. We want to write tests that prevent problems. This doesn't do that.

vbauer: Ok, I'll delete it. I just wanted to check my changes and increase test coverage.

vbauer:
PR was updated, thank you for review.

swankjesse:
Thanks!

vbauer:
Thank you!

Issue #639, opened 2015-04-21
SerializedName should allow multiple names to be matched to a single name for input


inder123:

ikkyu0319:
+1

fahimk:
@inder123 @ikkyu0319 Do you mean something like this (this isn't available yet) ?

```java
@SerializedName({"name", "firstName", "userName"})
public String name;

@SerializedName("id")
public String id;
```

where the idea is to re-use a model for multiple API endpoints but the attribute returned has a different name?

ikkyu0319:
Yes~ I need this~~

inder123:
To maintain backward compatibility, I revised it to:

@SerializedName(value="default_name", alternate={"name", "firstName", "userName"})
public String name;

Let me know if you see a problem with this.

JakeWharton:
This shipped in 2.4

fahimk:
@inder123 this is great, thanks for working on it.

inder123:
@fahimk great. Looking forward to learning how it worked for you.

GaniduAsh:
How can i use this in kotlin  ? 
 @SerializedName(value="name", alternate={"person", "user"})
this gives me unexpected token error

JakeWharton:
Use arrayOf

GaniduAsh:
yes found it thanks

iballan:
Can we let the alternative or on of the alternatives get chosen when serializing object that will be sent to server ?

sivabe35:
For Kotlin ` @SerializedName(value="firstName", alternate= ["userName"])`

sunil965:
> To maintain backward compatibility, I revised it to:
> 
> @SerializedName(value="default_name", alternate={"name", "firstName", "userName"})
> public String name;
> 
> Let me know if you see a problem with this.

Hi, does the "default_name" should me the one among {"name", "firstName", "userName"} ?

sunil965:
I am trying this, but it's not working.
@SerializedName(value = "text", alternate = {"text", "image", "document", "voice", "audio", "video", "location", "contacts"})
    private String content;

akinsete:
@sunil965 alternate should be an array not object.

AungHtetKyawWho:
What up? May i help u.?

rajatsharma16:
Doesn't seem to work if there is a space:
`@SerializedName(value = "american express", alternate = {"AmericanExpress", "AMERICAN EXPRESS"})`

[...]

PR #699, opened 2015-09-07
Implemented support for multiple values for SerializedName annotation.

inder123:


inder123:
@JakeWharton @joel-leitch @swankjesse Please review and comment on this new feature that I implemented. See https://github.com/google/gson/issues/639 for the original feature request.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -63,13 +63,13 @@ static boolean excludeField(Field f, boolean serialize, Excluder excluder) {
     return !excluder.excludeClass(f.getType(), serialize) && !excluder.excludeField(f, serialize);
   }
```
swankjesse: Not binary compatible!

inder123: I am wondering may be should do a release that breaks binary compatibility with this and a few additional features (such as implement List for JsonArray).

What's the big downside of breaking binary compatibility.

JakeWharton: Libraries and apps which depend on Gson will break unless they all are
recompiled.

swankjesse:
Maybe just add another field for alternate names?

inder123:
@JakeWharton @swankjesse Ok, I have revised this pull request to be binary-compatible by using an additional field to specify alternate values. Take another look.

gson/src/main/java/com/google/gson/internal/bind/SerializedNameValue.java
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gson.internal.bind;
+
+import java.util.Iterator;
+
+final class SerializedNameValue implements Iterable<String> {
```
swankjesse: is this class really necessary? Why not just call createBoundField with a `List<String>` ?

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
```diff
@@ -0,0 +1,53 @@
...
+    assertEquals("v1", gson.fromJson("{'name':'v1'}", MyClass.class).a);
+
+    // Both name1 and name2 gets deserialized to b
+    assertEquals("v11", gson.fromJson("{'name1':'v11'}", MyClass.class).b);
+    assertEquals("v2", gson.fromJson("{'name2':'v2'}", MyClass.class).b);
+    assertEquals("v3", gson.fromJson("{'name3':'v3'}", MyClass.class).b);
+  }
+
+  public void testMultipleNamesInTheSameString() {
+    // The last value takes precedence
```
swankjesse: Should this crash? This seems like a potential security bug?

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.gson.annotations.SerializedName;
+
+import junit.framework.TestCase;
+
+public final class SerializedNameTest extends TestCase {
+  private final Gson gson = new Gson();
+
+  public void testFirstNameIsChosenForSerialization() {
+    MyClass target = new MyClass("v1", "v2");
+    assertTrue(gson.toJson(target).contains("\"name1\":\"v2\""));
```
swankjesse: Also assert that the other names aren't included?

swankjesse:
This has ballooned into a pretty big feature. Do we really need it?

gson/src/main/java/com/google/gson/internal/bind/SerializedNameValue.java
inder123: I did this to avoid copying the array for annotation alternate names.

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
inder123: I think it is fine. Gson accepts multiple values mapping to the same name at other places.

gson/src/main/java/com/google/gson/internal/bind/SerializedNameValue.java
inder123: Revised the implementation to use List<String>

inder123:
@swankjesse Look at the entire diff again. I have reduced it to be of relatively small size. Does that allay your concerns?

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -213,9 +232,12 @@ private Adapter(ObjectConstructor<T> constructor, Map<String, BoundField> boundF
...
       out.beginObject();
       try {
-        for (BoundField boundField : boundFields.values()) {
-          if (boundField.writeField(value)) {
-            out.name(boundField.name);
+        for (Map.Entry<String, BoundField> entry : boundFields.entrySet()) {
+          String name = entry.getKey();
+          BoundField boundField = entry.getValue();
+          String defaultName = boundField.name;
+          if (name.equals(defaultName) && boundField.writeField(value)) {
```
swankjesse: Wouldn't it be simpler to just create a boundField that has `serialized = false` for each of the alternates? Also a lot less string comparison in the default encoding loop.

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import junit.framework.TestCase;
+
+public final class SerializedNameTest extends TestCase {
+  private final Gson gson = new Gson();
+
+  public void testFirstNameIsChosenForSerialization() {
+    MyClass target = new MyClass("v1", "v2");
+    String json = gson.toJson(target);
+    assertTrue(json.contains("\"name1\":\"v2\""));
```
swankjesse: Why not just assert the exact contents of the emitted JSON? it's deterministic!

swankjesse:
This is Gson's inner loop on encoding, so I'm anxious about doing extra work that makes that slower. In this case I think we should avoid doing an extra `String.equals()` call for every single field, whether or not it uses this new annotation!

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
inder123: That is a great suggestion. Implementing it in the next commit.

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
inder123: Yes, will do that.

inder123:
Ok, all these changes have further shrunk the overall diff. Good stuff.
Please take another look.

swankjesse:
LGTM

Please rebase and squash (http://rebaseandsqua.sh) so our git log has only one commit!

inder123:
Rebased and squashed (learned something new today, neat!). Take another look.

swankjesse:
Cool!

PR #700, opened 2015-09-08
Expose serializeNulls with an accessor.

swankjesse:
Useful for frameworks that build on Gson.


JakeWharton:
Your local copy is so old (how old is it?) that the SHA on which this branch is based didn't have a `.travis.yml`.

swankjesse:
Yeah, something is unhappy here. Won't happen again!

PR #704, opened 2015-09-26
Expose newJsonWriter wrapper instead of just options.

JakeWharton:
This is more future proof to ensure that consumers holding on to TypeAdapters for efficiency can always encode JSON in the same way as if the Gson instance was used directly.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -247,11 +247,6 @@ public final class Gson {
     this.factories = Collections.unmodifiableList(factories);
   }
 
-  /** Returns true if this Gson instance writes fields with null values. */
-  public boolean serializeNulls() {
-    return serializeNulls;
-  }
-
   private TypeAdapter<Number> doubleAdapter(boolean serializeSpecialFloatingPointValues) {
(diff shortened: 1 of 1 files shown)


swankjesse:
LGTM

inder123:
@JakeWharton  When did we introduce serializeNulls()? I presume it hasn't been released yet.

inder123:
I don't understand the use-case of newJsonWriter(writer)? What would programmers use it for? Why can't they just use the Gson instance to do the same?

JakeWharton:
When you cache `TypeAdapter` instances you have no way of creating a `JsonWriter` whose configuration matches that of what the `Gson` instance would create. Instead, it has unsatisfactory behaviors like serializing nulls by default. This allows those caching (or bringing their own) `TypeAdapter` instances to serialize with the same settings as if they had just called one of the convenience methods on `Gson`.

inder123:
@JakeWharton Thanks for the explanation. Yes, I have run into that as well.
Also, checked that serializeNulls was recently introduced and hence can be safely removed.

JakeWharton:
Yeah Jesse added that method for this very same use-case. But it only
encompasses one of a few options that can be configured.

Issue #627, opened 2015-03-19
LazilyParsedNumber does not implement eqals and hashCode methods


GoogleCodeExporter:
```
test case to reproduce the issue

package com.google.gson.internal;

import junit.framework.TestCase;

public class LazilyParsedNumberTest extends TestCase {

    public void testHashCode(){
        LazilyParsedNumber n1 = new LazilyParsedNumber("1");
        LazilyParsedNumber n1Another = new LazilyParsedNumber("1");
        assertEquals(n1.hashCode(), n1Another.hashCode());
    }

    public void testEquals(){
        LazilyParsedNumber n1 = new LazilyParsedNumber("1");
        LazilyParsedNumber n1Another = new LazilyParsedNumber("1");
        assertTrue(n1.equals(n1Another));
    }


What is the expected output? What do you see instead?
Any class should implement hashCode and equals when implementing Number 
interface.

What version of the product are you using? On what operating system?
2.3

Please provide any additional information below.
https://code.google.com/p/google-gson/issues/detail?id=523
Similar one.

I could work on these, but first I would like to know if there was any reason 
you decided not to implement these methods? 
```

Original issue reported on code.google.com by `vrab...@gmail.com` on 4 Feb 2015 at 12:06

PR #705, opened 2015-09-27
Implemented equals and hashcode for LazilyParsedNumber to fix

inder123:
https://github.com/google/gson/issues/627


gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
```diff
@@ -75,4 +75,26 @@ public String toString() {
   private Object writeReplace() throws ObjectStreamException {
     return new BigDecimal(value);
   }
-}
\ No newline at end of file
+
+  @Override
+  public int hashCode() {
+    return value == null ? 0 : value.hashCode();
```
swankjesse: value is never null

gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
```diff
@@ -75,4 +75,26 @@ public String toString() {
...
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null || !(obj instanceof LazilyParsedNumber)) {
```
swankjesse: the instanceof check includes a null check

gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
```diff
@@ -75,4 +75,26 @@ public String toString() {
...
+      return false;
+    }
+    LazilyParsedNumber other = (LazilyParsedNumber) obj;
+    return equals(value, other.value);
+  }
+
+  private static boolean equals(Object obj1, Object obj2) {
+    if (obj1 == null) return obj2 == null;
```
swankjesse: value is never null

swankjesse:
LGTM with some nitpicks

inder123:
Great comments! incorporated them in the pull request.

PR #710, opened 2015-10-07
Updating the ProtoTypeAdapter to support custom field names/values annotations

Bhasmithal:
The current ProtoTypeAdapter is quite old, doesn't work for a few things correctly (I don't remember what exactly but it had some failures when I tried to use it), so I decided to update it and to make it better.

This version of ProtoTypeAdapter now:
- Lets you customize the proto field name serialization per field
- Lets you use different cases for serialization (e.g. you could use UpperCamelCase for your proto field names and willing to serialize them using LowerHyphen)
- Lets you choose to serialize enum values with their value (0, 1, 2, ...) instead of their name.

We've been using it in prod for a while, I added quite a few unit tests to make sure everything works properly.


proto/pom.xml
```diff
@@ -91,6 +109,7 @@
                 <pathconvert pathsep=" " property="proto.files" refid="proto.path" />
                 <exec executable="protoc" failonerror="true">
                   <arg value="--java_out=src/main/java" />
+                  <arg value="--proto_path=/usr/include" />
```
Bhasmithal: I'm not sure if this is the best way to set this value. I was looking for something that would point the maven-antrun-plugin to /usr/include, but I coudln't find any plugin/project parameter that would allow me to do so.

This probably needs to be updated to a value that depends on the OS. What's the best way to do that?

Bhasmithal:
I signed it!

proto/pom.xml
```diff
@@ -52,21 +55,35 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>1.7.1</version>
+      <version>2.3.1</version>
```
inder123: Change it to 2.4, the latest released version.

Bhasmithal: Done.

inder123:
:+1:

PR #715, opened 2015-10-20
Moved the .gitignore file to a more appropriate place

Bhasmithal:
That will avoid committing the generated/ folder that is useless in the repository.


inder123:
:+1:

PR #716, opened 2015-10-21
Perf: Moving string length calculation to local variable.

arriolac:
Performance improvement by moving string length calculation to local variable. Previous implementation calculated the string's length for each each loop iteration.


arriolac:
I signed it!

swankjesse:
This method is intrinsic.

arriolac:
Yikes. I thought `length()` was O(n). However, this is still a very minor pref improvement because of the cost of making a method call.

swankjesse:
The cost of the method call is like, actually zero. You should send us performance improvements, but they should be based on performance problems, not speculation.

swankjesse:
Thanks for the PR though. I want you to send me something that makes a measured improvement, with before/after benchmarks!

PR #717, opened 2015-10-21
Support for anonymous class serialization

jedediah:
Add a flag to `GsonBuilder` that lets anonymous classes through the serialization pipeline. Among other things, this allows for convenient creation of ad-hoc documents e.g.:
```
gson.toJson(new Object() {
    final int things = 17;
    final String stuff = "fun";
});
```

### Rationale

Currently, anonymous classes are silently dropped or replaced with null, unconditionally. This is needlessly restrictive, and likely to take developers by surprise. The Gson design document already says it best, though on a slightly different subject:

> We could have chosen to restrict the serialization to support only generic collections, but chose not to. This is because often the user of the library are concerned with either serialization or deserialization, but not both. In such cases, there is no need to artificially restrict the serialization capabilities.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -251,6 +251,17 @@ public final class GsonBuilder {
   }
 
   /**
+   * Configures Gson to include anonymous and/or local classes during serialization.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 1.3
+   */
+  public GsonBuilder enableAnonymousClassSerialization() {
+    excluder = excluder.enableAnonymousClassSerialization();
+    return this;
+  }
+
+  /**
    * Configures Gson to apply a specific serialization policy for {@code Long} and {@code long}
    * objects.
    *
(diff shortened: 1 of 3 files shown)


swankjesse:
I'm not a fan. If you want this, just embed the raw JSON.

jedediah:
Like this?

``` "{\"things\":17,\"stuff\":\"fun\"}" ```

Pretty gross IMHO, particularly the manual string escaping. And if the values are anything non-trivial, you have to serialize them and splice them in, which is even more gross. And you get no syntax checking or anything else you might have in the serialization pipeline. And any serialization API layers you have on top of Gson have to accept raw strings as well as objects.

But the ad-hoc objects were just the simplest example I could think of. Here is the reason *we* want this IRL: We use interfaces to define JSON messages, like this:

```java
interface NameChange {
    String name();
}

```
Custom serializers know to call the getter method. We use interfaces instead of classes because they are composable, and to avoid copying values, and to ensure that messages are complete. It's a nice system.

It would be much nicer though if we could implement messages inline, like this:
```java
void changeName(User user) {
    send(gson.toJson(new NameChange() {
        @Override String name() { return user.getName(); }
    }));
}
```

Unfortunately, Gson makes it absolutely impossible to pass anonymous classes through the pipeline, no matter how much we are willing to customize it. There's no reason we shouldn't be able to opt-in to allow them.

swankjesse:
How would you deserialize this?

jedediah:
Nothing per-se prevents Gson from deserializing an anonymous class. If it's also a non-static inner class, i.e. bound to an outer instance, then it can't be deserialized without a custom instance creator, but that is equally true for named inner classes, which Gson is happy to work with. An anonymous class from a static context can be deserialized just fine, as far as I know.

Regardless, you would have to write some pretty strange looking code to even attempt to deserialize an anonymous class. If someone actually goes to the trouble of doing it, there's no reason for Gson to stand in their way.

swankjesse:
I think you might have better luck with another framework. Gson is opinionated and the pattern you've invented that causes this requirement doesn't fit with the rest of the project.

Issue #415, opened 2015-03-19
Support stream access to long string literals


GoogleCodeExporter:
```
Sometimes the strings in your JSON document are so darn long you want to access 
them with Reader/Writer rather than Strings.

http://groups.google.com/group/google-gson/browse_thread/thread/3d73c3084864fac4
/ac8bf9cb4223d457?show_docid=ac8bf9cb4223d457
```

Original issue reported on code.google.com by `limpbizkit` on 25 Feb 2012 at 4:03

PR #718, opened 2015-10-21
added method to fetch next string into OutputStream

rodnaph:
Related open issue: https://github.com/google/gson/issues/415

This PR is without much thought or tests yet, but I'm opening it to get some feedback on if this would be a potential addition, and how i'd go about doing that.

The use-case this patch solves (as described in the other ticket) is that sometimes JSON documents can contain very large values (base64 encoded file content in my case) which you don't want to have to realise in memory at once.  This allows reading these sections into an OutputStream which can write it to a file for example.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -21,7 +21,9 @@ import com.google.gson.internal.bind.JsonTreeReader;
 import java.io.Closeable;
 import java.io.EOFException;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Reads a JSON (<a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627</a>)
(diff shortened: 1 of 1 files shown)


rodnaph:
I signed it!

swankjesse:
I like the idea, but as-is I want two changes:

1. Strings should be a character stream, not as a byte stream.
2. The stream should be something you pull from, not something you push into.

So the API probably wants to be this:

```
public Reader nextStringReader();
```

That reader returns the characters of the JSON string, unescaping them as necessary. We’ll also need tests, particularly tests that check things like `\t` and `\u0020` are properly unescaped.

rodnaph:
1. Sure
2. With `nextStringReader`, I assume it would then be the responsibility of the caller to ensure they've consumed the entire stream before fetching the next token? For example...

```
Reader r = reader.nextStringReader();
int i = reader.nextInt();
```

When calling `nextInt` the Reader needs to be been drained or skipped to avoid buffering it in memory.  I guess buffering it in memory could be the default behaviour if you have not consumed it already yourself?

swankjesse:
I think we should throw an IllegalStateException if you attempt to access the stream after the underlying JsonReader has advanced.

PR #719, opened 2015-10-21
Fix type hierarchy adapters to do a runtime check.

swankjesse:
Otherwise if we have a type hierarchy adapter for Vehicle, and we
attempt to decode a JSON string as a Car, we get the right exception
if the JSON string is actually decoded as a Truck.


inder123:
:+1:

PR #723, opened 2015-10-28
Use base64 encoding to serialize bytestrings

cconroy:


proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
```diff
@@ -0,0 +1,192 @@
...
+   * @param in A character array containing the Base64 encoded data.
+   * @return An array containing the decoded data bytes.
+   * @throws IllegalArgumentException If the input is not valid Base64 encoded data.
+   */
+  public static byte[] decode(char[] in) {
+    return decode(in, 0, in.length);
+  }
+
+  /**
+   * Decodes a byte array from Base64 format. No blanks or line breaks are allowed within the Base64
```
swankjesse: It’s weird that this class calls itself _lenient_ and then goes and makes a claim like this.

cconroy: agreed. @dragonsith i think has more context on this.

proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * <p> This class is used to encode and decode data in Base64 format as described in RFC 1521.
+ *
+ * <p> Project home page: <a href="http://www.source-code.biz/base64coder/java/">www.source-code.biz/base64coder/java</a><br>
+ * Author: Christian d'Heureuse, Inventec Informatik AG, Zurich, Switzerland<br> Multi-licensed: EPL
+ * / LGPL / GPL / AL / BSD / MIT.
+ *
+ * Modified to accept URL encoding, some apis removed from the original.
+ */
+public class LenientBase64Coder {
```
swankjesse: Are you particularly attached to this base64 coded? I’m more experienced with the one in okio & it has a few little benefits:
https://github.com/square/okio/blob/master/okio/src/main/java/okio/Base64.java 

It ignores whitespace.
It doesn't create an extra copy of the char array, instead operating on the `String` directly
It can decode base64url
It isn't broken for inputs longer than `(Integer.MAX_VALUE / 3)`

cconroy: FWIW, this is what we've been using in our internal fork of ProtoTypeAdapter. Assuming the output is equivalent, we could definitely switch to the okio base64coder.

proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
dragonsinth: I think it's just that it accepts either standard or url encoded form, and trailing '=' chars are optional.

proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
dragonsinth: I grabbed it off the web originally mostly because it was fast. :)

swankjesse: Yeah, and it looks fast, with an unrolled loop! But the allocations might dominate, particularly for small values.

proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
swankjesse: Yup you're right! It'll encode standard and decode either. In this use case that's good enough!

cconroy:
rebased to resolve merge conflicts with the threadsafe PR that landed.

proto/src/main/java/com/google/gson/protobuf/LenientBase64Coder.java
cconroy: Ok so @swankjesse should we keep this coder or add a dep on okio instead?

swankjesse: Copy the base64 coder from Okio? Or fix the 2GiB bug

PR #724, opened 2015-10-28
Thread safe method cache

cconroy:


swankjesse:
Is `MapMaker` major overkill? Would adding `synchronized` on the requesting method be sufficient?

swankjesse:
Upon further inspection, this entire thing is doing things the Gson 1.x way which is slow and bad. The right long term fix is to migrate this to `TypeAdapterFactory` which would mean we'd look up the method once, rather than once per field being decoded.

In the interim, this seems reasonable enough.

swankjesse:
If you and @dragonsinth wanna port this to `TypeAdapterFactory`, everything will be faster & we won't need this memory-hungry cache!

cconroy:
yeah, perhaps i'll tackle that during hackweek

proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
```diff
@@ -380,18 +384,23 @@ private EnumValueDescriptor findValueByNameAndExtension(EnumDescriptor desc,
     }
   }
 
-  private static Method getCachedMethod(Class<?> clazz, String methodName)
-      throws NoSuchMethodException {
-    if (!mapOfMapOfMethods.containsKey(methodName)) {
-      mapOfMapOfMethods.put(methodName, new HashMap<Class<?>, Method>());
-    }
+  private static Method getCachedMethod(Class<?> clazz, String methodName,
+      Class<?>... methodParamTypes) throws NoSuchMethodException {
```
dragonsinth: I take it the method param types don't factor into the key?  Why didn't you need the methodParamTypes before?  Seems a little odd.

proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
```diff
@@ -190,6 +191,9 @@ public static Builder newBuilder() {
   private static final com.google.protobuf.Descriptors.FieldDescriptor.Type ENUM_TYPE =
       com.google.protobuf.Descriptors.FieldDescriptor.Type.ENUM;
 
+  private static final ConcurrentMap<String, Map<Class<?>, Method>> mapOfMapOfMethods =
+      new MapMaker().makeMap();
```
dragonsinth: Does it matter whether it's String -> Map<Class, Method> vs. Class -> Map<String, Method>?  Just wondering since if it could be the latter, you could make the keys weak refs.

swankjesse: Weak keys won't help generally, the Gson instance strongly references the classes its configured to encode and decode.

proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
swankjesse: Yeah, I think we're relying on the fact that the caller never uses overloads. It's cheating, but not bad cheating.

proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
dragonsinth: I have to disagree, since you're shoving them into a static map.  Putting class/method references into a static map is bad smell, because you can end up pinning entire classloaders forever.  Releasing the Gson instance should ultimately allow for cleanup.

dragonsinth: By the way, wouldn't Class -> Map generally allow for less garbage?  In the case where you serialize one class type, you end up with 1 outer map and 1 inner map of string -> method.  But with the String -> Map formulation, serializing one type would end you up with 1 outer map of String -> Map and N inner maps of String -> Class, one map per method.

swankjesse: Oh, you're right that this is static. Ugh. I think the fix we want to ship is one where we switch this to `TypeAdapterFactory`, where we don't need any maps at all.

dragonsinth: https://gist.github.com/dragonsinth/4d5ed3a96fc411d4dc48

PR #728, opened 2015-11-05
minor modification

wildpeanut:
for internal IOException throw correct mapping Exception(JsonIOException)


wildpeanut:
I signed it!

swankjesse:
Unfortunately, I don't think this is compatible with what callers currently expect.

PR #729, opened 2015-11-05
updated minimum JDK version to 1.6.

inder123:
Added Overrides for methods implementing an interface.


swankjesse:
`String.isEmpty()` !

Issue #624, opened 2015-03-19
Error desirialization of ConcurrentNavigableMap 


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Create POJO with filled ConcurrentNavigableMap field
2. Sirialize to json string
3. Desirialize from json string

What is the expected output? What do you see instead?
Original state is expected. IllegalArgumentException is thrown

What version of the product are you using? On what operating system?
Gson 2.3.1, java 8, windows 7

Please provide any additional information below.
Gson works only with class type field such as ConcurrentSkipListMap.
Good programming style is using interface type such as ConcurrentNavigableMap. 
See attach for log and unit-test

```

Original issue reported on code.google.com by `dkhomya...@gmail.com` on 29 Jan 2015 at 8:34

Attachments:
* [gsonErrLog.TXT](https://storage.googleapis.com/google-code-attachments/google-gson/issue-624/comment-0/gsonErrLog.TXT)
* [JsonUtilsTest.java](https://storage.googleapis.com/google-code-attachments/google-gson/issue-624/comment-0/JsonUtilsTest.java)

inder123:
ConcurrentNavigableMap was introduced in JDK 1.6.
@swankjesse @JakeWharton @joel-leitch  Shall we abandon support for JDK 1.5?

JakeWharton:
I'm in favor. JVM users on Java 5 aren't likely going to be updating their Gson instances at any frequent pace. The stronger realization of this change is that it eliminates Android 2.2 and older. This accounts for 0.2% of the 1.some billion devices that talk to Google Play in the last 2 weeks. As far as I know, Google has stopped officially supporting it with their apps, and this was well behind others who have long since moved on from this far outdated platform.

swankjesse:
Yeah, those users can continue to use earlier releases.

inder123:
Ok. Agreed. Let's switch to Java 6 as the minimum version once we introduce support for ConcurrentNavigableMap and other Java 6 standard classes.

PR #730, opened 2015-11-05
Added support to serialize/deserialize ConcurrentMap and ConcurrentNa…

inder123:
…vigableMap. This fixes https://github.com/google/gson/issues/624

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -16,10 +16,6 @@
 
 package com.google.gson.internal;
 
-import com.google.gson.InstanceCreator;
-import com.google.gson.JsonIOException;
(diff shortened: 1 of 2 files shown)


swankjesse:
Isn't this like, particularly sketchy? Because it might be changing as we encode?

inder123:
@swankjesse Please explain what you mean by "encode". 
We already do something similar for Maps (use LinkedHashMap), and List. This is just an extension to support additional java.util.concurrent classes.

swankjesse:
It's very weird to serialize a value that's under active modification. When would you do this?

inder123:
Where is it under active modification? Are you saying it because it is a concurrent map?

inder123:
For that matter, ANY java object can be under active modification while Gson is serializing it.

swankjesse:
Works for me. I still think code that does this is weird, but I can imagine non-weird use cases.

PR #733, opened 2015-11-07
Supported Currency and added additional tests for Vector, Stack and Properties.

inder123:


inder123:
@swankjesse  review please?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -576,6 +577,18 @@ public void write(JsonWriter out, UUID value) throws IOException {
 
   public static final TypeAdapterFactory UUID_FACTORY = newFactory(UUID.class, UUID);
 
+  public static final TypeAdapter<Currency> CURRENCY = new TypeAdapter<Currency>() {
```
swankjesse: Is anyone asking for this?!

swankjesse:
LGTM, though I'm quite surprised that you're adding such a bogus type.

Also, please squash.

inder123:
No one asking for it, but I noticed that the default serialization was not correct.

swankjesse:
Got it. Slightly scary, this change is not interoperable to anyone (accidentally) relying on the current broken behavior. I like Moshi's behavior here, which is to throw an exception! `Platform class java.util.Vector annotated [] requires explicit JsonAdapter to be registered`

inder123:
@swankjesse  I only added support for Currency, not vector.
Yes, it breaks with the current behavior, but I think this is what any normal user will expect. In the past, we have added default adapters for classes from java.util.* packages over time, technically breaking the serialization semantics.

inder123:
merging based on earlier LGTM

PR #734, opened 2015-11-10
support contains(string),contains(int) in jsonarray class

cloudbow:
This change is to support searching string in JsonArray since a novice developer may not really know about JsonPrimitves

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -151,6 +151,28 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
   public boolean contains(JsonElement element) {
     return elements.contains(element);
   }
+  
+  /**
+   * Overloaded convenience method which searches a string in JsonArray and returns true if it exists
+   * Assumes that the JsonArray contains string elements
+   * @return true if this array contains the specified element.
+   * @param element whose presence in this array is to be tested
+   */
+  public boolean contains(String element) {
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -151,6 +151,28 @@ public JsonElement remove(int index) {
...
+     return contains(new JsonPrimitive(element));     
+  }
+  
+  /**
+   * Overloaded convenience method which searches a Integer in JsonArray and returns true if it exists
+   * Assumes that the JsonArray contains integer elements
+   * @return true if this array contains the specified element.
+   * @param element whose presence in this array is to be tested
+   */
+  public boolean contains(Integer element) {
```
inder123: what about other primitive types?

inder123:
marginal value for this PR, adds 2 extra methods.
Sorry, I am inclined to ignore it.

Thank you for the effort nonetheless. Also, appreciate the fact that you wrote a test.

gson/src/main/java/com/google/gson/JsonArray.java
cloudbow: Json array can contain only integer, string , json objects. Most of us use contains for checking String,Integer. Hence only 2 primitives.

inder123: also boolean

swankjesse:
Agreed.

PR #738, opened 2015-11-16
Added details and causes while throwing AssertionError

inder123:


swankjesse:
LGTM. Rebase & squash.

inder123:
Rebase and squash script no longer works for me.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -796,7 +796,7 @@ public EnumTypeAdapter(Class<T> classOfT) {
           constantToName.put(constant, name);
         }
       } catch (NoSuchFieldException e) {
-        throw new AssertionError();
+        throw new AssertionError("Missing field in " + classOfT.getName(), e);
```
inder123: This is JDK 1.7-specific constructor, so can not be used.

Issue #650, opened 2015-06-04
Default Deserialization Strategy for Dates


oleersoy:
Currently an exception is thrown when the date is specified in milliseconds.  It would be great if gson would assume that this is the default, or would at least provide a simple builder option for it.  More details here:

http://stackoverflow.com/questions/30652314/gson-datetypeexception-when-converting-date-in-typed-in-milliseconds/30653568#30653568

Cheers,
- Ole

oleersoy:
Just wanted to add some more information.  I'm testing a REST update method.  First I submit a new contact.  Then I update the contact and use gson to create the JSON string sent in the update.  When I do this gson serializes the date to:

'Jun 5, 2015 10:38:13 AM'

This results in the following Spring Boot RestController Exception:

{"timestamp":1433519208523,"status":400,"error":"Bad Request","exception":"org.springframework.http.converter.HttpMessageNotReadableException","message":"Could not read JSON: Can not construct instance of java.util.Date from String value 'Jun 5, 2015 10:38:13 AM': not a valid representation (error: Failed to parse Date value 'Jun 5, 2015 10:38:13 AM': Can not parse date \"Jun 5, 2015 10:38:13 AM\": not compatible with any of standard forms (\"yyyy-MM-dd'T'HH:mm:ss.SSSZ\", \"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\", \"EEE, dd MMM yyyy HH:mm:ss zzz\", \"yyyy-MM-dd\"))\n at [Source: java.io.PushbackInputStream@446e776b; line: 1, column: 123] (through reference chain: com.mydemandchain.site.models.Contact[\"created\"]); nested exception is com.fasterxml.jackson.databind.exc.InvalidFormatException: Can not construct instance of java.util.Date from String value 'Jun 5, 2015 10:38:13 AM': not a valid representation (error: Failed to parse Date value 'Jun 5, 2015 10:38:13 AM': Can not parse date \"Jun 5, 2015 10:38:13 AM\": no* Closing connection 0
[...]

PR #744, opened 2015-11-27
Specify serialization date format type. + Milliseconds and unix date formats.

IgorMiller:
* Refactored the DateTypeAdapter and DefaultDateTypeAdapter to support more date formats. DefaultDateTypeAdapter can be configured for a different serialization date format.

* GsonBuilder has 2 new methods:
 * setDateFormat(DateFormat,boolean useAsOuputFormat): the one can pass a DateFormat object to the builder. On Gson creation, this date format will be cloned to avoid any changes. If useAsOutputFormat is set to true, this format will be used to serialize dates.
 * setOutputDateFormatType(DateFormatType): the date format type can be set with an enum. See DateFormatType.

* Added more tests, + concurrency tests with dates

This should help to solve this issue: https://github.com/google/gson/issues/650

Check it and let me know what you think. Thanks!


gson/src/main/java/com/google/gson/DateFormatType.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.google.gson;
+
+public enum DateFormatType {
+	/**
+	 * Use custom {@link java.text.DateFormat} to format Date at serialization if
+	 * was set. Otherwise is same as {@code DateFormatType.DEFAULT}
+	 */
+	CUSTOM,
```
inder123: Not sure if we need this level of customization. We could instead just provide a few sample DateTypeAdapters that match up with these that users can register.
Why add more complexity to the API?

IgorMiller: Add few sample DateTypeAdapters was my first idea, but then I thought that maybe we would like the adapter to be able to parse multiple date formats, and also be able to specify the output date format. That's how I came up with an idea to create DateFormatters and simply specify to the adapter which of them to use. I agree that too much customization adds more complexity.

Maybe we could provide an EnumSet of DateFormatTypes for parsing, and one Enum DateFormatType for output. ex: 
`new DefaultDateTypeAdapter(EnumSet.of(ISO8601, MILLIS, EN_US), DateTypeFormat.MILLIS);`
That way the adapter will use ISO8601, MILLIS and EN_US formatters to parse, and MILLIS for output.

I can remove the CUSTOM and DEFAULT format types, so only standard date formats can be used for output.

In the case nothing was specified, the adapter will use the defaults: EN_US, LOCAL and ISO8601 to parse and EN_US for output, as it does now.

What do you think?

PR #751, opened 2015-11-28
Remove explicit step numbers; use autonumbering.

mbrukman:
Markdown supports automatic list numbering; using this approach makes it easier
to add or remove new items without having to update all the following entries.


swankjesse:
I think the current format reads better as plaintext.

swankjesse:
Inder?

inder123:
Thanks for checking with me, @swankjesse.
Auto-numbering is fine by me. Thanks for the change @mbrukman

Issue #722, opened 2015-10-27
Serialization of generic types


dtodor:
I have a weird issue with the serialisation of generic types. Consider the following program:

```
public class Service {
	private static class Characteristic<T> {
		private final int id = 1;
	}
	private static class StringCharacteristic extends Characteristic<String> {
		private final String format = "string";
	}

	private final List<Characteristic<?>> characteristics = Arrays
			.asList(new StringCharacteristic(), new Characteristic<Integer>());

	public static void main(String[] args) {
		final Gson gson = new Gson();
		final Service service = new Service();
		System.out.println("service: " + gson.toJson(service));
		System.out.println("characteristics only: " + gson.toJson(service.characteristics));
	}
}
```

The output (unexpectedly) is :
service: {"characteristics":[{"id":1},{"id":1}]}
characteristics only: [{"format":"string","id":1},{"id":1}]

i.e. the "format" attribute of the string characteristic is omitted when serializing the service object. However if I change the definition of the list to:
```
private final List<Characteristic> characteristics = ...;
```

everything works as expected:
service: {"characteristics":[{"format":"string","id":1},{"id":1}]}
characteristics only: [{"format":"string","id":1},{"id":1}]

Any thoughts?

grayjay:
I just encountered this issue when serializing guava's `Optional` type.

```java
import com.google.common.base.Optional;
import com.google.gson.Gson;

public final class GsonTest {
  static final Optional<String> OPTIONAL = Optional.of("string");

  public static final class GenericAttribute {
    final Optional<String> optional = OPTIONAL;
  }

  public static final class RawAttribute {
    final Optional optional = OPTIONAL;
  }

  public static void main(String[] args) {
    System.out.println(new Gson().toJson(OPTIONAL));
    System.out.println(new Gson().toJson(new GenericAttribute()));
    System.out.println(new Gson().toJson(new RawAttribute()));
  }
```

`Optional.of` creates an instance of `Optional`'s subclass `Present`, which has a `reference` attribute.  However, `reference` is missing in the second line of output:

    {"reference":"string"}
    {"optional":{}}
    {"optional":{"reference":"string"}}

Here is my understanding of the behavior:  Say `Y` is a subclass of `X`.  `O` is an object with an attribute of type `Z`.  `X` appears in `Z`, and `X` is parameterized.  When gson serializes `O`, it serializes the attributes of `X`, but not the attributes of `Y`.

inder123:
[...]

PR #755, opened 2015-12-06
Added failing test for #722

grayjay:
This commit adds three test cases created from the examples in https://github.com/google/gson/issues/722#issuecomment-162139672, as suggested in https://github.com/google/gson/issues/722#issuecomment-162250053.  The test `testParameterizedTypeSubclassInstanceVariableSerialization` currently fails with `expected:<{"optional":{"value":"string"}}> but was:<{"optional":{}}>`

The tests use a simplified version of guava's `Optional` type to avoid a dependency on guava.  I noticed that all three tests pass when they use the `Optional` type from Java 8, because Java 8 implements `Optional` as one final class instead of a class hierarchy.


inder123:
Closing the pull request as it is reproduced in https://github.com/google/gson/pull/757

PR #757, opened 2015-12-13
Using Raw class for serialization in case a parameterized type has no…

inder123:
… fields.


inder123:
This solution has performance overhead of counting of class fields.
It is also an edge case, so I am closing the pull request.

PR #761, opened 2015-12-27
Remove synthetic accessors from being generated.

JakeWharton:
Removes 14 methods being generated for trampolining to private members which brings the total to 1294 from 1308.
```
-rw-r--r--   1 jw  jw   176K Dec 27 01:39 after.dex
-rw-r--r--   1 jw  jw   221K Dec 27 01:38 after.jar
-rw-r--r--   1 jw  jw   177K Dec 27 01:39 before.dex
-rw-r--r--   1 jw  jw   223K Dec 27 01:39 before.jar
```

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
+++ b/gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
@@ -762,6 +762,9 @@ public final class LinkedHashTreeMap<K, V> extends AbstractMap<K, V> implements 
     Node<K, V> lastReturned = null;
     int expectedModCount = modCount;
 
+    LinkedTreeMapIterator() {
+    }
+
     public final boolean hasNext() {
       return next != header;
     }
--- a/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
+++ b/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
@@ -120,7 +120,7 @@ public enum FieldNamingPolicy implements FieldNamingStrategy {
    * Converts the field name that uses camel-case define word separation into
    * separate words that are separated by the provided {@code separatorString}.
    */
-  private static String separateCamelCase(String name, String separator) {
+  static String separateCamelCase(String name, String separator) {
     StringBuilder translation = new StringBuilder();
     for (int i = 0; i < name.length(); i++) {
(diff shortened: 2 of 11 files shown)


gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
```diff
@@ -762,6 +762,9 @@ void add(Node<K, V> node) {
     Node<K, V> lastReturned = null;
     int expectedModCount = modCount;
 
+    LinkedTreeMapIterator() {
```
inder123: are default constructors needed to remove synthetic methods?

inder123:
Excellent, I have been meaning to do the same after seeing a similar pull request in Moshi.
How do we add a comment explaining why we are making things package private?

gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
swankjesse: ... because this is a private class. Alternative would be to remove `private` on the `LinkedTreeMapIterator` class declaration.

swankjesse:
LGTM

gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
JakeWharton: Right. A default constructor always exists, it just matches the class visibility by default. So in this case a second constructor had to be generated for the other class to create an instance. The change just promotes the implicit default constructor to a visibility that can be called directly.

inder123: Ah. Good learning. Thanks

inder123:
How do you know these are the only culprits for synthetic methods? Did you use a tool to find this out?

JakeWharton:
I used IntelliJ's "Private member access between outer and inner classes" inspection.

You can also use the `dx` tool in the Android SDK to compile the .jar to .dex and then use `dexdump` to list all methods and match on ones which start with 'access$':
```
$ dexdump -f before.dex | grep 'access\$' | wc -l
      12

$ dexdump -f after.dex | grep 'access\$' | wc -l
       0
```

PR #765, opened 2016-01-05
Consolidation of Date/Time type adapters

inder123:


gson/src/main/java/com/google/gson/internal/bind/DateTimeTypeAdapters.java
```diff
@@ -0,0 +1,163 @@
+/*
+ * Copyright (C) 2011-2016 Google Inc.
```
swankjesse: Is this a new format?

swankjesse:
I’m very thumbs down on anything that changes behavior of date adapters.
 expected:<"Sep 1[1, 2011 10:55:03 P]M"> but was:<"Sep 1[2, 2011 5:55:03 A]M">

gson/src/main/java/com/google/gson/internal/bind/DateTimeTypeAdapters.java
inder123: what do you mean? 2011-2016 part?

I thought that was the standard way to include the year a file is written.

inder123:
@swankjesse  Yes, I have no intention of changing the behavior of date adapters.
Where are you seeing the new behavior?

swankjesse:
The test case failures suggest unintended behavior changes.

inder123:
@swankjesse  Can you help me reproduce the bug? The tests pass on my machine. I also tried rewriting the failing tests with setting timezone to somewhere else, but to no help.

gson/src/main/java/com/google/gson/internal/bind/DateTimeTypeAdapters.java
swankjesse: I think you should pick either 2011 or 2016? As is it looks like the copyright expires at the end of 2016!

swankjesse: Off-topic, since Google is no longer involved in any way with Gson, what do we do?

inder123: Ok, I am fine with 2011 (though I think the first public release of Gson was in 2009).

inder123: We could change the copyright line to Gson Authors

swankjesse: Yeah, either is fine with me.

Issue #764, opened 2016-01-02
Rare exception during POJO serialization


huguesb:
We're seeing what appears to be some sort of concurrency-related bug in our CI system. It occurs very rarely and only when multiple threads are serializing similar objects.

```
java.lang.IllegalStateException: null
	at com.google.gson.Gson$FutureTypeAdapter.write(Gson.java:890) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:68) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write(ReflectiveTypeAdapterFactory.java:89) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write(ReflectiveTypeAdapterFactory.java:195) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:68) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:96) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:60) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:68) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write(ReflectiveTypeAdapterFactory.java:89) ~[gson-2.2.4.jar:na]
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write(ReflectiveTypeAdapterFactory.java:195) ~[gson-2.2.4.jar:na]
	at com.google.gson.Gson.toJson(Gson.java:593) ~[gson-2.2.4.jar:na]
```

The hierarchy being serialized is as follows:

```
public class ChildrenList {
    public final @Nullable String parent;
    public final List<Folder> folders;
    public final List<File> files;
}

public class CommonMetadata {
    public final String id;
[...]

PR #767, opened 2016-01-07
Fix race for concurrent serialization of recursive type

pietern:
This race happens when two or more threads try to serialize the same recursive type for the first time at the same time.

The fix makes sure that TypeAdapter instances are not cached in the instance cache until all FutureTypeAdapters in a re-entrant call to getAdapter() have their delegates set.

Also see: #764


pietern:
This commit breaks a couple of tests, so please don't consider this final.

Pushed to have a starting point.

swankjesse:
My guess is that the best fix will be to change FutureTypeAdapter to look up the type adapter in Gson when it suffers a miss.

pietern:
@swankjesse I think that would be quite an elaborate fix since you would have to synchronize between the two threads. Add a latch so that the thread that misses can wait on the one that sets the delegate.

pietern:
Changed the fix to overwrite the future type adapters with the real one in the threadCalls map. This leads to the same result as before, where the newly built types are only externalized to other threads once the last one has been created, unlike the situation before this fix.

pietern:
The CLA is taken care of. See the vmware-google-contributors Google Group.

@swankjesse This is ready to merge, if this is the way you want to go for the fix (it's the most straightforward way to fix it in my opinion, by not leaking any adapters into the cache before the whole tree is traversed).

ttddyy:
Hi there,

I was looking #764 and came this PR.
Is this fix going to be merged, or already solved by other way?

Thanks,

itsgrimetime:
Ping. Any chance of this getting merged in?

yrodiere:
Hello,

The fix looks good, any particular reason to delay merging?

Marcono1234:
Thanks a lot for you work on this! This has now be fixed by #1832, but thank you nonetheless!

PR #771, opened 2016-01-17
Add setting for leniency on Gson instance.

JakeWharton:
Add a JsonReader factory (for parity with the JsonWriter one) which provides a configured instance using the Gson settings.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -101,6 +101,12 @@ import com.google.gson.stream.MalformedJsonException;
  */
 public final class Gson {
   static final boolean DEFAULT_JSON_NON_EXECUTABLE = false;
+  static final boolean DEFAULT_LENIENT = false;
+  static final boolean DEFAULT_PRETTY_PRINT = false;
+  static final boolean DEFAULT_ESCAPE_HTML = true;
+  static final boolean DEFAULT_SERIALIZE_NULLS = false;
+  static final boolean DEFAULT_COMPLEX_MAP_KEYS = false;
+  static final boolean DEFAULT_SPECIALIZE_FLOAT_VALUES = false;
 
   private static final String JSON_NON_EXECUTABLE_PREFIX = ")]}'\n";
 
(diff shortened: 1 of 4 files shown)


swankjesse:
This is good. It’s a bit unfortunate that by default `Gson` is actually lenient in some cases and not-lenient in others. Oh well.

swankjesse:
LGTM. @inder123 any thoughts?

JakeWharton:
> It’s a bit unfortunate that by default Gson is actually lenient in some cases and not-lenient in others. Oh well.

Agree! Too late to change. Just trying to make the best of that unfortunate situation.

gson/src/test/java/com/google/gson/functional/LeniencyTest.java
```diff
@@ -0,0 +1,39 @@
...
+  private Gson gson;
+
+  @Override
+  protected void setUp() throws Exception {
+    super.setUp();
+    gson = new GsonBuilder().setLenient().create();
+  }
+
+  public void testLenientFromJson() {
+    String json = gson.fromJson("\"Hello\"", String.class);
```
swankjesse: Wanna use another example? It’s likely we’ll support top-level primitives.

JakeWharton: Yep yep. I was planning on porting the RFC 7159 change to here today anyway.

JakeWharton: Updated to use comments and single-quote string in otherwise valid JSON.

swankjesse:
LGTM!

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -705,6 +714,15 @@ public JsonWriter newJsonWriter(Writer writer) throws IOException {
   }
 
   /**
+   * Returns a new JSON writer configured for the settings on this Gson instance.
+   */
+  public JsonReader newJsonReader(Reader reader) {
```
inder123: The method body seems fairly straightforward. Why can't the developer do this themselves?
We could just expose an isLenient() method instead.

JakeWharton: Parity with `newJsonWriter` and protection against future configuration options being added.

inder123: Ok, that seems reasonable.

PR #772, opened 2016-01-17
Add nullsafe option to JsonAdapter annotation

chalup:


chalup:
CLA signed.

swankjesse:
I think this is a weird API. Should we just do null safety automatically with `@JsonAdapter`?

chalup:
That would be reasonable, but strictly speaking it's a backward
incompatible change. Not sure what's the project's policy.

swankjesse:
@inder123  ?

swankjesse:
@JakeWharton ?

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -51,18 +51,24 @@ public JsonAdapterAnnotationTypeAdapterFactory(ConstructorConstructor constructo
   static TypeAdapter<?> getTypeAdapter(ConstructorConstructor constructorConstructor, Gson gson,
       TypeToken<?> fieldType, JsonAdapter annotation) {
     Class<?> value = annotation.value();
+    TypeAdapter<?> typeAdapter;
     if (TypeAdapter.class.isAssignableFrom(value)) {
-          Class<TypeAdapter<?>> typeAdapter = (Class<TypeAdapter<?>>) value;
-      return constructorConstructor.get(TypeToken.get(typeAdapter)).construct();
-    }
-    if (TypeAdapterFactory.class.isAssignableFrom(value)) {
+          Class<TypeAdapter<?>> typeAdapterClass = (Class<TypeAdapter<?>>) value;
```
JakeWharton: nit: wrong indent

gson/src/main/java/com/google/gson/annotations/JsonAdapter.java
```diff
@@ -94,4 +94,6 @@
   /** Either a {@link TypeAdapter} or {@link TypeAdapterFactory}. */
   Class<?> value();
 
+  /** Should the created {@link TypeAdapter} be null tolerant */
```
JakeWharton: nit: missing period.

JakeWharton:
> Should we just do null safety automatically with `@JsonAdapter`?

Yeah this is probably best since it's in line with the behavior without being annotated. Or at least for fields it is. Not sure about classes in that case.

chalup:
Nits fixed.

So we really have two questions: should nullsafe wrap be configurable and what should be the default behavior. I can't think of the non-contrived use case for not nullsafe adapter, so I'd be fine with making it non-configurable default.

Let me know what you think, and I'll update my PR.

swankjesse:
Not configurable by default is my preference.

swankjesse:
So please update your PR!

chalup:
Updated.

swankjesse:
LGTM

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -51,18 +51,20 @@ public JsonAdapterAnnotationTypeAdapterFactory(ConstructorConstructor constructo
...
           .construct()
           .create(gson, fieldType);
+    } else {
+      throw new IllegalArgumentException(
+          "@JsonAdapter value must be TypeAdapter or TypeAdapterFactory reference.");
     }
 
-    throw new IllegalArgumentException(
-        "@JsonAdapter value must be TypeAdapter or TypeAdapterFactory reference.");
+    return typeAdapter.nullSafe();
```
inder123: I think we should trust the developer to do the null check.

swankjesse: The catch is that with `@JsonAdapter` there’s no easy way to do `nullSafe()`

inder123: Correct. However, the developer can do the null check in code too.
Thinking again, I think it is ok for us to wrap the developer's adapter in nullSafe(). It is slightly inefficient if developer already has written the null check.

JsonSerializer/JsonDeserializer are actually similar to this as they don't require a null check.

PR #784, opened 2016-02-07
Added a unit test and removed unreachable code.  The while loop befor…

daveb1040:
…e it will only ever let index get as high as name.lenth - 1.
I wanted to add Lower_Camel_Case FieldNamingPolicy because of an issue I had the other day at work, but when I started to refactor the code, I saw a line that I couldn't get to with a unit test, so I figured I'd just add the unit test, and remove the line.
Also, I'm new to git and github, and wanted to try making a contribution to a great package.  The is my first pull request, so please bear with eh newbie if I'm not following some sort of protocol.
Thanks
David

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -54,7 +54,13 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>3.8.2</version>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>pl.pragmatists</groupId>
+      <artifactId>JUnitParams</artifactId>
+      <version>1.0.4</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
+++ b/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
@@ -56,6 +56,7 @@ public enum FieldNamingPolicy implements FieldNamingStrategy {
     }
   },
 
+ 
   /**
    * Using this naming policy with Gson will ensure that the first "letter" of the Java
    * field name is capitalized when serialized to its JSON form and the words will be
@@ -147,10 +148,6 @@ public enum FieldNamingPolicy implements FieldNamingStrategy {
 
       fieldNameBuilder.append(firstCharacter);
       firstCharacter = name.charAt(++index);
-    }
-
-    if (index == name.length()) {
-      return fieldNameBuilder.toString();
     }
 
     if (!Character.isUpperCase(firstCharacter)) {
--- a/gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
+++ b/gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+
+/**
+ * Unit test for the {@link FieldNamingPolicy} class
+ * 
+ * @author David Betancourt
+ *
+ */
+@RunWith(JUnitParamsRunner.class)
+public class FieldNamingPolicyTest {
+
+  private Object[] getUpperCaseFirstLetterParams() {
+    return new Object[] { getUpperCaseFirstLetterParamArgs("someFieldName", "SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("SomeFieldName", "SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("_someFieldName", "_SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("_SomeFieldName", "_SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("_999someFieldName", "_999SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("_999SomeFieldName", "_999SomeFieldName"),
+        getUpperCaseFirstLetterParamArgs("x", "X"),
+        getUpperCaseFirstLetterParamArgs("X", "X"),
+        getUpperCaseFirstLetterParamArgs("555", "555"),
+        getUpperCaseFirstLetterParamArgs("_", "_") };
+
+  }
+
+  private Object[] getUpperCaseFirstLetterParamArgs(String fieldName, String expectedFieldName) {
+    return new Object[] { fieldName, expectedFieldName };
+  }
+
+  @Test
+  @Parameters(method = "getUpperCaseFirstLetterParams")
+  public void testUpperCaseFieldLetter(String fieldName, String expectedFieldName) {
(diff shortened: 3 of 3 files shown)


daveb1040:
I signed it!
Thanks!
David Betancourt
GitHub id: daved1040

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,78 @@
+package com.google.gson;
```
swankjesse: nit: copy & paste google copyright

swankjesse: or Gson Authors copyright

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Test;
+import org.junit.runners.Parameterized.Parameters;
+
+import static org.junit.Assert.assertEquals;
+
+public class FieldNamingPolicyTest {
+	
+
+	@Test
+	public void testUpperCaseFirstLetterNonUpperCaseLetter() {
```
swankjesse: nit: please format this to match the rest of the project

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Test
+	public void testUpperCaseBlank() {
+		verifyUpperCaseFirstLetter(" ", " ");
+	}
+
+	/**
+	 * Verifies the upperCaseFirstLetter is working as expected
+	 * @param fieldName
```
swankjesse: you can delete this doc

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	@Test
+	public void testUpperCaseFirstJustUnderscore() {
+		verifyUpperCaseFirstLetter("_", "_");
+	}
+
+	@Test
+	public void testUpperCaseBlank() {
+		verifyUpperCaseFirstLetter(" ", " ");
```
swankjesse: you can probably collapse all of these tests into a single test method. It’s extremely unlikely we’ll ever want to run these tests independently, or that they’ll fail independently. That way if/when we ever want to add tests for more policies, we don’t need to write thousands of lines of code.

gson/src/main/java/com/google/gson/FieldNamingPolicy.java
```diff
@@ -149,10 +150,6 @@ static String upperCaseFirstLetter(String name) {
       firstCharacter = name.charAt(++index);
     }
 
-    if (index == name.length()) {
```
swankjesse: I’m anxious that this will break compatibility with code already deployed. At this point that’s much more important than actual correctness, which is both depressing and true.

daveb1040: I wanted to leverage this behavior for my LOWER_CAMEL_CASE fieldNamingPolicy, so before refactoring, I surrounded it with unit tests.  (I'm a unit test junkie)  I did an emma code coverage test, and realized it was impossible for me to put anything into any of my unit tests.  (The pit mutation test also flagged it as unreadable with my existing permutations) 

That said, understandable on being concerned with such a highly used class, so if you want, I'll put it back in there.  (Although, I was hoping my unit tests could mitigate the risk).  Let me know how you want to proceed, and I'll commit those changes.

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
daveb1040: These will be added in the next commit

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
daveb1040: Just installed Mars....  auto formatting wasn't set.  I pulled the formatted in from lib.  that will be in, in the next commit.

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
daveb1040: I'll make parameterized tests for these, and commit those.  The reason I don't stack them all in one is because when I do put in some code which breaks the tests, I find it desirable to have the summary of what I broke, where as, if they're all in one method, once 1 assertion fails, the test executes.  So, did I do something that broke 10 of 15 scenarios, or just 1.  I'll get this in, in the next commit

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
daveb1040: Will do

daveb1040:
I have addressed the peer review comments.  The only thing pending is whether or not to drop the if statement.  Thanks!
David

gson/src/main/java/com/google/gson/FieldNamingPolicy.java
```diff
@@ -20,10 +20,11 @@
 import java.util.Locale;
 
 /**
- * An enumeration that defines a few standard naming conventions for JSON field names.
- * This enumeration should be used in conjunction with {@link com.google.gson.GsonBuilder}
```
swankjesse: why reformat all this code? Please revert

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,60 @@
...
+import junitparams.Parameters;
+
+/**
+ * Unit test for the {@link FieldNamingPolicy} class
+ * 
+ * @author David Betancourt
+ *
+ */
+@RunWith(JUnitParamsRunner.class)
+public class FieldNamingPolicyTest {
```
swankjesse: Dig.

gson/src/test/java/com/google/gson/FieldNamingPolicyTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private Object[] getUpperCaseFirstLetterParams() {
+    return new Object[] { getUpperCaseFirstLetterParamArgs("someFieldName", "SomeFieldName"),
```
swankjesse: might be more legible if every call to `getUpperCaseFirstLetterParamArgs` was on its own line?

daveb1040: I imported the eclipse code formatter that was in the project, and had it autoformat on save...  I assumed it was required.  I'll turn it off, put that on it's own line, and commit tonight.

gson/src/main/java/com/google/gson/FieldNamingPolicy.java
daveb1040: I imported the eclipse code formatter that was in the project, and had it autoformat on save...  I assumed it was required.  I'll turn it off and revert this.  Thanks

daveb1040:
OK...  the should do it.  Thanks
David

daveb1040:
Are there any other issues I should address?
Thanks!
David

gson/pom.xml
```diff
@@ -54,7 +54,13 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>3.8.2</version>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>pl.pragmatists</groupId>
```
swankjesse: Any way you can implement this without adding another 3rd party dependency? We’ve worked hard to limit our dependencies. Gson is often embedded in other software & adding dependencies is a huge burden downstream.

gson/src/main/java/com/google/gson/FieldNamingPolicy.java
swankjesse: This file still has tons of unnecessary changes. Please really revert.

swankjesse: (the GitHub diff on this file should be nothing)

gson/pom.xml
swankjesse: I think you can accomplish something very similar to this with these classes:

```
import org.junit.runners.Parameterized;
import org.junit.runners.Parameterized.Parameter;
```

swankjesse:
A few more comments. Plus you need to rebase and squash. See http://rebaseandsqua.sh for tips.

eamonnmcmanus:
The most recent comments, from February 2016, went unaddressed so I think we can consider this obsolete.

PR #787, opened 2016-02-11
Simplify Gson poms.

swankjesse:
A lot of the configuration in these poms has been taken over by improvements
to the Maven release plugin.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -1,246 +1,36 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>com.google.code.gson</groupId>
-  <artifactId>gson</artifactId>
-  <packaging>jar</packaging>
-  <version>2.6-SNAPSHOT</version>
-  <inceptionYear>2008</inceptionYear>
-  <name>Gson</name>
+
   <parent>
-    <groupId>org.sonatype.oss</groupId>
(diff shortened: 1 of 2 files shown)


JakeWharton:
LGTM. I would do a `mvn clean javadoc:jar sources:jar` to verify those tasks do what you expect and generate the output we expect.

gson/pom.xml
```diff
@@ -1,246 +1,36 @@
...
-            <goals>
-              <goal>jar</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <archive>
-            <addMavenDescriptor>false</addMavenDescriptor>
-            <manifestEntries>
-              <Bundle-RequiredExecutionEnvironment>J2SE-1.5, JavaSE-1.6, JavaSE-1.7, JavaSE-1.8</Bundle-RequiredExecutionEnvironment>
```
inder123: This was added specifically for OSGI. How will that requirement be met?
Disclaimer: I don't fully understand what the requirement was.

pom.xml
```diff
@@ -1,249 +1,97 @@
...
   <url>https://github.com/google/gson</url>
-  <description>Google Gson library</description>
+
+  <modules>
+    <module>gson</module>
+  </modules>
+
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <java.version>1.7</java.version>
```
inder123: Why 1.7? I thought we went up to 1.6?

swankjesse: Oh wow, that’s a big bug. Will fix.

gson/pom.xml
swankjesse: Yeah, I don’t think this is the current way that OSGi works anyways. (There’s a special bundle target now).

pom.xml
swankjesse: https://github.com/google/gson/pull/789

PR #789, opened 2016-02-12
Use the Java 7 language but limit to Java 6 APIs.

swankjesse:
This way we get nice features like ARM blocks and diamonds, but still work
on Android.


swankjesse:
https://github.com/google/gson/pull/790

PR #791, opened 2016-02-14
Fixed code style to remove explicit casting op.

haskellcamargo:
· Separated binary operator

This is a very, very, **very** small change, but that unnecessary explicit `(double)` casting with joined expressions on binary operator was affecting my OCD.


haskellcamargo:
I signed it!

Issue #372, opened 2015-03-19
GsonBuilder.setLenient(false)


GoogleCodeExporter:
```
The gson JsonParser is final and it's #parse(JsonReader json) does this:
json.setLenient(true); 

In my unit tests I validate the output of my json rest api and I would like to 
have a parser with strict parsing. Creating my own StrictJsonReader also is not 
possible unless I put it in the com.google.gson package.

Maybe you could add a setter or constructor param for lenient or strict parsing?
```

Original issue reported on code.google.com by `franci...@gmail.com` on 26 Oct 2011 at 2:11

sgbrown:
I think Gson.lenient flag probably does more harm than good in the current state.  The comments on the GsonBuilder.setLenient() method say 

> By default, Gson is strict and only accepts JSON as specified by <a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627</a>.

The comments on Gson.setLenient() also references the JsonReader.setLenient(boolean) method which has a great explanation of what should and shouldn't be considered when using the lenient flag.

The unfortunate fact here is that due to the Gson.fromJson(JsonReader, Type) method always setting the lenient flag to true when parsing, almost none of what is commented in GsonBuilder about the default behavior of Gson is true.  The only affect that setting the lenient flag on Gson will do is to allow comments to be at the end of your json buffer when calling Gson.fromJson(Reader, Type) since the assertFullConsumption(Object, JsonReader) method will be called from here and only checks for data at the end of the buffer having not been consumed (e.g. comments at the end).

Consider, for example, the test in com.google.gson.functional.LeniencyTest
```
    @Override
   protected void setUp() throws Exception {
     super.setUp();
     gson = new GsonBuilder().setLenient().create();
   }
 
   public void testLenientFromJson() {
     List<String> json = gson.fromJson(""
         + "[ # One!\n"
         + "  'Hi' #Element!\n"
         + "] # Array!", new TypeToken<List<String>>() {}.getType());
     assertEquals(singletonList("Hi"), json);
   }
```
If you were to remove the comment at the end of the String ("# Array!"), then it would not make any difference at all whether or not the lenient flag had been set.  The following would also pass:
```
     List<String> json = new GsonBuilder().create().fromJson(""
         + "[ # One!\n"
         + "  'Hi' #Element!\n"
         + "]", new TypeToken<List<String>>() {}.getType());
     assertEquals(singletonList("Hi"), json);
```

[...]

PR #795, opened 2016-02-23
Gson.fromJson methods now respect the lenient flag

sgbrown:
Gson.fromJson(JsonReader, Type) would previously override the lenient flag in the JsonReader to always be true while being used from here.  This would cause all instances of Gson to act as though the lenient flag is set to true regardless of the actual state in the Gson object.  I have corrected this to respect the lenient flag.

also updated JsonReader.nextDouble() method to deserialize infinite values, following the comments
in GsonBuilder.serializeSpecialFloatingPointValues() method which claims:
"Gson always accepts these special values during deserialization"

These fixes should also address issue #372

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -869,8 +869,6 @@ public final class Gson {
   @SuppressWarnings("unchecked")
   public <T> T fromJson(JsonReader reader, Type typeOfT) throws JsonIOException, JsonSyntaxException {
     boolean isEmpty = true;
-    boolean oldLenient = reader.isLenient();
-    reader.setLenient(true);
     try {
       reader.peek();
       isEmpty = false;
@@ -892,8 +890,6 @@ public final class Gson {
     } catch (IOException e) {
       // TODO(inder): Figure out whether it is indeed right to rethrow this as JsonSyntaxException
       throw new JsonSyntaxException(e);
-    } finally {
-      reader.setLenient(oldLenient);
     }
   }
 
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -888,7 +888,7 @@ public class JsonReader implements Closeable {
    *
    * @throws IllegalStateException if the next token is not a literal value.
    * @throws NumberFormatException if the next literal value cannot be parsed
-   *     as a double, or is non-finite.
+   *     as a double
    */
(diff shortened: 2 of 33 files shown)


swankjesse:
Unfortunately, as the tests suggest, this is massively incompatible. If you'd like this behavior in your own code, avoid toJson/fromJson and use a TypeAdapter instead.

sgbrown:
I think Gson.lenient flag probably does more harm than good in the current state.  The comments on the GsonBuilder.setLenient() method say 

> By default, Gson is strict and only accepts JSON as specified by <a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627</a>.

The comments on Gson.setLenient() also references the JsonReader.setLenient(boolean) method which has a great explanation of what should and shouldn't be considered when using the lenient flag.

The unfortunate fact here is that due to the Gson.fromJson(JsonReader, Type) method always setting the lenient flag to true when parsing, almost none of what is commented in GsonBuilder about the default behavior of Gson is true.  The only affect that setting the lenient flag on Gson will do is to allow comments to be at the end of your json buffer when calling Gson.fromJson(Reader, Type) since the assertFullConsumption(Object, JsonReader) method will be called from here and only checks for data at the end of the buffer having not been consumed (e.g. comments at the end).

Consider, for example, the test in com.google.gson.functional.LeniencyTest
```
    @Override
   protected void setUp() throws Exception {
     super.setUp();
     gson = new GsonBuilder().setLenient().create();
   }
 
     List<String> json = gson.fromJson(""
         + "[ # One!\n"
         + "  'Hi' #Element!\n"
         + "] # Array!", new TypeToken<List<String>>() {}.getType());
     assertEquals(singletonList("Hi"), json);
   }
```
If you were to remove the comment at the end of the String ("# Array!"), then it would not make any difference at all whether or not the lenient flag had been set.  The following would also pass:
```
     List<String> json = new GsonBuilder().create().fromJson(""
         + "[ # One!\n"
         + "  'Hi' #Element!\n"
         + "]", new TypeToken<List<String>>() {}.getType());
     assertEquals(singletonList("Hi"), json);
```

It seems counterintuitive that with the lenient flag unset, comments in the middle of the JSON data would be ignored but comments at the end would cause a failure when the Gson parser is supposed to be "strict".

If it is considered undesirable to change the default behavior (to the behavior that is documented) by actually respecting a default lenient flag being set to false (i.e. default to strict), then perhaps the correct answer is to respect the lenient flag, default Gson to have the lenient flag set to true, and update the javadocs to reflect that Gson by default is lenient.

Do you have any thoughts on this alternate approach?

PR #797, opened 2016-02-23
Restore OSGi manifest headers

### LINKED ISSUE #796: OSGi manifest headers removed?
The OSGi headers were removed as part of https://github.com/google/gson/commit/2016e9553eb946a09a9e70d83d31f4e8da675805

These should be restored

seanbright:
Commit 2016e95 removed the necessary MANIFEST.MF headers for Gson to
resolve in an OSGi environment. This patch restores them.

Fixes #796

### DIFF
--- a/gson/bnd.bnd
+++ b/gson/bnd.bnd
@@ -1,0 +1,14 @@
+Bundle-SymbolicName: com.google.gson
+Bundle-Name: ${project.name}
+Bundle-Description: ${project.description}
+Bundle-Vendor: Google Gson Project
+Bundle-ContactAddress: ${project.parent.url}
+Bundle-RequiredExecutionEnvironment: J2SE-1.5, JavaSE-1.6, JavaSE-1.7, JavaSE-1.8
+
+-removeheaders: Private-Package
+
+-exportcontents:\
+    com.google.gson,\
+    com.google.gson.annotations,\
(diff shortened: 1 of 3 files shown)


gson/bnd.bnd
```diff
@@ -0,0 +1,16 @@
...
+Bundle-Description: ${project.description}
+Bundle-Vendor: Google Gson Project
+Bundle-ContactAddress: https://github.com/google/gson
+Bundle-RequiredExecutionEnvironment: J2SE-1.5, JavaSE-1.6, JavaSE-1.7, JavaSE-1.8
+
+Import-Package: !com.google.gson.*
+Export-Package: \
+    com.google.gson, \
+    com.google.gson.annotations, \
+    com.google.gson.internal, \
```
swankjesse: wanna omit the internal packages? These aren't intended to be exported

swankjesse:
Thanks!

seanbright:
The 2.5 bundle exported the `com.google.gson.internal.*` packages, so for backwards compatibility I was doing the same here. I've updated the PR to exclude them.

swankjesse:
Awesome, thanks!

swankjesse:
(I like this approach much better than the OSGi plugins I’ve encountered also)

PR #800, opened 2016-02-25
Fixed a regression in Gson 2.6 where Gson caused NPE if the TypeAdapt…

inder123:
…erFactory.create() returned null.

### DIFF
--- a/gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
+++ b/gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
@@ -85,9 +85,6 @@ public class JsonAdapterNullSafeTest extends TestCase {
           }
           private JsonElement parseString(JsonParser parser, String json, String prevJson)
               throws IOException { // called recursively
-            if (json == null || json.trim().isEmpty()) {
-              return null;
-            }
             JsonElement root = parser.parse(json);
             if (root instanceof JsonPrimitive) {
               prevJson = json;
--- a/gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
+++ b/gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
@@ -15,54 +15,37 @@
  */
 package com.google.gson.regression;
 
-import java.io.IOException;
-import java.util.Objects;
-
 import com.google.gson.Gson;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonParser;
-import com.google.gson.JsonPrimitive;
-import com.google.gson.JsonSyntaxException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
 import com.google.gson.annotations.JsonAdapter;
 import com.google.gson.reflect.TypeToken;
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonWriter;
 
 import junit.framework.TestCase;
 
 public class JsonAdapterNullSafeTest extends TestCase {
-  // The recursiveCall in {@link Device.JsonAdapterFactory} is the source of this bug
-  // because it returns a null type adapter.
-
   private final Gson gson = new Gson();
 
   public void testNullSafeBugSerialize() throws Exception {
-    Device device = new Device("ec57803e", 2);
+    Device device = new Device("ec57803e");
     gson.toJson(device);
   }
 
   public void testNullSafeBugDeserialize() throws Exception {
-    String json = "\"{\\\"id\\\":\\\"ec57803e2\\\",\\\"category\\\":2}\"";
-    Device device = gson.fromJson(json, Device.class);
+    Device device = gson.fromJson("{'id':'ec57803e2'}", Device.class);
     assertEquals("ec57803e2", device.id);
-    assertEquals(2, device.category);
   }
 
   @JsonAdapter(Device.JsonAdapterFactory.class)
   private static final class Device {
     String id;
-    int category;
-    Device(String id, int category) {
+    Device(String id) {
       this.id = id;
-      this.category = category;
     }
 
-    /**
-     * Write the value as a String, not JSON.
-     */
     static final class JsonAdapterFactory implements TypeAdapterFactory {
+      // The recursiveCall in {@link Device.JsonAdapterFactory} is the source of this bug
+      // because we use it to return a null type adapter on a recursive call.
       private static final ThreadLocal<Boolean> recursiveCall = new ThreadLocal<Boolean>();
 
       @SuppressWarnings({"unchecked", "rawtypes"})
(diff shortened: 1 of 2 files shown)


inder123:
I would like to cut Gson 2.6.2 with this change. Let me know if you have any concerns @swankjesse and @JakeWharton

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,186 @@
...
+import com.google.gson.annotations.JsonAdapter;
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+import junit.framework.TestCase;
+
+public class JsonAdapterNullSafeTest extends TestCase {
```
swankjesse: Any way you could write a simpler test for this? There’s a lot of stuff here

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+      deviceJson = "{'id':'device1','controlData':'\\\"{}\\\"'}";
+      device = gson.fromJson(deviceJson, Device.class);
+
+      try {
+          deviceJson = "{'id':'device1','controlData':'a'}";
+          device = gson.fromJson(deviceJson, Device.class);
+          assertNotNull(device.controlData);
+          fail();
+      } catch (JsonSyntaxException expected) {}
```
swankjesse: nit: weird syntax, including 4 space indents in this file

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,186 @@
...
+      this.controlData = controlData;
+    }
+  }
+
+  @JsonAdapter(ControlData.JsonAdapterFactory.class)
+  private static final class ControlData {
+    String id;
+    int category;
+    boolean alwaysOn;
+    @SerializedName("codeset_id") int codesetId;
```
swankjesse: Seems like a distraction from the test

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,186 @@
...
+      this.alwaysOn = alwaysOn;
+      this.codesetId = codesetId;
+    }
+
+    /**
+     * DeviceControlData is received as String in JSON instead of proper JSON.
+     * So, we need to write a special type adapter.
+     */
+    static final class JsonAdapterFactory extends StringifiedJsonAdapterFactory<ControlData> {
+      private static final ThreadLocal<Boolean> recursiveCall = new ThreadLocal<Boolean>();
```
swankjesse: Is a thread local really necessary to trigger this?

swankjesse:
Fix looks reasonable but the test case looks like you just pulled it from another codebase. Could you write a targeted test case instead?

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
inder123: ok

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
inder123: ok, will simplify by merging Device with ControlData.

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
inder123: Yes. As I debugged the problem, I realized that the constructorConstructor.get() resulted in a recursive call to the JsonAdapterFactory.  Since we don't want to be in an infinite loop, I had used a ThreadLocal to avoid that problem in Gson 2.5.
Now, that ThreadLocal returning null was causing this regression.

inder123:
Done, take another look.

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Device(String id, int category) {
+      this.id = id;
+      this.category = category;
+    }
+
+    /**
+     * Write the value as a String, not JSON.
+     */
+    static final class JsonAdapterFactory implements TypeAdapterFactory {
+      private static final ThreadLocal<Boolean> recursiveCall = new ThreadLocal<Boolean>();
```
swankjesse: Does this really need a recursive call to reproduce? Weird!

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
```diff
@@ -0,0 +1,106 @@
...
+          }
+          @Override public Device read(JsonReader in) throws IOException {
+            String json = in.nextString();
+            JsonParser parser = new JsonParser();
+            JsonElement root = parseString(parser, json, null);
+            return root == null ? null : delegate.fromJsonTree(root);
+          }
+          private JsonElement parseString(JsonParser parser, String json, String prevJson)
+              throws IOException { // called recursively
+            if (json == null || json.trim().isEmpty()) {
```
swankjesse: Do these cases happen in tests?

swankjesse:
I still don’t really understand this test. Would end users use ThreadLocals ?

inder123:
@swankjesse  This is production code written by me for a specific use-case.
I was interacting with a REST service that, unfortunately, stored JSON as s String in a field. To support that with JSON, I wrote this adapter class. I didn't start with ThreadLocals, but the recursive invocation from JsonAdapterFactory forced my hand.

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
inder123: Yes, it does because the thread-local is used to prevent an infinite recursion.

gson/src/test/java/com/google/gson/regression/JsonAdapterNullSafeTest.java
inder123: You mean, the null and isEmpty checks? No. I have removed it.

swankjesse:
Rather than starting with the test you used to trigger this in your application, can you build a test case from scratch specifically to reproduce this bug?

swankjesse:
(I don't imagine that parsing JSON as a string is necessary for the fix made here.)

swankjesse:
Erm, you shouldn't need to convert a string from JSON within another type adapter.

inder123:
Ok, revised the test to eliminate string to JSON conversion.

swankjesse:
Yeah, that's more obvious. Thanks!

Issue #604, opened 2015-03-19
JsonReader.nextInt() doesent work if p == PEEKED_UNQUOTED


GoogleCodeExporter:
```
What steps will reproduce the problem?
This unit test describes the problem:

    @Test public void test() {
        Map<Integer, Integer> expected = new HashMap<Integer, Integer>() {{ put(0, 1); }};
        Map<Integer, Integer> actual = new Gson().fromJson("{0:1}", new TypeToken<Map<Integer, Integer>>() {}.getType());
        assertEquals(expected, actual);
    }


What version of the product are you using? On what operating system?
2.3 (version 2.1 works OK)
```

Original issue reported on code.google.com by `feathoro...@gmail.com` on 28 Oct 2014 at 4:32

Marcono1234:
Looks like this was fixed by #809.

Issue #524, opened 2015-03-19
Parsing a Map field with unquoted Integer keys fails in Gson 2.2.3 and above


GoogleCodeExporter:
```
What steps will reproduce the problem?
(see Java code below)
1. Create a class with a field of type: Map<Integer, someClass>
2. Create a JSON string for that class where the key values of the map field 
are unquoted
3. Parse the above string with the Gson.fromJson method

What is the expected output? What do you see instead?
An exception is thrown:
java.lang.IllegalStateException: Expected an int but was STRING at line 1 
column 15
In the code below this exception is thrown only when parsing the string 
noQuotes, any only when using Gson versions 2.2.3 and 2.2.4.

What version of the product are you using? On what operating system?
The issue appears in Gson 2.2.3 and Gson 2.2.4, but not in Gson 2.2.2.
I don't know if this was a bug in 2.2.2 which was fixed, or a bug in the new 
versions. Anyway this is a backward-incompatibility and I didn't find any 
reference to it in the release notes of version 2.2.3.
OS: Win8 64bit

Please provide any additional information below.
Unquoted integers are parsed just fine when they are intended for primitive int 
fields, reference Integer fields, lists of Integers or Map values. As far as I 
can tell this issue appears only for Integers which are Map keys.

---- Java code ---

import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import java.util.Map;

public class GsonTest {

    public static class TestClass {
        public Map<Integer, Integer> mapField = null;
    }


    public static void main(String[] args) throws Exception {
        Gson gson = new GsonBuilder().create();

        String quotes = "{\"mapField\": {\"5\": \"6\", \"7\": \"8\"}}";
        String keyQuotes = "{\"mapField\": {\"5\": 6, \"7\": 8}}";
        String noQuotes = "{\"mapField\": {5: 6, 7: 8}}";

        TestClass res = null;
        res = gson.fromJson(quotes, TestClass.class);
        res = gson.fromJson(keyQuotes, TestClass.class);
        res = gson.fromJson(noQuotes, TestClass.class);

    }
}

```

Original issue reported on code.google.com by `yonmost` on 18 Jun 2013 at 6:47

Marcono1234:
Looks like this was fixed by #809.

PR #809, opened 2016-03-12
allow unquoted long and integer keys

sgbrown:
Fixes a bug in JsonReader.nextInt() and JsonReader.nextLong() whereby PEEKED_UNQUOTED is not handled.  This bug caused failure to deserialize maps with Long or Integer keys when the key is unquoted.

This fixes issues #604 and #524


swankjesse:
Nice find!

PR #813, opened 2016-03-24
Improving InterceptorFactory to allow usage without Intercept annotation.

thetoster:
I wan't to use Intercept functionality in my project, however it requires usage of annotation. In fact this is not something which I want to put into my code, this will break my abstraction. So I decided to extend current solution to allow developer "inject" binding without need to annotate classes.


thetoster:
I signed it!

thetoster:
I signed it!

PR #814, opened 2016-03-24
Improving InterceptorFactory to allow usage without Intercept annotation.

thetoster:
I wanted to use intercept extras, however it looks like I need to add annotation to my classes. From my project perspective this was unacceptable step, since it breaks my abstraction. So I added possibility to "inject" post process binding by configuration method rather then by annotation.


inder123:
@joel-leitch You had an idea of generalized interceptors. Do you want to think through how to revise this pull request or write a new one?

PR #821, opened 2016-03-31
Refactoring: Keep track of registered JsonAdapters in ThreadLocals.

inder123:


gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -40,33 +58,69 @@ public JsonAdapterAnnotationTypeAdapterFactory(ConstructorConstructor constructo
...
-      Class<TypeAdapterFactory> typeAdapterFactory = (Class<TypeAdapterFactory>) value;
-      typeAdapter = constructorConstructor.get(TypeToken.get(typeAdapterFactory))
-          .construct()
-          .create(gson, fieldType);
+      TypeAdapterFactory factory = getTypeAdapterFactory(rawType, annotation, constructorConstructor);
+      typeAdapter = factory == null ? null : factory.create(gson, type);
     } else {
       throw new IllegalArgumentException(
-          "@JsonAdapter value must be TypeAdapter or TypeAdapterFactory reference.");
+          "@JsonAdapter value must be TypeAdapter, TypeAdapterFactory, JsonSerializer or JsonDeserializer reference.");
```
swankjesse: Is the JsonSerializer / JsonDeserializer change true?

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -40,33 +58,69 @@ public JsonAdapterAnnotationTypeAdapterFactory(ConstructorConstructor constructo
...
+      return null;
+    }
+    return getTypeAdapterFactory(annotatedClass, annotation, constructorConstructor);
+  }
+
+  @SuppressWarnings({ "unchecked", "rawtypes" }) // Casts guarded by conditionals.
+  TypeAdapter<?> getTypeAdapter(ConstructorConstructor constructorConstructor, Gson gson,
+      TypeToken<?> type, JsonAdapter annotation) {
     Class<?> value = annotation.value();
+    Class<?> rawType = type.getRawType();
```
swankjesse: It looks like this used to work for parameterized types and now it doesn't. Is `@JsonAdapter Foo<Bar>` supported?

swankjesse:
This is a big complicated change with no motivating test case.

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
inder123: I am not sure what you mean. JsonAdapter always took a class as value.
```
@JsonAdapter(MyTypeAdapter.class)
Foo foo;
```
You can also use JsonAdapter on a parameterized field if you wanted. But that shouldn't get affected here.

inder123:
@swankjesse @JakeWharton I have revised this pull request to support JsonSerializer and JsonDeserializer in JsonAdapter. Please take a look.

gson/src/test/java/com/google/gson/functional/JsonAdapterSerializerDeserializerTest.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (C) 2008 Google Inc.
```
swankjesse: Nit: 2008

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
swankjesse: What happens if I have a `List<Foo>` and a `List<Bar>`? Do I get cache pollution?

swankjesse: (Because List.class is the cache key for both)

inder123: Aha, Good point. This was indeed a bug. See the revised commit that fixes this problem and has a regression test.

gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
inder123: Yes, it is now implemented.

swankjesse:
What’s the motivation? ThreadLocals make me nervous.

inder123:
Threadlocal for TypeAdapter is not really needed here, it just provide caching for reusing a JsonAdapter.

It is needed for a TypeAdapterFactory to support skipPast.

What's your concern with ThreadLocals.

swankjesse:
I don’t see why the ThreadLocals are necessary to solve this problem. Which is the test case that needs it?

inder123:
JsonAdapterNullSafeTest.testNullSafeBugSerialize will fail if we remove the TypeAdapterFactory ThreadLocal.

inder123:
@swankjesse  Please review (note the test that fails if we remove ThreadLocal for the factory. I can remove the other threadlocal, but it may have a performance benefit).

swankjesse:
I looked at this and it’s a bit difficult to trace. It’s unfortunate that the test that fails seems particularly contrived. A simpler test would just fail with a StackOverflowError rather than using a thread local to try to defeat the recursion.

What should the precedence rules be with `@JsonAdapter` ? That is, if I say `@JsonAdapter` on a type that also has a user-configured adapter, does that adapter get used with `getDelegateAdapter()` ? Or does it always delegate to the reflective implementation?

One thing that’s potentially surprising is that `getDelegateAdapter()` can yield different results for different `Gson` instances. That’s probably okay, but it’s awkward to spread the configuration in both the callsite code (at field declaration) and also in the GsonBuilder.

I think it’s pretty straightforward to avoid either of the new ThreadLocals by making a tweak to `getDelegateAdapter()` like this:
```
    if (!factories.contains(skipPast)) {
      skipPast = jsonAdapterFactory;
    }
```
But that impacts the precedence of the delegate adapter. It also causes potentially surprising behavior in the non-`@JsonAdapter` case because an unknown `skipPast` is treated as a sentinel value.

inder123:
@swankjesse 
precedence rules. I think with GsonBuilder setting should always take precedence since that allows a user to override a pre-configured behavior. Obviously, JsonAdapter on a field is NOT overridable in GsonBuilder, but the user has the option of attaching a TypeAdapter to the enclosing class.

I think you are saying that you don't want me to use your solution of avoid threadlocals. Anything else that needs to be done for this pull request?

inder123:
BTW, those precedence rules are asserted in JsonAdapterAnnotationOnClassesTest

inder123:
@swankjesse  Once this is merged, I would like to cut a new Gson release.

swankjesse:
I think avoiding ThreadLocals would be great. At least for me they’re difficult to reason about.

inder123:
Ok. Removed one of the ThreadLocals. Can't remove the other ThreadLocal without breaking (admittedly convoluted, but from real production code) JsonAdapterNullSafeTest.testNullSafeBugSerialize test.

Since this line of code is only invoked when JsonAdapter annotation is present, I think it is OK to leave it as is. If you are worried about performance impact, I can make instantiation of ThreadLocal conditional.

inder123:
@swankjesse I don't see any further enhancements in this PR. Please complete this review, or provide alternate code that I can use the ThreadLocal with.

swankjesse:
Here’s a branch that shows one approach to solving this problem without a ThreadLocal.
https://github.com/google/gson/compare/master...jwilson.0601.get_delegate_adapter?expand=1#files_bucket

inder123:
This approach looks good to me. Let me issue a pull request to you from
this branch.

On Wed, Jun 1, 2016 at 9:34 PM, Jesse Wilson <notifications@github.com>
wrote:

> Here’s a branch that shows one approach to solving this problem without a
> ThreadLocal.
>
> https://github.com/google/gson/compare/master...jwilson.0601.get_delegate_adapter?expand=1#files_bucket
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/google/gson/pull/821#issuecomment-223193208>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABCwypOrLi3LQj_033Prly-pht_DOueAks5qHl1XgaJpZM4H8Wzm>
> .
>



-- 
http://www.singhinderjeet.com/

inder123:
As #873 has been merged and is a duplicate of this, closing this.

PR #827, opened 2016-04-11
Add the interface, GsonInterface, to the Gson object to enable cleaner unit testing

artnaseef:
Since the Gson object is final, it is not possible to mock it, making unit testing more difficult for users of the class.

Adding an interface to the class eliminates the problem as users of the Gson object can operate on the interface instead, and thereby enable mocks to be injected for testing purposes.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -100,7 +100,7 @@ import com.google.gson.stream.MalformedJsonException;
  * @author Joel Leitch
  * @author Jesse Wilson
  */
-public final class Gson {
+public final class Gson implements GsonInterface {
   static final boolean DEFAULT_JSON_NON_EXECUTABLE = false;
   static final boolean DEFAULT_LENIENT = false;
   static final boolean DEFAULT_PRETTY_PRINT = false;
@@ -372,6 +372,7 @@ public final class Gson {
    * @throws IllegalArgumentException if this GSON cannot serialize and
    *     deserialize {@code type}.
    */
+  @Override
   @SuppressWarnings("unchecked")
   public <T> TypeAdapter<T> getAdapter(TypeToken<T> type) {
     TypeAdapter<?> cached = typeTokenCache.get(type == null ? NULL_KEY_SURROGATE : type);
@@ -465,6 +466,7 @@ public final class Gson {
    *
    * @since 2.2
    */
+  @Override
   public <T> TypeAdapter<T> getDelegateAdapter(TypeAdapterFactory skipPast, TypeToken<T> type) {
     boolean skipPastFound = false;
     // Skip past if and only if the specified factory is present in the factories.
@@ -494,6 +496,7 @@ public final class Gson {
    * @throws IllegalArgumentException if this GSON cannot serialize and
    *     deserialize {@code type}.
    */
+  @Override
   public <T> TypeAdapter<T> getAdapter(Class<T> type) {
     return getAdapter(TypeToken.get(type));
   }
@@ -511,6 +514,7 @@ public final class Gson {
    * @return Json representation of {@code src}.
    * @since 1.4
    */
+  @Override
   public JsonElement toJsonTree(Object src) {
     if (src == null) {
       return JsonNull.INSTANCE;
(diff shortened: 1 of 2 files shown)


inder123:
Thanks for taking the time to write a pull request. Such concrete feedback is much more helpful than just a feature request.
Unfortunately, however, we are not yet ready to expand Gson API. Can you use a mocking library instead?

artnaseef:
Mocking libraries cannot mock final classes, thus the benefit and purpose of this pull request.

Note that the changes in this pull request do not change the API itself.  It simple adds an interface with the same API (method signatures are exactly the same) so that users can choose to make accesses through the interface rather than directly against the class.  Doing so makes mocking possible.

Like this:

    // ORIGINAL 
    public final class MyUtil {
        public int add(int x, int y) {
            return x + y;
        }
    }

=====

    // UPDATED
    public final class MyUtil implements MyUtilInterface {
        @Override
        public int add(int x, int y) {
            return x + y;
        }
    }

    public interface MyUtilInterface {
        int add(int x, int y);
    }

Would you please reconsider?  Without this, unit testing code that use Gson become significantly harder.

JakeWharton:
Why are you mocking Gson? Seems like a huge anti-pattern.

artnaseef:
Hey Jake - mocking is a unit testing technique that allows one to limit all
test interactions to the class-under-test.  In order for that to work
properly, all other objects with which that class operates need to be
mocked.  Without mocking, those classes are being tested too, which leads
to more complicated tests that are more brittle (i.e. break more easily),
and making it far harder to track down bugs as bugs manifest themselves
across multiple unit tests instead of only showing up in the one for the
class that introduced the bug.

I can clarify and provide examples if that helps.

There's a lot of good information on mocking out there.

If you can explain the concern and why it feels like an anti-pattern, I'll
address the same.

Art

JakeWharton:
I know what mocking is. Why are you wanting to mock the Gson object instead
of just letting it serialize?

artnaseef:
Jake - it sounds like you and I have a different approach to unit testing.

In the approach I use and advocate, letting Gson "serialize" (or perform any of the other operations for which we us it) during unit testing of our own classes is not appropriate.  Perhaps some day we'll have the opportunity to grab a beer and discuss our various approaches.

JakeWharton:
In that case you should be completely abstracting your serialization
library in an application-defined type. It's not the library's
responsibility to provide that abstraction.

artnaseef:
That is true - the library doesn't have to do so.  I was hoping it would,
though, as this is a very simple and minimal update.  Making a thin layer
in my own code increases the amount of code to solve the problem (compared
with this solution), and the complexity of the same, by a fair amount.
And, on every product on which I work, I need to repeat the same exercise
(i.e. the complexity is distributed).

Honestly, I was at first discouraged to see that all the Gson classes are
marked as final, but when I read that the reason is to prevent extending of
classes by users of the library - i.e. that the classes are "final" in
order to create the contract "we don't support extending this class," I
realized that's a great approach.  So then I thought about how I could make
the code easy-to-mock and this hit me:

   1. The new interface does not affect anybody that does not use it.
   2. All that is needed from the library is to add the interface and
   maintain the list of methods.
   3. Folks that want "mockable" use can then easily transition to the use
   of "GsonInterface" in place of "Gson" where needed.
   4. As the API of Gson changes across versions, the GsonInterface would
   track along quite simply as well.

This increased ease-of-use will increase my adoption of the library as
well.  So, win/win for very little effort.  Writing a layer in-between my
application and third-party libraries that is nothing more than a
pass-through to the third-party library is not a practice that I want to
adopt.  Creating code that's easy to unit test already leads to too much
boiler-plate code.  Honestly, I keep wishing the JVM offered a "test mode"
in which interactions with all classes could easily be mocked - especially
built-in classes interacting with system resources (such as files and
sockets).

Art

JakeWharton:
Adding this interface actively hurts the public API. What happens when we need to add a new method to `Gson` in the future? Adding methods to interfaces is not source or binary compatible so some hypothetical (and illogical) user which created another implementation of this interface would have their application break.

You also haven't justified the need for mocking. `Gson` is deterministic and side-effect free type so there's no reason tests should avoid calling it. If anything, providing a fake or mock implementation is likely to cause _more_ problems than it solves. You may inadvertently create behavior differences between what your tests code against and what actually happens at runtime. This is usually a sign of overly-white box tests that should actually be seeking more black box behavior. You're testing _how_ the code interacts with the library instead of testing _what_ the code does (e.g., what outputs certain inputs create).

I would have been :-1: on this change even if it wasn't closed.

Issue #817, opened 2016-03-28
JNI Error in Android


abhijith-gururaj:
I am getting a 'JNI DETECTED AN ERROR IN APPLICATION" while using gson 2.5.jar in Android. Can anyone tell me how to solve this? Appreciate it.
Here is the stack trace: 

`03-28 14:01:15.445: E/art(28683): JNI DETECTED ERROR IN APPLICATION: can't make objects of type java.util.concurrent.locks.Lock: 0x70dd7858
03-28 14:01:15.445: E/art(28683):     in call to AllocObject
03-28 14:01:15.445: E/art(28683):     from java.lang.Object sun.misc.Unsafe.allocateInstance(java.lang.Class)
03-28 14:01:15.445: E/art(28683): "main" prio=5 tid=1 Runnable
03-28 14:01:15.445: E/art(28683):   | group="main" sCount=0 dsCount=0 obj=0x7617aef8 self=0x7fa917c400
03-28 14:01:15.445: E/art(28683):   | sysTid=28683 nice=0 cgrp=default sched=0/0 handle=0x7face5d2c0
03-28 14:01:15.445: E/art(28683):   | state=R schedstat=( 657346072 17671229 611 ) utm=52 stm=13 core=1 HZ=100
03-28 14:01:15.445: E/art(28683):   | stack=0x7fe28ac000-0x7fe28ae000 stackSize=8MB
03-28 14:01:15.445: E/art(28683):   | held mutexes= "mutator lock"(shared held)
03-28 14:01:15.445: E/art(28683):   at sun.misc.Unsafe.allocateInstance(Native method)
03-28 14:01:15.445: E/art(28683):   at java.lang.reflect.Method.invoke!(Native method)
03-28 14:01:15.445: E/art(28683):   at com.google.gson.internal.UnsafeAllocator$1.newInstance(UnsafeAllocator.java:48)
03-28 14:01:15.445: E/art(28683):   at com.google.gson.internal.ConstructorConstructor$14.construct(ConstructorConstructor.java:223)
03-28 14:01:15.445: E/art(28683):   at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:207)
03-28 14:01:15.445: E/art(28683):   at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:117)
[...]

PR #829, opened 2016-04-14
Check if class can be instantiated based on class modifiers. Fix for #817

ionspin:
If class can't be instantiated throw an unsupported operation exception. On different android devices attempt to instantiate interface or abstract class would crash the VM sometimes with only a segmentation fault message and no other useful logs (Issue #817). Checking for interface or abstract class modifiers before trying to allocate instance and throwing an exception will help debugging.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java
+++ b/gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java
@@ -20,6 +20,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectStreamClass;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 /**
  * Do sneaky things to allocate objects without invoking their constructors.
@@ -45,6 +46,7 @@ public abstract class UnsafeAllocator {
         @Override
         @SuppressWarnings("unchecked")
         public <T> T newInstance(Class<T> c) throws Exception {
+          assertInstantiable(c);
           return (T) allocateInstance.invoke(unsafe, c);
         }
       };
@@ -68,6 +70,7 @@ public abstract class UnsafeAllocator {
         @Override
         @SuppressWarnings("unchecked")
         public <T> T newInstance(Class<T> c) throws Exception {
+          assertInstantiable(c);
           return (T) newInstance.invoke(null, c, constructorId);
         }
       };
@@ -87,6 +90,7 @@ public abstract class UnsafeAllocator {
         @Override
         @SuppressWarnings("unchecked")
         public <T> T newInstance(Class<T> c) throws Exception {
+          assertInstantiable(c);
           return (T) newInstance.invoke(null, c, Object.class);
         }
       };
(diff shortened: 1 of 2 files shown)


gson/src/test/java/com/google/gson/internal/UnsafeAllocatorInstatiationTest.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2016 Ugljesa Jovanovic
```
swankjesse: Sorry, we won’t accept pull requests unless you assign copyright to either Google or Gson Authors. (We don’t want to have to get your permission later if we need to change the license of the whole project.)

gson/src/test/java/com/google/gson/internal/UnsafeAllocatorInstatiationTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+package com.google.gson.internal;
+
+import junit.framework.TestCase;
+
+/**
+ * Test unsafe allocator instantiation
+ */
+public class UnsafeAllocatorInstatiationTest extends TestCase {
```
swankjesse: please reformat this so that it’s consistent with the rest of the project

gson/src/test/java/com/google/gson/internal/UnsafeAllocatorInstatiationTest.java
```diff
@@ -0,0 +1,79 @@
...
+        } catch (Exception e) {
+            assertEquals(e.getClass(), UnsupportedOperationException.class);
+        }
+    }
+
+    /**
+     *  Ensure that the {@link java.lang.UnsupportedOperationException} is thrown when  trying
+     *  to instantiate an abstract class
+     */
+    public void testAbstractClassInstantiation(){
```
swankjesse: nit: here and elsewhere, add a space before `{`

swankjesse:
Looks good but I’ve got some comments

ionspin:
@swankjesse Should I squash this into one commit?

gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java
```diff
@@ -101,4 +105,18 @@ public static UnsafeAllocator create() {
       }
     };
   }
+
+  /**
+   * Check if the class can be instantiated by unsafe allocator. If the instance has interface or abstract modifiers
+   * throw an {@link java.lang.UnsupportedOperationException}
+   * @param c instance of the class to be checked
+   */
+  private static void assertInstantiable(Class c) {
```
swankjesse: supernit: should be `Class<?>` because `Class` is a raw type.

gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java
```diff
@@ -101,4 +105,18 @@ public static UnsafeAllocator create() {
...
+    if (Modifier.isInterface(c.getModifiers())) {
```
swankjesse: nit: can you extract a variable and keep `c.getModifiers()` ? It’s not necessarily a fast method on all VMs

swankjesse:
LGTM!

swankjesse:
Yes, please squash into one commit.

ionspin:
@swankjesse Done! Thanks for the comments!

swankjesse:
Thanks!

PR #830, opened 2016-04-15
fixed throwing RuntimeException instead of JsonIOException

gavlyukovskiy:
Javadoc says that JsonIOException will be thrown if any problem happened, but actually I'm getting RuntimeException only in that method.


gavlyukovskiy:
I signed it!

Issue #831, opened 2016-04-19
JsonAdapter annotation ignored for primitive fields


mintern:
`ReflectiveTypeAdapterFactory` correctly detects the `@JsonAdapter` annotation and registers the custom adapter ([source](https://github.com/google/gson/blob/6f6af8050799bec5321d2c06cd3230daadbb6535/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L133)), but its `write` method wraps that type adapter in a `TypeAdapterRuntimeTypeWrapper` ([source](https://github.com/google/gson/blob/6f6af8050799bec5321d2c06cd3230daadbb6535/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L111)), which overrides the adapter with the default Gson adapter ([source](https://github.com/google/gson/blob/6f6af8050799bec5321d2c06cd3230daadbb6535/gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java#L65)).

Here's a test that demonstrates the behavior:

```diff
diff --git a/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java b/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest
index 4c745ec..8cae980 100644
--- a/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
+++ b/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
@@ -220,4 +220,43 @@ public final class JsonAdapterAnnotationOnFieldsTest extends TestCase {
       this.part = part;
     }
   }
+
+  public void testPrimitiveFieldAnnotationTakesPrecedenceOverDefault() {
+    Gson gson = new Gson();
+    String json = gson.toJson(new GadgetWithPrimitivePart(42));
+    assertEquals("{\"part\":\"42\"}", json);
+    GadgetWithPrimitivePart gadget = gson.fromJson(json, GadgetWithPrimitivePart.class);
+    assertEquals(42, gadget.part);
+  }
+
+  private static final class GadgetWithPrimitivePart {
+    @JsonAdapter(LongToStringTypeAdapterFactory.class)
+    final long part;
+
+    private GadgetWithPrimitivePart(long part) {
+      this.part = part;
+    }
+  }
+
[...]

PR #832, opened 2016-04-20
Ensuring that JsonAdapter annotation works correctly for primitive fi…

inder123:
…elds.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
@@ -15,14 +15,15 @@
  */
 package com.google.gson.internal.bind;
 
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
+
 import com.google.gson.Gson;
 import com.google.gson.TypeAdapter;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
-import java.io.IOException;
-import java.lang.reflect.Type;
-import java.lang.reflect.TypeVariable;
 
 final class TypeAdapterRuntimeTypeWrapper<T> extends TypeAdapter<T> {
   private final Gson context;
--- a/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
+++ b/gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
@@ -63,7 +63,6 @@ public final class JsonAdapterAnnotationOnFieldsTest extends TestCase {
         @Override public void write(JsonWriter out, Part part) throws IOException {
           throw new AssertionError();
         }
-
         @Override public Part read(JsonReader in) throws IOException {
           throw new AssertionError();
         }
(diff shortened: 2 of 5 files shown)


inder123:
Fix for https://github.com/google/gson/issues/831

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
```diff
@@ -51,7 +52,7 @@ public void write(JsonWriter out, T value) throws IOException {
 
     TypeAdapter chosen = delegate;
     Type runtimeType = getRuntimeTypeIfMoreSpecific(type, value);
-    if (runtimeType != type) {
+    if (runtimeType != type && !Primitives.areEquivalentTypes(runtimeType, type)) {
```
swankjesse: This doesn’t seem like the right location for the fix. I’d expect it to be nearby the `@JsonAdapter` handling code.

inder123: There is no other way that I can think of. The problem is that ReflectiveTypeAdapterFactory (line 109) field.get(value) returns Long even though the field is of type long. This results in the runtime type being incorrectly characterized for all primitive types.  Hence the TypeAdapterRuntimeTypeWrapper needs to be smarter when dealing with primitive wrapper types.

swankjesse: The problem is triggered by primitive vs. wrapper types, but that isn't the root of the problem. The root of the problem is that Gson currently lets the runtime type take precedence over the `@JsonAdapter` annotation. This is weird!

For example, consider a JSON adapter that calls `toString()` on the object and writes that string to the `JsonWriter`:

```
  private static class ToStringAdapter extends TypeAdapter<Object> {
    @Override public Object read(JsonReader in) throws IOException {
      throw new UnsupportedOperationException();
    }

    @Override public void write(JsonWriter out, Object value) throws IOException {
      out.value(value.toString());
    }
  }
```

We can request this adapter like so:

```
  private static final class Box {
    @JsonAdapter(ToStringAdapter.class)
    Object o;
  }
```

But unfortunately due to a mistake in how Gson precedence works, the `ToStringAdapter` is ignored whenever `o`’s type isn’t `Object.class`. For example, this test fails because the `ToStringAdapter` is never used.

```
  public void test() {
    Gson gson = new Gson();

    Box box = new Box();
    box.o = Arrays.asList(5, 6, 7);

    String json = gson.toJson(box);
    assertEquals("{\"o\":\"" + box.o.toString() + "\"}", json);
  }
```

Fixing the precedence is [possible](https://github.com/google/gson/commit/d1770c9f77cd7761add943d81b773b2df4594f3b), but dangerous and potentially incompatible with currently deployed code.

inder123: I see your point. I think we should fix the precedence for JsonAdapter as that is the expected behavior by clients. I think JsonAdapter is new enough to not have serious incompatibility issues (as we discovered with this primitive-related behavior).
I will update the pull request.

JakeWharton: We've fixed precedence before: 7f8f490fdc95fb087e735dc582a2af07cee46078

swankjesse: Sounds good!

inder123:
@swankjesse I have revised the code per feedback. Now if JsonAdapter annotation is present, we ignore  the run-time type specific adapter.

Also, rebased and squashed. Please review.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
swankjesse: Is this change still necessary?

gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
```diff
@@ -220,4 +220,44 @@ private GadgetWithOptionalPart(Part part) {
       this.part = part;
     }
   }
+
+  /** Regression test contributed through https://github.com/google/gson/issues/831 */
+  public void testPrimitiveFieldAnnotationTakesPrecedenceOverDefault() {
```
swankjesse: Worth having a test for the non-primitive case?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
```diff
@@ -51,7 +52,7 @@ public void write(JsonWriter out, T value) throws IOException {
 
     TypeAdapter chosen = delegate;
     Type runtimeType = getRuntimeTypeIfMoreSpecific(type, value);
-    if (runtimeType != type) {
+    if (runtimeType != type && !Primitives.areEquivalentTypes(runtimeType, type)) {
       TypeAdapter runtimeTypeAdapter = context.getAdapter(TypeToken.get(runtimeType));
```
swankjesse: If it is, in theory it should just happen in here

inder123: Good point. Will make that change.

gson/src/test/java/com/google/gson/functional/JsonAdapterAnnotationOnFieldsTest.java
inder123: Yes, was being lazy.. :)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
inder123: On second thoughts, I don't really see an advantage of moving the code here. Can you explain what you mean by providing a snippet?

swankjesse: What I intended to say was that this line will do the primitive/wrapper work automatically so the check above is unnecessary.

inder123:
@swankjesse  Ok. I have removed that check. Please review.

swankjesse:
Thumbs up

PR #833, opened 2016-04-20
Add annotation that enables null serialization per field

markuspfeiffer:
This adds the ability to specify whether to serialize a `null` value on a per-field basis, using the new `@Serialize` annotation.

If the `@Serialize` annotation is not specified, or its value is set to `Inclusion.DEFAULT` the behavior of Gson remains unchanged. A class member's value will be included in the JSON if it has a value other than `null`. If Gson has been configured to always include null values the member's value will be included even if it's `null`.

By adding `@Serialize(Inclusion.ALWAYS)` to a class member, it will always be included in the JSON even if its value is `null`. Basically this annotation makes it possible to enable inclusion of nulls on a per-field basis. If Gson has been configured to always include null values adding the annotation in this form does nothing.

By adding `@Serialize(Inclusion.NON_NULL)` to a class member, it will always be excluded from the JSON if its value is `null`. Normally this matches Gson's default behavior, unless Gson has been configured to include null values. In this case the annotation allows to exclude nulls on a per-field basis.

**Be aware** that this does not change any other of Gson's rules. If a member is excluded for any other reason, for example if it has a `transient` modifier, this annotation will not change this. Also if Gson has been configured to exclude fields that do not have an `@Expose` annotation you will still need to add @Expose to class members that should be serialized. In other words the `@Serialize` annotation only works on members that are serialized under normal circumstances and only changes the behavior of Gson in regards to `null`.

**Examples**
```
public class Main {

    public static class Example {

        @Serialize(Inclusion.ALWAYS)
        public String field1 = null;

		 @Serialize(Inclusion.NON_NULL)
        public String field2 = null;

		 //@Serialize(Inclusion.DEFAULT)
		 public String field3 = null;

        public String field4 = "Field 4";
    }

    public static void main(final String[] args) {

        final Gson gson1 = new GsonBuilder().create();
        final Gson gson2 = new GsonBuilder().serializeNulls().create();

        final Example example1 = new Example();
        
        final String json1 = gson1.toJson(example1);
        System.out.println(json1);
        
        final String json2 = gson2.toJson(example1);
        System.out.println(json2);
    }
}
```

Without support for `@Serialize` the output of Gson would be:
```
// Output for gson1 (don't serialize nulls)
{ "field4": "Field 4" }

// Output for gson2 (serialize nulls)
{ "field1": null, "field2": null, "field3": null, "field4": "Field 4" }
```

**With** support for `@Serialize` its output instead becomes:
```
// Output for gson1 (don't serialize nulls)
{ "field1": null, "field4": "Field 4" }

// Output for gson2 (serialize nulls)
{ "field1": null, "field3": null, "field4": "Field 4" }
```

### DIFF
--- a/gson/src/main/java/com/google/gson/annotations/Serialize.java
+++ b/gson/src/main/java/com/google/gson/annotations/Serialize.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2016 Markus Pfeiffer.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * An annotation that defines additional conditions to apply when serializing a member that is
+ * {@code null}.
+ * <p>
+ * This annotation has no effect if a member is excluded from serialization. A member may be
+ * excluded from serialization if it has a {@code transient} modifier or if it does not have an
+ * {@link Expose} annotation and {@link com.google.gson.Gson} has been built with a
+ * {@link com.google.gson.GsonBuilder} and the
+ * {@link com.google.gson.GsonBuilder#excludeFieldsWithoutExposeAnnotation()} method has been
+ * invoked.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface Serialize {
+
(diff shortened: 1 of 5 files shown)


markuspfeiffer:
I signed it!

JakeWharton:
This is a very heavyweight solution to a problem I have trouble understanding. When would you need such granular control vs. always emitting or always omitting `null` literals?

inder123:
Agree with Jake. Can there be a much lighter-weight solution to this problem. Something that extends the existing annotations (such as SerializedName), for example.

inder123:
BTW, appreciate the detailed tests and the pull request.

markuspfeiffer:
> This is a very heavyweight solution to a problem I have trouble understanding. When would you need such granular control vs. always emitting or always omitting `null` literals?

Example: You want to omit `null` in general to keep the size of your JSON to the minimum. However there are some fields where `null` and _not-present_ are treated differently.

When you send a value to the server, the server will update its field to that value. When you send `null` to the server, the server will clear its field. When you do not send a value, the server will leave its field as it is.

So in general you want to omit `null` for reasons of size, but for some models you require `null` values to be able to reset fields.

Instead of a separate annotation this could probably be added to either `SerializedName` or `Expose`. Using `SerializedName` feels just wrong to me, because its name suggests no functionality of such kind. `Expose` on the other hand already controls serialize/deserialize, so adding a more fine-grained control for serialize makes sense to me (e.g. serialize=[true|false], deserialize=[true|false], serializeNulls=[default, always, never]).

inder123:
@markuspfeiffer  Agree. This is also needed for partial updates.

PR #836, opened 2016-04-22
Add boxed boolean value() overload.

JakeWharton:
When calling value() with a Boolean, overload resolution would choose value(boolean) which would throw an NPE on null. The other boxed types are all numbers which would resolve to value(Number) and behave correctly.

Proof this happens: https://github.com/bugsnag/bugsnag-android/pull/42


swankjesse:
Sigh.

PR #837, opened 2016-04-23
 Documentation Error

zzz40500:
String anotherStr = gson.fromJson("[\"abc\"]", String.class); => String[] anotherStr = gson.fromJson("[\"abc\"]", String[].class);


JakeWharton:
Thanks!

PR #838, opened 2016-04-25
Adds getters for config fields.

rharter:
This adds simple getters for certain config fields that would be helpful in custom `TypeAdapter`s to deal with situations like this: https://github.com/rharter/auto-value-gson/issues/18

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -127,6 +127,8 @@ public final class Gson {
   private final List<TypeAdapterFactory> factories;
   private final ConstructorConstructor constructorConstructor;
 
+  private final Excluder excluder;
+  private final FieldNamingStrategy fieldNamingStrategy;
   private final boolean serializeNulls;
   private final boolean htmlSafe;
   private final boolean generateNonExecutableJson;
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -244,11 +248,27 @@ public Gson() {
...
+        constructorConstructor, fieldNamingStrategy, excluder));
 
     this.factories = Collections.unmodifiableList(factories);
   }
 
+  public Excluder getExcluder() {
+    return excluder;
+  }
+
+  public FieldNamingStrategy getFieldNamingStrategy() {
```
JakeWharton: I would drop the 'get' prefix on these two.

gson/src/test/java/com/google/gson/GsonTest.java
```diff
@@ -0,0 +1,40 @@
+package com.google.gson;
```
JakeWharton: Needs copyright header

rharter: added.

swankjesse:
Merged manually. I updated your copyright header!

swankjesse:
Merged commit is a851569ab916b5f9466949ed106d2abff1d1650c

rharter:
Hah, I just copied that from another file and didn't pay attention.  I didn't actually write this in 2008 :)

PR #839, opened 2016-04-27
Optimize list allocation for every reflective field.

JakeWharton:
The common cases (no @SerializedName and a single-value @SerializedName) now use a specialized one-element collection. Cases with alternate names use a much more space efficient ArrayList which is exactly sized.

This is also a performance win for adapter initialization since an index-based loop was used on the returned type for which LinkedList is not suited for.


inder123:
👍

PR #844, opened 2016-05-04
Deserialization aware

manishgo:
This feature will enable the user to provide hook in Java object to perform additional steps after that object is deserialized from JSON.

### DIFF
--- a/gson/src/main/java/com/google/gson/aware/DeserializationAware.java
+++ b/gson/src/main/java/com/google/gson/aware/DeserializationAware.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


JakeWharton:
You can do this already with a `TypeAdapter.Factory` and your own callback interface. Why should it be added to Gson?

manishgo:
Wouldn't it be easier to just implement this interface and write the required code in the class' method itself rather than writing it in another class (TypeAdapter) in which case user has to deal with TypeAdapter syntax. Also user needs to write additional logic for performing default deserialization in the adapter and then calling interface method which can be hidden in GSON library.

JakeWharton:
You can have the exact same behavior as this PR by checking for an
interface and calling a method on the type though.

manishgo:
Agreed. But doesn't it make sense to provide a convenient feature to the user so that he can just implement this interface and thats it!

JakeWharton:
Not unless there's a clear value proposition desired by large chunks of
users and even then it needs a well-designed and thought out API that we
are comfortable maintaining. It only takes 2 seconds to send a PR but we
have to live with the ramifications and compatibility concerns for years.

PR #846, opened 2016-05-05
adding @DateFormat annotation to accept date formats

jainsahab:
this field annotation will allow passing date formats to GSON in order to parse/format
date strings while deserialization/serialization. It will work for `java.util.Date` and `java.sql.Date` type fields.

  
    public class Trip {
        @DateFormat("yyyy-MM-dd")
        private Date startDate;

        @DateFormat("yyyy/MM/dd")
        private Date endDate;

        public Trip(Date startDate, Date endDate) {
          this.startDate = startDate;
          this.endDate = endDate;
        }
    }

that's how we can pass date formats for different fields to Gson.

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -24,7 +24,6 @@ import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Locale;
-import java.util.TimeZone;
 
 import com.google.gson.internal.bind.util.ISO8601Utils;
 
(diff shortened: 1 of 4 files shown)


inder123:
Thanks for the concrete proposal.
There is no need for a new annotation. You can write a TypeAdapter with this format, and use the JsonAdapter annotation. Overall, we set a very high bar to expand the Gson API. Simple short APIs have a smaller conceptual weight and appreciated by user.

jainsahab:
Apologies for the late reply.
I thought this change could be of generic use like in our case an external vendor develops the API and those guys don't really care about the date format in the JSON response. Different API's have different date formats. So we thought instead of writing different TypeAdapter for different date formats, we can just write a GSON annotation that can tell the expect date format for a field.

sacot41:
I need this feature. I use Gson not because it's quick library or for his "simple short APIs"; I use Gson because it save development time and it make code really clean. Writing a TypeAdapter each time a backend dev use a different date format take time, add a lot of boilerplate code and it's very frustrating. 

Please, consider this feature.

PR #850, opened 2016-05-10
made nullSafe wrapper of JsonAdapter optional

testcenter:
as there could be the need of handling `null` values within a custom adapter (which is added via the `@JsonAdapter` annotation), the `nullSafe` wrapper is made optional now.

as the annotation has a default value, it is no problem to specify the adapter as before
`@JsonAdapter(MyCustomAdapter.class)`

to remove the nullSafe wrapper, call
`@JsonAdapter(value = MyCustomAdapter.class, nullSafe = false);


testcenter:
I signed it!

PR #851, opened 2016-05-10
made nullSafe wrapper of JsonAdapter optional

testcenter:
as there could be the need of handling `null` values within a custom adapter (which is added via the `@JsonAdapter` annotation), the `nullSafe` wrapper is made optional now.

as the annotation has a default value, it is no problem to specify the adapter as before
`@JsonAdapter(MyCustomAdapter.class)`

to remove the nullSafe wrapper, call
`@JsonAdapter(value = MyCustomAdapter.class, nullSafe = false);

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Change Log
 ==========
 
+## Version 2.7
+_2016-06-14_  [GitHub Diff](https://github.com/google/gson/compare/gson-parent-2.6.2...gson-parent-2.7)
+ * Added support for JsonSerializer/JsonDeserializer in @JsonAdapter annotation
+ * Exposing Gson properties excluder(), fieldNamingStrategy(), serializeNulls(), htmlSafe()
+ * Added JsonObject.size() method
+ * Added JsonWriter.value(Boolean value) method
+ * Using ArrayDeque, ConcurrentHashMap, and other JDK 1.6 features
+ * Better error reporting
+ * Plenty of other bug fixes
+
 ## Version 2.6.2
 _2016-02-26_  [GitHub Diff](https://github.com/google/gson/compare/gson-parent-2.6.1...gson-parent-2.6.2)
  * Fixed an NPE bug with @JsonAdapter annotation


inder123:
Doesn't seem like an important enough usecase.

testcenter:
@inder123 an important use case would be to be compatible with the JSON-API:

JSON API specifies setting a value like this:

* if any value is set it will be set / updated on the server
* if any value is `null` it will be deleted on the server
* if any value is not serialized (excluded from the request), it will not be updated on the server (stays as it is)

so how should i realize following scenario:
i have a very complex object, which means the object contains some other objects and a lot of primitive values. all in all about 40-50 elements.
one of these primitive values is a `Long` value: `deletedAt`.

on the one hand, i want to deserialize the timestamp `deletedAt`, because i have to delete a database entry on the phone if this timestamp is set.

on the other hand i am only allowed to serialize valid values (so the server entry gets deleted).
if i serialize null, the server throws an error (because the server tries to undelete the value).
-> so if i want to update this entry, i am not allowed to serialize this value

in this scenario i would add a special adapter to the given `deletedAt` value via the annotation `@JsonAdapter`.
this adapter does simply not serialize the value if set to `null`.
but this will not work, if the `nullSafe` wrapper is wrapping my adapter.

in this case, i would have to create an adapter for the whole complex object, handling 40-50 primitives, just because of one value...
the way better solution would be, to simply set a `@JsonAdapter` to this one specific value.

**by the way**: this was the behaviour before you upgraded to version 2.6. which means passing `null` values to a `@JsonAdapter` adapter worked on versions <= 2.5

swankjesse:
Can you do this without @JsonAdapter? It's a relatively recent addition and it should be possible to accomplish your goals without it.

testcenter:
@swankjesse it would be possible, but i would have to write a custom serializer for several complex objects which all have about 20-50 members.... which is really a pain in the a... :smile: 
other than that, our structure would not be easily extendable.

i already saw, that this addition is relatively new (if i can recall correctly, the change was made this february).
this new behaviour breaks our system since the upgrade to the version 2.6 (which contains this change).
the solution i propose in this pull request does not affect the new behaviour at all.
but if the developer does set the additional `nullSafe` flag in the annotation the "old" behaviour is recreated.

swankjesse:
I defer to inder on this, @JsonAdapter is his API.

testcenter:
@inder123 any updates on this after the conversation with @swankjesse ?

inder123:
@testcenter ok, let me think it through more.

inder123:
@swankjesse I am fine with this change. Never quite liked the fact that we wrapped the adapter with nullSafe() to guard against sloppy code. 
Any thoughts?

swankjesse:
Works for me.

inder123:
If you want this merged, please update pull request to enable merging.
Also do http://rebaseandsqua.sh

inder123:
Can you do http://rebaseandsqua.sh ?

inder123:
The pull request has changes from other branches. You probably forgot to pull master correctly.
Can you clone gson git repository directly, create a branch and make changes in it directly? I can then help you do rebase and squash if needed. Thanks.

testcenter:
@inder123 i am sorry for the inconvenience. it seems as i am too stupid for github (rebase and squash)... :P

i accidentally closed this pull request by force pushing the current state of the upstream/master branch to the master of my fork.
so i created a new pull request:

see pull request #881

Issue #171, opened 2015-03-19
Custom deserializer not called for explicit null value


GoogleCodeExporter:
```
When a json object contains an explicit null for a field, the custom
deserializer for that field type is never called, rather the field is
simply set to null. This makes it impossible to map json nulls to some
non-null value in a consistent way. It also makes it impossible to have
guaranteed non-null fields in objects which are deserialized from json.

The handling of explicit nulls and missing values is also inconsistent: in
the case of an explicit null, the Java object's field is set to null. In
the case of a missing value, the field is left alone (as initialized by the
no-arg constructor).

The behavior for missing values is reasonable, but the behavior for
explicit nulls is very difficult to work around. Explicit null values
should (perhaps optionally) be passed through the custom deserializer.

What steps will reproduce the problem?
1. compile the attached file
2. run it

What is the expected output? What do you see instead?
Expected output:
  NullInner should equal NullInner
  NullInner should equal NullInner
OR:
  NullInner should equal NullInner
  InitInner should equal NullInner
I see:
  null should equal NullInner
  InitInner should equal NullInner

What version of the product are you using? On what operating system?
gson 1.4


```

Original issue reported on code.google.com by `hell...@gmail.com` on 3 Nov 2009 at 11:33

Attachments:
- [Main.java](https://storage.googleapis.com/google-code-attachments/google-gson/issue-171/comment-0/Main.java)

PR #852, opened 2016-05-10
Add option to let nulls through custom deserializers

joffrey-bion:
This option fixes the "bug" that custom deserializers don't get called for null values, but without breaking backwards compatibility.
The issue #171 was already marked as closed but the problem was never truly adressed.

Resolves:
https://github.com/google/gson/issues/171

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -106,6 +106,7 @@ public final class Gson {
   static final boolean DEFAULT_PRETTY_PRINT = false;
   static final boolean DEFAULT_ESCAPE_HTML = true;
   static final boolean DEFAULT_SERIALIZE_NULLS = false;
+  static final boolean DEFAULT_LET_NULLS_THROUGH_DESERIALIZERS = false;
   static final boolean DEFAULT_COMPLEX_MAP_KEYS = false;
   static final boolean DEFAULT_SPECIALIZE_FLOAT_VALUES = false;
 
@@ -130,6 +131,7 @@ public final class Gson {
   private final Excluder excluder;
   private final FieldNamingStrategy fieldNamingStrategy;
   private final boolean serializeNulls;
+  private final boolean letNullsThroughDeserializers;
   private final boolean htmlSafe;
   private final boolean generateNonExecutableJson;
   private final boolean prettyPrinting;
@@ -146,6 +148,8 @@ public final class Gson {
    *   kept as is since an array is an ordered list. Moreover, if a field is not null, but its
    *   generated JSON is empty, the field is kept. You can configure Gson to serialize null values
    *   by setting {@link GsonBuilder#serializeNulls()}.</li>
+   *   <li>When nulls are serialized, Gson does not let them through custom deserializers by
+   *   default. You can allow that by setting {@link GsonBuilder#letNullsThroughDeserializers()}</li>
    *   <li>Gson provides default serialization and deserialization for Enums, {@link Map},
    *   {@link java.net.URL}, {@link java.net.URI}, {@link java.util.Locale}, {@link java.util.Date},
    *   {@link java.math.BigDecimal}, and {@link java.math.BigInteger} classes. If you would prefer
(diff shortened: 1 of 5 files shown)


joffrey-bion:
I signed the CLA!

JakeWharton:
The linked issue has a workaround without needing any changes though. Why
is this better?

joffrey-bion:
Because, as you said, this is a workaround. It doesn't really solve the problem.
The problem is that the current behaviour is not what is expected by users. As stated in the first answer on the issue, the only reason preventing the fix was backwards compatibility, which is kept in this PR.
I'm sure that other users like me and the reporter of the issue will stumble upon this problem while using custom deserializers.

joffrey-bion:
Also, I may be missing something, but is there any reason not to make this change? I don't find that very costly to provide such an option.

swankjesse:
If you want null you should use a TypeAdapter instead of a Deserializer. That already does what you want.

I'm opposed to this change because it adds a global setting to configure local behavior. In particular I don't expect all of my deserializers want this behavior, and enabling it means they all must be made null safe.

joffrey-bion:
Using a TypeAdapter, what is the equivalent of the following:

    @Override
    public Property<?> deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws
            JsonParseException {
        if (json.isJsonNull()) {
            return new SimpleObjectProperty<>(null);
        }
        Type typeParam = ((ParameterizedType) typeOfT).getActualTypeArguments()[0];
        Object obj = context.deserialize(json, typeParam);
        return new SimpleObjectProperty<>(obj);
    }

This is a custom deserializer for JavaFX Property<T>.

Maybe I'm not familiar enough with TypeAdapters, but it looks like using it would require me to know the whole object hierarchy inside my object, which I don't.

JakeWharton:
```java
public final class PropertyTypeAdapterFactory implements TypeAdapterFactory {
  @Override public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
    if (type.getRawType() != Property.class) return null;
    Type param = ((ParameterizedType) type.getType()).getActualTypeArguments()[0];
    TypeAdapter<?> delegate = gson.getDelegateAdapter(this, TypeToken.get(param));
    return (TypeAdapter<T>) new PropertyTypeAdapter<>(delegate);
  }

  static class PropertyTypeAdapter<V> extends TypeAdapter<Property<V>> {
    private final TypeAdapter<V> delegate;

    PropertyTypeAdapter(TypeAdapter<V> delegate) {
      this.delegate = delegate;
    }

    @Override public void write(JsonWriter out, Property<V> value) throws IOException {
      delegate.write(out, value.getValue());
    }

    @Override public Property<V> read(JsonReader in) throws IOException {
      return in.peek() == JsonToken.NULL
          ? new SimpleObjectProperty<V>(null)
          : new SimpleObjectProperty<>(delegate.read(in));
    }
  }
}
```

joffrey-bion:
Thanks a lot Jake for the adaptation, so in fact I need not only a `TypeAdapter` but a whole `TypeAdapterFactory`. 

Now I understand the workaround better, but anyway, don't you think `Serializers` should be able to handle it as well?
I mean, just because a workaround exist doesn't make the actual behaviour of the serializers more obvious to the users of Gson. I still believe null is a valid use case for deserialization. Maybe I just don't have the correct mindset.

joffrey-bion:
Interestingly, in the rush of giving sample code to convert, I reproduced the very thing I'm fighting against: I didn't let the chance to the delegate serializer to handle null as it pleases. It should instead have been:

```
@Override
public Property<?> deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws
        JsonParseException {
    if (json.isJsonNull()) {
        return new SimpleObjectProperty<>(null);
    }
    Type typeParam = ((ParameterizedType) typeOfT).getActualTypeArguments()[0];
    Object obj = context.deserialize(json, typeParam);
    return new SimpleObjectProperty<>(obj);
}
```

AlexeyGorovoy:
Tried to use a code snippet provided @JakeWharton (except that I replaced types with my own), but unfortunately the `TypeAdapter.read()` method never gets called for null values.

PR #691, opened 2015-08-25
Feature: No built-in type adapter for Iterable #672

warabei14:

Issue #672, opened 2015-07-29
No built-in type adapter for Iterable


collinsauve:
This: https://github.com/google/gson/issues/428

Iterable is a very fundamental type that sees extensive use.   It really should be supported out of the box.

kaqqao:
> added `Priority-High`  label on Sep 27, 2015
😂

Issue #428, opened 2015-03-19
No built-in type adapter for Iterable


GoogleCodeExporter:
```
What steps will reproduce the problem?
See the attached file.

What is the expected output?
[{"a":1},{"a":2}]
[{"a":1},{"a":2}]
[1, 2]
[1, 2]

What do you see instead?
[{"a":1},{"a":2}]
{}
[1, 2]
Exception in thread "main" java.lang.RuntimeException: Unable to invoke no-args 
constructor for java.lang.Iterable<com.ash2k.test1.App$A>. Register an 
InstanceCreator with Gson for this type may fix this problem.
    at com.google.gson.internal.ConstructorConstructor$8.construct(ConstructorConstructor.java:166)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:163)
    at com.google.gson.Gson.fromJson(Gson.java:755)
    at com.google.gson.Gson.fromJson(Gson.java:721)
    at com.google.gson.Gson.fromJson(Gson.java:670)
    at com.ash2k.test1.App.main(App.java:31)
Caused by: java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
    at java.lang.reflect.Method.invoke(Method.java:597)
    at com.google.gson.internal.UnsafeAllocator$1.newInstance(UnsafeAllocator.java:48)
    at com.google.gson.internal.ConstructorConstructor$8.construct(ConstructorConstructor.java:163)
    ... 5 more
Caused by: java.lang.InstantiationException: java.lang.Iterable
    at sun.misc.Unsafe.allocateInstance(Native Method)
    ... 11 more

What version of the product are you using? On what operating system?
2.1 on Ubuntu 11.10 x64

```

Original issue reported on code.google.com by `Ash2kk@gmail.com` on 1 Apr 2012 at 7:38

Attachments:
- [App.java](https://storage.googleapis.com/google-code-attachments/google-gson/issue-428/comment-0/App.java)

inder123:
Hi Folks.
What is the intended behavior here? Do you want us to pick a random sub-class of Iterable (such as List) and deserialize the collection into it?

[...]

PR #854, opened 2016-05-14
issue#672.add IterableTypeAdapterFactory

ianate:
add IterableTypeAdapterFactory for Iterable. no enough testcase.some exceptions unhandled


inder123:
please write tests as well.

ikus060:
I see many issues related to IterableTypeAdapterFactory: #672 #691 #428 

In recent version of Gson 2.8.0, I still don't see the IterableTypeAdapterFactory class. So whats is the status of it. Is it implemented in master or not ?

PR #860, opened 2016-05-19
Update UserGuide.md

jachinte:
Correct code for two examples:
- Custom exclusion strategy. The final field `clazzField` is not initialized in the example. Removing the final modifier makes necessary to modify the json output to `{"longField":1234,"clazzField":null}`.
- Custom instance creator. The static method `Id.get` does not exist; instead of that, a regular object instantiation should be used: `return new Id((Class)idType, 0L);`

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -421,7 +421,7 @@ class IdInstanceCreator implements InstanceCreator<Id<?>> {
   public Id<?> createInstance(Type type) {
     Type[] typeParameters = ((ParameterizedType)type).getActualTypeArguments();
     Type idType = typeParameters[0]; // Id has only one parameterized type T
-    return Id.get((Class)idType, 0L);
+    return new Id((Class)idType, 0L);
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


jachinte:
I signed it!

UserGuide.md
```diff
@@ -421,7 +421,7 @@ class IdInstanceCreator implements InstanceCreator<Id<?>> {
   public Id<?> createInstance(Type type) {
     Type[] typeParameters = ((ParameterizedType)type).getActualTypeArguments();
     Type idType = typeParameters[0]; // Id has only one parameterized type T
-    return Id.get((Class)idType, 0L);
+    return new Id((Class)idType, 0L);
```
Marcono1234, line 424:
```suggestion
    return new Id<>((Class<?>)idType, 0L);
```
Would get raw type warnings otherwise

eamonnmcmanus:
Looks like these issues got fixed in the intervening decade.

Issue #824, opened 2016-04-07
gson parse success for json strings that have an invalid escape sequence


nkarkala:
I am basically doing the following,

`JsonElement jsonElement = new JsonParser().parse(config);  
`

When the config string is for example : {"userName": "ab\cd"} which has an invalid escape sequence (\c), instead of getting a parse error, it appears that the escape character is just being removed. Hence the resulting jsonElement has "userName": "abcd". 

Looking at the code base in the JsonReader class, I see that when a valid escape sequence is not matched , it is simply ignored (default case) rather than throwing an invalid json error. Is this behavior intended?  This is critical since I fail to catch the error and end up using a modified string (i.e with escape character removed) . 

```
private char readEscapeCharacter() throws IOException {
    if (pos == limit && !fillBuffer(1)) {
      throw syntaxError("Unterminated escape sequence");
    }

    char escaped = buffer[pos++];
    switch (escaped) {
    case 'u':
      if (pos + 4 > limit && !fillBuffer(4)) {
        throw syntaxError("Unterminated escape sequence");
      }
      // Equivalent to Integer.parseInt(stringPool.get(buffer, pos, 4), 16);
      char result = 0;
      for (int i = pos, end = i + 4; i < end; i++) {
        char c = buffer[i];
        result <<= 4;
        if (c >= '0' && c <= '9') {
          result += (c - '0');
        } else if (c >= 'a' && c <= 'f') {
          result += (c - 'a' + 10);
        } else if (c >= 'A' && c <= 'F') {
          result += (c - 'A' + 10);
        } else {
          throw new NumberFormatException("\\u" + new String(buffer, pos, 4));
        }
      }
      pos += 4;
      return result;

    case 't':
      return '\t';

    case 'b':
      return '\b';

    case 'n':
      return '\n';

    case 'r':
      return '\r';

    case 'f':
      return '\f';

    case '\n':
      lineNumber++;
      lineStart = pos;
      // fall-through

    case '\'':
    case '"':
    case '\\':
   default:
      return escaped;
    }
  } 
```

JMVM:
Related to this issue, there is another one when having for example a date with double-slash escaped solidus character (in my case: "date":"27\\/4\\/2016").

Having the look **JsonReader class ** too, the escaped character is **/**, being returned by default. So far so good, having the same behavior than escaped **\a**, returning character **a**. However, in this case, getting an exception from **Gson class**:

[...]

PR #862, opened 2016-05-23
Enhancing the json reader, adding corresponding test case for it; #824

deyboy90:
As discussed in the issue #824, when invalid escape sequences are provided the parse is still successful.

eg. " \"h\\ello\": true",  the parser never used to complain; but after making the below enhancement when ever an unknown escape is encountered then the code will throw syntax error and rightfully inform the user rather than making assumptions.

Also added a test case to demonstrate this. Have executed all the existing tests to avoid any regression bugs.

### DIFF
--- a/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
+++ b/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
@@ -173,7 +173,22 @@ public final class JsonReaderTest extends TestCase {
     reader.endObject();
     assertEquals(JsonToken.END_DOCUMENT, reader.peek());
   }
+  
+  public void testInvalidJsonInput() throws IOException {
+	  String json = "{\n" +
+		        "   \"h\\ello\": true,\n" +
+		        "   \"foo\": [\"world\"]\n" +
+		        "}";
 
+	  try {
+		  JsonReader reader = new JsonReader(reader(json));
+		  reader.beginObject();
+		  assertEquals("hello", reader.nextName());
+	      fail();
+	    } catch (IOException expected) {
+	    }
+  }
+  
   public void testNulls() {
     try {
       new JsonReader(null);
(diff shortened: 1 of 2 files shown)


deyboy90:
I signed it!

swankjesse:
LGTM. Some small style issues.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -173,7 +173,22 @@ public void testHelloWorld() throws IOException {
     reader.endObject();
     assertEquals(JsonToken.END_DOCUMENT, reader.peek());
   }
-
+  
+  public void testInvalidJsonInput() throws IOException {
+	  String json = "{\n" +
```
swankjesse: indent this +2 everywhere please

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -173,7 +173,22 @@ public void testHelloWorld() throws IOException {
...
+		        "   \"h\\ello\": true,\n" +
+		        "   \"foo\": [\"world\"]\n" +
+		        "}";
+
+	  try {
+		  JsonReader reader = new JsonReader(reader(json));
+		  reader.beginObject();
+		  assertEquals("hello", reader.nextName());
```
swankjesse: the assertEquals() isn’t necessary, just calling nextName() should trigger it to throw

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -173,7 +173,22 @@ public void testHelloWorld() throws IOException {
...
-
+  
+  public void testInvalidJsonInput() throws IOException {
+	  String json = "{\n" +
+		        "   \"h\\ello\": true,\n" +
+		        "   \"foo\": [\"world\"]\n" +
+		        "}";
+
+	  try {
+		  JsonReader reader = new JsonReader(reader(json));
```
swankjesse: this line and the line following shouldn't be in the try/catch block, it's only the nextName call that throws

deyboy90:
Have made the necessary changes. Hope the indentation is fine now.

swankjesse:
Merged manually!

PR #863, opened 2016-05-23
handle explicit null values in JSON

nykolaslima:
Fix error when trying deserialize JSON with explicit `NULL` value.

```
{msg:'foo', count:null}
```


nykolaslima:
I signed it!

PR #879, opened 2016-06-27
Added flush() to toJson() method when using Appendable writer

gaborantal:
If you use Appendable writer (e.g.: FileWriter), it can happen that the Appendable object never gets flushed. Example:

```
List<MyClassType> current = /* */;
Gson gson = /* */;
gson.toJson(current, new FileWriter("MYFILE.JSON"));
```

If you do something like this, it can happen, that the created file is empty. It is not the best solution, as the opened file never gets closed, and so, but  if you create some really simple app to the work for you, maybe a fast solution.  After adding an extra flush() call, after write finished, all the contents will must be in the file.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -646,6 +646,7 @@ public final class Gson {
     try {
       JsonWriter jsonWriter = newJsonWriter(Streams.writerForAppendable(writer));
       toJson(src, typeOfSrc, jsonWriter);
+      jsonWriter.flush();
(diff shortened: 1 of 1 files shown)


gaborantal:
I signed it!

JakeWharton:
Since `JsonWriter` doesn't buffer I'm tempted to say that Gson has no obligation here. Any reason you're not using something like this?
```java
try (Writer writer = new FileWriter("MYFILE.JSON")) {
  gson.toJson(current, writer);
}
```

gaborantal:
True point, i can live with it.

PR #882, opened 2016-06-30
Nullable checker

ntrrgc:
Crazy idea: ensure data marked as not null actually is not null when you deserialize it.

You can either use Java annotations (e.g. JetBrains, javax.validation, Java 8 `NotNull`) or any other metadata accesible with reflection. 

The following example uses Kotlin reflection:

```kotlin
data class MyNestedObject(
    val mayNotBeNull: String
)

data class MyEntity(
    val mayBeNull: String?,
    val nestedObject: MyNestedObject
)

fun main(args: Array<String>) {
    val kotlinNotNullCheker = object: NullableChecker {
        override fun fieldIsNullable(field: Field): Boolean {
            val property = field.kotlinProperty
            if (property != null) {
                return property.returnType.isMarkedNullable
            } else {
                return true;
            }
        }
    }

    val gson = GsonBuilder()
        .setNullableChecker(kotlinNotNullCheker)
        .create()

    val entity = gson.fromJson("""
        {"mayBeNull": null, "nestedObject": {"mayNotBeNull": null}}
        """, MyEntity::class.java)
}
```

The parser is able to detect the error:

```
Exception in thread "main" com.google.gson.JsonSyntaxException:
Non-nullable field was read as null: mayNotBeNull
```

Missing non-nullable properties are also catched. For instance, `{"mayBeNull": null}` throws:

```
Exception in thread "main" com.google.gson.JsonSyntaxException:
Missing required fields: nestedObject
```

### DIFF
--- a/gson/src/main/java/com/google/gson/DummyNullableChecker.java
+++ b/gson/src/main/java/com/google/gson/DummyNullableChecker.java
@@ -4,6 +4,8 @@ import java.lang.reflect.Field;
 
 /**
  * Dummy NullableChecker that does not check anything.
+ *
+ * @author Juan Luis Boya García
  */
 class DummyNullableChecker {
   private static NullableChecker nullableChecker = null;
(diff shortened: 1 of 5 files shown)


ntrrgc:
I signed it.

Issue #714, opened 2015-10-19
Add JsonElement.getAsDate


Queatz:

Marcono1234:
Probably unlikely that this will be implemented given that the corresponding pull request was rejected: https://github.com/google/gson/pull/883#issuecomment-230132558

(Note that I am not a member of this project.)

PR #883, opened 2016-07-02
Add JsonElement.getAsDate

izzabudaka:
See #714

Added convenience method `public Date getAsDate(String pattern)` to get JsonPrimitive Date Strings as java.util.Date and corresponding JsonArray implementation.


izzabudaka:
I signed it!

JakeWharton:
This is a strange thing to specialize, and the asymmetry of the
implementation makes me even less excited about the prospect of its
addition.

inder123:
Please indicate why this must be done inside Gson.

izzabudaka:
Using Gson in multiple projects, I have found myself implementing similar Date parsing/validation functionality. I believe this change would offer greater convenience to the users. Any comments on how to make it more symmetric are welcomed :)

PR #885, opened 2016-07-06
Give FieldNamingStrategy the ability to translate Strings

lndbrg:
This opens up the possibility of translating objects that aren't
fields but when there is need to know the JSON field representation
of it. For example data classes where the methods are annotated with
@SerializedName.

Currently I have to copy paste code and modify it to take strings:
https://github.com/danielnorberg/auto-matter/pull/39
This is unfortunate as that code will diverge as soon as something changes in gsons FieldNamingPolicies.


lndbrg:
I signed it!

todoooo:
I think it would be better if the FieldNamingStrategy interface only contained the translateName(string s) method.

PR #888, opened 2016-07-13
Fix for RuntimeTypeAdapterFactory.

leangl:
RuntimeTypeAdapterFactory stopped working since the 2.0 release. This should fix it, I can provide the corresponding test cases if needed.


JakeWharton:
Test cases are required, yes.

swankjesse:
This is working as designed. If you have a problem with the design, we can address that.

Basically RuntimeTypeAdapterFactory only triggers when the supertype is declared:

```
   Shape shape;
```

It doesn't trigger when a subtype is declared:

```
   Square square;
```

There’s not much reason to use RuntimeTypeAdapterFactory in that case.

todoooo:
This pull request change set makes sense. The adapter factory interrogates the class type of the instance during serialisation, not the declared class type. The original implementation only works for deserialisation.

lglossman:
I signed it!

Marcono1234:
This was resolved by #2160.

Issue #884, opened 2016-07-04
Feature Request: JsonArray#getAsString() should return an empty string if the array is empty


sibennayak:
Currently **JsonArray#getAsString()** returns the first element of the array as a string if it contains only one element. Perhaps we could also handle empty arrays by returning a empty string.

apranav19:
Hi @sibennayak 
I have submitted a PR for this suggestion.

maysale01:
Curious why JsonArray.getAsString() wouldn't return the entire serialized array? I would expect it to be valid JSON, i.e. "[]".

PR #889, opened 2016-07-14
Allow JSONArray to return a blank string if empty

apranav19:
The purpose of this PR is to satisfy the issue #884 

Changes include:
- Returning a blank string if a `JSONArray` instance does not have any elements in it.
- The respective unit test for it.


apranav19:
I signed it!

PR #890, opened 2016-07-19
Use json element tree directly in type adapters

todoooo:
The Tree type adapter and Runtime type adapter use a json element to perform serialization/deserialization. Parsing a json reader into a tree structure and then back into a reader for consumption by other adapters is inefficient. e.g the RuntimeTypeAdapter uses the tree structure to find the class "type" and then finds another registered (possibly the Tree type) adapter to serialize/deserialize the descendant class. 

These changes allow the tree structure to be consumed directly by a type adapter.

### DIFF
--- a/gson/src/main/java/com/google/gson/TypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/TypeAdapter.java
@@ -191,12 +191,29 @@ public abstract class TypeAdapter<T> {
           TypeAdapter.this.write(out, value);
         }
       }
+
+      @Override public JsonElement toJsonTree(T value) {
+        if (value == null) {
+          return JsonNull.INSTANCE;
+        } else {
+          return TypeAdapter.this.toJsonTree(value);
+        }
+      }
+
       @Override public T read(JsonReader reader) throws IOException {
         if (reader.peek() == JsonToken.NULL) {
           reader.nextNull();
           return null;
         }
         return TypeAdapter.this.read(reader);
+      }
+
+      @Override public T fromJsonTree(JsonElement jsonTree) {
+        if (jsonTree.isJsonNull()) {
+          return null;
+        } else {
+          return TypeAdapter.this.fromJsonTree(jsonTree);
+        }
       }
     };
   }
--- a/gson/src/main/java/com/google/gson/TypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/TypeAdapter.java
@@ -228,7 +228,7 @@ public abstract class TypeAdapter<T> {
    * @return the converted JSON tree. May be {@link JsonNull}.
    * @since 2.2
    */
-  public final JsonElement toJsonTree(T value) {
+  public JsonElement toJsonTree(T value) {
     try {
       JsonTreeWriter jsonWriter = new JsonTreeWriter();
       write(jsonWriter, value);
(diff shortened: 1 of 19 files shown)


todoooo:
I signed it.

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -191,13 +191,30 @@ public final void toJson(Writer out, T value) throws IOException {
           TypeAdapter.this.write(out, value);
         }
       }
+
+      @Override public JsonElement toJsonTree(T value) {
+        if (value == null) {
+          return JsonNull.INSTANCE;
```
swankjesse: I believe type adapters are allowed to convert null values into non-null JSON

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -191,13 +191,30 @@ public final void toJson(Writer out, T value) throws IOException {
...
       @Override public T read(JsonReader reader) throws IOException {
         if (reader.peek() == JsonToken.NULL) {
           reader.nextNull();
           return null;
         }
         return TypeAdapter.this.read(reader);
       }
+
+      @Override public T fromJsonTree(JsonElement jsonTree) {
+        if (jsonTree.isJsonNull()) {
```
swankjesse: Same

swankjesse:
Not sure I really follow this. JsonTreeReader is pretty efficient because it doesn’t actually decode anything. Currently fromJsonTree() delegates to read(); this flips the order of delegation which could be problematic.

gson/src/main/java/com/google/gson/TypeAdapter.java
todoooo: @swankjesse The "if null" code was moved from the preceding read()/write() methods, not added.

todoooo:
@swankjesse 
Nothing has changed for TypeAdapters that read/write from the reader/writer objects. 

I'm considering the situation where an adapter (i.e. RunTimeTypeAdapter) specifically converts the Reader/Writer into json elements and then calls a delegate jsonFromTree/jsonToTree method. When a developer uses custom JsonSerializer/JsonDeserializer classes they are wrapped by TreeTypeAdapters which convert the json reader/writer into json elements for processing. In this case the RunTimeTypeAdapter would convert a reader/writer into json elements and then its call to "delegate.to/fromJsonTree()" would convert them back into a reader/writer for a TreeTypeAdapter to convert them back into json elements. Inefficient!

swankjesse:
Have you looked at how JsonTreeReader works?

todoooo:
Yes. It wraps the json elements which is fine. However, the TreeTypeAdapter then converts that reader/writer back into a new tree of json elements for serializing/deserializing.

swankjesse:
One more question – what happens if the type adapter mutates the JsonObject? I believe the current implementation gets a copy, and your change makes it so that there’s a single instance that’s shared. That could be surprising?

todoooo:
I can't think of a reason for not wanting to mutate the shared tree instance, but at least the new implementation gives the developer a choice. Either mutate the existing tree or clone it, as the RunTimeTypeAdapter does during serialization. (Although I'm not sure why, since it doesn't clone the tree during deserialisation.)

Issue #269, opened 2015-03-19
Add support for custom actions on annotations


GoogleCodeExporter:
```
See the thread: 
http://groups.google.com/group/google-gson/browse_thread/thread/1ea949a5e50cc507

by supporting a custom action on an annotation, we will enable support for 
JPA/Hibernate and other kinds of annotations.
```

Original issue reported on code.google.com by `inder123` on 26 Nov 2010 at 7:00

inder123:
https://github.com/google/gson/pull/930

fwonce:
> One simple example: I'm using Objectify on AppEngine, and so one of my entity 
> classes is like so:
> 
> public class MyThing {
>   @Id Long id;
>   String title;
> }
> 
> I want this to map to:
> {"id":443,"title":"Thing"}
> 
> If I could write a type adapter that triggers off annotations, I could just use 
> the existing @Id annnotation to ensure the Long is serialized as a String

Exactly the same need here. Another reason: if `MyThing` in the example above is located in a common package meant to provide as a 3rd party jar and will be serialized to JSON format in some application at runtime. Provided that for good reason it won't include `gson` in its dependencies, to archive this goal, the application can register a custom `TypeAdapter` or `TypeAdapterFactory` serving the fields annotated with `@Id`. Simple and clean for the client code, right?

amaksoft:
Hi, i want to share my implementation of custom annotation support for gson, which is backwards compatible:
new `interface ParametrizedTypeAdapterFactory` extends `TypeAdapterFactory` and new create() method only gets called if field is annotated with specific annotation `@Parametrized`

I mostly wanted it for serializing some dates by specifying custom pattern in annotation like this:

```java
class Person {

  @Expose
  @SerializedName("birth_date")
  @Parametrized
  @CustomDateFormat("MMMM d, yyyy h:mm:ss a z")
  private Date birthDate;

}
```

Here is a draft implementation:
https://github.com/amaksoft/gson/commit/b264334451c487d79241483103853c3fff27cbd4
Can you please review it?

amaksoft:
Just wanted to explain my position:
`ParamterizedTypeAdapterFactory` is just a special case of `TypeAdapterFactory`. In most cases `Gson` don't even need to know that your factory accepts annotations. But in special cases like `@CustomDateFormat` on `Date` or `@Id` on `int`, you return a customized `TypeAdapter` only for this case, (we don't cache it).
[...]

PR #893, opened 2016-07-27
Added a GsonBuilder method to register a post deserialization annotat…

inder123:
…ion.

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/PostDeserializationAnnotationRegistrationTest.java
+++ b/gson/src/test/java/com/google/gson/functional/PostDeserializationAnnotationRegistrationTest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2008 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson.functional;
+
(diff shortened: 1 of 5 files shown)


inder123:
This addresses Issue https://github.com/google/gson/issues/269

gson/src/test/java/com/google/gson/functional/PostDeserializationAnnotationRegistrationTest.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2008 Google Inc.
```
cgruber: This probably is a copy-pasta error.  2016?

inder123: yes

JakeWharton:
This is similar to what https://github.com/google/gson/pull/844 wanted.

You can currently write and register a delegating TypeAdapterFactory which checked for annotated methods post-deserialization without needing support in Gson itself. Is that enough?

inder123:
@JakeWharton  I don't understand the delegating TypeAdapterFactory solution. Can you illustrate? Thanks.

inder123:
This specific solution is good in one specific way: You can just map the javax.annotation.PostConstruct annotation and that will benefit users of frameworks such as JPA.

inder123:
Can you suggest a better name for the GsonBuilder method, or even a different signature?
For example, should this method apply the annotation to All classes or match them with a regular expression?
Are there actions beyond post-deserialization where we could look for annotation? For example, pre/post serialization.

swankjesse:
Here’s an example that does it standalone:
https://gist.github.com/swankjesse/20df26adaf639ed7fd160f145a0b661a

swankjesse:
One other difference with this PR vs the gist above – this only applies to field bindings; that applies to any type.

JakeWharton:
Nice @swankjesse. That's exactly what I was thinking.

inder123:
Thanks @swankjesse for writing this up. I will try it out.
I think overall something like this needs to be part of Gson. We have talked about it for a very long time but without acting.

inder123:
Closing this PR since the specified GIST is a workaround.

PR #900, opened 2016-08-12
Add static factories for array and parameterized type tokens.

JakeWharton:
These are useful when creating TypeAdapterFactories that delegate to others with more complex types. They also are useful when writing dynamic code that deals with types that cannot be fully reified using the normal subclass technique.

Equivalent methods are exposed on Moshi's `Types` utility.

Some examples of where these are needed:
 * Writing a Retrofit `Converter.Factory` which automatically wraps your model object type in an envelope type (like JSON-RPC). For example, `TypeToken.getParameterized(JsonRpcRequest.class, userType)`.
 * When doing compile-time code generation of `TypeAdapter` implementations that support generics. To handle `T[]` you need to generate code that combines the statically known type with the runtime generic type. For example, `TypeToken.getArray(runtimeType)`.

### DIFF
--- a/gson/src/main/java/com/google/gson/reflect/TypeToken.java
+++ b/gson/src/main/java/com/google/gson/reflect/TypeToken.java
@@ -302,4 +302,19 @@ public class TypeToken<T> {
   public static <T> TypeToken<T> get(Class<T> type) {
     return new TypeToken<T>(type);
   }
+
+  /**
+   * Gets type literal for the parameterized type represented by applying {@code typeArguments} to
+   * {@code rawType}.
+   */
+  public static TypeToken<?> getParameterized(Type rawType, Type... typeArguments) {
+    return new TypeToken<Object>($Gson$Types.newParameterizedTypeWithOwner(null, rawType, typeArguments));
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -302,4 +302,19 @@ private static boolean matches(Type from, Type to, Map<String, Type> typeMap) {
...
+   * {@code rawType}.
+   */
+  public static TypeToken<?> getParameterized(Type rawType, Type... typeArguments) {
+    return new TypeToken<Object>($Gson$Types.newParameterizedTypeWithOwner(null, rawType, typeArguments));
+  }
+
+  /**
+   * Gets type literal for the array type whose elements are all instances of {@code componentType}.
+   */
+  public static TypeToken<?> getArray(Type componentType) {
```
inder123: How about array of parameterized type?

JakeWharton: You can do `getArray(getParameterized(List.class, String.class).getType())`

inder123:
Looks good to me.

PR #901, opened 2016-08-14
Enable subclassing of ReflectiveTypeAdapterFactory

F43nd1r:
Usecase: I need to serialize a subclass of a collection which adds additional fields, see https://github.com/F43nd1r/Multitool/blob/master/app/src/main/java/com/google/gson/internal/bind/CollectionReflectiveTypeAdapter.java .
As you can see, this requires a lot of duplicated code from ReflectiveTypeAdapter.
To prevent this, it would be nice to be able to access this method and inner class from subclasses.
(ReflectiveTypeAdapterFactory requires a ConstructorConstructor in its constructor, which is why it is necessary that gson exposes its ConstructorConstructor.)

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -255,6 +255,10 @@ public final class Gson {
     this.factories = Collections.unmodifiableList(factories);
   }
 
+  public ConstructorConstructor constructorConstructor() {
+    return constructorConstructor;
+  }
+
   public Excluder excluder() {
     return excluder;
   }
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -139,7 +139,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     };
   }
 
-  private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type, Class<?> raw) {
+  protected Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type, Class<?> raw) {
     Map<String, BoundField> result = new LinkedHashMap<String, BoundField>();
     if (raw.isInterface()) {
       return result;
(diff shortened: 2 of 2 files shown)


F43nd1r:
I signed it!

swankjesse:
You can accomplish this with delegation rather than inheritance. Look at TypeAdapterFactory and its APIs to compose adapters from other adapters.

F43nd1r:
@swankjesse That is not possible in this case, as Gson will always return a `CollectionTypeAdapterFactory.Adapter` if i try to access a delegate adapter for a type extending collection.
But I need a `CollectionTypeAdapterFactory.Adapter` and a `ReflectiveTypeAdapterFactory.Adapter` _for the same type_ if I would solve this with composition.

F43nd1r:
If there is a way to enable compostion in this case I'd happily use it, but I could not think of a way of doing that without seriously messing with the existing API.

swankjesse:
Got it. What you're doing is pretty weird! I don’t think it's a good idea to depend on Gson’s internal package. In this case your best bet is probably to change your JSON or Java models to be less weird!

F43nd1r:
I found that suggestion on [stackoverflow](http://stackoverflow.com/questions/27506989/how-do-i-gson-serialize-an-object-extending-a-collection) too. 
However, my datamodel is enforced by the connected view.
This is why I'm trying to simulate composition on the JSON model, while my Java model uses inheritance. (As Gson/Json does not support properties on an array.)

F43nd1r:
I just thought of an alternative:
A method like this in `Gson` would do it too: 

  
```java
  public <C, T extends TypeAdapter<C>>  T getAdapterOfType(Class<? super T> adapterClass, TypeToken<C> forType){
    for (TypeAdapterFactory factory : factories) {

      TypeAdapter<C> candidate = factory.create(this, forType);
      if (candidate != null && adapterClass.isAssignableFrom(candidate.getClass())) {
        //noinspection unchecked
        return (T) candidate;
      }
    }
      throw new IllegalArgumentException("GSON cannot find adapter of class " + adapterClass + " for type " + forType);
  }
```

Does that look better?

swankjesse:
The Stack Overflow answer with the checkmark is what you want. Your value should have a list, not be a list.

F43nd1r:
> However, my datamodel is enforced by the connected view.
I'd need to create a deep copy of my datastructure each time I want to persist it, as changing the datamodel in usage is not an option.

swankjesse:
Got it. Too bad. In that case I recommend copying the code you need into your project. I don’t want to encourage anyone to pull code from Gson’s `internal` package.

F43nd1r:
Oh well, not what I wanted. But I understand the notion to keep internals hidden.

PR #902, opened 2016-08-17
Add an option to operate parser in strict mode.

chajath:
I can't help but feeling that hard coding leniency setting at the top level was a mistake. However, given changing that will introduce too big a change in behavior, I propose that we introduce a separate flag to enforce strict parsing at the outermost level of the element.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -108,6 +108,7 @@ public final class Gson {
   static final boolean DEFAULT_SERIALIZE_NULLS = false;
   static final boolean DEFAULT_COMPLEX_MAP_KEYS = false;
   static final boolean DEFAULT_SPECIALIZE_FLOAT_VALUES = false;
+  static final boolean DEFAULT_STRICT_PARSING = false;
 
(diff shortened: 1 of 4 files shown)


chajath:
This will be a workaround for google/gson#372

swankjesse:
You can workaround by getting a TypeAdapter first. It doesn't do anything lenient.

chajath:
Not quite sure what you mean by that. Is there any code snippet that I can look at? Workaround-able or not, we might as well offer this option to the user anyway to be more user friendly.

JakeWharton:
Call getAdapter with your type first and then use that TypeAdapter for
calls to toJson or fromJson.

chajath:
I see. I guess the workaround is relatively simple. I still think there is some value in this pull request, because this change makes it more obvious to the users what gson fromJson is doing by inspecting API of the gson builder.

I also see #804 which will make fromJson more consistent, but I guess this will introduce breaking change.

Marcono1234:
This seems to be obsolete with the `Strictness` mode API introduced by #2437 for the next Gson version.

Thanks nonetheless for the work you have put into this!

PR #903, opened 2016-08-22
Fix BigInteger equals bug, and its tests.

ybank:
Dear GSON developers, this pull request is a fix to solve the existed big integer equals problem: when an integer is really large it can be regarded as equal to some small integer, because of the loss of precision when converting big integer to long. The commit also fixed and improved the test. Thanks!

### DIFF
--- a/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
+++ b/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
@@ -226,12 +226,21 @@ public class JsonPrimitiveTest extends TestCase {
     MoreAsserts.assertEqualsAndHashCode(new JsonPrimitive(Float.NaN), new JsonPrimitive(Double.NaN));
   }
 
-  public void testEqualsIntegerAndBigInteger() {
-    JsonPrimitive a = new JsonPrimitive(5L);
-    JsonPrimitive b = new JsonPrimitive(new BigInteger("18446744073709551621")); // 2^64 + 5
-    // Ideally, the following assertion should have failed but the price is too much to pay 
-    // assertFalse(a + " equals " + b, a.equals(b));
-    assertTrue(a + " equals " + b, a.equals(b));
+  public void testEqualsForBigIntegers() {
+    BigInteger limit =         // 2^64
+        new BigInteger("18446744073709551616");
+    JsonPrimitive one = new JsonPrimitive(1L);
+    JsonPrimitive lp1 =        // limit + 1
+        new JsonPrimitive(limit.add(new BigInteger("1")));
+    JsonPrimitive lp1c =       // limit + 1, a different object
+        new JsonPrimitive(limit.add(new BigInteger("1")));
+    JsonPrimitive lp2 =        // limit + 2
+        new JsonPrimitive(limit.add(new BigInteger("2")));
+    // compare 1, limit + 1, limit + 2, etc.
+    assertFalse("limit + 1 = 1", lp1.equals(one));
+    assertFalse("1 = limit + 1", one.equals(lp1));
+    assertFalse("limit + 1 = limit + 2", lp1.equals(lp2));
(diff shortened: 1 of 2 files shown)


gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
```diff
@@ -169,6 +169,25 @@ public void testLongEqualsBigInteger() {
     assertEquals(p1.hashCode(), p2.hashCode());
   }
 
+  public void testHashCodeBigIntegerEqualsBigInteger() {
+    // within long range
+    BigInteger smallBigInt = new BigInteger("10");
+    JsonPrimitive p1 = new JsonPrimitive(smallBigInt);
+    JsonPrimitive p2 = new JsonPrimitive(smallBigInt);
+    assertTrue("should be different objects: " + p1 + ", " + p2, p1 != p2);
```
Marcono1234, line 177: Not needed because above `p1` and `p2` are assigned two new (therefore not-same) instances

gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
```diff
@@ -169,6 +169,25 @@ public void testLongEqualsBigInteger() {
...
+    assertTrue("p1 and p2 should have the same value", p1.equals(p2));
+    assertTrue("p1 and p2 should have the same hash code", p1.hashCode() == p2.hashCode());
+
+    // out of long range
+    BigInteger limitPlus5 =  // 2^64 + 5
+      new BigInteger("18446744073709551616").add(BigInteger.valueOf(5));
+    JsonPrimitive p3 = new JsonPrimitive(limitPlus5);
+    JsonPrimitive p4 = new JsonPrimitive(limitPlus5);
+    assertTrue("should be different objects: " + p3 + ", " + p4, p3 != p4);
```
Marcono1234, line 186: Not needed because above `p1` and `p2` are assigned two new (therefore not-same) instances

gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
```diff
@@ -169,6 +169,25 @@ public void testLongEqualsBigInteger() {
...
   }
 
+  public void testHashCodeBigIntegerEqualsBigInteger() {
+    // within long range
+    BigInteger smallBigInt = new BigInteger("10");
+    JsonPrimitive p1 = new JsonPrimitive(smallBigInt);
+    JsonPrimitive p2 = new JsonPrimitive(smallBigInt);
+    assertTrue("should be different objects: " + p1 + ", " + p2, p1 != p2);
+    assertTrue("p1 and p2 should have the same value", p1.equals(p2));
+    assertTrue("p1 and p2 should have the same hash code", p1.hashCode() == p2.hashCode());
```
Marcono1234, line 179: Use `assertEquals(int, int)` to get a more descriptive error message in case of mismatch

gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
```diff
@@ -169,6 +169,25 @@ public void testLongEqualsBigInteger() {
...
+
+    // out of long range
+    BigInteger limitPlus5 =  // 2^64 + 5
+      new BigInteger("18446744073709551616").add(BigInteger.valueOf(5));
+    JsonPrimitive p3 = new JsonPrimitive(limitPlus5);
+    JsonPrimitive p4 = new JsonPrimitive(limitPlus5);
+    assertTrue("should be different objects: " + p3 + ", " + p4, p3 != p4);
+    assertTrue("p3 and p4 should have the same value", p3.equals(p4));
+    assertTrue("p3 and p4 should have the same hash code", p3.hashCode() == p4.hashCode());
```
Marcono1234, line 188: Use `assertEquals(int, int)` to get a more descriptive error message in case of mismatch

eamonnmcmanus:
I think this was fixed by #2311.

PR #905, opened 2016-08-22
BigInteger equals bugfix

ybank:
The BigInteger conversion may lose precision so when compare with a small integer it can incorrectly regard them as equal to each other. This fix removes this bug and updated tests to prevent regression.


ybank:
Sorry, duplicate!

PR #916, opened 2016-08-30
Add the exports for the internal packages for backwards compatibility

nacx:
Exports for the `com.google.gson.internal.*` packages where not added back in https://github.com/google/gson/pull/797.

Apache jclouds heavily relies on Gson and does many complex (de)serialization stuff at its core to leverage clean interfaces for the supported providers. It relies on [some classes in the internal packages](https://github.com/jclouds/jclouds/search?utf8=%E2%9C%93&q=%22com.google.gson.internal%22&type=Code) and it is now broken in OSGi, which is a real blocker to upgrade Apache jclouds to the latest version of Gson.

The classes and methods being used are all public, so I'm wondering whether they are really internal (as they provide some features that could be used by downstream consumers). Could the internal packages be added back again to the OSGi exports, to keep backwards compatibility?

### DIFF
--- a/gson/bnd.bnd
+++ b/gson/bnd.bnd
@@ -11,4 +11,6 @@ Bundle-RequiredExecutionEnvironment: J2SE-1.5, JavaSE-1.6, JavaSE-1.7, JavaSE-1.
     com.google.gson,\
     com.google.gson.annotations,\
     com.google.gson.reflect,\
-    com.google.gson.stream
+    com.google.gson.stream,\
+    com.google.gson.internal,\
+    com.google.gson.internal.bind


nacx:
I signed it!

swankjesse:
I don't get it. What’s the point of OSGi if not to strongly encapsulate implementation details?

nacx:
Absolutely.

The problem here is the deprecation path and the inconsistencies between regular apps and OSGI deployed ones. Those packages were *explicitly* exported in previous versions and have been removed directly, without a reference in the changelog or release notes. There is also no deprecation warning or migration path for the downstream users so they can upgrade Gson and their apps. In one release, those packages literally *disappeared* in OSGi without any warning or alternative.

I agree that internal classes should not be exposed, but with the current setup, a standalone Java app can use those classes, apps deployed in servlet containers can use them too, but apps deployed in OSGi containers can't. And that is not consistent. Visibility and scope of internal classes shouldn't be enforced by a package name, but the visibility of their constructors and methods, and those internal classes are all public. Otherwise one should expect downstream users to be using them (just like people uses `sun.misc.Unsafe`).

I'd suggest a friendly upgrade path, so users of old Gson versions affected by this can upgrade without issues:

1. Add the exports back, and add a warning in the release notes that those packages won't be no longer exported in future releases, and cut a bugfix release exporting those packages.
2. Make all internal classes constructors package-private and limit the visibility of their methods so they can't be instantiated/used from the outside.
3. Consider providing a public interface to leverage the functionality that will be hidden, so users have an upgrade path.

I think this makes sense. There is no actual impact on Gson itself for adding the exports back and helps users upgrade properly.

JakeWharton:
> In one release, those packages literally disappeared in OSGi without any warning or alternative.

But they were not part of the public API. This is exactly what normal Java users experience when they depend on `internal` packages and upgrade to find that they changed.

> Visibility and scope of internal classes shouldn't be enforced by a package name, but the visibility of their constructors and methods, and those internal classes are all public.

Java has no mechanism of sharing internal classes among different packages before Jigsaw so that's not really a viable option.

nacx:
This is sensible, as in the end it is all enforced by naming, and this leads to interpretations.

The `com.google.gson.internal` package is explicitly exported since Gson 1.7, and that means that it is *explicitly* made part of the public API in OSGi environments. 

I'm not arguing about the need to hide those classes, but highlighting the fact that if they were internal, they weren't properly tagged as such (at least in OSGi). Someone seeing an explicit export of a package may think that package is OK to be used (why exporting it if not?).

Since it's been there for 15 releases, I'm just asking for a proper deprecation path. From an OSGi perspective, removing an export is like removing an entire package of a public API, and that has a big impact if not advertised properly or a deprecation path is provided.

demobox:
> The com.google.gson.internal package is explicitly exported since Gson 1.7

@JakeWharton @swankjesse Would either of you be in a position to comment as to whether that was intentional or perhaps inadvertent?

From an Apache jclouds perspective (just echoing Ignasi's point here) I think we're in full agreement that using internal classes is not "the right thing to do". The question is whether it's possible to provide a deprecation path that would allow us to use the latest minor whilst preparing for an upgrade to the next major by removing use of those classes.

JakeWharton:
I'm betting someone external contributed the required bits for OSGi and exposed everything. Since none of us use OSGi, it was likely merged without a close enough inspection or understanding as to what the implications were.

demobox:
> I'm betting someone external contributed the required bits for OSGi and exposed everything.

@JakeWharton Ah, OK, thanks for clarifying. Would it be possible to consider restoring those exports in the 2.6.x only, assuming there are plans for another minor release there?

That would leave master "clean" but allow us and potentially other consumers to at least use the latest 2.6.x version - it would be great if we could at least get up to 2.6.

swankjesse:
Can somebody just fix jclouds? Maybe send a PR that copy-pastes the relevant `internal` classes from Gson into that project?

demobox:
> Maybe send a PR that copy-pastes the relevant internal classes from Gson into that project?

This is definitely one of the options being considered - see thread [here](http://markmail.org/message/olgebygfgdy3hwtm). The concern is that that's more of a patch than a real fix, since we obviously would rather use Gson cleanly and depend only on public and recommended classes.

@swankjesse Any input on that thread, and especially any suggestions as to what the right way would be to access the needed functionality, would be very much appreciated.

@nacx is probably in a better position to comment on why we'd like to avoid copying the classes if possible, but that's a discussion perhaps better taken to the thread? Of course, also happy to continue here if preferred.

nacx:
Thanks for all the feedback. I'll just file an issue in jclouds to stop relying on the internal packages and discard this PR.
I don't have enough knowledge of gson to safely replace that code, but I agree this is the right path. Let's work on removing the offending bits from jclouds.

Thanks!

Issue #640, opened 2015-04-21
For an enum, it should be possible to have a designated default catchall.


inder123:
For enum Platform {  IOS, ANDROID, WEB, OTHER }

I should be able to indicate that OTHER is a catch-all for all strings that don't match as specific name.

prcaen:
:+1:

jpbetz:
+1 This would ease the rollouts of a change where a new enum symbol is added and not not all readers are yet aware of it.

pzhangleo:
+1

gmale:
The [linked PR](https://github.com/google/gson/issues/919) suggests adding this functionality via a `GsonBuilder` method. It seems like it would be nicer to leverage annotations, instead. Configuring every single GSON instance with all the types that it may or may not parse does not feel as clean as annotating the model, itself, with meaningful attributes that are "automatically" interpreted–in the same way SerializedName works.

So it feels like `SerializedName` may be the way to go but perhaps a less complex approach would be a good compromise (i.e. use a simple boolean, rather than a regex match):

```java
@SerializedName(value="UNKNOWN", isDefault=true) UNKNOWN
```

inder123:
Good points @gmale 
isDefault is less complicated than the current pattern approach.
Feel free to submit an alternate PR.

nateboxer:
+1

PR #919, opened 2016-08-31
Added SerializedName.pattern to enable regex mapping for an enum valu…

inder123:
…e to a field.

Currently, support for this beyond Enum isn't implemented.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -39,6 +39,7 @@ import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.regex.Pattern;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
@@ -789,6 +790,7 @@ public final class TypeAdapters {
   private static final class EnumTypeAdapter<T extends Enum<T>> extends TypeAdapter<T> {
     private final Map<String, T> nameToConstant = new HashMap<String, T>();
     private final Map<T, String> constantToName = new HashMap<T, String>();
+    private List<Pair<Pattern, T>> patternToConstant;
 
     public EnumTypeAdapter(Class<T> classOfT) {
       try {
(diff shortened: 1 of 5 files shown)


inder123:
Implements https://github.com/google/gson/issues/640

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -800,6 +802,14 @@ public EnumTypeAdapter(Class<T> classOfT) {
             for (String alternate : annotation.alternate()) {
               nameToConstant.put(alternate, constant);
             }
+            String regex = annotation.pattern();
+            if (regex != "") {
```
swankjesse: nit: almost always want `.equals()` on strings. (These will be interned, but it’s still awkward)

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
```diff
@@ -51,4 +53,26 @@ public void testMultipleNamesInTheSameString() {
...
+    assertEquals(ErrorCode.AUTH_FAILURE, target.error1);
+    assertEquals(ErrorCode.UNKNOWN, target.error2);
+    assertEquals(ErrorCode.SERVER_ERROR, target.error3);
+  }
+
+  private enum ErrorCode {
+    AUTH_FAILURE,
+    SERVER_ERROR,
+    /** match all non-whitespace characters. Note that pattern is applied last. */
+    @SerializedName(value="UNKNOWN", pattern="\\S*") UNKNOWN
```
swankjesse: Do you anticipate anyone using patterns for anything other than setting a default?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
inder123: Yes, thought about that. However, this is getting the annotation value, which is defaulted to "", so used ==. 
I will switch to equals()

gson/src/test/java/com/google/gson/functional/SerializedNameTest.java
inder123: Primary use-case is to set default. I can imagine someone using a pattern to map a collection of names to a single field.

Would love to see a more user-friendly proposal.

swankjesse:
```
Gson gson = new GsonBuilder()
    .enumFallback(ErrorCode.UNKNOWN)
    .create();
```

inder123: I presume we will supply the target enum class as well.

How do I do this within the Enum class itself? Either as a JsonAdapter or SerializedName?

swankjesse: We know the target enum ’cause you can call `e.getDeclaringClass()`. To implement this, we’d just install a type adapter like `EnumWithFallbackAdapter`. Ie. this would be sugar for:

```
Gson gson = new GsonBuilder()
    .registerTypeAdapter(ErrorCode.class, new EnumFallbackAdapter(ErrorCode.UNKNOWN))
    .create();

```

Moi0 (commented):
:)

Moi0 (approved):
Boo :)

PR #922, opened 2016-09-05
Added @Documented to Gson annotations.

inder123:
This is useful since it shows the JSON format for a class which the user of that class should typically know.

### DIFF
--- a/gson/src/main/java/com/google/gson/annotations/Expose.java
+++ b/gson/src/main/java/com/google/gson/annotations/Expose.java
@@ -16,6 +16,7 @@
 
 package com.google.gson.annotations;
 
+import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/annotations/JsonAdapter.java
```diff
@@ -87,6 +89,7 @@
  * @author Jesse Wilson
  */
 // Note that the above example is taken from AdaptAnnotationTest.
+@Documented
```
JakeWharton: I don't think it should be on this one. The adapter in use for a class or field is an implementation detail.

JakeWharton:
lgtm

gson/src/main/java/com/google/gson/annotations/JsonAdapter.java
inder123: I was torn on this one. How about Since, Until and Expose? Should we have Documented only for SerializedName?

I am new to Documented, so you can guide me where it is appropriate to be used.

JakeWharton: I'm not a user of those annotations, but they seem relevant to be documented.

inder123:
Removed Documented from JsonAdapter.

inder123:
Merging since got LGTM

PR #936, opened 2016-09-28
Fix a test error in gson/extra 

thaibui:
GraphAdapterBuilderTest fails on master when execute `mvn -e clean test` under json/extras directory; this PR should fix the issue.

### DIFF
--- a/extras/src/test/java/com/google/gson/graph/GraphAdapterBuilderTest.java
+++ b/extras/src/test/java/com/google/gson/graph/GraphAdapterBuilderTest.java
@@ -76,7 +76,7 @@ public final class GraphAdapterBuilderTest extends TestCase {
         .registerOn(gsonBuilder);
     Gson gson = gsonBuilder.create();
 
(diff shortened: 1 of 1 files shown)


thaibui:
I signed it!

swankjesse:
This test case used to pass, but it was regressed with this change:
https://github.com/google/gson/commit/8d5a41329ec2f4db3aed02c20d379baf92840b37

@inder123 thoughts on self-references?

Marcono1234:
Probably obsolete now due to https://github.com/google/gson/pull/1959#discussion_r707676817.

Though at least personally I think it might be worth revisiting the self-reference detection of `ReflectiveTypeAdapterFactory` again at some point. It only detects direct self-references so this only helps for a very specific case. Additionally, if this was only done to support serialization of `Throwable`, then removing the direct self-reference check would not be so problematic because you should not serialize JDK classes using reflection.

eamonnmcmanus:
Closing as obsolete. If we do want to revisit the self-reference detection of `ReflectiveTypeAdapterFactory` then that should be a separate issue.

PR #937, opened 2016-10-04
Json primitive lru cache

speters-f5:
This addresses performance problems found in our message passing system daemon that uses Gson for JSON serialization.  The pattern followed is the same as in the Java VM for string interning or StringPool by using an LRU cache to reduce allocations of the same immutable objects.  

In this particular use case, we are processing multiple concurrent requests with large JSON payloads.  The process was crashing with the exception: “java.lang.OutOfMemoryError: GC overhead limit exceeded”.  We found by use of a profiler that Gson JsonPrimitive object creation was overloading the GC with allocations.   After these changes this use case is successful.   

By adding a small, optionally enabled, LRU Cache we found that for strings we get 100:1 hit to miss ratio resulting in significantly fewer allocations by reusing the same immutable JsonPrimitive objects.  JsonPrimitive of Numbers type are less reused, though our system is strongly weighted towards using strings.

The JsonPrimitive of type Boolean is cached in static objects given there are only two permutations.

Preferably I wanted to make the JsonPrimitive fully immutable but didn’t want to break the backwards compatibility for those not wanting to enable the cache.

The cache:
1.  Can be optionally enabled/disabled
void JsonPrimitive.setCacheEnabled(boolean enableCache)
2.  Has a settable max size for caches for String and Number types
void JsonPrimitive.setCacheSizes(boolean string, boolean number)
3.  Has settable max size of JsonPrimitive of type String to be cached
void setMaxStringCacheSize(int maxStringSize)
4.  Get statistics for hit/miss ratio
CacheStats getCacheStats()

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -70,7 +70,7 @@ final class DefaultDateTypeAdapter implements JsonSerializer<Date>, JsonDeserial
   public JsonElement serialize(Date src, Type typeOfSrc, JsonSerializationContext context) {
     synchronized (localFormat) {
       String dateFormatAsString = enUsFormat.format(src);
-      return new JsonPrimitive(dateFormatAsString);
+      return JsonPrimitive.allocate(dateFormatAsString);
     }
   }
 
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -55,7 +55,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @param bool the boolean that needs to be added to the array.
    */
   public void add(Boolean bool) {
-    elements.add(bool == null ? JsonNull.INSTANCE : new JsonPrimitive(bool));
+    elements.add(bool == null ? JsonNull.INSTANCE :  JsonPrimitive.allocate(bool));
   }
 
   /**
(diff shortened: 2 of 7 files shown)


inder123:
If you are concerned about memory consumption, you should not use Gson to parse into a DOM of JsonElements.

speters-f5:
@inder123 The LRU cache is less about overall memory consumption and more about the fact that small objects are created and destroyed quickly.  The Java Object to JSON String format conversion introduces many small short lived objects.  The creation of these can outpace the garbage collector's ability to clean them up.  Making the JsonPrimitive objects (the key contributor) to use a cache makes it such that the same JsonPrimitive for a String or Number can be reused and just referenced instead of allocated and de-allocated in the heap.  This presumes that your object model has some repetition in it, the cache would not be useful if all your JSON name/values were all unique in every use.  For example, we have a reference field our object model called "selfLink".  When users access our REST API the same cached JsonPrimitive("selfLink") can be referenced over and over again by different requests without adding any allocation / de-allocation beyond the initial one that gets cached.

The trade-offs here are memory consumed by the cache vs CPU usage in allocation / de-allocation.  This is disabled by default (existing behavior), each consumer of Gson would need to make the choice to enable this new feature or not and tune for their program's and environment.

inder123:
Thanks for the details @speters-f5 
In general, we much prefer Pull Requests with no new APIs. Is there a way to do smaller scale version of this PR that interns the repeated objects?

inder123:
Also, note that Gson internally doesn't use JsonElement based DOM anymore for toJson() calls. So, the memory consumption is there only when you explicitly ask Gson to parse to a DOM.

speters-f5:
@inder123 I wasn't aware of the shift to no longer use JsonElements internally when going from Java object to JSON string in toJson().  That will likely make this pull request irrelevant.  Please disregard this pull request for now.  We will need to re-factor our code and profile it with the latest Gson to see if a cache is even still necessary and keep performance where we need it.  Can you confirm that fromJson() also bypasses JsonElement objects now when going from string to Java Object?

Thank you.

inder123:
Yes, fromJson also bypasses JsonElements.

PR #938, opened 2016-10-04
Make GsonBuilder non-final

gagern:
This allows for projects to store common project-specific settings in the constructor of a derived class, which in some scenarios feels more natural than a static method which creates a new builder instance.

Or is there a reason the builder has to be final? Unfortunately the public history doesn't extend that far into the past, so the git log doesn't help answering that question.


inder123:
As a design principle, Gson doesn't support extension through inheritence. The reason is to avoid unforseen API commitments that we will need to continue supporting in future.

You will need to indicate a substantial benefit before we will accept this.

Issue #941, opened 2016-10-07
Get JsonObject Keys?


hessmjr:
Is there specific reason a JsonObject doesn't return a set of keys as Strings such as Java's Map data structure?  Seems like a simple and useful method especially considering the similarities between the two data structures.

hessmjr:
Submitted a PR for this since its fairly simple. #942

hessmjr:
This is done.

PR #942, opened 2016-10-07
Add keySet method and test

hessmjr:
# Overview

Adds keySet method similar to that of Java Map.  Since all Map.entries utilize strings as a key.  Test method also included.


hessmjr:
This addresses the issue I opened, #941 .

inder123:
👍

PR #954, opened 2016-11-02
Allow existing type field name if it serializes to the same value

kaoskorobase:
Instead of failing in `RuntimeTypeAdapterFactory` when there is an existing `type` field name, only fail if its value differs from the computed value. This allows to store the type name in the serialized object's type field.


kaoskorobase:
Signed.

inder123:
Assigned to @swankjesse, the author of this feature.

kaoskorobase:
I need the `type` field to be included in deserialization; AFAIU `transient` excludes the field from both serialization and deserialization, doesn't it?

What makes you think it's not a good idea? I've just noticed that an additional check is required before calling `getAsString()`, I can adapt the PR if needed.

PR #962, opened 2016-11-19
Add a convenience builder method to register a collection of TypeAdapterFactories

ZacSweers:
To use us as an example: We have cases where we hand down a set of factories (via Dagger 2's multibindings) downstream, so this is a nice convenience.

Note that I didn't see where the appropriate place to add tests for this were, as `GsonBuilderTest` seems more focused on its construction semantics rather than verifying that its builder methods behave correctly. Let me know if there's a place where I should add any.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Type;
 import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


inder123:
I don't find it much useful. Build a stronger case please.

hzsweers:
It's a convenience method on a builder. While it might not be personally useful to you, it would be useful to others, and comes at no real cost. Dagger multibindings seem like a fair enough example, it comes in handy if you have multiple modules supplying their own factories into a multibinding.

What usefulness bar do you think it needs to meet your requirements?

inder123:
The cost is bloating of GsonBuilder API, we don't add methods just because it is a minor convenience to some.
Dagger multibinding seems interesting but I have never used something like that (I haven't used Dagger). May be @JakeWharton or @swankjesse can opine?

Issue #957, opened 2016-11-11
GSON release 2.8.0 JAR is not an OSGI bundle


iamrakesh:
Hi,
GSON release 2.8.0 JAR published to/available from maven repository is not an OSGI bundle; MANIFEST file doesn't contain OSGI specific info.

2.8.0 JAR MANIFEST.MF contains

`Manifest-Version: 1.0
Archiver-Version: Plexus Archiver
Built-By: jwilson
Created-By: Apache Maven 3.2.3
Build-Jdk: 1.8.0_92
`

where as 2.7.0 JAR MANIFEST.MF contains

`Manifest-Version: 1.0
Bundle-Description: Gson JSON library
Bundle-SymbolicName: com.google.gson
Archiver-Version: Plexus Archiver
Built-By: inder
Bnd-LastModified: 1465948546904
Bundle-ManifestVersion: 2
Bundle-RequiredExecutionEnvironment: J2SE-1.5, JavaSE-1.6, JavaSE-1.7,
  JavaSE-1.8
Bundle-Vendor: Google Gson Project
Import-Package: com.google.gson.annotations
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Bundle-ContactAddress: https://github.com/google/gson
Tool: Bnd-3.1.0.201512181341
Export-Package: com.google.gson;uses:="com.google.gson.reflect,com.goo
 gle.gson.stream";version="2.7.0",com.google.gson.annotations;version=
 "2.7.0",com.google.gson.reflect;version="2.7.0",com.google.gson.strea
 m;version="2.7.0"
Bundle-Name: Gson
Bundle-Version: 2.7.0
Created-By: 1.8.0_91 (Oracle Corporation)
Build-Jdk: 1.8.0_91
`
Regards,
Rakesh.A

cameronoxley:
+1

roedll:
+1

inder123:
Please send a pull request @iamrakesh

Singleton06:
Is there any reason that the manifest generated through the bnd-maven-plugin[1] isn't getting used in the released JAR?  I'm not overly familiar with all that GSON does for their development practice, but I have noticed that there are certain versions of the gson do not have a manifest at all (such as the above).

[1] - https://github.com/google/gson/blob/master/gson/pom.xml#L34-L47

iamrakesh:
[...]

PR #963, opened 2016-11-20
#957 Added maven-bundle-plugin and changed packaging to 'bundle'.

iamrakesh:
Fix for #957

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -7,6 +7,7 @@
     <version>2.8.1-SNAPSHOT</version>
   </parent>
 
+  <packaging>bundle</packaging>
   <artifactId>gson</artifactId>
   <name>Gson</name>
 
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,13 @@
     </pluginManagement>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.5.3</version>
+        <inherited>true</inherited>
+        <extensions>true</extensions>
+      </plugin>
+      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-release-plugin</artifactId>
         <version>2.4.2</version>


iamrakesh:
I signed it

gson/pom.xml
```diff
@@ -7,6 +7,7 @@
     <version>2.8.1-SNAPSHOT</version>
   </parent>
 
+  <packaging>bundle</packaging>
```
inder123: what is the impact of this change? By default the packaging is Jar. Is the output still Jar? If so, does the size change in anyway?

swankjesse:
When we did this in OkHttp it broke our users.
https://github.com/square/okhttp/issues/2243

gson/pom.xml
jwcarman: It still spits out a jar file, but the MANIFEST.MF file will have more entries in it, making this thing a real OSGi bundle.  So, yes, the size changes, somewhat.

pom.xml
```diff
@@ -82,6 +82,13 @@
     </pluginManagement>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>2.5.3</version>
+        <inherited>true</inherited>
```
jwcarman: If this is inherited, where is it inheriting it from?  You probably want to set up some of the other headers here.  Take a look at what the Savoir parent project does here:

https://github.com/savoirtech/savoir-parent/blob/master/pom.xml

We use this all over the place and it works well for us.  You can customize imports/exports and stuff using the osgi.* properties.

iamrakesh: As per the documentation inherited property allows plugin configuration propagation to child poms.
See - https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Using_the_inherited_Tag_In_Build_Plugins

Regarding .properties file - not much configuration is needed here I think.

iamrakesh:
@swankjesse - Issue you mentioned (ClassNotFoundException for OkHostnameVerifier) is due to usage of class from 'okhttp3.internal.tls' package, which seems internal and was not supposed to be directly used (if I've to go by package name); Same case with gson as well, 'internal' package is not supposed to be directly used, it is not exported by default, and is not available in an OSGI environment.

iamrakesh:
It seems [this](https://github.com/google/gson/commit/08c7bffba0e93611162cadfaf2ee03e3d143e7d1#diff-b870e20e5600e8e051854b3fa6a8a1b5) check-in tries to solve this in a different way.

Though I have one observation on the generated MANIFEST.MF -
'Import-Package' manifest entry with value 'com.google.gson.annotations' is not correct, I think it should be 'Private-Package'.

cristcost:
Hi All,

changing the packaging to bundle may have other side effects on the .class packaged inside the final jar.

I instead suggest that the maven-bundle-plugin is used in it form that does not change the "packaging" like it is described [here](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#adding-osgi-metadata-to-existing-projects-without-changing-the-packaging-type). In synthesis, it is required the manifest goal bound to the process-classes phase, and configure the maven-jar-plugin to package the processed manifest:

```
<plugin>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>  
      <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
    </archive> 
  </configuration>
</plugin>  
<plugin>   
  <groupId>org.apache.felix</groupId>
  <artifactId>maven-bundle-plugin</artifactId>
  <executions>
    <execution>
      <id>bundle-manifest</id>
      <phase>process-classes</phase>
      <goals>    
        <goal>manifest</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

cristcost:
Another note:
I suggest to use a more recent version of the, [maven-bundle-plugin](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.apache.felix%22%20AND%20a%3A%22maven-bundle-plugin%22) 

last release is the one month old 3.3.0.

iamrakesh:
> changing the packaging to bundle may have other side effects on the .class packaged inside the final jar.
Could you please elaborate or provide a link that I can refer ?

cristcost:
Finally, 
the <Export-Package> should be specified explicitly so to export also packages `com.google.gson.internal` and `com.google.gson.internal.bind`, which are excluded by the [maven-bundle-plugin defaults](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#default-behavior)

GSON 2.5 was exporting these packages:
image
(and I used these packages to develop a TypeAdapter that supports polymorphism)

In GSON 2.6 and 2.7 these <Export-Package> are disappeared:
image
(and my TypeAdapter does not work anymore)

cristcost:
Hi @iamrakesh 

> Could you please elaborate or provide a link that I can refer ?

check the [instructions of Export-Package](http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#instructions):

> _...These packages are copied into the resulting bundle JAR file from the available classes (i.e., project classes, dependencies, and class path); thus, it is possible to include classes into your bundle that are not associated with source files in your project..._

with the plugin defaults this **inclusion of classes from dependencies** won't happen, but for this reason, when developing general purpose libraries used also outside of OSGi context, I always prefer to keep the packaging of type "jar" and include OSGi headers as described in the previous message.

JakeWharton:
Excluding internal is intended. You shouldn't have used those types since
they're clearly marked internal.

jwcarman:
I don't think using the bundle packaging will cause issues in this case, especially if you use the plugin's default settings (which work for most cases).  Also, the plugin will *not* (by default) export any packages with "impl" or "internal" in them.  This is by design as these are typically classes not meant for folks to use outside the bundle.

cristcost:
> Excluding internal is intended. You shouldn't have used those types since
they're clearly marked internal.

There is difference between "I should use" and "I can't use": for an open source project, an internal package is discouraged to be used. For an OSGi project it is different.
As a result, my library works with GSON 2.6 and 2.7 everywhere except inside OSGi...

GSON is not designed as bundle and it won't be maintained as such. Also, in the recent history of GSON, OSGi headers have been removed and re-added many times, probably by people unaware of OSGi technology. 

I think a less intrusive approach (packaging of type **jar** AND exporting all packages including **com.google.gson.internal** and  **com.google.gson.internal.bind**) would have avoided these problems and made GSON more stable for OSGi users.

jwcarman:
It's easy to export the other packages, but the packaging type really doesn't make a difference

JakeWharton:
By choosing to use OSGi you are at the mercy of the project owners creating
the bundle, and since it's us and since the internal package isn't for
public use, marking it as not exposed seems like a good thing.

jwcarman:
@JakeWharton totally agree with you.  I'd rather not expose the "internal" stuff.  With OSGi, you have the opportunity to make sure folks don't use the classes they're not supposed to be using.  If it were me, I'd take advantage of that fine-grained control that OSGi provides you.

jwcarman:
@JakeWharton since you and @swankjesse are the two GSON contributors who have commented on this issue, how do you guys propose we move forward?  Would you like to see this PR changed?  Would you like us to provide an alternative approach?  Our team loves GSON and we're an OSGi shop, so we've been stuck with 2.6.2 for a while now.  This prevents us from upgrading to security patches if they come out, so we would really like to see OSGi support come back.

JakeWharton:
I would love to get it fixed for 2.8 so that it just works for OSGi users
and we don't have to think about it. None of us are experts here, we just
know what our intent is. So if this affords you a working library in OSGi
and it represents our intent with 'internal' than I'm very supportive.

iamrakesh:
I too agree that 'internal' package not exported, this also gives flexibility for authors to change implementation when required even if it is kind of API break (in classes/methods of 'internal' package).

> @iamrakesh do you want to modify your PR with these settings so we can get this thing moving forward?

Latest master already has fix, which uses 'maven-jar-plugin', see [this commit](https://github.com/google/gson/commit/08c7bffba0e93611162cadfaf2ee03e3d143e7d1#diff-b870e20e5600e8e051854b3fa6a8a1b5)
'mave-jar-plugin' solution does the required job, any one is fine as long as it works.

inder123:
Since this hasn't been resolved so far, I am just going to close this for now.

PR #964, opened 2016-11-24
Allow deserialization of a Number represented as a String

egor-n:
This works:
```
gson.fromJson("\"15\"", int.class)
```

This doesn't:
```
gson.fromJson("\"15\"", Number.class)
```

This PR makes it so the second case works too.


inder123:
👍

PR #967, opened 2016-11-27
Update RuntimeTypeAdapterFactory.java

waqasmoazzam:
Any other type adapter registered to typeFieldName does not work as its removed.
Using get instead allows any type adapter registered for this field to do its job.
For example register runtime type adapter on a parent class with "type" to be an enum and have a type adapter registered on the enum to convert string value in response to appropriate enum in java class.
It won't work as the field "type" was removed.
Order of registering type adapter and type adapter factory does not help.

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
@@ -199,7 +199,7 @@ public final class RuntimeTypeAdapterFactory<T> implements TypeAdapterFactory {
     return new TypeAdapter<R>() {
       @Override public R read(JsonReader in) throws IOException {
         JsonElement jsonElement = Streams.parse(in);
-        // using get instead of remove lets other typeadater registered for this field in json to work.
+        // using get instead of remove lets other typeadpaters registered for this field in json to work.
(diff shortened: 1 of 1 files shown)


waqasmoazzam:
I Signed CLA.

inder123:
Assiging to @swankjesse since he wrote this. I support this change.

Codex-:
There appears to be a typo in your comment: "typeadater" instead of "typeadapter" :)

xMikeTx:
I think this is a good idea. I would however utilize the builderpattern to configure that:
```
public RuntimeTypeAdapterFactory<T> keepTypeValueInSubclass() {
        keepClassTypeValue = true;
        return this;
    }

and in read()
 JsonElement labelJsonElement = keepClassTypeValue ? jsonElement.getAsJsonObject().get(typeFieldName) : jsonElement.getAsJsonObject().remove(typeFieldName);
```

swankjesse:
I think that if you want this behavior your best bet is to create your own fork of this TypeAdapter. This will break any type adapter that’s configured to fail on unexpected fields.

xMikeTx:
That is why i think that adding an option to enable that behavior for the current approach is better.
Nothing breaks if you do not use the keepValueInSubclass() addon.

Marcono1234:
It looks like this has been superseeded by #1342.

PR #983, opened 2016-12-22
Added method to register a defaultType for parsing

xMikeTx:
The current behavior of the RuntimeTypeAdapterFactory is to fail if a received subtype is not registered for security reasons.
In our project the call must not fail if a new subtype suddenly appears, because the app should still handle all other subtypes.


xMikeTx:
I signed it!

Issue #559, opened 2015-03-19
integer values of array become a float values


GoogleCodeExporter:
```
public class Filter extends Object implements Serializable{

    private static final long serialVersionUID = 4944267946319467028L;  
    SparseArray<List<Integer>> catalogue = new SparseArray<List<String>>();

        ....
}

I'm filling catalogue variable with next values : {0=>{1,2,3,4}, 3=>{20,21,22} 
...}

Then I'm saving Filter object like this:
   ...
   ArrayList<Filter> filterList;
   .....

   public void saveFilter(Filter filterModel){      

      filterList.add(filterModel);

      Gson gson = new Gson();
      String jsonStr = gson.toJson(filterList);

      // For example save as
      preference = activity.getSharedPreferences("SJ_MAIN_STORAGE", Context.MODE_PRIVATE);
      Editor editor = preference.edit();
      editor.putString(key, jsonStr);       
      editor.commit();
   }    

I'm getting Filter object this way :
   public ArrayList<Filter> getFilter{
      Gson gson = new Gson();
      ArrayList<Filter> arrayList = new ArrayList<Filter>();    
      String json = preference.getString(key, "");
      if (hStrings.isEmpty(json)) return arrayList;

      JsonParser parser = new JsonParser();
      JsonArray array = parser.parse(json).getAsJsonArray();

      try{
    for(int i = 0; i < array.size(); i++)
    {
        arrayList.add(gson.fromJson(array.get(i), Filter.class));
    }
      }catch(Exception e){

      }
      return arrayList ;
   }

I have the next result of catalogue value: {0=>{1.0,2.0,3.0,4.0}, 
3=>{20.0,21.0,22.0} ...}

(integer values of array become a float values).
Is it right behavior?

Thank you.
```

Original issue reported on code.google.com by `troshkov...@gmail.com` on 12 Feb 2014 at 3:00

jeffreyzh:
I have the same problem，does anyone fixed  it ？

BK-Choi:
```java
((new Gson()).fromJson("{\"test\":630}", Object.class))
// returns 630.0
```
maybe this is the same problem
why does gson treat an int type json value as if it is an int-frac type?

eamonnmcmanus:
Addressed by #1290.

Issue #525, opened 2015-03-19
number double or integer


GoogleCodeExporter:
```
com.google.gson.internal.bind.ObjectTypeAdapter
line 78;

    case NUMBER:
      double tmp = in.nextDouble();
      if(tmp>Math.floor(tmp))
          return tmp;
      else
          return new Double(tmp).intValue();

```

Original issue reported on code.google.com by `sunju...@gmail.com` on 20 Jun 2013 at 6:59

PR #989, opened 2016-12-29
enhance ObjectTypeBinder to works with number type 'int'

BK-Choi:
- number double or integer #525
- integer values of array become a float values #559

### the reason I pr this:
- Spring's `@RequestBody` parses integer params to double values
- this behavior is not right when you parse HTTP request body parameters. 
  - Web App never knows whether the parameter type is wrong or not
  - `int` should be parsed `int`, not `double`
- [RFC-4627](http://www.ietf.org/rfc/rfc4627.txt) and [JSON](http://json.org) defines type `int`,
so I hope this PR does not violate any rules.

### related issues
- number double or integer #525
- integer values of array become a float values #559

### DIFF
--- a/gson/src/test/java/com/google/gson/ObjectTypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/ObjectTypeAdapterTest.java
@@ -27,9 +27,9 @@ public final class ObjectTypeAdapterTest extends TestCase {
   private final TypeAdapter<Object> adapter = gson.getAdapter(Object.class);
 
   public void testDeserialize() throws Exception {
-    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2,null],\"c\":{\"x\":\"y\"}}");
-    assertEquals(5.0, map.get("a"));
-    assertEquals(Arrays.asList(1.0, 2.0, null), map.get("b"));
+    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2.0,null],\"c\":{\"x\":\"y\"}}");
+    assertEquals(5, map.get("a"));
+    assertEquals(Arrays.asList(1, 2.0, null), map.get("b"));
     assertEquals(Collections.singletonMap("x", "y"), map.get("c"));
     assertEquals(3, map.size());
   }
--- a/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
@@ -76,7 +76,12 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
       return in.nextString();
 
     case NUMBER:
-      return in.nextDouble();
+      String number = in.nextString();
+      if (number.contains(".")) {
+        return new Double(number);
+      } else {
+        return new Integer(number).intValue();
+      }
 
     case BOOLEAN:
       return in.nextBoolean();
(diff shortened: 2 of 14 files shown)


BK-Choi:
CLA signed (individual)

jobh:
A few issues. Most importantly, check for integer overflow (NumberFormatException), and exponent notation ("e" in addition to "."). Maybe use Long instead of Integer?

Now, even Long and Double may overflow, in the latter case losing precision. Also, doubles typically lose precision when converting from decimal, even if not overflowing. I think the best option would be to return BigDecimal instead of Double, and BigInteger instead of Integer. Opinions?

SergeyKoval:
When it will be merged to muster and released?

BK-Choi:
@jobh 
my opinion is,
- integer should be ok with common number use
- Long sounds good to me
- BigDecimal sounds not good
  - waste of memory and time for some people; performance issue
  - maybe what we need is an option for BigDecimal parsing.
- exponential expression
  - oh nice pointing out. thank you.

I can change the "integer" code to "long" one anytime,
however one GSON contributor said GSON is never going to support integer(or long).
(couldn't find the reference, short-term-memory :( )
So, the contributors' affirmitive opinion first, then I can use my time to fix codes.

SergeyKoval:
I believe not only me wait for this fix...

inder123:
@swankjesse  Looks like a reasonable change to me. I plan to merge it in the next few days unless I hear a concern from you.

swankjesse:
Changing the type of a thing is extremely incompatible. If we merge this, nobody will be able to upgrade safely.

gson/src/test/java/com/google/gson/ObjectTypeAdapterTest.java
```diff
@@ -27,9 +27,9 @@
   private final TypeAdapter<Object> adapter = gson.getAdapter(Object.class);
 
   public void testDeserialize() throws Exception {
-    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2,null],\"c\":{\"x\":\"y\"}}");
-    assertEquals(5.0, map.get("a"));
-    assertEquals(Arrays.asList(1.0, 2.0, null), map.get("b"));
+    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2.0,null],\"c\":{\"x\":\"y\"}}");
+    assertEquals(5, map.get("a"));
```
swankjesse: Specifically, this is not at all backwards-compatible.

SergeyKoval:
@swankjesse does it mean that in gson int will be never converted to correct type if it is not specified in end object directly?

swankjesse:
> RFC-4627 and JSON defines type int,
so I hope this PR does not violate any rules.

Though the label `int` is used to describe the number type, an `int` is not a standalone value. That’s `number` and JSON doesn’t offer different types depending on whether a `.0` is in the encoded form.

```
 value = false / null / true / object / array / number / string 
```

Making `.0` significant would be not unlike making whitespace significant.

SergeyKoval:
@swankjesse but with current implementation if we deserialize json and serialize it back we will get different jsons

jobh:
This class looks like it was designed for the purpose - 

[com.google.gson.internal.LazilyParsedNumber](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
)

- it is an abstract number (like JSON), not a specific type
- it roundtrips perfectly, since the string rep is stored
- converts to whatever is needed, even BigDecimal etc

BK-Choi:
@swankjesse 
In GSON world, you said that `0` and `0.0` are the same.
So no need to fix this.
please close this PR (or should I?)

@jobh
thank you!

PR #994, opened 2017-01-11
#957 OSGi headers from the bnd.bnd will be included in the jar

spegelref:
This would ensure that the headers created by the bnd maven plugin where included in the jar (as per the documentation found here https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin#important-note).


inder123:
👍  Thanks

Issue #1005, opened 2017-01-28
Default Value for Missing fields in Json


jainsahab:
I know, We can achieve this by just setting up the field with the expected value, but  people may not always remember that. Can Gson provide us with some mechanism of handling the fields which were not present in JSON?

lyubomyr-shaydariv:
Are there any advantages for such an approach?

jainsahab:
@lyubomyr-shaydariv if your question is why one need such functionality as part of Gson library, then my answer would be, my library should be able to initialize default values in the objects.
Ex: one of the use case I thought, I can keep my field as Optional and gson should be able to initialize it with `Optional.present` or `Optional.absent` based on the node present in JSON string.

If your question is around to change the approach to achieve the same functionality, I am open to suggestions and we can improve upon the approach.

FYI, I created a [pull request](https://github.com/google/gson/pull/1006) for this as well.

lyubomyr-shaydariv:
All you need to cover such a case is a custom type adapter: missing `Optional` fields can be automatically set to `Optional.empty()` without changes in the core. An example inspired by the [interceptors demo](https://github.com/google/gson/tree/master/extras/src/main/java/com/google/gson/interceptors):

```java
final class PostProcessingTypeAdapterFactory
		implements TypeAdapterFactory {

	private final Iterable<? extends Consumer<Object>> postProcessors;

	private PostProcessingTypeAdapterFactory(final Iterable<? extends Consumer<Object>> postProcessors) {
		this.postProcessors = postProcessors;
	}

	static TypeAdapterFactory getPostProcessingTypeAdapterFactory(final Consumer<Object>... postProcessors) {
		return new PostProcessingTypeAdapterFactory(asList(postProcessors.clone()));
	}

	static TypeAdapterFactory getPostProcessingTypeAdapterFactory(final Iterable<? extends Consumer<Object>> postProcessors) {
		return new PostProcessingTypeAdapterFactory(postProcessors);
	}

	@Override
	public <T> TypeAdapter<T> create(final Gson gson, final TypeToken<T> typeToken) {
		final TypeAdapter<T> delegateAdapter = gson.getDelegateAdapter(this, typeToken);
		return new TypeAdapter<T>() {
			@Override
			public void write(final JsonWriter out, final T value)
					throws IOException {
				delegateAdapter.write(out, value);
			}

			@Override
			public T read(final JsonReader in)
					throws IOException {
				final T value = delegateAdapter.read(in);
[...]

PR #1006, opened 2017-01-31
#1005 Gson default value handling

jainsahab:
Making gson capable of populating default values in Java models if the corresponding nodes are missing in Json.


jainsahab:
@inder123 @JakeWharton any feedback on this PR?

drekbour:
Shame that you spent all that time without checking if Google bother with 3rd party submissions.

FireController1847:
Wondering the status of this PR because I could really use this in my project right now

PR #1008, opened 2017-02-03
Update TypeAdapters.java

m-rahimy:
This will made it legal to use char primitive (like 'f') as input, instead of Character.valueOf('f').
It's because we should consider using primitives as long as we can, the Character class is far bigger than a 2 bytes char in Java,

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -391,7 +391,15 @@ public final class TypeAdapters {
       }
       String str = in.nextString();
       if (str.length() != 1) {
-        throw new JsonSyntaxException("Expecting character, got: " + str);
+        // maybe the input string is a number representing a char (like 102 for 'f')
+        try {
+          // so try returning a Character from its value, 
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -391,7 +392,15 @@ public Character read(JsonReader in) throws IOException {
       }
       String str = in.nextString();
       if (str.length() != 1) {
-        throw new JsonSyntaxException("Expecting character, got: " + str);
+        // maybe the input string is a number representing a char (like 102 for 'f')
+        try {
+          // so try returning a Character from its value, 
+          // first make in a real int, because JsonReader.nextString() returns String
+          return Character.valueOf((char) Integer.parseInt(str)); 
+        } catch (IllegalFormatException ex) {
```
Marcono1234, line 400: In case this pull request is still relevant: This should probably catch `NumberFormatException` not `IllegalFormatException`.

I am also not too sure whether the behavior of the default adapter should be changed. Maybe it would be better if users just registered their own adapter which has this behavior, if desired. Though I am not a member of this project and they might think differently about this.

PR #1015, opened 2017-02-15
Memory optimisation in JsonReader

yasirmhd:
The default constructor of StringReader gives a capacity of 16. It expands and does an array copy in the event of the capacity being exceeded while appending characters.

JsonReader.nextQuotedValue was initializing the default constructor. If the entire string was in buffer and there were no characters to be escaped then we can directly create a string object and return it back (based on the profiling for our sample JSON payload this was the flow this code goes into for 90% of the times).

In the event of it going into other code paths trying to estimate a length to be the double of the current length and ensure it is not less than 16.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -1024,7 +1024,6 @@ public class JsonReader implements Closeable {
 
       if(null == builder) {
         int len = (p - start) * 2;
-        initialLength = len;
         builder = new StringBuilder(len < 16 ? 16 : len);
       }
       builder.append(buffer, start, p - start);
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -997,7 +997,7 @@ public class JsonReader implements Closeable {
         if (c == quote) {
           pos = p;
           int len = p - start - 1;
-          if(null == builder) {
+          if (builder == null) {
             return new String(buffer, start, len);
           } else {
             builder.append(buffer, start, len);
@@ -1007,9 +1007,9 @@ public class JsonReader implements Closeable {
           pos = p;
           int len = p - start - 1;
           char escapeChar = readEscapeCharacter();
-          if(null == builder) {
+          if (builder == null) {
             int estimatedLength = (len + pos - p) * 2;
-            builder = new StringBuilder(estimatedLength < 16 ? 16 : estimatedLength);
+            builder = new StringBuilder(Math.max(estimatedLength, 16));
           }
           builder.append(buffer, start, len);
           builder.append(escapeChar);
@@ -1022,9 +1022,9 @@ public class JsonReader implements Closeable {
         }
       }
 
-      if(null == builder) {
-        int len = (p - start) * 2;
-        builder = new StringBuilder(len < 16 ? 16 : len);
+      if (builder == null) {
+        int estimatedLength = (p - start) * 2;
+        builder = new StringBuilder(Math.max(estimatedLength, 16));
       }
       builder.append(buffer, start, p - start);
       pos = p;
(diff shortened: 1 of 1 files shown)


yasirmhd:
I signed it!

inder123:
Looks good to me. But @swankjesse  and @JakeWharton understand the implications better.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -996,12 +996,23 @@ private String nextQuotedValue(char quote) throws IOException {
 
         if (c == quote) {
           pos = p;
-          builder.append(buffer, start, p - start - 1);
-          return builder.toString();
+          int len = p - start - 1;
+          if(null == builder) {
```
JakeWharton: nit: missing space after `if`. We also traditionally put `null` on the right.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -996,12 +996,23 @@ private String nextQuotedValue(char quote) throws IOException {
...
+            builder.append(buffer, start, len);
+            return builder.toString();
+          }
         } else if (c == '\\') {
           pos = p;
-          builder.append(buffer, start, p - start - 1);
-          builder.append(readEscapeCharacter());
+          int len = p - start - 1;
+          char escapeChar = readEscapeCharacter();
+          if(null == builder) {
```
JakeWharton: nit: missing space, order

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -996,12 +996,23 @@ private String nextQuotedValue(char quote) throws IOException {
...
+            int estimatedLength = (len + pos - p) * 2;
+            builder = new StringBuilder(estimatedLength < 16 ? 16 : estimatedLength);
```
JakeWharton: Maybe use `Math.max(estimatedLength, 16)` instead?

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1011,6 +1022,10 @@ private String nextQuotedValue(char quote) throws IOException {
         }
       }
 
+      if(null == builder) {
```
JakeWharton: nit: missing space, order

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1011,6 +1022,10 @@ private String nextQuotedValue(char quote) throws IOException {
...
+        int len = (p - start) * 2;
+        builder = new StringBuilder(len < 16 ? 16 : len);
```
JakeWharton: `Math.max` again

gson/src/main/java/com/google/gson/stream/JsonReader.java
yasirmhd: done.

gson/src/main/java/com/google/gson/stream/JsonReader.java
yasirmhd: done

yasirmhd:
@JakeWharton Thank you for the review. I have made the changes requested

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -996,12 +996,23 @@ private String nextQuotedValue(char quote) throws IOException {
...
+            return builder.toString();
+          }
         } else if (c == '\\') {
           pos = p;
-          builder.append(buffer, start, p - start - 1);
-          builder.append(readEscapeCharacter());
+          int len = p - start - 1;
+          char escapeChar = readEscapeCharacter();
+          if (builder == null) {
+            int estimatedLength = (len + pos - p) * 2;
```
swankjesse: Why pos-p? Isn't it always 1?

yasirmhd: Agreed. Raising a PR with the required change in a bit.

yasirmhd:
@JakeWharton / @swankjesse when is the next release of gson planned? this seems to give us performance gains upto 25%


image

PR #1016, opened 2017-02-17
Incorporating the review comments in #1015

yasirmhd:
Incorporating the review comments in #1015. 

While readEscapeCharacter might read more than one character from the buffer it will also write 1 to the output. So the right thing to do is len+1


yasirmhd:
@swankjesse Thanks for catching this in #1015 . This should fix that issue.

PR #1017, opened 2017-02-18
GsonBuilder Perf optimization

guptasourabh:
In HierarchFactories we are keep inserting the items at first position due to all the pointers in the list have to be moved up one position to before the new one can be inserted into the list.

With collection.reverse it will only run for the one time , rather than keep moving pointer every time whenver a insert is happening.


guptasourabh:
I signed it!

guptasourabh:
@JakeWharton Please review

PR #1018, opened 2017-02-18
JsonPrimitive code correction and Perf changes

guptasourabh:
Changes done -:
1) FieldNamingPolicy -: Validating the condition before starting while loop.
2) GsonBuilder -: Instead of keep adding item at first position in hierarchyList , its more performance efficient to add all the items in last and then reverse at once while adding them all in the factories. As reversing a list is 15-16 times more faster than adding the item at first position and moving all the pointers when ever a new insert happens.
Along with that datePattern.trim.length() is much faster to check as compare to equal which will check if both objects are instanceof String then check length and then iterate over while loop.
3) JsonArray -: While making deepCopy of array , we know the size of the array that needs to be copy.
4) JsonNull = Unnecessary check is added in equal condition , only the check of other instance is instanceof JsonNull is needed.
5) JsonPrimitve -: we are calling getAsString when the object is not a number , which again checks whether the object is number or not which results in redundant checks.
Apart from it getAsString is creating new instance of LazilyParsedNumber which is not needed , if we want to only access the string.

### DIFF
--- a/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
+++ b/gson/src/main/java/com/google/gson/FieldNamingPolicy.java
@@ -140,13 +140,14 @@ public enum FieldNamingPolicy implements FieldNamingStrategy {
     int index = 0;
     char firstCharacter = name.charAt(index);
 
-    while (index < name.length() - 1) {
-      if (Character.isLetter(firstCharacter)) {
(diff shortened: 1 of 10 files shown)


guptasourabh:
@JakeWharton @swankjesse for review

PR #1019, opened 2017-02-18
gson builder perf improvement and code corrections

guptasourabh:
**Changes done** -:
**FieldNamingPolicy** -: Validating the condition before starting while loop.
**GsonBuilder** -: Instead of keep adding item at first position in hierarchyList , its more performance efficient to add all the items in last and then reverse at once while adding them all in the factories. As reversing a list is 15-16 times more faster than adding the item at first position and moving all the pointers when ever a new insert happens.
Along with that datePattern.trim.length() is much faster to check as compare to equal which will check if both objects are instanceof String then check length and then iterate over while loop.
**JsonArray/JsonObject** -: While making deepCopy of array , we know the size of the array that needs to be copy.
**JsonElement** -: No closure of JsonWriter.
**JsonPrimitve** -: getAsString is creating new instance of LazilyParsedNumber which is not needed , if we want to only access the string.


guptasourabh:
@JakeWharton @swankjesse for review

PR #1023, opened 2017-02-23
don't throw JsonSyntaxException where it should be IOException instead

sp-1234:
This is (was) just insane. When underlying stream breaks (due to socket timeout for example), how it's a syntax error?


sp-1234:
>Thanks for your pull request. It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Come on, it's a **SINGLE WORD** change.
If you are so stubborn, just treat it as an issue and do it yourself. Okay?

PR #1026, opened 2017-03-01
Eliminating Code Overhead

anirudhramanan:
* Calculating length and size beforehand in cases of for loops
* Avoid creation of StringBuilder object if type argument's length is 0 ($Gson$Types.java : 485)
* Removed boilerplate null checks from TypeAdapter, and replaced with nullSafe()


anirudhramanan:
I signed it!

JakeWharton:
Thanks!

PR #1029, opened 2017-03-02
Support custom annotation processing

fwonce:
This [discuss](https://groups.google.com/forum/#!topic/google-gson/Y-NoyrqEpgo) and issue #269 talks about the same problem I want to tackle with. And I commented under issue #269 to state my reason. Generally speaking, I'd like to see a feature that let me use custom annotations in my data object freely without exposing Gson stuff there, and change the behaviors of reading/writing the fields with those annotations by Gson.

PS: I'm not so confident with the javadoc I put on added methods. I'll be glad to discuss with the reviewer to make it more comprehensive.

### DIFF
--- a/gson/src/main/java/com/google/gson/FieldAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/FieldAdapterFactory.java
@@ -1,0 +1,32 @@
+package com.google.gson;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+
+import com.google.gson.annotations.JsonAdapter;
+
+/**
+ * Factory for producing {@link TypeAdapter} that only read / write specific fields
+ * with some given annotation.
+ * <p>
+ * Configured through {@link GsonBuilder#registerFieldAdapterFactory}.
+ * <p>
+ * Normally a field needs only one adapter, so it doesn't make sense if both this
+ * and {@link JsonAdapter} are applied with a single field. In that case, this will
+ * be ignored.
+ *
+ * @author Floyd Wan
+ * @since 2.8.1
+ */
+public interface FieldAdapterFactory {
+
+  /**
+   * Create a {@link TypeAdapter} for a field with a special annotation.
+   *
+   * @param annotation the annotation instance on the given {@code field}
+   * @param field the field needs to be investigated
+   * @param <T> the target type of {@link TypeAdapter}
+   * @return a {@link TypeAdapter} to read / write this very field.
+   */
+  <T> TypeAdapter<T> create(Annotation annotation, Field field);
+}
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -22,6 +22,8 @@ import java.io.Reader;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
 import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -135,6 +137,7 @@ public final class Gson {
   private final boolean prettyPrinting;
   private final boolean lenient;
   private final JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory;
+  private final Map<Class<? extends Annotation>, FieldAdapterFactory> fieldAdapterFactories;
 
   /**
    * Constructs a Gson object with default configuration. The default configuration has the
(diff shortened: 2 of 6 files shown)


fwonce:
I signed it!

inder123:
Overall, I like where you are going with this.
A few questions:
1. Why not support annotation on classes as well?
2. Would be great if we can demonstrate some real use cases. For example, taking JPA or JAXB annotations, and automatically using them as SerializedName.

Thanks

fwonce:
Hi @inder123 Thanks for your inspiring reply.
Your first advise makes sense to me so I add a few lines to do this. And with considering this, I've realized that in some cases a `FieldAdapterFactory` should be versatile for both kind of annotation as a best practice (like d722d40).
I'll try to add test cases for real use cases in a couple of days because I do this in spare time:)

inder123:
You can add JPA/JAXB as test dependencies. We will factor them out in a separate project as this progresses. Thank you!

fwonce:
Speaking of JPA/JAXB, I think the most wanted feature is using the name specified by `@Column` (JPA) or `@XmlElement` (JAXB) as the serialized name. So I wrote a test. Before that I had to add the Gson instance as the first argument to [`FieldAdapterFactory#create`](https://github.com/fwonce/gson/blob/field-adapter/gson/src/main/java/com/google/gson/FieldAdapterFactory.java). And then I did something like:

```java
Column column = (Column) annotation;
final name = column.name();
return new TypeAdapter<T>() {
  @Override public void write(JsonWriter out, T value) throws IOException {
    // ... nullcheck ...
    out.name(name); // >>>> Unfortunately this will raise a IllegalStateException <<<<
    TypeAdapter adapter = gson.getAdapter(value.getClass());
    adapter.write(out, value);
  }
  // ... read ...
}
```

The 6th line will raise a `IllegalStateException` because `ReflectiveTypeAdapterFactory.Adapter#write` has already called `out.name()` right before it and probably there is no way to get around unless a major refactor is taken. So I think custom annotation processing can't be used for serialized name related purpose at this point and maybe a custom `FieldNamingStrategy` is more suitable. On the other hand, I wrote some meaningful tests regarding JPA indeed, like ignoring [`@Transient`](http://www.objectdb.com/api/java/jpa/Transient) and [`@Version`](http://www.objectdb.com/api/java/jpa/Version) when serialization. But I'm not sure if that meets your expectation. Or you may give more information about what kind of use cases you would like to see? @inder123 

Thanks
Floyd

inder123:
Great progress. @JakeWharton  @swankjesse  Can you review this as well please?
I think this generic way of binding annotation processing is what we want.

gson/src/main/java/com/google/gson/FieldAdapterFactory.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (C) 2008 Google Inc.
```
swankjesse: Two small problems here. It’s not 2008, and it’s not Google. Please use “Copyright (C) 2017 Gson Authors”.

gson/src/main/java/com/google/gson/FieldAdapterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+ * with some given annotation.
+ * <p>
+ * Configured through {@link GsonBuilder#registerFieldAdapterFactory}.
+ * <p>
+ * Normally a field needs only one adapter, so it doesn't make sense if both this
+ * and {@link JsonAdapter} are applied with a single field. In that case, this will
+ * be ignored.
+ *
+ * @author Floyd Wan
+ * @since 2.8.1
```
swankjesse: unlikely that we’ll release new APIs in a patch release.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -512,6 +518,36 @@ static void checkValidFloatingPoint(double value) {
...
   /**
+   * This method is used to get the {@link TypeAdapter} for the specified field. The type
+   * adapter needs to be configured by {@link GsonBuilder#registerFieldAdapterFactory}.
+   *
+   * @param field the field needs to be investigated
+   * @return the {@link TypeAdapter} for the given {@code field}
+   * @since 2.8.1
+   */
+  public <T> TypeAdapter<T> getFieldAdapter(Field field) {
+    if (fieldAdapterFactories == null || fieldAdapterFactories.isEmpty()) {
```
swankjesse: I do like this shortcut

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -512,6 +518,36 @@ static void checkValidFloatingPoint(double value) {
...
+      return null;
+    }
+    for (Annotation fieldAnnotation : field.getAnnotations()) {
+      Class<? extends Annotation> annotationType = fieldAnnotation.annotationType();
+      if (fieldAdapterFactories.containsKey(annotationType)) {
```
swankjesse: might be more efficient to loop over the factories and call `field.isAnnotationPresent()` rather than looping over the field’s annotations. Only because `isAnnotationPresent()` doesn’t allocate, but `getAnnotations()` needs to build instances of all of the field’s annotations.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -512,6 +518,36 @@ static void checkValidFloatingPoint(double value) {
...
+        FieldAdapterFactory fieldAdapterFactory = fieldAdapterFactories.get(annotationType);
+        return fieldAdapterFactory.create(fieldAnnotation, field);
+      }
+    }
+    // If both this field and its declaring class matched, ignore the latter one
+    for (Annotation classAnnotation : field.getDeclaringClass().getAnnotations()) {
```
swankjesse: precedence rules? ick

fwonce: Removed the 'misleading' comment, but I think the logic here is natural - firtst look at the field annotation, and then the class annotations...

gson/src/main/java/com/google/gson/FieldAdapterFactory.java
fwonce: Sure. I copied the license lines from the bottom of google/gson's readme page, felt weird at first, but didn't think about it again. Maybe you guys should update it too.

gson/src/main/java/com/google/gson/Gson.java
fwonce: Thanks! I'll pay attention to that every time in the future.

fwonce:
Hi @swankjesse First thanks for your review. I have updated according to your inline comments but I'm not sure this is necessary, since you said you are reluctant about this whole idea.

fwonce:
Hi @swankjesse , you haven't made any further comment. I'd like to know if this means there is no chance this get accepted. In fact, your last comment seemed quite ambiguous to me and I didn't even know what exactly your point is. The "style" in which to archive this goal can be discussed, if only new features can be assessed and accepted, w/o assuming anything too powerful is not welcomed to Gson. Do you suggest that Gson is on its process of going into an inactive development phase and the maintenance team is moving onto something new called Moshi (sorry I haven't heard about it before)?

@inder123 What's your attitude?

swankjesse:
I’m reluctant to add large new features to Gson at this time. Large features need careful design and I’m not interested in spending that time to review them.

It’s a popular library and it is very difficult to anticipate the compatibility consequences of any change.

Note that although the project originated at Google, no Google employees are involved at this time.

fwonce:
Got it. For now my only choice is to maintain my own fork. Nevertheless, I have to say I'm quite confident that this PR is backward compatible and solid enough:) In my company, I have applied this version of Gson to many production application for months and it works fine. There are several interesting use cases enabled by this. Hope you guys will have a second look in the future.

By the way, does Moshi support such kind of feature?

inder123:
@swankjesse  I think we have been looking for such a powerful feature for quite some time. This may be the fourth way, but for now, it is the only way to make existing annotations work with Gson. I am highly supportive of this, but am fine with looking at an alternate design.

PR #1032, opened 2017-03-10
RuntimeFieldName: Add annotation and strategy class for runtime field

damienbiggs:
Some users have a use case where they would like to set the serialized name for certain fields at
runtime. In my own company's case, there are custom field names for certain Jira projects.
We needed a way to make these configurable as different teams might have different custom field names.

Testing Done: Added test class to show how serialization and deserialization works.


damienbiggs:
I signed the CLA

eamonnmcmanus:
This seems kind of special-purpose and is something that people can easily build themselves if needed.

PR #1035, opened 2017-03-13
Added a new public method TypeAdapter.name()

### LINKED ISSUE #1011: Enum annotation serialization
`SerializedName` annotations are ignored for enums.
[stack overflow](http://stackoverflow.com/q/42085301/3648107)

inder123:
This new method allows a type adapter writer to customize the JSON name.
Previously, we were using String.valueOf() to translate a key to name.
This caused a problem for using enums as keys (If a SerializedName is set, then it is ignored and toString() value is used instead).
This fixes https://github.com/google/gson/issues/1011

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.google.code.gson</groupId>
     <artifactId>gson-parent</artifactId>
-    <version>2.8.1-SNAPSHOT</version>
+    <version>2.9.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>gson</artifactId>
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -39,7 +39,13 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
   public JsonArray() {
     elements = new ArrayList<JsonElement>();
   }
-  
+
+  /**
+   * Initialize the JsonArray expecting to hold the specified number of items
+   *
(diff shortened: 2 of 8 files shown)


swankjesse (approved):
I’m opposed to this as-is because it’s backwards-incompatible in the encoded data. I’m using Gson extensively across many services and so changes like this are very scary to me. It’s difficult to know if this would break any programs, or whether such breaks would trigger crashes or data corruption.

So suppose we decided to enable this but only with a configuration option to opt-in. In that case, the flag already exists! It’s called `enableComplexMapKeySerialization()` and it shipped in Gson 1.7. Your test case even passes with this option enabled.

```
    gson = new GsonBuilder()
        .enableComplexMapKeySerialization()
        .create();
```

swankjesse (changes requested):
(I’m not in favor of this PR)

eamonnmcmanus:
It might have been better if `Map` instances had been serialized by serializing the key of each entry as a JSON string, then using that string as the key in a JSON object, instead of just calling `String.value` on it. However, making that change now would be massively incompatible. We would certainly need to introduce some sort of configuration option to enable it, and then it's probably almost as easy for users to register their own `TypeAdapterFactory` for `Map` with the exact semantics they want. I'll also note that you would certainly want a special case for `String` because you wouldn't want `Map.of("foo", 23)` to be serialized as `{"\"foo\"":23}`.

As @swankjesse noted in a comment from Ancient Times, the alternative encoding for `Map` is more appropriate for use cases like these. Then you don't have the issue of double encoding and decoding that you get if you try to stuff JSON serialized values into a JSON string.

PR #1037, opened 2017-03-15
elimiating code overhead

guptasourabh:
Changes are -:
1) Calculating length and size beforehand in cases of for loops.
2) index can never be equal to length in FieldNamingPolicy.


guptasourabh:
@inder123 for review

gson/src/main/java/com/google/gson/FieldNamingPolicy.java
```diff
@@ -150,10 +150,6 @@ static String upperCaseFirstLetter(String name) {
       firstCharacter = name.charAt(++index);
     }
 
-    if (index == length) {
```
guptasourabh: it will never get execute as index will always be at max length-1.

guptasourabh:
@JakeWharton for review

PR #1038, opened 2017-03-16
list addition optimization

guptasourabh:
Changes -:
1) Instead of always adding item at 0th position and then shifting the list every time with the new addition, add the item in the last and then reverse the complete list.
2) Correctly allocation of array sizes.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -16,13 +16,6 @@
 
 package com.google.gson;
 
-import com.google.gson.internal.$Gson$Preconditions;
-import com.google.gson.internal.Excluder;
-import com.google.gson.internal.bind.TreeTypeAdapter;
-import com.google.gson.internal.bind.TypeAdapters;
-import com.google.gson.reflect.TypeToken;
-import com.google.gson.stream.JsonReader;
-
 import java.lang.reflect.Type;
 import java.sql.Timestamp;
 import java.text.DateFormat;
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -560,9 +559,10 @@ public GsonBuilder serializeSpecialFloatingPointValues() {
    * @return an instance of Gson configured with the options currently set in this builder
    */
   public Gson create() {
-    List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>();
+    List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>(this.factories.size() + this.hierarchyFactories.size() + 3);
     factories.addAll(this.factories);
     Collections.reverse(factories);
+    Collections.reverse(this.hierarchyFactories);
```
inder123: what's the point of adding this reverse?

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -560,9 +559,10 @@ public GsonBuilder serializeSpecialFloatingPointValues() {
    * @return an instance of Gson configured with the options currently set in this builder
    */
   public Gson create() {
-    List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>();
+    List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>(this.factories.size() + this.hierarchyFactories.size() + 3);
```
inder123: why size() + 3?

gson/src/main/java/com/google/gson/GsonBuilder.java
guptasourabh: @inder123  earlier we are adding items at 0th position instead of that adding item at end and while adding in factories reversing that .

gson/src/main/java/com/google/gson/GsonBuilder.java
guptasourabh: + 3 is for addTypeAdaptersForDate(String datePattern, int dateStyle, int timeStyle,
      List<TypeAdapterFactory> factories) {

inder123:
Thanks for the explanation. 👍

PR #1039, opened 2017-03-17
optimized Redundant isNumber() instance checks.

guptasourabh:
In JsonPrimitive getAsString getting called when its not a number , but inside that also its again checking if it's a number or not which is redundant

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -94,7 +94,8 @@ public final class JsonPrimitive extends JsonElement {
     if (primitive instanceof Character) {
       // convert characters to strings since in JSON, characters are represented as a single
       // character string
-      this.value = String.valueOf(((Character) primitive).charValue());
+      char c = ((Character) primitive).charValue();
+      this.value = String.valueOf(c);
     } else {
       $Gson$Preconditions.checkArgument(primitive instanceof Number
               || isPrimitiveOrString(primitive));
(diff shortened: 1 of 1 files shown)


guptasourabh:
@inder123 for review

gson/src/main/java/com/google/gson/JsonPrimitive.java
```diff
@@ -173,13 +172,16 @@ public boolean isString() {
...
-    }
+    return isNumber() ? getAsNumber().toString() : getAsStringForNonNumber();
+  }
+  
+  /**
+   * convenience method to get this element as a String given the case it is not an instanceof number.
+   *
+   * @return get this element as a String.
+   */
+  public String getAsStringForNonNumber() {
```
inder123: make it private.

inder123:
Overall, I don't know if this really improves anything. Probably the resulting code is a little less readable.

guptasourabh:
@inder123 yes but when we knew its a instanceof Number then calling getAsNumber and then checking if it's instance of String doesn't make sense .

gson/src/main/java/com/google/gson/JsonPrimitive.java
```diff
@@ -94,8 +94,7 @@ void setValue(Object primitive) {
     if (primitive instanceof Character) {
       // convert characters to strings since in JSON, characters are represented as a single
       // character string
-      char c = ((Character) primitive).charValue();
-      this.value = String.valueOf(c);
+      this.value = String.valueOf(((Character) primitive).charValue());
```
JakeWharton: Revert?

swankjesse:
Yeah, I don’t think this is an improvement.

PR #1042, opened 2017-03-20
Simplified access of getSimpleName

guptasourabh:
instead of calling getClass.getSimpleName() that will check too many conditions inside , we can make it as final String and use it directly.

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -37,11 +37,12 @@ import com.google.gson.internal.bind.util.ISO8601Utils;
 final class DefaultDateTypeAdapter implements JsonSerializer<Date>, JsonDeserializer<Date> {
 
   // TODO: migrate to streaming adapter
-
+  
+  private static final String SIMPLE_NAME = "DefaultDateTypeAdapter";
+  
   private final DateFormat enUsFormat;
   private final DateFormat localFormat;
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
```diff
@@ -40,6 +40,7 @@
 
   private final DateFormat enUsFormat;
   private final DateFormat localFormat;
+  private final String SIMPLE_NAME = "DefaultDateTypeAdapter";
```
JakeWharton: static

guptasourabh:
@inder123 for review

gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
```diff
@@ -40,6 +40,7 @@
 
   private final DateFormat enUsFormat;
   private final DateFormat localFormat;
+  private final static String SIMPLE_NAME = "DefaultDateTypeAdapter";
```
JakeWharton: Static fields go at the top above instance fields.

guptasourabh:
@JakeWharton requested changes done

inder123:
👍

gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
```diff
@@ -111,7 +113,7 @@ private Date deserializeToDate(JsonElement json) {
   @Override
   public String toString() {
```
inder123: toString() is mainly used for debugging.

PR #1043, opened 2017-03-21
Optimized no of iterations for copying values.

guptasourabh:
For copying the values we are iterating 3 times over the stackSize with the use of arrayCopy individually for stack,patchIndices and pathNames instead of that we can run a single for loop that will iterate till stackSize and copy all the values of stack,pathIndixes and pathNames in new array.


swankjesse:
Please don’t contribute micro-optimizations without motivating benchmarks. In this case it might be a pessimization because arraycopy is intrinsic.

PR #1055, opened 2017-03-27
Added failing test for RuntimeTypeAdapterFactoryTest

mgerhardy:
Provided a few testcases that shows where the RuntimeTypeAdapterFactory fails for Collections.

**testRuntimeTypeAdapterCollectionSerializationAndDeserialization**: doesn't put the type into the json
**testRuntimeTypeAdapterCollectionDeserializationViaTypeToken**: works fine because TypeToken is used.
**testRuntimeTypeAdapterCollectionDeserializationViaCollectionClass**: doesn't work because the class of the Collection is used.

### DIFF
--- a/extras/src/test/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactoryTest.java
+++ b/extras/src/test/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactoryTest.java
@@ -20,9 +20,61 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonParseException;
 import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import junit.framework.TestCase;
 
 public final class RuntimeTypeAdapterFactoryTest extends TestCase {
+
+  public void testRuntimeTypeAdapterCollectionSerializationAndDeserialization() {
+    final RuntimeTypeAdapterFactory<BillingInstrument> rta = RuntimeTypeAdapterFactory.of(
+        BillingInstrument.class)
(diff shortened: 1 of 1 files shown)


mgerhardy:
My problem is that I need to deserialize Collections at runtime - thus I can't use the new TypeToken<List<MyType>>() {}.getType() syntax. I only have a Class instance. Do I have to register some special type of Collection adapter or JsonDeserializer? I tried to add a deserializer but this wasn't possible because I lack needed types for fromJson calls.

`	private class RunTimeTypeCollectionAdapter implements JsonDeserializer<Collection> {
		@Override
		@SuppressWarnings({ "rawtypes", "unchecked" })
		public Collection deserialize(final JsonElement json, final Type typeOfT, final JsonDeserializationContext context)
				throws JsonParseException {
			if (json == null) {
				return null;
			}
			final TypeToken<?> typeToken = TypeToken.get(typeOfT);
			try {
				Class<?> collectionClass = typeToken.getRawType();
				if (collectionClass == List.class) {
					collectionClass = ArrayList.class;
				} else if (collectionClass == Set.class) {
					collectionClass = HashSet.class;
				} else if (collectionClass == Map.class) {
					collectionClass = HashMap.class;
				}
				final Collection collection = (Collection) Class.forName(collectionClass.getTypeName()).newInstance();
				final Class<?> rawType = typeToken.getRawType();
				if (json.isJsonArray()) {
					final JsonArray array = json.getAsJsonArray();
					for (int i = 0; i < array.size(); ++i) {
						final JsonElement element = array.get(i);
						final Object deserialize = context.deserialize(element, rawType);
						collection.add(deserialize);
					}
				} else {
					// final Object deserialize = context.deserialize(json,
					// rawType);
					// collection.add(deserialize);
				}
				return collection;
			} catch (final Exception e) {
				logger.error(e.getMessage(), e);
			}
			return null;
		}
	}`

PR #1056, opened 2017-03-27
Add constructor to UtcDateTypeAdapter

greg-pendlebury:
…  to allow setting of the boolean 'millis' flag that was present in the format() method.


greg-pendlebury:
Oh well, sorry, but I can't agree to the CLA: "Grant of Copyright License."

The apache license it seriously not enough?

greg-pendlebury:
You should probably also check the source code for that class. I think there's clear references there to code lifted from the Jackson libraries which may not comply with the way that CLA is written.

JakeWharton:
Yes, Apache is not enough.

Issue #1063, opened 2017-04-12
Gson static initialisation takes close to 27 ms on a Nexus 5 device


yasirmhd:
![image](https://cloud.githubusercontent.com/assets/13655724/24973102/3209df38-1fdb-11e7-98cf-3f1310a334f1.png)

This is because of the static initialisation of **NULL_KEY_SURROGATE** in Gson class. 
Given this is quite expensive and this is required only when someone passes in a null for the type parameter (which would be very rare, we should look at lazy initialisation of this variable)

PR #1064, opened 2017-04-12
Performance fix to do lazy initialization of null key surrogate

yasirmhd:
Fixes issue #1063

image

This is because of the static initialisation of NULL_KEY_SURROGATE in Gson class.
Given this is quite expensive and this is required only when someone passes in a null for the type parameter (which would be very rare, we should look at lazy initialisation of this variable)

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -394,6 +394,7 @@ public final class Gson {
       if(null == sNullKeySurrogate) {
         sNullKeySurrogate = new TypeToken<Object>() {};
       }
+      return sNullKeySurrogate;
     }
   }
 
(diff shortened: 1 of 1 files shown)


swankjesse:
I suspect a measurement error. What’s the latency to parse a value before and after this change?

yasirmhd:
@swankjesse This is the cost that was being paid by the app just to create an instance of the Gson

yasirmhd:
btw, this was on a 4.4 Nexus 5 device

yasirmhd:
The complete call stack : 

image

JakeWharton:
This should use `TypeToken.get(Object.class)`. There's no need for the subclass and reflection hacks.

yasirmhd:
@JakeWharton : Agreed. Will make the changes in a bit.

yasirmhd:
Thanks @JakeWharton . It will be great if we get a SNAPSHOT version of the library with the changes done for this as well as #1015.

JakeWharton:
I don't have the bits to push a SNAPSHOT. Perhaps @swankjesse can.

PR #1070, opened 2017-04-21
Migrate DefaultDateTypeAdapter to streaming adapter

lyubomyr-shaydariv:
* Migrated `DefaultDateTypeAdapter` to streaming adapter, removing one item from the in-source TODO list.
* `DefaultDateTypeAdapter` does not introduce its own type parameter like `<T extends Date>` in order to keep compatiblity.
* `DefaultDateTypeAdapter` now requires a `Class<? extends Date>` to be passed to the primary constructor: all package-protected constructors always accept the `dateType` parameter; however, there was one public constructor accepting no `dateType`, so now there are two constructors and another one accepts `dataType` in order not to break public interface.
* `dateType` default value is `java.util.Date.class`.
* `dateType` must be one of: `java.util.Date.class`, `java.sql.Timestamp.class`, or `java.sql.Date.class`. Otherwise a fail-fast `IllegalArgumentException` that is migrated from the former `deserialize` method (`AssertionError` is thrown instead) and now thrown in the new constructor. I hope this is acceptable however I cannot recall fail-fast constructors in Gson (factory methods only).

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -16,7 +16,7 @@
 
 package com.google.gson;
 
-import java.lang.reflect.Type;
+import java.io.IOException;
 import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.text.ParseException;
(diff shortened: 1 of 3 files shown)


lyubomyr-shaydariv:
I signed it. Please verify.

inder123:
Thank you! 👍

Issue #1084, opened 2017-05-23
Gson doesn't deserialise Long numbers correctly


danieleguiducci:
Gson shouldn't cast a number to a Double if a number does not have decimal digits. It's clearly wrong.
```java
public class GsonVsJackson {


    public static void main(String[] args) throws Exception {
        testGson();
        System.out.println("========================");
        testJackson();
    }

    public static void testGson() {
        System.out.println("Testing Gson 2.8");
        Gson gson = new Gson();
        HashMap<String, Object> newTest = new HashMap<>();
        newTest.put("first", 6906764140092371368L);
        String jsonString = gson.toJson(newTest);
        System.out.println(jsonString); // output ok: {"first":6906764140092371368}
        Map<String, Object> mapFromJson = gson.fromJson(jsonString, Map.class);
        Number numberFromJson = (Number) mapFromJson.get("first");
        System.out.println(numberFromJson.getClass() + " = " + numberFromJson); // java.lang.Double val 6.9067641400923709E18
        long longVal = numberFromJson.longValue();
        System.out.println(longVal); // output rounded: 6906764140092370944
    }

    public static void testJackson() throws Exception {
        System.out.println("Testing Jackson");
        ObjectMapper jackson = new ObjectMapper();
        HashMap<String, Object> newTest = new HashMap<>();
        newTest.put("first", 6906764140092371368L);
        String jsonString = jackson.writeValueAsString(newTest);
        System.out.println(jsonString); // output ok: {"first":6906764140092371368}
        Map<String, Object> mapFromJson = jackson.readValue(jsonString, Map.class);
        Number numberFromJson = (Number) mapFromJson.get("first");
        System.out.println(numberFromJson.getClass() + " = " + numberFromJson); // java.math.BigInteger = 6906764140092371368
        long longVal = numberFromJson.longValue();
        System.out.println(longVal); // output OK: 6906764140092371368
    }

}
```

Kind Regards,
Daniele

lyubomyr-shaydariv:
It's not a bug, and gson does it fully legit: JSON does not distinguish between integers or floats, and does not care the size of numbers. Numerics in JSON are just numbers, and `java.lang.Double` is the best and largest primitive-counterpart candidate to hold a JSON number.

If you need a long value at a call-site, then just call its `longValue()` method. If, for any particular reason, you need a behavior you are talking about, then you have to implement a custom type adapter factory. Say, something like this (not sure if it's implemented right, though):

```java
[...]

PR #1092, opened 2017-05-31
Deserialize JSON non-fp literals to Long by default

amogilev:
See issue #1084.

Before this fix, ObjectTypeAdapter deserialized both fp and non-fp
literals to Double. In case of very big numbers, it caused undesired
rounding, as Double cannot fit large integers without loosing precision.

From now on, a JSON integer literal (without a decimal point) is mapped
to Long if it fits into Long ranges. The JSON literals with decimal
point, and out-of-range integer literals are still mapped to Double.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
@@ -203,6 +203,24 @@ public final class JsonTreeReader extends JsonReader {
     }
   }
 
+  @Override public Number nextNumber() throws IOException {
+    JsonToken token = peek();
+    if (token != JsonToken.NUMBER && token != JsonToken.STRING) {
+      throw new IllegalStateException(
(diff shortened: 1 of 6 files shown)


lyubomyr-shaydariv:
Doesn't it break backwards compatibility?

amogilev:
In practice, there are two changes which may affect backward compatibility:

1) Some numbers de-serialized as Objects (i.e. by ObjectTypeAdapter) may change the resulting type from Double to Long. That is exactly what fixes the issue, so I guess it is fine. Although toString() may change (e.g. from "5.0" to "5"), I cannot imagine a reason to compare this other than in Unit tests. 

2) Numbers de-serialized as Number (by TypeAdapters.NUMBER) change from LazilyParsedNumber to Long or Double. This part is optional, but made for consistency. And I guess nobody would really cast to LazilyParsedNumber in their code.

swankjesse:
This is not backwards compatible and I’m not willing to burden Gson’s userbase with the costs of this change.

inder123:
Sorry, we can't break backward compatibility

Issue #232, opened 2015-03-19
Read and write Json properties using methods (ie. getters & setters)


GoogleCodeExporter:
```
if your object has an int getId(); method. Maybe you want to have a 

@Field("id")
public int getId();

When the Json is generated the getId() method is called and the result is 
placed in "id"
I know that I could have a int field in my class, but, in my case the Id is 
generate by Hibernate once the object is saved.

```

Original issue reported on code.google.com by `Germanattanasio` on 7 Sep 2010 at 2:04

jansohn:
This would be very helpful when working with JavaFX properties, e.g. `javafx.beans.property.StringProperty`, as these properties cannot be instantiated with an empty constructor.

schlm3:
GSon is useless for me without this feature. Wanted to use it for client server communication together with a GWT app while using a common Data interface. The point is, that the server already has it's very own object model and likes only to provide a view for the json serialisation which should look like the data required inside the GWT app. Would have worked perfectly if gson had supported getter/setter like serialisation.

rajsrivastav1919:
I have the same requirement, I want Gson to use setters for setting the field instead of reflection.
I am deserializing a thrift object's json (which does not contain isSet vector details). If setter methods were called, isSet vector will be formed automatically.
(It can be argued that isSet vector should have been present in the json in the first place. Agreed, but it is an internal field for use by thrift and thrift takes the pain of maintaining it every time some fields are edited.)

With current Gson implementation, thrift objects can't be deserialized "effectively". Other JSON & XML deserializing libraries (like Jackson, CastorXML) call setters, and hence the objects formed are compatible with thrift. For now, I am using Gson by making changes in it for calling setter methods instead of using reflection.

From the thread above, I see that you don't want to give native support in Gson for this feature.
My suggestion is to add a flag in GsonBuilder which indicates Gson to use getter/setter methods. This will be set to false by default so the current functionality does not change. If set to true, Gson will try to call getter/setter methods if present, otherwise fallback to reflection for getting/setting the values. This will make Gson compatible to be used for Thrift objects also.

For 99% POJO classes current implementation works I agree. But, having as an optional feature shouldn't harm anyone I guess.

inder123:
@rajsrivastav1919 Would be great for someone to fork Gson with support for this. As we understand the solution better, we will consider adding it.

rajsrivastav1919:
Thanks @inder123 for the consideration. I will try to point out some genuine use cases for this if I find any.

[...]

PR #1094, opened 2017-05-31
Adding support for using getter/setter methods based on user choice #232

rajsrivastav1919:
This PR contains changes to use getter/setter methods instead of field level reflection methods for serialization/deserialization of objects, if the user has specified this preference while instantiating Gson.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -108,6 +108,7 @@ public final class Gson {
   static final boolean DEFAULT_SERIALIZE_NULLS = false;
   static final boolean DEFAULT_COMPLEX_MAP_KEYS = false;
   static final boolean DEFAULT_SPECIALIZE_FLOAT_VALUES = false;
+  static final boolean DEFAULT_USE_GETTER_SETTER = false;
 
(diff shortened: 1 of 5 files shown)


rajsrivastav1919:
I signed it!

thiagotom:
Hello @rajsrivastav1919 any chance this will be incorporated into a new gson release? It would be great to have this feature. I'm already using 2.8.5 and it doesn't look like there is a way to force it to go through getters and setters.

rajsrivastav1919:
@thiagotom The community was not in favour of adding this support officially in Gson library as per the last discussion. You can comment on the actual issue's thread to discuss this further with them.

mrydeen:
Thank you Raj for making these changes.  We were originally looking at Gson for use in our company, but since Gson works at the member level, we could not easily overwrite enumerations/non Gson capable classes.  Yes we could have written our own deserializers, but that would have been painful for all our objects.  We have existing objects that were used when we were converting from XML and we wanted to use the same objects, just different source.  The objects are a combination of primative types and classes.  And to make it works, the XML to JSON is not a one to one mapping.  But by being able to use the setters and getters, we can create extending classes which can set the correct values in the parents.

teodor-mysko:
Is there any update on the decision to merge this fix?

Issue #1098, opened 2017-06-06
Gson 2.8.1 DefaultDateTypeAdapter is not null safe.


NightlyNexus:
```java
public void testNullSafe() throws IOException {
  TypeAdapter<Date> date =
    new GsonBuilder().setDateFormat("yyyy-MM-dd HH:mm:ss Z").create().getAdapter(Date.class);
  assertNull(date.fromJson("null"));
}
```
fails.
The recent change to the streaming adapter unfortunately broke some backwards compatibility for some of my projects.

JakeWharton:
Dupe of #1096

Issue #1096, opened 2017-06-05
2.8.1 can't serialize and deserialize date null (2.8.0 works fine)


xanscale:
```
W/System.err: com.google.gson.JsonParseException: The date should be a string value
W/System.err:     at com.google.gson.DefaultDateTypeAdapter.read(DefaultDateTypeAdapter.java:96)
W/System.err:     at com.google.gson.DefaultDateTypeAdapter.read(DefaultDateTypeAdapter.java:40)
W/System.err:     at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:129)
W/System.err:     at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:220)
W/System.err:     at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:41)
W/System.err:     at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:82)
W/System.err:     at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:61)
```

https://github.com/google/gson/blob/b8f616c939c652b8540c95fa2b377b8c628ef3ff/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java#L95

this will trows JsonParseException
```
    if (in.peek() != JsonToken.STRING) {
      throw new JsonParseException("The date should be a string value");
    }
```

this is my conf
`new GsonBuilder().serializeNulls().setDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS").create()`

inder123:
Can you create a PR with a failing test? Thanks.

tylerOrMe:
Having similar issue in serialization.
```
Gson gson = new GsonBuilder().setDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").create()
gson.toJson(myObjectWhoContainsNullDates)
```
throws `NullPointerException` from Gson 2.8.1
2.8.0 works fine.

SergeyTrukhachev:
Faced the same issue with 2.8.1.
2.8.0 works well.

bmarty:
Hi,

This is my workaround for this bug, waiting for 2.8.2:

[...]

PR #1099, opened 2017-06-07
Add failing DefaultDateTypeAdapterTest case.

NightlyNexus:
referencing #1096 and #1098.
I'm not sure where the correct place to fix the problem is. In GsonBuilder or in the DefaultDateTypeAdapter itself?


lyubomyr-shaydariv:
>I'm not sure where the correct place to fix the problem is. In GsonBuilder or in the DefaultDateTypeAdapter itself?

It looks like the `DefaultDateTypeAdapter` itself.  All internal type adapters in `com.google.gson.internal.bind.TypeAdapters` use the nullability check. However, the nullability case was not tested in older pre-stream tests, therefore I missed the nullability point in b8f616c939c652b8540c95fa2b377b8c628ef3ff.

NightlyNexus:
Great! Thank you!

PR #1105, opened 2017-06-23
Fix well formed exception

StefMa:
The previous exception was styled in the following way:
... ["DATE']...
This commit will fix the different quote marks and replace the " with '


inder123:
👍

PR #1119, opened 2017-07-18
ProtoTypeAdapter uses wrong case format to find repeated field's generic type via reflection

orischwartz:
`fieldNameSerializationFormat` is a converter from protobuf case format -> json case format.

When deserializing json -> protobuf, `ProtoTypeAdapter` uses reflection to determine the Java field name of a repeated protobuf field (to obtain its generic type).

The Java protobuf implementation always uses lower camel case for its field names, but `ProtoTypeAdapter` was using the caller-specified proto->json converter. 

Before this PR, if the JSON case format didn't produce the same name as lower camel, deserialization would fail.

This PR:
* Removes the `Converter` and instead stores both proto and json `CaseFormat`s so the json `CaseFormat` can be referenced.
* Fixes the field name translation used in reflection so it always converts to lower camel.
* Adds deserialization and serialization unit tests on a protobuf with a repeated field name that exposes the error.

### DIFF
--- a/proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
+++ b/proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
@@ -19,7 +19,6 @@ package com.google.gson.protobuf;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.CaseFormat;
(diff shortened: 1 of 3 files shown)


orischwartz:
Anything I can do to help get this merged/reviewed? Thanks!

inder123:
👍

PR #1128, opened 2017-07-31
Fix StackOverflowError on resolving types with TypeVariable recursion

amogilev:
Sample failing code:
  private static class TestType<X> {
    TestType<? super X> superType;
  }
  ...
  new Gson().getAdapter(TestType.class);


inder123:
👍

inder123:
Thank you for a test that demonstrates the problem!

PR #1130, opened 2017-08-01
support hidden fields, fix IllegalArgumentException "... declares multiple JSON fields named ..."

amogilev:
Before this fix, an attempt to serialize a class with a hidden field
(e.g. class C2 defined as below:
  class C1 { String f; }
  class C2 extends C1 { String f; }
)
failed with IllegalArgumentException with a message like "C2 declares multiple JSON fields named f".

This is fixed by mangling the hidden fields with a "^i" suffix, where i is the number of the field-declaring class in the chain of super-classes, e.g. "f^1" stands for the hidden field "f" declared in the first super-class of the class being serialized.

So, JSON for C2 now may look like {"f":"c2Field", "f^1":"c1Field"}


inder123:
This seems like a non-standard convention. We wouldn't want to support it unless it was accepted practice in some JSON standard.

PR #1141, opened 2017-08-14
Make GsonBuilder.create() factory order idempotent.

naturalwarren:
Calling `GsonBuilder.create()` multiple times results in the builder's underlying hierarchy factory list being reversed.

This change copies hierarchy factories to a new local list when `GsonBuilder.create()` is called so subsequently calls don't result in a change in factory ordering.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -562,8 +562,12 @@ public final class GsonBuilder {
     List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>(this.factories.size() + this.hierarchyFactories.size() + 3);
     factories.addAll(this.factories);
     Collections.reverse(factories);
-    Collections.reverse(this.hierarchyFactories);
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -562,8 +562,12 @@ public Gson create() {
     List<TypeAdapterFactory> factories = new ArrayList<TypeAdapterFactory>(this.factories.size() + this.hierarchyFactories.size() + 3);
     factories.addAll(this.factories);
     Collections.reverse(factories);
-    Collections.reverse(this.hierarchyFactories);
-    factories.addAll(this.hierarchyFactories);
+
+    List<TypeAdapterFactory> hierarchyFactories = new ArrayList<TypeAdapterFactory>(this.hierarchyFactories.size());
+    hierarchyFactories.addAll(this.hierarchyFactories);
```
JakeWharton: Can you pass this to the constructor directly?

naturalwarren: Good idea. a0edf14c

Issue #1139, opened 2017-08-10
newBuilder() API


naturalwarren:
For the pour souls that aren't on Moshi would a PR be welcomed that adds a `newBuilder` API to `Gson`?

inder123:
Sure, good idea.

kaqqao:
Shouldn't this be closed since #1142 was merged and released already?

PR #1142, opened 2017-08-14
Add newBuilder() API

naturalwarren:
Addresses #1139 

`Gson.newBuilder()` returns `GsonBuilder`

Few changes here:
- Increases visibility of `Gson` member variables so they can be seen by `GsonBuilder`
- Adds additional member variables to `Gson` to save state required to transform `Gson` back to `GsonBuilder`

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -105,6 +105,31 @@ public final class GsonBuilder {
   }
 
   /**
+   * Constructs a GsonBuilder instance from a Gson instance. The newly constructed GsonBuilder
+   * has the same configuration as the previously built Gson instance.
+   *
+   * @param gson the gson instance whose configuration should by applied to a new GsonBuilder.
+   */
+  public GsonBuilder(Gson gson) {
+    this.excluder = gson.excluder;
+    this.fieldNamingPolicy = gson.fieldNamingStrategy;
+    this.instanceCreators.putAll(gson.instanceCreators);
+    this.serializeNulls = gson.serializeNulls;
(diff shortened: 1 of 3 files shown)


naturalwarren:
@inder123 What do you think about this?

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -105,6 +105,31 @@ public GsonBuilder() {
   }
 
   /**
+   * Constructs a GsonBuilder instance from a Gson instance. The newly constructed GsonBuilder
+   * has the same configuration as the previously built Gson instance.
+   *
+   * @param gson the gson instance whose configuration should by applied to a new GsonBuilder.
+   */
+  public GsonBuilder(Gson gson) {
```
NightlyNexus: probably a redundant API with newBuilder(). make this package-private?

naturalwarren: Good idea. Will do.

naturalwarren: 8058ab76

naturalwarren:
This built fine locally, CI is having trouble:

```
0.01s$ mvn -f gson install -DskipTests=true
Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java
The command "mvn -f gson install -DskipTests=true" failed and exited with 1 during .
```

inder123:
@JakeWharton  This looks interesting. I am inclined to accept, what do you think?

JakeWharton:
I'm fine with it, yes.

inder123:
👍

PR #1143, opened 2017-08-16
Enum with abstract methods

jithinrajtnr:
##  Fix on creating TypeAdapter for Enum having abstract methods.

For enums having abstract methods, the `Cass.getClass()` will not return the Class object corresponding to the `enum `constant. Instead of it will return a runtime `Enum `class of the enum variable. This fails when creating custom `TypeAdapter `for that enum. So the Type of the enum can be obtained by `Enum.getDeclaringClass()` call.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -573,12 +573,21 @@ public final class Gson {
    * {@link #toJson(Object, Type)} instead. If you want to write out the object to a
    * {@link Writer}, use {@link #toJson(Object, Appendable)} instead.
    *
+   * For enums having  abstract methods , {@link Class#getClass()} will not return the Class
+   * object corresponding to the enum constant. Instead of it will return a runtime Enum class
+   * of the enum variable. This fails when creating custom {@link TypeAdapter} for that enum.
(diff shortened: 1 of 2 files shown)


jithinrajtnr:
Hi, I signed it!

jithinrajtnr:
Okay, I committed the changes from my secondary machine. Now I added and verified the email in the commits to GitHub. is it okay ?

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -573,13 +573,22 @@ public JsonElement toJsonTree(Object src, Type typeOfSrc) {
    * {@link #toJson(Object, Type)} instead. If you want to write out the object to a
    * {@link Writer}, use {@link #toJson(Object, Appendable)} instead.
    *
+   * For enums having  abstract methods , {@link Class#getClass()} will not return the Class
```
inder123: This should be a comment inside the method. It is not a javadoc comment since Gson user can't do anything with this info.

Marcono1234:
@jithinrajtnr, could you please address the comment above? Additionally it might be good to also apply this change for `Gson.toJsonTree(Object)`.

Issue #1107, opened 2017-06-26
Gson deserializes wildcards to LinkedHashMap


egor-n:
This issue is a successor to #1101.

Models:
```java
// ? extends causes the issue
class BigClass { Map<String, ? extends List<SmallClass>> inBig; }

class SmallClass { }
```

Json:
```json
{
  "inBig": {
    "key": [
      { "inSmall": "hello" }
    ]
  }
}
```

Gson call:
```java
SmallClass small = new Gson().fromJson(json, BigClass.class).inBig.get("inSmall").get(0);
```

This call will fail with a `ClassCastException` exception –
 `com.google.gson.internal.LinkedTreeMap cannot be cast to Entry`. If we remove `? extends` then everything works fine.

Marcono1234:
Could you please test if this still occurs with the latest Gson version? I think #1146 should have fixed this. At least the originally provided sample code seems to work, except that the member name seems to be incorrect: <code>fromJson(json, BigClass.class).inBig.get("<del>inSmall</del><ins>key</ins>").get(0)</code>

If you still experience this please provide a small self contained example, and mention which Gson version you are using.

mhdsuhail172:
Hi, <br>

Im using Gson 2.10.1. I think the issue is still present. I was trying to write a class that can
parse json array from a json file stored in the android assets folder.

The following is the class for the parser. This takes in a Generic Type.

``` 
class AssetJsonParser<T>  {  
    companion object{  
        private const val TAG = "AssetJsonParser"  
  }  
  
    fun getDataAsList(context: Context, filePath : String): List<T> {  
        lateinit var jsonString : String  
        try {  
            jsonString = context.assets.open(filePath).bufferedReader().use {  
 it.readText()  
            }  
  }catch (ioException: IOException){  
            Log.d(TAG, "parse: failed due to : ${ioException.message}" )  
        }  
        val listOfData = object : TypeToken<List<T>>() {}.type  
  return Gson().fromJson(jsonString,listOfData)  
    }  
}

```

I was expecting the Gson parser to return a list of the generic type (here TypeA or TypeB).

I was expecting this test case to pass but I see that the type returned is actually a
LinkedTreeMap type and not what is expected (TypeA or TypeB). Is it an issue with my
way of using the Gson library or is it same as the issue #1107 mentioned here ?

```
@Test  
fun typeTest() {  
    // Context of the app under test.  
  val appContext = InstrumentationRegistry.getInstrumentation().targetContext  
  
[...]

PR #1146, opened 2017-08-19
fix issue #1107: resolve element type in wildcard collection types

amogilev:
This change fixes of detecting element types by method $Gson$Types.getCollectionElementType() when the generic collection type is declared using wildcards. For example, for the collection type "? extends List&lt;Something&gt;" it shall extract "Something" as the element type. 

This change fixes issue #1107.


inder123:
👍

PR #1147, opened 2017-08-27
workaround for enums corrupted by ProGuard

amogilev:
Default ProGuard configuration removes valueOf() and values() from the
processed enum classes. As a result, type.getEnumConstants() == null for
such classes, and NullPointerException was thrown from the constructor
of EnumTypeAdapter.

As an example of the corrupted enum classes, see Predicates.ObjectPredicates
from jersey-guava (up to 2.25.1 so far).
Also see https://www.guardsquare.com/en/proguard/manual/examples#enumerations

No unit tests are provided now, but you can reproduce the exception using
code like this:
  import jersey.repackaged.com.google.common.base.Predicates;
  ...
  new Gson().toJson(Predicates.alwaysTrue());

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -814,6 +814,10 @@ public final class TypeAdapters {
       if (!rawType.isEnum()) {
         rawType = rawType.getSuperclass(); // handle anonymous subclasses
       }
+      if (rawType.getEnumConstants() == null) {
+        // invalid enum class, probably corrupted by ProGuard (e.g. with missing values() method)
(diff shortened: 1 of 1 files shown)


Marcono1234:
Would it maybe be better to throw a descriptive exception? With this proposed change it would fall back to the reflection based approach which is likely not going to help in this situation and might make debugging difficult.

Or is there a use case where you do not want any exception at all and an improperly (de-)serialized enum is better?

Marcono1234:
Also similar to #1495 maybe a more reliable solution would be to go through the declared fields and filter by `Field.isEnumConstant()`.

Marcono1234:
Looks like this has been superseded by #1495, thank you nonetheless for the pull request!

PR #1169, opened 2017-10-11
Clarify handling of non-object omitted fields

elharo:
fix ##1166

@inder123 While deserializing, a missing entry in JSON results in setting the corresponding field in the object to its default value: null for object types, zero for numeric types, and false for booleans.

Also added some missing periods and corrected one word.


elharo:
Looks like there's a flaky test. Someone with commit rights may need to restart it.

inder123:
Thank you!

PR #1178, opened 2017-10-24
Fixes #485

### LINKED ISSUE #485: Let JsonElement implement Cloneable and Serializable
```
the gson API is neither Serializable nor Cloneable. I'd imagine that both 
features should be relatively easy to implement...?
```

Original issue reported on code.google.com by `lukas.eder@gmail.com` on 8 Nov 2012 at 3:51

mvysny:
The `JsonElement` class now implements Serializable; I've added tests for all four implementors so that the serialization works. A special hack is added to `JsonNull` so that the deserialization preserves its singleton property; a workaround for `JsonObject` has been added, since `LinkedTreeMap` serializes to `LinkedHashMap` and using the default serialization causes `java.lang.ClassCastException: cannot assign instance of java.util.LinkedHashMap to field com.google.gson.JsonObject.members of type com.google.gson.internal.LinkedTreeMap in instance of com.google.gson.JsonObject`

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonElement.java
+++ b/gson/src/main/java/com/google/gson/JsonElement.java
@@ -19,6 +19,7 @@ package com.google.gson;
 import com.google.gson.internal.Streams;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
+import java.io.Serializable;
 import java.io.StringWriter;
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -30,7 +31,7 @@ import java.math.BigInteger;
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
-public abstract class JsonElement {
+public abstract class JsonElement implements Serializable {
   /**
    * Returns a deep copy of this element. Immutable elements like primitives
    * and nulls are not copied.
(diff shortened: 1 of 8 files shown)


mvysny:
I signed the CLA.

inder123:
Thanks for the PR. 
Can you explain the use-case please? Why serialize when you can just use JSON?

mvysny:
Sure. We plan to store the JsonElement tree in the web session, which requires serializable objects. Vaadin needs to do that, to track changes done in the client web component. Currently we're using Json structure from gwt jar and using simple in-house object-to-json but it would be great to replace that with something more solid.

Please let me know if this is clear or whether further explanation is needed.

philip-searle:
@mvysny has essentially the same use-case we have: we need to store JsonElement objects as part of a larger structure in conversation/session scope. Converting the JsonElement to a JSON string defeats our objective of having an easily manipulated in-memory representation of our JSON objects -- the (short-lived) serialization is required to integrate with the existing server lifecycle. We can work around the lack of Serializable support by writing our own writeObject/readObject methods on the enclosing classes, but it would be much easier if they supported serialization without any additional work.

motinis:
Is there any reason not to have this merged?

PR #1179, opened 2017-10-26
Java 9 module

michaelsavich:
I've added a module-info.java file. Right now it's disabled by default-- I'm no maven wizard, so I didn't want to mess with the pom more than absolutely necessary. To try compiling the GSON as a jigsaw module, delete `<excludes>**/module-info.java</excludes>` and set java.version to 1.9, then run mvn compile.

### DIFF
--- a/gson/src/main/java/com.google.gson/module-info.java
+++ b/gson/src/main/java/com.google.gson/module-info.java
@@ -1,0 +1,8 @@
+module com.google.gson {
+	exports com.google.gson;
+	exports com.google.gson.annotations;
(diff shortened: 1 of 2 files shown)


michaelsavich:
Hey, can someone check what's going on with Travis CI? It seems like everybody's builds are failing…

michaelsavich:
I signed it!

Foorack:
Can Travish be re-run? It looks like the error is with the CI system (JAVA_HOME not set) rather than with this pull request.

PR #1186, opened 2017-11-01
Enables distinction between Double and Integer when auto parsing

yaziza:


JakeWharton:
This isn't a backwards-compatible change and will break user code. And since it's entirely untested (and likely breaks existing tests) we cannot merge this.

Marcono1234:
In addition to the comment above, this might also have been superseded by #1290.

PR #1198, opened 2017-11-30
Optimised the create() call for Excluder TypeAdapterFactory

anirudhramanan:
excludeClass used to get called twice, which internally used to check if class is a innerClass, or an anonymous or local class. Attaching the CPU Profiler snapshot.

screen shot 2017-11-30 at 2 51 22 pm

With this change, we only check once and return if the class has to be excluded.


anirudhramanan:
Closing this, breaking test cases. Will raise a new one

PR #1199, opened 2017-11-30
Optimised the create() call for Excluder TypeAdapterFactory

anirudhramanan:
excludeClass used to get called twice, which internally used to check if class is a innerClass, or an anonymous or local class. Attaching the CPU Profiler snapshot.

screen shot 2017-11-30 at 2 51 22 pm

With this change, we only check once and return if the class has to be excluded.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -287,6 +287,17 @@ public final class GsonBuilder {
   }
 
   /**
+   * Configures Gson to exclude anonymous and local classes during serialization.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.8.2
+   */
+  public GsonBuilder disableAnonymousAndLocalClassSerialization() {
+    excluder = excluder.disableAnonymousAndLocalClassSerialization();
+    return this;
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


yasirmhd:
@inder123 kindly review this. 

This has given us good gains as the number of times this gets called is high.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -286,6 +286,17 @@ public GsonBuilder disableInnerClassSerialization() {
     return this;
   }
 
+  /**
+   * Configures Gson to exclude anonymous and local classes during serialization.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.8.2
+   */
+  public GsonBuilder disableAnonymousAndLocalClassSerialization() {
```
SimonVT: Afaik this is already disallowed by design.
If you're trying to add it as a feature it should have its own pull request, since it's a completely separate change.

yasirmhd: @SimonVT , thanks for your review. I agree with you that we can break this into two PRs.

I agree that this variable being inverse. @anirudhramanan let us fix this.

anirudhramanan: @SimonVT thanks for pointing it out. It has been fixed

anirudhramanan:
@inder123 Shall I break this into 2 separate PR's ?

PR #1209, opened 2017-12-19
JsonArray now can return the List of JsonElement

ayarullin:
I think it may be very convenient: for example it will be very helpful, if we need to sort this JsonArray with custom comparator (Using Collections.sort())

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -372,6 +372,15 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
     throw new IllegalStateException();
   }
 
+  /**
+   * Returns this array as List of JsonElement
+   *
+   * @return this array as List of JsonElement
+     */
+  public List<JsonElement> getAsList() {
+    return elements;
+  }
+
   @Override
   public boolean equals(Object o) {
     return (o == this) || (o instanceof JsonArray && ((JsonArray) o).elements.equals(elements));
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -373,11 +373,9 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
   }
 
   /**
-   * Returns this array as List of JsonElement
-   *
-   * @return this array as List of JsonElement
-     */
-  public List<JsonElement> getAsList() {
+   * @return this array as {@link List} of {@link JsonElement}
+   */
+  public List<JsonElement> asList() {
     return elements;
   }
 


ayarullin:
I signed it!

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -372,6 +372,15 @@ public boolean getAsBoolean() {
     throw new IllegalStateException();
   }
 
+  /**
+   * Returns this array as List of JsonElement
+   *
+   * @return this array as List of JsonElement
+     */
+  public List<JsonElement> getAsList() {
```
NightlyNexus: how about `asList`()?

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -372,6 +372,15 @@ public boolean getAsBoolean() {
     throw new IllegalStateException();
   }
 
+  /**
+   * Returns this array as List of JsonElement
+   *
+   * @return this array as List of JsonElement
```
NightlyNexus: redundant doc.

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -372,6 +372,15 @@ public boolean getAsBoolean() {
...
+     */
+  public List<JsonElement> getAsList() {
+    return elements;
```
NightlyNexus: `return Collections.unmodifiableList(new ArrayList<>(elements));`

ayarullin: Hm... Why? The idea of this metod is modify JsonArray by modifying List (for example for sorting).

> That said, it seems like an asList does belong here, since this already implements Iterable.
We cannot convert Iterable to List without going over all elements. This is not optimal way.

NightlyNexus: ah, i misunderstood the use case.

ayarullin:
Hey guys?
I commited fixes, is it ok?

lyubomyr-shaydariv:
What if the `JsonArraỳ could implement `List<JsonElement>` rather just being an iterable?

lyubomyr-shaydariv:
By the way, there's 3b7bc45212ffe7bb7d62ede1bbe11552c95910d6 at the `JsonArrayImplementsList` branch but it's not merged to `master` for some reason. I think that it breaks `JsonArray.add` return type, but in general it also looks like unfinished work. For example, `JsonObject` might implement `Map<String, JsonElement>` as well?

ayarullin:
@lyubomyr-shaydariv 
No, this is because is breaks binary compatibility - look #683 
By the way, JsonObject has entrySet method which returns the set of elements in the object.
Maybe returning the Map<String, JsonElement> may be redundant.
What do you think @NightlyNexus

lyubomyr-shaydariv:
@ayarullin Oh, I see. Thanks.

Marcono1234:
Note that directly exposing the underlying ArrayList allows inserting Java `null` values which should not be possible (see also #1771). So maybe this pull request should either change the the list implementation of `JsonArray.elements` (which would always decrease performance), or better have `asList()` wrap the underlying list in a list which prevents `null` values.

Marcono1234:
This has been superseded by #2225. Thank you nonetheless for effort on this!

Issue #1210, opened 2017-12-20
Java9 DateFormat changes


amogilev:
The default formats used in DateFormat has changes in Java 9. For example:

- Java8 (and below): "Jan 1, 1970 12:00:00 AM"
- Java9: "Jan 1, 1970, 12:00:00 AM"
- Java8 (and below): "Thursday, January 1, 1970 12:00:00 AM UTC"
- Java9: "Thursday, January 1, 1970 at 12:00:00 AM Coordinated Universal Time"

As a result, a number of Gson unit tests currently fail when run on Java 9. Additionally, this causes some sort of a backward incompatibility: a date written by Gson in Java 8 cannot be parsed back in Java 9. 

In order to fix it, I suggest the following changes:
- provide a way for Gson to read default "old" formats. As Gson by default uses only US date formats, it is possible;
- change "write" unit tests so that they check the results depending on the current Java version. If run on Java 9 or later, expect the new formats to be used; otherwise, expect the "old" results

PR #1211, opened 2017-12-20
fix Java9 DateFormat changes

amogilev:
Adopts Gson for Java9 DateFormat changes, as described in issue #1210

### DIFF
--- a/gson/src/main/java/com/google/gson/util/VersionUtils.java
+++ b/gson/src/main/java/com/google/gson/util/VersionUtils.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2017 The Gson authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson.util;
+
+/**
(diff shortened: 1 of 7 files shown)


gson/src/main/java/com/google/gson/util/VersionUtils.java
```diff
@@ -0,0 +1,49 @@
...
+   * @return the major Java version, i.e. '8' for Java 1.8, '9' for Java 9 etc.
+   */
+  public static int getMajorJavaVersion() {
+    return majorJavaVersion;
+  }
+
+  /**
+   * @return {@code true} if the application is running on Java 9 or later; and {@code false} otherwise.
+   */
+  public static boolean isJava9OrLater() {
```
inder123: Does this logic work on Android?

amogilev: Good question! For Android, "java.version" is "0", so the method getMajorJavaVersion() returns "0", and isJava9OrLater() returns "false". Guess it is fine until Android is actually moved to Java 9 (if it will ever happen).

jansohn:
When do you plan to release a new version which includes this fix?

gson/src/main/java/com/google/gson/util/VersionUtils.java
seventhmoon: https://developer.android.com/about/versions/13/features#core-libraries

> Android 13 starts the work of refreshing Android's core libraries to align with the OpenJDK 11 LTS release with both library updates and Java 11 language support for application and platform developers. The core library changes introduced in Android 13 will also be available to Android 12 devices through a Google Play system update to the ART Mainline Module.

Android is moving to newer Java implementations

Marcono1234: @seventhmoon, thanks for pointing that out. Would you mind creating a separate issue so that we can investigate there if / which effect this will have on Gson and its `VersionUtils.isJava9OrLater()` method?

shahirshann:
Vh

PR #1213, opened 2017-12-26
Fixes #1188,Update ReflectiveTypeAdapterFactory.java

### LINKED ISSUE #1188: Problem while parsing List(or array) as String
I need to parse a list as String form. Why is it giving an exception while parsing a list as a simple string?
While an int or a float variable is easily accessible as String.

carson2440:
Single attribute parsing errors do not affect other attributes。


carson2440:
I signed it!

carson2440:
I signed it!

Marcono1234:
I am not sure if changing this is a good idea because (if I see that correctly) it would completely disable error reporting.

eamonnmcmanus:
This seems very unfinished in any case. If the @carson2440 is still interested, perhaps they can clean the PR up a bit and explain why they think it is a good idea.

Issue #1216, opened 2017-12-29
JSON-serialization causes "Illegal reflective access" warning on JDK 9


emulov:
Hi,

Since upgrading to JDK 9 (Gson v2.8.1), we keep getting warning messages like the following when doing JSON-serialization using Gson: 

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.gson.internal.ConstructorConstructor (./lib/gson-2.8.1.jar) to constructor java.text.NumberFormat()
WARNING: Please consider reporting this to the maintainers of com.google.gson.internal.ConstructorConstructor
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

It seems that with the introduction of modularization in Java 9, reflective access is only permitted to classes within the same module. However, Gson apparently accesses classes in the module java.base (in this example java.text.NumberFormat) which causes above warning message (by default, it is only displayed once on the first illegal reflective access). Since illegal access of this form will be prohibited in future releases (probably in Java 10), and because of this annoying warning message, this should be fixed as soon as possible.

Thanks!

inder123:
Can you propose a solution, or may be send a PR? Thanks.

JakeWharton:
Don't serialize types in java.*. Write a custom type adapter which converts
them to JSON primitives rather than serializing internals of the JDK which
are not guaranteed across versions.

inder123:
@JakeWharton Do we do that somewhere in Gson? If so, we should fix that.

JakeWharton:
Gson allows users to include types from java.* in their models and Gson
happily serializes their internal contents. In this case, NumberFormat,
which should never be in your model without a custom serializer to convert
it to some other representation (like a String).

inder123:
I see. One solution will be to explicitly add type-adapters for each of the java.* classes in Gson. 
But we will only be able to do that for classes up to JDK 6 (since we can't include classes for JDK 7+ given JDK 6 being the minimally supported version).

Your suggestion to @emulov is right: don't serialize java.* classes without writing an explicit type-adapter.

@emulov if you do write type adapters for java.* classes (especially JDK 9), feel free to contribute them back in the extras package for the benefit of others.

JakeWharton:
I would rather see an option to prevent the serialization. Sadly we'll
never be able to turn it on by default.

amogilev:
[...]

PR #1218, opened 2018-01-01
Java 9 support: use Unsafe-based reflection in Java 9+

amogilev:
Fixes "illegal reflective access" warnings and exceptions, like one in #1216


gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -40,13 +40,16 @@
...
+import com.google.gson.reflect.ReflectionAccessUtils;
+import com.google.gson.reflect.ReflectionAccessor;
 import com.google.gson.reflect.TypeToken;
 
 /**
  * Returns a function that can construct an instance of a requested type.
  */
 public final class ConstructorConstructor {
   private final Map<Type, InstanceCreator<?>> instanceCreators;
+  private final static ReflectionAccessor accessor = ReflectionAccessUtils.getReflectionAccessor();
```
inder123: This doesn't need to be static. There is just one instance of ConstructorConstructor in Gson

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -49,6 +51,7 @@
   private final FieldNamingStrategy fieldNamingPolicy;
   private final Excluder excluder;
   private final JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory;
+  private final static ReflectionAccessor accessor = ReflectionAccessUtils.getReflectionAccessor();
```
inder123: Again doesn't need to be static

gson/src/main/java/com/google/gson/reflect/ReflectionAccessUtils.java
```diff
@@ -0,0 +1,61 @@
...
+ * appeared in Java 9, like {@link java.lang.reflect.InaccessibleObjectException} thrown or
+ * warnings like
+ * <pre>
+ *   WARNING: An illegal reflective access operation has occurred
+ *   WARNING: Illegal reflective access by ...
+ * </pre>
+ * <p/>
+ * Works both for Java 9 and earlier Java versions.
+ */
+public class ReflectionAccessUtils {
```
inder123: Can you delete this class, and move this code to a getInstance() method under ReflectionAccessor?

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package com.google.gson.reflect;
+
+import java.lang.reflect.AccessibleObject;
+
+/**
+ * Provides a replacement for {@link AccessibleObject#setAccessible(boolean)}, useful when that basic operation is
+ * prohibited, e.g. throws {@link java.lang.reflect.InaccessibleObjectException} in Java 9.
+ */
+public interface ReflectionAccessor {
```
inder123: This can be made an abstract class, no need of interface

gson/src/main/java/com/google/gson/reflect/impl/PreJava9ReflectionAccessor.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.lang.reflect.AccessibleObject;
+
+/**
+ * A basic implementation of {@link ReflectionAccessor} which is suitable for Java 8 and below.
+ * <p>
+ * This implementation just calls {@link AccessibleObject#setAccessible(boolean) setAccessible(true)}, which worked
+ * fine before Java 9.
+ */
+public class PreJava9ReflectionAccessor implements ReflectionAccessor {
```
inder123: make final?

gson/src/main/java/com/google/gson/reflect/impl/UnsafeReflectionAccessor.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Field;
+
+/**
+ * An implementation of {@link ReflectionAccessor} based on {@link Unsafe}.
+ * <p>
+ * NOTE: This implementation is designed for Java 9. Although it should work with earlier Java releases, it is better to
+ * use {@link PreJava9ReflectionAccessor} for them.
+ */
+public class UnsafeReflectionAccessor implements ReflectionAccessor {
```
inder123: make final

gson/src/main/java/com/google/gson/reflect/impl/UnsafeReflectionAccessor.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final Unsafe theUnsafe = getUnsafeInstance();
```
inder123: remove static?

inder123:
Thanks for the PR

JakeWharton:
All of these classes should be in `internal`. They're not public API.

amogilev:
@inder123 Thanks for the review, I have changed the code according to your comments.
@JakeWharton Do you mean that the code shall be moved to the package 'com.google.gson.internal.reflect'? Guess it makes sense...

gson/src/main/java/com/google/gson/reflect/package-info.java
```diff
@@ -1,5 +1,6 @@
 /**
- * This package provides utility classes for finding type information for generic types.
+ * This package provides utility classes for finding type information for generic types,
+ * as well as  some utilities for working with Reflection API.
```
liach: two spaces between `as` and `some`

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
```diff
@@ -15,18 +15,52 @@
...
+   * In such a case, use {@link ReflectionAccessor#makeAccessible(AccessibleObject)} on a field, method or constructor
+   * (instead of basic {@link AccessibleObject#setAccessible(boolean)}).
+   */
+  public static ReflectionAccessor getInstance() {
+    return ReflectionAccessorHolder.instance;
+  }
+
+  // singleton holder
+  private static class ReflectionAccessorHolder {
+    private static final ReflectionAccessor instance = createReflectionAccessor();
```
inder123: can you get rid of ReflectionAccessorHolder class, and just place the instance in ReflectionAccessor?

gson/src/main/java/com/google/gson/reflect/impl/PreJava9ReflectionAccessor.java
```diff
@@ -25,7 +25,7 @@
  * This implementation just calls {@link AccessibleObject#setAccessible(boolean) setAccessible(true)}, which worked
  * fine before Java 9.
  */
-public class PreJava9ReflectionAccessor implements ReflectionAccessor {
+public final class PreJava9ReflectionAccessor extends ReflectionAccessor {
```
inder123: This class can be package private

gson/src/main/java/com/google/gson/reflect/impl/UnsafeReflectionAccessor.java
```diff
@@ -27,7 +27,7 @@
  * NOTE: This implementation is designed for Java 9. Although it should work with earlier Java releases, it is better to
  * use {@link PreJava9ReflectionAccessor} for them.
  */
-public class UnsafeReflectionAccessor implements ReflectionAccessor {
+public final class UnsafeReflectionAccessor extends ReflectionAccessor {
```
inder123: package private?

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
liach: If you get the holder, the instance field should not be private because of synthetic method (jvm does not allow such access so that java will generate extra methods to this field)

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
```diff
@@ -0,0 +1,66 @@
...
+ * avoid reflective access issues appeared in Java 9, like {@link java.lang.reflect.InaccessibleObjectException}
+ * thrown or warnings like
+ * <pre>
+ *   WARNING: An illegal reflective access operation has occurred
+ *   WARNING: Illegal reflective access by ...
+ * </pre>
+ * <p/>
+ * Works both for Java 9 and earlier Java versions.
+ */
+public abstract class ReflectionAccessor {
```
liach: package-private constructor please

gson/src/main/java/com/google/gson/reflect/impl/PreJava9ReflectionAccessor.java
```diff
@@ -0,0 +1,37 @@
...
+ * <p>
+ * This implementation just calls {@link AccessibleObject#setAccessible(boolean) setAccessible(true)}, which worked
+ * fine before Java 9.
+ */
+public final class PreJava9ReflectionAccessor extends ReflectionAccessor {
+
+  /**
+   * {@inheritDoc}
+   */
+  public void makeAccessible(AccessibleObject ao) {
```
liach: `@Override` annotation

gson/src/main/java/com/google/gson/reflect/impl/UnsafeReflectionAccessor.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public final class UnsafeReflectionAccessor extends ReflectionAccessor {
+
+  private static final Unsafe theUnsafe = getUnsafeInstance();
+  private static final Field overrideField = getOverrideField();
+
+  /**
+   * {@inheritDoc}
+   */
+  public void makeAccessible(AccessibleObject ao) {
```
liach: `@Override` annotation

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
amogilev: Why adding a constructor here? The class is abstract, so the default (public) constructor still cannot be used by anyone except of implementing classes

gson/src/main/java/com/google/gson/reflect/impl/UnsafeReflectionAccessor.java
amogilev: It implies that 'reflect.impl' classes need to be moved up to the 'reflect' package. Seems fine after moving to a separate 'internal' package though...

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
JakeWharton: Because it prevents subclasses outside the package.

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
amogilev: @inder123 You are right, there is no need of laziness for this singleton, so the holder can be removed

gson/src/main/java/com/google/gson/internal/reflect/ReflectionAccessor.java
```diff
@@ -0,0 +1,62 @@
...
+   * You may need one a reflective operation in your code throws {@link java.lang.reflect.InaccessibleObjectException}.
+   * In such a case, use {@link ReflectionAccessor#makeAccessible(AccessibleObject)} on a field, method or constructor
+   * (instead of basic {@link AccessibleObject#setAccessible(boolean)}).
+   */
+  public static ReflectionAccessor getInstance() {
+    return instance;
+  }
+
+  private static ReflectionAccessor createReflectionAccessor() {
+    if (VersionUtils.getMajorJavaVersion() < 9) {
```
inder123: Could be inlined as 
```ReflectionAccessor instance = VersionUtils.getMajorJavaVersion() < 9 ? new PreJava9ReflectionAccessor() : new UnsafeReflectionAccessor();```

File: gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java [hunk 16]
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * An implementation of {@link ReflectionAccessor} based on {@link Unsafe}.
+ * <p>
+ * NOTE: This implementation is designed for Java 9. Although it should work with earlier Java releases, it is better to
+ * use {@link PreJava9ReflectionAccessor} for them.
+ */
+final class UnsafeReflectionAccessor extends ReflectionAccessor {
+
+  private static final Unsafe theUnsafe = getUnsafeInstance();
```
inder123: do these need to be static? Please make them non-static

inder123:
Thanks for patiently incorporating all the feedback

gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java
amogilev: I am not sure why you prefer instance fields over static ones, but sure, these may be change to non-static. What about static methods, like getUnsafeInstance()? If needed, they may be changed to non-static as well.

gson/src/main/java/com/google/gson/reflect/ReflectionAccessor.java
amogilev: Yes, but why shall we prevent such subclasses? I am not against it, but just do not understand...

gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java
inder123: static fields will get initialized when class is loaded which is an unnecessary penalty in memory and loading time.

Static methods are fine (and slightly preferred by me).

swankjesse:
I think this might be the wrong solution to the problem. Unfortunately, the right solution to this problem is a lot of work.

Instead of using a bigger, more powerful weapon to read & write platform types, I think Gson should have a new java-platform-typeadapters module.

Wherever Gson’s users are relying on reflection to read+write a platform type like java.util.UUID, we should handwrite a type adapter. That way we avoid getting into an arms race with the JDK maintainers who really don’t want their fields to be reflected upon.

swankjesse:
... anyone who wants to use Gson with Java 9 to read+write platform types would need this module. And the Gson-using community would have to built the module to cover as many types as is reasonable.

JakeWharton:
Strongly agree https://github.com/google/gson/issues/1216#issuecomment-354657771. The module system is the opportunity to enforce something which never should have been allowed!

inder123:
@swankjesse Agree with you. We can start on adding some of these type adapters, and build upon it gradually.

PR #1221, opened 2018-01-03
Added method to include anonymous and local classes in excluder

anirudhramanan:


inder123:
How about some tests that show that this works

anirudhramanan:
@inder123 sure, will add a test case for the same.

anirudhramanan:
@inder123 Added the test cases for both Local and Anonymous Class Exclusion. Please review

PR #1222, opened 2018-01-06
Added a check for use of type variable to TypeToken constructor.

JornVernee:
A pull request that addresses my earlier issue #1219.

This adds a check to the protected `TypeToken` constructor that will now throw a runtime exception if the `TypeToken` uses a type variable. So that this design time error produces a more clear error earlier during execution.

### DIFF
--- a/gson/src/main/java/com/google/gson/reflect/TypeToken.java
+++ b/gson/src/main/java/com/google/gson/reflect/TypeToken.java
@@ -16,14 +16,15 @@
 
 package com.google.gson.reflect;
 
-import com.google.gson.internal.$Gson$Types;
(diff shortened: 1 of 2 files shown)


JornVernee:
I've used a `RuntimeException`  because `getSuperclassTypeParameter` does the same for missing type parameters. So I think that issue can be ignored.

I've also made the accessibility of the `usesTypeVariable` method the same as `getSuperclassTypeParameter`. That could also be made private not sure what to do there.

I also haven't added any javadoc yet. I see you are not using javadoc annotations, so I'm not entirely sure  what to do there...

JornVernee:
Closing old PR

PR #1229, opened 2018-01-16
Allow JsonValue writer without String usage

d-fournier:
The "jsonValue" API was created to write a raw JSON content into the output writer.
However, the API depends on String and the goal of this MR would be to remove String dependency. (To allow zeroisation on specific values)

Another way to proceed would be to use a "CharSequence" instead of a String but the writer would convert it to a String later.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -438,6 +438,23 @@ public class JsonWriter implements Closeable, Flushable {
   }
 
   /**
+   * Writes {@code value} directly to the writer without quoting or
+   * escaping.
+   *
+   * @param value the char array value to write, or null to encode a null literal.
+   * @return this writer.
+   */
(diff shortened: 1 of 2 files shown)


d-fournier:
I signed it!

Marcono1234:
This seems to mix two concepts:
- writing raw JSON (for which `JsonWriter.jsonValue` already exists)
- using `char[]` to allow zeroisation of sensitive data

I would assume that users which require only the second feature (zeroisation of sensitive data) don't really need to be able to write raw JSON, in fact it might even cause them to write malformed JSON, e.g. when a password contains a `"`. So I think it would make sense to keep this separate. Though maybe the project members think differently about this.
In #2012 I have implemented that for `CharSequence`, but as you mention the `Writer` implemention passed to `JsonWriter` might call `CharSequence.toString()`, so a user would have to provide a custom `Writer` implementation which does not do that.

eamonnmcmanus:
I'm not seeing a strong motivation for this change. I'm closing it for now but feel free to reopen with justification for use cases that would be helped by this and are difficult with the existing API.

PR #1246, opened 2018-02-03
Fix typo

JamesZBL:
I think it should be modified here. Thank you.


Marcono1234:
@eamonnmcmanus, could you please merge this typo fix? Though might be good to rename the commit message before merging because the current PR title is rather misleading.

PR #1250, opened 2018-02-09
Add tests that shows JsonReader#skipValue() in an infinite loop.

garydgregory:
This PR adds tests that shows JsonReader#skipValue() in an infinite loop. I should not be able to cause GSON to loose its marbles.


eamonnmcmanus:
I tried running the test and it passed. So I think whatever the problem was here, it has been fixed in the last 8 years.

PR #1254, opened 2018-02-16
add possibility to set default subtype in RuntimeTypeAdapterFactory

serveravt:
use case: server released new version of API and added new subtypes to json. All not updated clients will 

`throw new JsonParseException("cannot deserialize " + baseType + " subtype named "+ label + "; did you forget to register a subtype?");`

this change adds possibility to set default type for using in such cases


serveravt:
I signed it!

Issue #1252, opened 2018-02-15
Possible dereference of a null value in JsonWriter.java from static analysis tool.


ashishrana160796:
While doing a static code analysis from CheckerFramework tool, field `stack` might lead to possible null dereference. Here is comparison of suggested code and original code. 
Code change suggestion for `push` method when it called from anonymous block. Here is original code:
```
  private void push(int newTop) {
    if (stackSize == stack.length) {
```
Suggested changes based on class instance still being in under-initialization state. Hence, dereferencing  of `stack` can be problematic.
Error message :
```
JsonWriter.java:360: error: [dereference.of.nullable] dereference of possibly-null reference stack
    if (stackSize == stack.length) {    // Annotation added from nullness checker analysis.
```
Suggested Changes :
```
  private void push(/*>>>@UnknownInitialization(java.lang.Object.class) JsonWriter this,*/ int newTop) {
    if (stack!=null && stackSize == stack.length) {    // Annotation added from nullness checker analysis.
``` 
If this issue is needed to be considered a `pr` can be made for update.
Here annotated file is also attached in reference to this issue with txt extension.
[JsonWriter.txt](https://github.com/google/gson/files/1727619/JsonWriter.txt)

Thanderpyae:
Good

JakeWharton:
Looks like a bug in the tool. This can never be null. What is its justification?

ashishrana160796:
Yes your analysis from `pr` comment is right and I have raised an issue for this regarding this problem. Considering a justification for such behavior is this following case :

@NonNull @UnderInitialization C b; | b | @NonNull
-- | -- | --
  | b.f | @MonotonicNonNull
---
Where b is class instance and f can be stack field. Hence, it gets treated as null type object when under initialization.  
Thanks.

PR #1255, opened 2018-02-16
JsonWriter null dereference issue #1252 resolved.

ashishrana160796:
Based on static tool analysis of Issue #1252.


ashishrana160796:
I signed it!

JakeWharton:
This variable can never be null. It's initialized upon construction and only overwritten with `newStack` which also is never null. This class is not thread safe so there's no issues with cross-thread visibility.

PR #1262, opened 2018-03-01
Add "Ignore" annotation to ignore some fields during serialization or d…

raomuyang:
`Ignore` is an annotation that indicates this member should be ignored for JSON
serialization or deserialization.

This annotation has no effect unless you build `com.google.gson.Gson`
with a ` com.google.gson.GsonBuilder` and invoke `GsonBuilder.excludeFieldsWithIgnoreAnnotation()` method.

* An example:

```java
public class Account {
  private String accountName;
  @Ignore(deserialize = false) private String info;
  @Ignore private String password;
}
```

If you create Gson via ` new Gson()`, the `@code toJson()` and `fromJson()`
methods will use all fields for serialization and deserialization. However, you can create
Gson via `new GsonBuilder().excludeFieldsWithIgnoreAnnotation().create()`, then the
` toJson()` will exclude the `info` field and the`password` field,
because those fields were marked with the`@Ignore` annotation and set
`serialize = true`. Similarly, the `fromJson()` will exclude `password` since `deserialize` is set to false.

---

What's the difference from`Expose` annotation?
The`Expose` annotation indicates this member should be exposed for JSON serialization
or deserialization, it means all the fields that you wanna expose, you should be marked with `Expose`:
```java
public class Student {
  @Expose private Account account;
  @Expose private String class;
}
```
You can Create Gson with `Gson gson = new GsonBuilder().excludeFieldsWithoutExposeAnnotation().create()`,
but the fields of `Account` class will be excluded because they are not marked by `Expose`.


NightlyNexus:
If you need this functionality, you can mark the fields `transient`.

raomuyang:
@NightlyNexus  Thanks for you review. But `transient` can not replace `@Ignore(deserialize = false)`

kyle-cackett:
I home-brewed the functionality in this pull request because `transient` was not sufficient for my use case. The problem with `transient` is that it has implications outside of GSON. In particular, I needed a particular field to be serialized by Java when writing objects to streams but not by GSON -- this is not possible with the `transient` field.

Johnny850807:
Agreed, @Expose makes class so ugly.

PR #1263, opened 2018-03-03
return Long instead of Double if number has no dot

liweys:
e.g., "1.0" is parsed as Double while "1" is parsed as Long.


liweys:
I signed it!

liweys:
tests fail

PR #1264, opened 2018-03-03
return Long instead of Double if number has no dot

liweys:
e.g., "1.0" is parsed as Double while "1" is parsed as Long.


artem-krutov:
IMHO it requires more logic to be added. Instead of just processing `case NUMBER`, it could be better to introduce new JsonToken parameter, say separate `NUMBER` into `DOUBLE_NUMBER` and `LONG_NUMBER`, and then process more straight.

liweys:
I think it will be better than using Double for all. But I think it's a good idea of `DOUBLE_NUMBER` and `LONG_NUMBER`.

DiogoRMax:
I do not understand the reason for adding BigInteger, without adding BigDecimal too

zenglian:
I think Double is able to hold arbitrary big decimal, but Long is limited to MAX_LONG.

PR #1265, opened 2018-03-08
Gradle config snippet to use `implementation`

serv-inc:
As of https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation, 

> The `compile` configuration still exists but should not be used as it will not offer the guarantees that the `api` and `implementation` configurations provide.

See also https://stackoverflow.com/questions/44493378/whats-the-difference-between-implementation-and-compile-in-gradle and https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
   * To use Gson in Android
 ```gradle
 dependencies {
-    compile 'com.google.code.gson:gson:2.8.2'
+    implementation 'com.google.code.gson:gson:2.8.2'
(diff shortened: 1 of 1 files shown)


artem-krutov:
IMHO it requires more logic to be added. Instead of just processing `case NUMBER`, it could be better to introduce new JsonToken parameter, say separate `NUMBER` into `DOUBLE_NUMBER` and `LONG_NUMBER`, and then process more straight.

JakeWharton:
Another PR was merged for this.

PR #1266, opened 2018-03-14
Added in Type Adapters to be able to deal with LocalDate and LocalDat…

reegz:
Added in Type Adapters to be able to deal with serialization and deserialization of LocalDate and LocalDateTime objects.


gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -16,6 +16,15 @@
 
 package com.google.gson.internal.bind;
 
+import com.google.gson.*;
```
JakeWharton: No star imports please

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -16,6 +16,15 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.internal.LazilyParsedNumber;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import guru.mmp.common.util.ISO8601;
```
JakeWharton: I have no idea where this comes from but we can't use it

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -24,39 +33,15 @@
...
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.StringTokenizer;
-import java.util.UUID;
+import java.text.ParseException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.util.*;
```
JakeWharton: No star imports

gson/src/main/java/guru/mmp/common/util/ISO8601.java
```diff
@@ -0,0 +1,231 @@
...
+import java.time.temporal.ChronoField;
+import java.time.temporal.TemporalAccessor;
+
+/**
+ * The <code>ISO8601</code> class provides a helper class for handling ISO 8601 strings of the
+ * following format: "2008-03-01T13:00:00+01:00". It also supports parsing the "Z" timezone.
+ *
+ * @author Marcus Portmann
+ */
+public final class ISO8601
```
JakeWharton: This entire file should not be in the PR

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -693,6 +678,104 @@ public void write(JsonWriter out, Locale value) throws IOException {
     }
   };
 
+  public static final TypeAdapter<LocalDateTime> LOCAL_DATE_TIME = new TypeAdapter<LocalDateTime>() {
```
JakeWharton: These aren't being registered anywhere so they're not really useful. Moreover, we cannot register them automatically because the library has to support Java 6 and 7.

gurkerl83:
Any update on this, would be an urgent feature.

reegz:
> Any update on this, would be an urgent feature.

Kinda forgot about this. I'll try making the requested fixes asap and resubmit for review.

reegz:
> I signed it!

I signed it!

inder123:
Let's spin into a separate library for JDK 8 type adapters for Gson.

gson/src/main/java/com/google/gson/util/ISO8601Util.java
```diff
@@ -0,0 +1,180 @@
...
+            new ThreadLocal<DateTimeFormatter>() {
+                @Override public DateTimeFormatter initialValue() {
+                    return DateTimeFormatter.ISO_DATE;
+                }
+            };
+
+    private static final ThreadLocal<DateTimeFormatter> threadLocalTimeFormatter =
+            new ThreadLocal<DateTimeFormatter>() {
+                @Override public DateTimeFormatter initialValue() {
+                    return DateTimeFormatter.ISO_DATE;
```
jpeer: shouldn't that be DateTimeFormatter.ISO_TIME?

inder123:
Please contribute these in https://github.com/google/gson/tree/master/extras/src

reegz:
> Please contribute these in https://github.com/google/gson/tree/master/extras/src

Being new to this - can I get clarification on what you're requesting please?

Do I need to submit this PR to the referenced repository?

inder123:
@reegz I am suggesting that we provide these as sample type adapters instead of the main Gson distribution.  Our extras directory holds such sample code, so you can move these classes there.

Probably a better location would be a java8 specific sub-directory where we add providers for Java 8. But that isn't setup yet.

PR #1267, opened 2018-03-18
Supporting Long numbers during desiralization

artem-krutov:
**Problem**: every number is being deserialized as Double, even if it doesn't have floating point: https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java#L79

**Solution**: added supporting for Long numbers:
- `1.0` and `1.2` will be deserialized as Doubles
- `1` will be deserialized as Long.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
@@ -75,6 +75,9 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
     case STRING:
       return in.nextString();
 
+    case LONG_NUMBER:
+      return in.nextLong();
+
     case NUMBER:
       return in.nextDouble();
 
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -90,6 +90,7 @@ public final class TypeAdapters {
       while (tokenType != JsonToken.END_ARRAY) {
         boolean set;
         switch (tokenType) {
+        case LONG_NUMBER:
         case NUMBER:
           set = in.nextInt() != 0;
           break;
@@ -350,6 +351,7 @@ public final class TypeAdapters {
       case NULL:
         in.nextNull();
         return null;
+      case LONG_NUMBER:
       case NUMBER:
       case STRING:
         return new LazilyParsedNumber(in.nextString());
@@ -700,6 +702,7 @@ public final class TypeAdapters {
       switch (in.peek()) {
       case STRING:
         return new JsonPrimitive(in.nextString());
+      case LONG_NUMBER:
       case NUMBER:
         String number = in.nextString();
         return new JsonPrimitive(new LazilyParsedNumber(number));
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -449,6 +449,7 @@ public class JsonReader implements Closeable {
     case PEEKED_BUFFERED:
       return JsonToken.STRING;
     case PEEKED_LONG:
+      return JsonToken.LONG_NUMBER;
     case PEEKED_NUMBER:
       return JsonToken.NUMBER;
     case PEEKED_EOF:
--- a/gson/src/main/java/com/google/gson/stream/JsonToken.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonToken.java
@@ -61,10 +61,14 @@ public enum JsonToken {
   STRING,
 
   /**
-   * A JSON number represented in this API by a Java {@code double}, {@code
-   * long}, or {@code int}.
+   * A JSON number represented in this API by a Java {@code double}.
    */
   NUMBER,
+
+  /**
+   * A JSON number represented in this API by a Java {@code long} or {@code int}.
+   */
+  LONG_NUMBER,
 
   /**
    * A JSON {@code true} or {@code false}.
--- a/gson/src/test/java/com/google/gson/ObjectTypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/ObjectTypeAdapterTest.java
@@ -27,11 +27,12 @@ public final class ObjectTypeAdapterTest extends TestCase {
   private final TypeAdapter<Object> adapter = gson.getAdapter(Object.class);
 
   public void testDeserialize() throws Exception {
-    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2,null],\"c\":{\"x\":\"y\"}}");
-    assertEquals(5.0, map.get("a"));
-    assertEquals(Arrays.asList(1.0, 2.0, null), map.get("b"));
+    Map<?, ?> map = (Map<?, ?>) adapter.fromJson("{\"a\":5,\"b\":[1,2.0,null],\"c\":{\"x\":\"y\"},\"d\":1.3}");
+    assertEquals(5L, map.get("a"));
+    assertEquals(Arrays.asList(1L, 2.0, null), map.get("b"));
     assertEquals(Collections.singletonMap("x", "y"), map.get("c"));
-    assertEquals(3, map.size());
+    assertEquals(1.3, map.get("d"));
+    assertEquals(4, map.size());
   }
 
   public void testSerialize() throws Exception {
--- a/gson/src/test/java/com/google/gson/functional/CollectionTest.java
+++ b/gson/src/test/java/com/google/gson/functional/CollectionTest.java
@@ -250,11 +250,10 @@ public class CollectionTest extends TestCase {
   }
 
   @SuppressWarnings("rawtypes")
-  public void testRawCollectionDeserializationNotAlllowed() {
+  public void testRawCollectionDeserializationAllowed() {
     String json = "[0,1,2,3,4,5,6,7,8,9]";
-    Collection integers = gson.fromJson(json, Collection.class);
-    // JsonReader converts numbers to double by default so we need a floating point comparison
-    assertEquals(Arrays.asList(0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0), integers);
+    Collection longs = gson.fromJson(json, Collection.class);
+    assertEquals(Arrays.asList(0L, 1L, 2L, 3L, 4L, 5L, 6L, 7L, 8L, 9L), longs);
 
     json = "[\"Hello\", \"World\"]";
     Collection strings = gson.fromJson(json, Collection.class);
@@ -263,7 +262,7 @@ public class CollectionTest extends TestCase {
   }
 
   @SuppressWarnings({"rawtypes", "unchecked"})
-  public void testRawCollectionOfBagOfPrimitivesNotAllowed() {
+  public void testRawCollectionOfBagOfPrimitivesAllowed() {
     BagOfPrimitives bag = new BagOfPrimitives(10, 20, false, "stringValue");
     String json = '[' + bag.getExpectedJson() + ',' + bag.getExpectedJson() + ']';
     Collection target = gson.fromJson(json, Collection.class);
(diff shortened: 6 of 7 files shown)


artem-krutov:
I signed it!

lyubomyr-shaydariv:
Would this break backward compatibility?

artem-krutov:
@lyubomyr-shaydariv I don't think so. Currently every number in a json document is being deserialized as a Double. So, people have the code like this in their projects:
```
double price = gson.fromJson(myData, myClass.class);
```
When we will start returning Longs, the same code will work. No problems in this code:
```
double a = 1L;
```

lyubomyr-shaydariv:
@artem-krutov Could you please give a more concrete example of the issue you're trying to fix?

artem-krutov:
@lyubomyr-shaydariv sure.

The issue reproduces when we have a POJO with a Map inside it. For example:
```
public class Person {
    public String name;
    public String lastName;
    public int age;
    public Map<String, Object> parameters; // The problem will be here

    // getters, setters
}
```
Imagine we have the following JSON object:
```
{
    "name": "Artem",
    "lastName": "Krutov",
    "age": 25,
    "parameters": {
        "childrenCount": 5,
        "numberOfFingersOnRightHand": 5
    }
}
```
What will we have once we deserialize?
```
Person person = new Gson().fromJson(jsonString, Person.class);
```
The age will be completely OK:
```
ing age = person.getAge(); // this is int
```
But what about parameters?
```
int childrenCount = person.getParameters().get("childrenCount"); // this line will fail
```
The line will fail, because Double will be returned. The same is for fingers. My PR fixes it – if we put double into a JSON, we will get double after deserializing. If we put Integer like in the example above, we will get integer after deserialization.

lyubomyr-shaydariv:
The issues I can see with the patch:

#### Issue: Introducing a new JSON token type: LONG_NUMBER

Gson follows the JSON standard that does not make a difference between integer and real numbers. That is why, I guess being not the inventor of Gson, `JsonToken.NUMBER` exists. Please note that the `JsonToken` enumeration covers all possible tokens that can occur in JSON documents. Adding a new specialized one would break a lot of code: any existing code that relies on `JsonToken.NUMBER` would just break in case of `JsonToken.LONG_NUMBER`; I usually use `switch`es in the form of the following pattern:

```java
switch ( jsonToken ) {
case NUMBER:
	// do something
case ...
	// do something
case ...
	// do something
default:
	throw new AssertionError(jsonToken); // This must never happen, an assertion, but all of my code will fail.
}
```

#### Issue: Possible misinterpreting of `java.lang.Number` and Gson `LazilyParsedNumber`

`java.lang.Number` allows mutual conversion between general numeric types. `LazilyParsedNumber`, that is a direct child of `java.lang.Number`, follows the contract of the latter as well. Holding a backing string that is the original value of a primitive numeric JSON value, it allows to convert the string into various numeric types depending on your particular needs. Please note that there is a low chance of losing precision since you are free to convert `LazilyParsedNumber` instances at call-site just by checking the value range.

For example, consider the following JSON to be deserialized:

```json
{
	"foo": 1,
	"bar": 2.3
}
```

Now the consider the following code:

```java
final Map<String, Number> fooBarMap = gson.fromJson(json, new TypeToken<Map<String, Number>>() {}.getType());

```

```java
private static class FooBar {
	final Number foo = null;
	final Number bar = null;
}
...
final FooBar fooBar = gson.fromJson(json, FooBar.class);
```

All four following lines:

```java
System.out.println(fooBarMap.get("foo").getClass());
System.out.println(fooBarMap.get("bar").getClass());
System.out.println(fooBar.foo.getClass());
System.out.println(fooBar.bar.getClass());
```

produces `class com.google.gson.internal.LazilyParsedNumber` that can be converted to any basic numeric value of *your* choice.

#### Issue: Deserializing a long-look-like value as `java.lang.Long`

This will also break backwards compatibility for cases like the below:

```java
final Map<String, ?> index = gson.fromJson(json, new TypeToken<Map<String, ?>>() {}.getType());
...
final Double d = index.get("foo"); // May throw:
                                   // java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Double
```

Also, `double d = 1L;` is not about a runtime-cast. In this case, `javac` converts a primitive long to a primitive `double` at the compile time. Check this out of `javap` output:

```javap
0: dconst_1
1: dstore_0
```

A similar output for `long l = (long) 1D;`

```javap
0: lconst_1
1: lstore_0
```

And this is what happens for `final Double d = (Double) (Object) 1L;` at runtime:

```javap
0: lconst_1
1: invokestatic  #2                  // Method java/lang/Long.valueOf:(J)Ljava/lang/Long;
4: checkcast     #3                  // class java/lang/Double
7: astore_0
```

#### Fix for your case

```java
int childrenCount = person.getParameters().get("childrenCount");
```

This cannot compile, but you can get the `childrenCount` property at least in two ways:

* `((Double) person.getParameters().get("childrenCount")).intValue()`
* `(int) (double) person.getParameters().get("childrenCount")` (this should work faster, as far as I understand)

And even making your mapping more convenient, you could just do (if possible for you case):

```java
Map<String, Number> parameters;

```

And then just get a proper type you want: `person.getParameters().get("childrenCount").intValue()`.

---

If I'd implemented such an implementation, I'd go with an option to disable the old functionality in `GsonBuilder` by implementing an internal type adapter factory that works for the `java.lang.Number` class hierarchy to work like you want it to be. But I wouldn't due to the issues I mentioned above, and I believe I missed some of them.

artem-krutov:
So, generally there are a lot of workarounds.

The following approach could work in some cases:
```
gson.fromJson(json, new TypeToken<Map<String, Number>>() {}.getType());
```
But it won't work in case when you don't know which class you need. For example:
```
String className = getClassName();
SomeBaseClass myObject = gson.fromJson(json, Class.forName(className));
```
I could agree it's not a super popular way to deserialize, and you usually know which exact class you need. But this flow is possible.

> any existing code that relies on JsonToken.NUMBER would just break in case of JsonToken.LONG_NUMBER

How's that? 
```
JsonToken token = JsonToken.NUMBER;
switch (token) {
    case JsonToken.NUMBER: {
        // this code will work;
    case JsonToken.LONG_NUMBER: {
        // we won't fall here
    }
}

...

JsonToken token = JsonToken.LONG_NUMBER;
switch (token) {
    case JsonToken.NUMBER: {
        // we won't fall here
    case JsonToken.LONG_NUMBER: {
        // this code will work
    }
}
```
We introduced LONG_NUMBER, and we process it in any place where we can see it.

Then, about `Number` and monster constructions like `(int) (double) person.getParameters().get("childrenCount")`.

The first doesn't look like a good option. When you develop an application you want to be as straight as it possible. If you can have only integer amount of children, why you need `Number`? Hacking your code just to fit into the 3rd party library's limitations doesn't look good.

Converting doubles into integers looks better to me. We can live with it, but it would be better if we don't need to do any convertions.

I do understand your reasons not to merge this PR. You sound confident and you are indeed, your comments looks solid. Honestly, I didn't take into account that Gson follows the JSON specifications.
I'm basically OK with closing this PR, but just wanted to point out that JSON is pretty simple and stupid. It is very basic. On the other hand, our enterprise code much more complex :) So, if we can make difference between int and double – why not to?

artem-krutov:
> I meant not your code, but mine, for example: the code relies the finite and known-ahead set of JSON tokens, so introducing LONG_NUMBER would just tear it apart.

I'm afraid your code could start failing on each and every change of the underlying library if it's so tied with it. You are basically not supposed to rely on the library's innards, you are supposed to use it's API. Or, use adapters ;)

Yes, these barriers between 3rd party and your own code – they are really cool, this is true. It's a good idea to have them for some complex peaces of code. But the whole idea of any library is to fit into your needs. If you want to do simple things like getting data from JSON object, I believe it's good when you just do it. Or, otherwise, your business logic would easily transform into a bunch of adapters for each and every thing.

> You can try to implement a TypeAdapterFactory

This is exactly how I workarond that double/int thing right now, thank you.

So, speaking generally, the biggest issue I can see with the current implementation – it is not consistent. You put integer and then you get double. We can debate for a long time and we both can propose reasonable solutions. Sometimes more reasonable and sometimes less :) But the fact is the fact – the behavior is not consistent. 

Don't include JavaScript stuff here. Gson is a barrier (or, adapter) between JavaScript's JSON and Java code. And I really want to get object exactly of the same type as I put into the structure.

lyubomyr-shaydariv:
Well, `JsonToken` is a public API enum and it's widely used for parsing JSON streams. And it's well designed for public use covering all possible JSON tokens. Sorry, when I was referring JavaScript, I meant that JSON originates its idea from it and the compromise for numeric values, and nothing else.

You're missing the fact that `Double` is probably the best way to store numbers except the case when Long may exceed the limits of precision of `Double` and `Double.longValue()` may return incorrect results. Especially when JSON documents do not store numbers type info. All of its, `Double`, operations are superfast from the JVM perspective. As an alternate solution that can be 100% precise is `LazilyParsedNumber` that holds the original JSON string but parses it every time its requested for a number. That's an option you can have. I don't know the exact reason why Gson prefers `Double` over `LazilyParsedNumber` in certain cases and vice versa, but both of them are `Number`s therefore giving you a chance to extract _a value of the type you need_. That must be either declared at declaration-site (field types, type tokens, etc) being as concrete as possible, or letting a user pick the best alternative at call-site. You're saying:

> it is not consistent. You put integer and then you get double.

True, it is not 100% consistent, and it cannot be such: how can you distinguish between bytes, shorts, ints, longs then? If you put an `Integer` and instruct Gson it's put into an `Object` of any sort, it writes a number, but this does not let you read it back to `Object` having exact type unless you declare it somehow.

Take a look: if you read `1`: is it a `long`, an `int` or a `short`? None! It was a `byte` while writing a single digit, but I didn't had any chance to specify that it is a `byte` and nothing else in JSON. That's it: this is the fundamentals of JSON where _every_ number is a value of double precision, and there is no way to specify a real type of the value it originated from. The best _I_ and _you_ can do is declaring the type before reading: `Byte` or `byte` for this example, but not an `Object`. The best _Gson_ can do is `Double` and `LazilyParsedNumber`, if more super types were used in declaration (`Object` or `Number`). `Number` allows to extract a `byte` in any case. The case you're addressing to is 100% supported with `LazilyParsedNumber` without breaking compatibility and being 100% precise.

Please don't get me wrong, but you should either have a proper mapping at declaration-site, or use `Number` public methods at call-site. The Gson internal parsers do the right job and don't intrude into heuristic of choosing for "what this number literal looks like the most" (`1` can be a `Float` too! -- I haven't mentioned this type today), but they obey _your_ mappings, and you totally control it. That's it.

> And I really want to get object exactly of the same type as I put into the structure.

100% working method is storing type information along with numbers. Say `[1, 1, 1, 1]` might be represented in the following form:

```json
[
{"type": "byte", "value": 1},
{"type": "short", "value": 1},
{"type": "long", "value": 1},
{"type": "double", "value": 1}
]
```

or purely hypothetical (and the highlighter detects it)

```json
[1B, 1S, 1L, 1D]
```

Looks crazy? It does. But this resolves the issue of not having the original type information. And this is not a very non-trivial case: take a look at `RuntimeTypeAdapterFactory` designed for exactly the same problem when you can't instantiate an instance of an abstract class without knowing the type of the original object the JSON has been generated from (and even the worst case: structurally such objects may look the equivalent having the same properties, but having no type discriminator will never help you to get the original type).

artem-krutov:
@lyubomyr-shaydariv thank you for detailed answers.

I agree that structures like `{"type": "byte", "value": 1}` and `[1B, 1S, 1L, 1D]` looks really crazy :) We tried to make things simpler but ended up with much more complex code, right.

I see your position. Though, a couple of comments.

> You're missing the fact that Double is probably the best way to store numbers except the case when Long may exceed the limits of precision of Double

I think this is handled by this code:
```
case PEEKED_LONG:
      return JsonToken.LONG_NUMBER;
```
(https://github.com/google/gson/pull/1267/files#diff-0776314ca8e083ca3943426740d1dd69R452)

GSON can understand when it peeks longs, there is `fitsInLong` flag: https://github.com/artem-krutov/gson/blob/3af5f69a97668ab6b72ed26c5e9c201a4eed0f45/gson/src/main/java/com/google/gson/stream/JsonReader.java#L732

So, there would not be (or should not be?) any mess with longs and doubles.

> Take a look: if you read 1: is it a long, an int or a short? None! It was a byte while writing a single digit, but I didn't had any chance to specify that it is a byte and nothing else in JSON.

Yes, that's true. But all these data formats are stand for numbers without floating part. It's too narrow case because people do not usually use these types at all. It could be confusing, because we don't need even `int` for fields like `fingersCount`, but anyway. I think that the only difference is between floating numbers and numbers without floating point.

I also want you to get me right: I don't want to argue, all I want is work together to make the library a bit better. I agree there are a bunch of moments that are not obvious to me and could be really obvious to you. I just want your help in some points (and already got nice-detailed answers).

As far as I understood, you stand for 2 points:
1. Leave the number-policy unified. We don't need unnecessary differences between number formats, because we can control everything on a caller side
2. We must not break backward compatibility, which this PR might potentially do.

Am I right? Also a question about the point 1: do you think we don't need it at all, or this particular implementation needs to be revisited and improved?

Marcono1234:
Do you think this change is still needed or has this been resolved by #1290?

PR #1278, opened 2018-04-02
Add FieldNamingPolicy.LOWER_CASE_WITH_DOTS

zenglian:
Dot separated property naming is one of the most common field naming policies, used in many products like Spark, hadoop, log4j etc, e.g., `spark.executors.memory`, which is mapped to `sparkExecutorsMemory` as class field name.
This is very similar to LOWER_CASE_WITH_DASHES, the only difference is using `.` instead of `-`.

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
+++ b/gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
@@ -63,6 +63,14 @@ public class NamingPolicyTest extends TestCase {
         + target.someConstantStringInstanceField + "\"}", gson.toJson(target));
   }
 
+  public void testGsonWithLowerCaseDotPolicySerialization() {
+    Gson gson = builder.setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_DOTS).create();
+    StringWrapper target = new StringWrapper("blah");
+    assertEquals("{\"some.constant.string.instance.field\":\""
+          + target.someConstantStringInstanceField + "\"}", gson.toJson(target));
+  }
+
+
   public void testGsonWithLowerCaseDashPolicyDeserialiation() {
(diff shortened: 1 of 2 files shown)


zenglian:
An example

```
{
  "monitor.interval": 120,
  "monitor.max.retries": 100,
  "monitor.emails": "",
  "server.address": "0.0.0.0",
  "server.port": 8091,
  "server.max.connections": 10,
  "server.auth": {
    "user": "user",
    "password": "password"
  },
  "jdbc.driver": "org.h2.Driver",
  "jdbc.url": "jdbc:h2:tcp://slc09woc:19092/h2-db",
  "jdbc.auth": {
    "user": "user",
    "password": "password"
  },
  "job.server.url": "http://slc09woc:8090",
  "job.server.auth": {
    "user": "user",
    "password": "password"
  },
  "proxy": {
    "host": "aaa",
    "port": 80,
    "auth": {
      "user": "user",
      "password": "password"
    }
  },
  "yarn.url": "http://slc09woc:8088/ws/v1/cluster",
  "yarn.auth": {
    "user": "user",
    "password": "password"
  }
}
```

gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
```diff
@@ -63,6 +63,14 @@ public void testGsonWithLowerCaseDashPolicySerialization() {
         + target.someConstantStringInstanceField + "\"}", gson.toJson(target));
   }
 
+  public void testGsonWithLowerCaseDotPolicySerialization() {
```
inder123: We need a deserialization test as well

inder123: Added the extra test in https://github.com/google/gson/pull/1303

Issue #1279, opened 2018-04-03
Add custom indent for pretty printing


WasylF:
It would be great if we can specify custom indent while call pretty printing. Currently pretty printing always use 2 spaces for indent. This customization also allows to reduce size of resulting json. 
Comparison table and code for generating json below:

| Indent \ setsCount | 100 | 2000 |
| --- | --- | --- |
| 2 spaces |  99,7 kb |  50,9 MB |
| 1 tab |  79,5 kb | 42,9 MB |
| nothing | 59,3 kb | 34,9 MB |

```
        List<Set<Integer>> sets = new ArrayList<Set<Integer>>();
        int setsCount = 2000;
        for (int setN = 0; setN < setsCount; setN++) {
            TreeSet<Integer> set = new TreeSet<Integer>();
            for (int i = 0; i < setsCount; i++) {
                set.add(setN * setsCount + i);
            }
            sets.add(set);
        }

        String json = gsonManager.getGson().toJson(sets);

```

JakeWharton:
What's the motivation? If you care about size you aren't using pretty printing.

WslF:
My team's project stores some large configuration data in the git repository as JSON. If I don't use pretty option the resulting file contains a single line with the whole JSON, so git couldn't show a difference between different versions. Currently, I use pretty JSON, but in my case, it increases the size of the JSON from 30 - 40 MB up to 100+ MB. Bigger resulting JSON also requires more RAM for converting. So, I suppose it would be cool to have such feature as a custom indent.
And the other reason is that war between people using tabs and spaces is endless:) So, it is great to have both options.

bonoboris:
I agree, if you are appending to an existing Json with either 2, 4 spaces or tab indent, it would be nice to have such a feature. I'm actually suprised there isn't.

yurymann:
+1 for configurable indent size

@JakeWharton , my motivation is being able to compare jsons generated with gson against json files formatted in other tools, in my case IntelliJ Idea.

nickgeorge:
+1 for configurability!

Conrad-T-Pino:
Already implemented in 2.8.7 release. 

Class `com.google.gson.stream.JsonWriter` has `public final void setIndent​(String indent);` which:

> Sets the indentation string to be repeated for each level of indentation in the encoded document. If `indent.isEmpty()` the encoded document will be compact. Otherwise the encoded document will be more human-readable.

This implementation indents with tab and appends final newline:

[...]

PR #1280, opened 2018-04-03
Add custom indent for pretty printing #1279

WslF:
Implementation for issue #1279

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -176,7 +176,6 @@ public class JsonWriter implements Closeable, Flushable {
    */
   private String indent;
 
-  private boolean prettyPrinting = false;
   /**
    * The name/value separator; either ":" or ": ".
    */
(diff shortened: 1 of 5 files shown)


gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -176,6 +176,7 @@
    */
   private String indent;
 
+  private boolean prettyPrinting = false;
```
lyubomyr-shaydariv: Is this field necessary? The `indent` field seems to cover this case.

lyubomyr-shaydariv:
I think this is a pretty useful feature to have in `GsonBuilder`. By the way, did you consider something like `.setPrettyPrintingByTabs(int count)` and `.setPrettyPrintingBySpaces(int count)` rather than allowing a user to specify an arbitrary indent string? I know that `JsonWriter` allows arbitrary indent strings, but it looks more of a design issue allowing to emit illegal JSON. Your thoughts?

gson/src/main/java/com/google/gson/stream/JsonWriter.java
WslF: I've removed it.

WslF:
@lyubomyr-shaydariv, Thanks for your review! Your proposal looks good. I could implement it. But what about just add check for indent string? smth like:

```
    if (indent != null && !indent.matches("\\s*")) {
        throw new IllegalArgumentException("Indent could contains ONLY whitespaces");
    }

```

ryber:
I'd love to see this PR merged (or some variation), this is pretty old now. Any possibility it will get merged?

inder123:
We would like to see a more general solution. For example, something that supplies a formatter that walks through a DOM tree and prints it.

lyubomyr-shaydariv:
@inder123 but does it really need to be such complicated, and doesn't this PR merely implement a minor tweak for an existing currently poorly customizable indentation feature (that's pretty old itself) rather than implementing a neat well-designed formatter from scratch?

inder123:
@lyubomyr-shaydariv It's not a minor tweak if you are introducing a new API.

PR #1281, opened 2018-04-03
updated README to use implementation instead of compile

greatjack1:
As of io 17, google reccomends using implementation when adding gradle dependencies instead of compile as compile is now depreciated.

I updated the readme to use implementation instead of compile.

See https://youtu.be/7ll-rkLCtyk?t=22m20s for more details.


greatjack1:
I confirm I am ok with the pull request

PR #1288, opened 2018-04-15
Parse JSON numbers to Number type instead of Double.

zenglian:
Parse JSON numbers to Java's Number type instead of Double. Subsequently it's parsed as Long and Double respectively and correctly. E.g., "1" is parsed as Long while "1.23", "1e10" are parsed as Double. This fix does not contradict with the original JSON design which does not have integer types. It also resolves the issue that "9223372036854775807" (Long.MAX_VALUE) is parsed as Double and precision is lost.


zenglian:
```
    public void testNumber() {
        Gson gson = new Gson();
        Object[] numbers1 = new Object[]{1, 1.0, 1.23, 9223372036854775807L, new BigInteger("9223372036854775808"),
                1.23E10, 1.23E20};
        String json = gson.toJson(numbers1);
        Object[] numbers2 = gson.fromJson(json, Object[].class);
        System.out.println(json);
        System.out.println(gson.toJson(numbers2));
        for (int i = 0; i < numbers2.length; i++) {
            System.out.println(numbers1[i] + " (" + numbers1[i].getClass().toString().substring(6)
                    + ")\t-->\t" + numbers2[i] + " (" + numbers2[i].getClass().toString().substring(6) + ")");
        }
    }

//result
[1,1.0,1.23,9223372036854775807,9223372036854775808,1.23E10,1.23E20]
[1,1,1.23,9223372036854775807,9.223372036854776E18,12300000000,1.23E20]

1 (java.lang.Integer)	-->	1 (java.lang.Long)
1.0 (java.lang.Double)	-->	1 (java.lang.Long)
1.23 (java.lang.Double)	-->	1.23 (java.lang.Double)
9223372036854775807 (java.lang.Long)	-->	9223372036854775807 (java.lang.Long)
9223372036854775808 (java.math.BigInteger)	-->	9.223372036854776E18 (java.lang.Double)
1.23E10 (java.lang.Double)	-->	12300000000 (java.lang.Long)
1.23E20 (java.lang.Double)	-->	1.23E20 (java.lang.Double)
```

zenglian:
**JSON does have "int"** (the same as Java though there is no restriction on its size/no base8/16), see www.json.org.  

    ...
    number -> int | int frac | int exp | int frac exp
    ...

astux7:
Can this PR be merge asap - it is an issue when i am trying to use json for long?

PR #1290, opened 2018-04-17
Support arbitrary Number implementation for Object and Number deserialization

lyubomyr-shaydariv:
[RFC 8259, 6 Numbers](https://tools.ietf.org/html/rfc8259#section-6) allows numbers of arbitrary range and precision specifying that a particular implementation may set some limits for numeric values. As of Gson 2.8.4 uses the following deserialization strategies:

* If the result type is declared as `Object`, Gson always returns `Double`s for all encountered JSON numbers. In this case, the minimum and maximum hold number value fits the limits of `Double` which cannot hold exact `Long` values that exceed the `Double` exact precision limits (`2^53+1`).
* ‎If the result type is declared as `Number` and there are no custom `Number` adapters, Gson always returns `LazilyParsedNumber`s that hold original string values parsing them to a requested type lazily allowing to use the whole range of `Long`. However, it does not allow to use numbers of arbitrary range and precision, and does not expose its hold string as a `BigDecimal`.

In order to fix these limitations and preserve backwards compatibility, some sort of "to-number" strategies might be accepted in `GsonBuilder` to override the default behavior of Gson. This pull-request introduces such a strategy interface to be used in the built-in `Object` and `Number` type adapters. There are also four strategies implemented in this PR (two standard to keep the backwards compatibility and two enhanced to overcome the limitations) using an enumeration:

* `ToNumberPolicy.DOUBLE` that implements the default behavior of the `Object` type adapter returning `Double`s only.
*  `ToNumberPolicy.LAZILY_PARSED_NUMBER` that implements the default behavior of the `Number` type adapter.
* `ToNumberPolicy.LONG_OR_DOUBLE` that tries to parse a number as a `Long`, otherwise then tries to parse it as a `Double`, if the number cannot be parsed as a `Long`.
* `ToNumberPolicy.BIG_DECIMAL` that can parse numbers of arbitrary range and precision.

Call-site is expected to extract proper values using the methods declared in `java.lang.Number`.

Examples of use:

#### Default behavior, backwards-compatible with the previous versions of Gson

```java
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(ToNumberPolicy.DOUBLE) // explicit default, may be omitted
  .create();
List<Object> actual = gson.fromJson("[null, 10, 10.0]", new TypeToken<List<Object>>() {}.getType());
List<Double> expected = Arrays.asList(null, 10.0, 10.0);
assertEquals(expected, actual);
```

```java
Gson gson = new GsonBuilder()
    .setNumberToNumberStrategy(ToNumberPolicy.LAZILY_PARSED_NUMBER) // explicit default, may be omitted
    .create();
List<Number> actual = gson.fromJson("[null, 10, 10.0]", new TypeToken<List<Number>>() {}.getType());
List<Object> expected = Arrays.<Object>asList(null, new LazilyParsedNumber("10"), new LazilyParsedNumber("10.0"));
assertEquals(expected, actual);
```

#### `Object`-declared numbers are always `LazilyParsedNumber`s

```java
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(ToNumberPolicy.LAZILY_PARSED_NUMBER)
  .create();
List<Object> actual = gson.fromJson("[null, 10, 10.0]", new TypeToken<List<Object>>() {}.getType());
List<LazilyParsedNumber> expected = Arrays.asList(null, new LazilyParsedNumber("10"), new LazilyParsedNumber("10.0"));
assertEquals(expected, actual);
```

#### `Number`-declared numbers are always `Double`s

```java
Gson gson = new GsonBuilder()
  .setNumberToNumberStrategy(ToNumberPolicy.DOUBLE)
  .create();
List<Number> actual = gson.fromJson("[null, 10, 10.0]", new TypeToken<List<Number>>() {}.getType());
List<Double> expected = Arrays.asList(null, 10.0, 10.0);
assertEquals(expected, actual);
```

#### `Object`-declared numbers are either `Long` or `Double`

```java
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
  .create();
List<Object> actual = gson.fromJson("[null, 10, 10.0]", new TypeToken<List<Object>>() {}.getType());
List<? extends Number> expected = Arrays.asList(null, 10L, 10.0);
assertEquals(expected, actual);
```

#### `Object`-declared numbers are `BigDecimal`s

```java
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(ToNumberPolicy.BIG_DECIMAL)
  .create();
List<Object> actual = gson.fromJson("[null, 3.141592653589793238462643383279, 1e400]", new TypeToken<List<Object>>() {}.getType());
List<BigDecimal> expected = Arrays.asList(null, new BigDecimal("3.141592653589793238462643383279"), new BigDecimal("1e400"));
assertEquals(expected, actual);
```

#### Custom bytes-only:

```java
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(new ToNumberStrategy() {
    @Override
    public Byte toNumber(final JsonReader in)
            throws IOException {
      return (byte) in.nextInt();
    }
  })
  .create();
List<Object> actual = gson.fromJson("[null, 10, 20, 30]", new TypeToken<List<Object>>() {}.getType());
List<Byte> expected = Arrays.asList(null, (byte) 10, (byte) 20, (byte) 30);
assertEquals(expected, actual);
```

#### Custom deserialization does not affect `Byte`-declared deserialization:

```java
ToNumberStrategy fail = new ToNumberStrategy() {
    @Override
    public Byte toNumber(final JsonReader in) {
      throw new AssertionError();
    }
  };
Gson gson = new GsonBuilder()
  .setObjectToNumberStrategy(fail)
  .setNumberToNumberStrategy(fail)
  .create();
List<Object> actual = gson.fromJson("[null, 10, 20, 30]", new TypeToken<List<Byte>>() {}.getType());
List<Byte> expected = Arrays.asList(null, (byte) 10, (byte) 20, (byte) 30);
assertEquals(expected, actual);
```

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -227,7 +227,7 @@ public final class Gson {
 
     // built-in type adapters that cannot be overridden
     factories.add(TypeAdapters.JSON_ELEMENT_FACTORY);
-    factories.add(objectAdapterFactory(objectToNumberStrategy));
+    factories.add(ObjectTypeAdapter.getFactory(objectToNumberStrategy));
 
     // the excluder must precede all adapters that handle user-defined types
     factories.add(excluder);
@@ -247,7 +247,7 @@ public final class Gson {
             doubleAdapter(serializeSpecialFloatingPointValues)));
     factories.add(TypeAdapters.newFactory(float.class, Float.class,
             floatAdapter(serializeSpecialFloatingPointValues)));
-    factories.add(numberAdapterFactory(numberToNumberStrategy));
+    factories.add(NumberTypeAdapter.getFactory(numberToNumberStrategy));
     factories.add(TypeAdapters.ATOMIC_INTEGER_FACTORY);
     factories.add(TypeAdapters.ATOMIC_BOOLEAN_FACTORY);
     factories.add(TypeAdapters.newFactory(AtomicLong.class, atomicLongAdapter(longAdapter)));
@@ -369,20 +369,6 @@ public final class Gson {
           + " is not a valid double value as per JSON specification. To override this"
           + " behavior, use GsonBuilder.serializeSpecialFloatingPointValues() method.");
     }
-  }
-
-  private static TypeAdapterFactory objectAdapterFactory(ToNumberStrategy objectToNumberStrategy) {
-    if (objectToNumberStrategy == ToNumberPolicy.DOUBLE) {
-      return ObjectTypeAdapter.FACTORY;
-    }
-    return ObjectTypeAdapter.newFactory(objectToNumberStrategy);
-  }
-
-  private static TypeAdapterFactory numberAdapterFactory(final ToNumberStrategy numberToNumberStrategy) {
-    if (numberToNumberStrategy == ToNumberPolicy.LAZILY_PARSED_NUMBER) {
-      return NumberTypeAdapter.FACTORY;
-    }
-    return NumberTypeAdapter.newFactory(numberToNumberStrategy);
   }
 
   private static TypeAdapter<Number> longAdapter(LongSerializationPolicy longSerializationPolicy) {
--- a/gson/src/main/java/com/google/gson/internal/bind/NumberTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/NumberTypeAdapter.java
@@ -36,7 +36,7 @@ public final class NumberTypeAdapter extends TypeAdapter<Number> {
   /**
    * Gson default factory using {@link ToNumberPolicy#LAZILY_PARSED_NUMBER}.
    */
-  public static final TypeAdapterFactory FACTORY = newFactory(ToNumberPolicy.LAZILY_PARSED_NUMBER);
+  private static final TypeAdapterFactory LAZILY_PARSED_NUMBER_FACTORY = newFactory(ToNumberPolicy.LAZILY_PARSED_NUMBER);
 
   private final ToNumberStrategy toNumberStrategy;
 
@@ -44,16 +44,22 @@ public final class NumberTypeAdapter extends TypeAdapter<Number> {
     this.toNumberStrategy = toNumberStrategy;
   }
 
-  public static TypeAdapterFactory newFactory(final ToNumberStrategy toNumberStrategy) {
+  private static TypeAdapterFactory newFactory(ToNumberStrategy toNumberStrategy) {
+    final NumberTypeAdapter adapter = new NumberTypeAdapter(toNumberStrategy);
     return new TypeAdapterFactory() {
       @SuppressWarnings("unchecked")
       @Override public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
-        if (type.getRawType() == Number.class) {
-          return (TypeAdapter<T>) new NumberTypeAdapter(toNumberStrategy);
-        }
-        return null;
+        return type.getRawType() == Number.class ? (TypeAdapter<T>) adapter : null;
       }
     };
+  }
+
+  public static TypeAdapterFactory getFactory(ToNumberStrategy toNumberStrategy) {
+    if (toNumberStrategy == ToNumberPolicy.LAZILY_PARSED_NUMBER) {
+      return LAZILY_PARSED_NUMBER_FACTORY;
+    } else {
+      return newFactory(toNumberStrategy);
+    }
   }
 
   @Override public Number read(JsonReader in) throws IOException {
--- a/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
@@ -40,7 +40,7 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
   /**
    * Gson default factory using {@link ToNumberPolicy#DOUBLE}.
    */
-  public static final TypeAdapterFactory FACTORY = newFactory(ToNumberPolicy.DOUBLE);
+  private static final TypeAdapterFactory DOUBLE_FACTORY = newFactory(ToNumberPolicy.DOUBLE);
 
   private final Gson gson;
   private final ToNumberStrategy toNumberStrategy;
@@ -50,7 +50,7 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
     this.toNumberStrategy = toNumberStrategy;
   }
 
-  public static TypeAdapterFactory newFactory(final ToNumberStrategy toNumberStrategy) {
+  private static TypeAdapterFactory newFactory(final ToNumberStrategy toNumberStrategy) {
     return new TypeAdapterFactory() {
       @SuppressWarnings("unchecked")
       @Override public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
@@ -60,6 +60,14 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
         return null;
       }
     };
+  }
+
+  public static TypeAdapterFactory getFactory(ToNumberStrategy toNumberStrategy) {
+    if (toNumberStrategy == ToNumberPolicy.DOUBLE) {
+      return DOUBLE_FACTORY;
+    } else {
+      return newFactory(toNumberStrategy);
+    }
   }
 
   @Override public Object read(JsonReader in) throws IOException {


lyubomyr-shaydariv:
Related pull-requests: #1267, #1288.

danieleguiducci:
Great!

Taipeek:
Still waiting for this to be released :)

Catfriend1:
lgtm

grieser-mcsh:
please, apply as soon as possible

douglasom:
Is this PR ever going to be applied? If not with this enhancement, what is the approach recommended by the Gson team for deserializing numbers?

If I'm deserializing as a Map<String, Object> and I have some Integer values how to avoid for them to be interpreted as a Double?

roycarser:
is this PR  acceptable now ？

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -325,6 +327,32 @@ public GsonBuilder setFieldNamingStrategy(FieldNamingStrategy fieldNamingStrateg
     return this;
   }
 
+  /**
+   * Configures Gson to apply a specific number strategy during deserialization for {@link Object}.
+   *
+   * @param objectToNumberStrategy the actual object-to-number strategy
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   *
+   * @see ToNumberPolicy#DOUBLE
```
Marcono1234, line 336: Maybe make it more clear here that this is the default (same for `setNumberToNumberStrategy`)? E.g. by mentioning it in the documentation and not adding a `@see`

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -325,6 +327,32 @@ public GsonBuilder setFieldNamingStrategy(FieldNamingStrategy fieldNamingStrateg
     return this;
   }
 
+  /**
+   * Configures Gson to apply a specific number strategy during deserialization for {@link Object}.
```
Marcono1234, line 331: Maybe "deserialization of" / "deserialization as" instead of "for" (same for `setNumberToNumberStrategy`)?

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+ * {@link Object} and {@link Number}.
+ *
+ * @see ToNumberStrategy
+ */
+public enum ToNumberPolicy implements ToNumberStrategy {
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link Double} values.
+   * This is the default strategy used during deserialization numbers as {@link Object}
+   * using {@link com.google.gson.internal.bind.ObjectTypeAdapter} in earlier versions
```
Marcono1234, line 38: This will create a dead link because internal classes are [not documented](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/pom.xml#L28)

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+  DOUBLE {
+    @Override public Double toNumber(JsonReader in) throws IOException {
+      return in.nextDouble();
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link LazilyParsedNumber}
+   * values. This is the default strategy using during deserialization numbers as
+   * {@link Number} using {@link com.google.gson.internal.bind.TypeAdapters#NUMBER}.
```
Marcono1234, line 50: This will create a dead link because internal classes are [not documented](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/pom.xml#L28)

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  DOUBLE {
+    @Override public Double toNumber(JsonReader in) throws IOException {
+      return in.nextDouble();
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link LazilyParsedNumber}
+   * values. This is the default strategy using during deserialization numbers as
```
Marcono1234, line 49:
```suggestion
   * values. This is the default strategy used during deserialization numbers as
```

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+  LONG_OR_DOUBLE {
+    @Override public Number toNumber(JsonReader in) throws IOException, JsonParseException {
+      String value = in.nextString();
+      try {
+        return Long.parseLong(value);
+      } catch ( NumberFormatException longEx ) {
+        try {
+          final Double d = Double.valueOf(value);
+          if (d.isInfinite() || d.isNaN()) {
+            throw new MalformedJsonException("JSON forbids NaN and infinities: " + d + in);
```
Marcono1234, line 75: Should this consult `JsonReader.isLenient()` and not throw an exception if the reader is lenient?

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+          }
+          return d;
+        } catch ( NumberFormatException doubleEx ) {
+          throw new JsonParseException("Cannot parse " + value);
```
Marcono1234, line 79: Maybe include cause `doubleEx` since it might explain why the number is invalid?

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as numbers of arbitrary length
+   * using {@link BigDecimal}.
+   */
+  BIG_DECIMAL {
+    @Override public BigDecimal toNumber(JsonReader in) throws IOException {
+      return new BigDecimal(in.nextString());
```
Marcono1234, line 91: Catch `NumberFormatException` and rethrow as `JsonParseException`, like for `LONG_OR_DOUBLE `

gson/src/main/java/com/google/gson/ToNumberStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * A strategy that is used to control how numbers should be deserialized for {@link Object} and {@link Number}
+ * when a concrete type of the deserialized number is unknown in advance. By default, Gson uses the following
+ * deserialization strategies:
+ *
+ * <ul>
+ * <li>{@link Double} values are returned for JSON numbers if the deserialization type is declared as
+ * {@link Object};</li>
+ * <li>{@link com.google.gson.internal.LazilyParsedNumber} values are returned if the deserialization type
```
Marcono1234, line 31: This will create a dead link because internal classes are [not documented](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/pom.xml#L28)

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link LazilyParsedNumber}
+   * values. This is the default strategy using during deserialization numbers as
+   * {@link Number} using {@link com.google.gson.internal.bind.TypeAdapters#NUMBER}.
+   */
+  LAZILY_PARSED_NUMBER {
+    @Override public LazilyParsedNumber toNumber(JsonReader in) throws IOException {
```
Marcono1234, line 53: `LazilyParsedNumber` is internal so `Number` should be used as return type to not exposed it here

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+   * of Gson.
+   */
+  DOUBLE {
+    @Override public Double toNumber(JsonReader in) throws IOException {
+      return in.nextDouble();
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link LazilyParsedNumber}
```
Marcono1234, line 48: This will create a dead link because internal classes are [not documented](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/pom.xml#L28)

gson/src/main/java/com/google/gson/ToNumberStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+ * is declared as {@link Number}.</li>
+ * </ul>
+ *
+ * <p>For historical reasons, Gson does not support arbitrary-length numbers deserialization as its stated in
```
Marcono1234, line 35:
```suggestion
 * <p>For historical reasons, Gson does not support deserialization of arbitrary-length numbers as its stated in
```

gson/src/main/java/com/google/gson/ToNumberStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+ *   implementations that expect no more precision or range than these
+ *   provide, in the sense that implementations will approximate JSON
+ *   numbers within the expected precision.  A JSON number such as 1E400
+ *   or 3.141592653589793238462643383279 may indicate potential
+ *   interoperability problems, since it suggests that the software that
+ *   created it expects receiving software to have greater capabilities
+ *   for numeric magnitude and precision than is widely available.
+ * </pre>
+ *
+ * <p>For example, {@link ToNumberPolicy#LONG_OR_DOUBLE} and {@link ToNumberPolicy#BIG_DECIMAL} to overcome
```
Marcono1234, line 53:
```suggestion
 * <p>Use for example {@link ToNumberPolicy#LONG_OR_DOUBLE} or {@link ToNumberPolicy#BIG_DECIMAL} to overcome
```
It looks like there is a verb missing in the first part of the sentence

gson/src/main/java/com/google/gson/ToNumberStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Reads a number from the given JSON reader. A strategy is supposed to read a single value from the
+   * reader, and the read value is guaranteed never to be {@code null}.
+   *
+   * @param in JSON reader to read a number from
+   * @return number read from the JSON reader.
+   * @throws IOException
+   */
+  public Number toNumber(JsonReader in) throws IOException;
```
Marcono1234, line 70: A method called `toNumber` with a parameter might suggest that it converts the parameter to a number, which is not the case here.
Would `readNumber` maybe be better? The interface name is probably not that problematic.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,95 @@
...
+ * strategies to overcome some historical Gson limitations while deserializing numbers as
+ * {@link Object} and {@link Number}.
+ *
+ * @see ToNumberStrategy
+ */
+public enum ToNumberPolicy implements ToNumberStrategy {
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link Double} values.
+   * This is the default strategy used during deserialization numbers as {@link Object}
```
Marcono1234, line 37:
```suggestion
   * This is the default strategy used during deserialization of numbers as {@link Object}
```

Marcono1234:
Could you please not force push next time? It requires looking at all the files again to find out what changed and it does not show any of the previous review comments in the code anymore.

For a clean commit history, all the commits can then later be squashed when the pull request is merged.

lyubomyr-shaydariv:
:)

> Could you please not force push next time? It requires looking at all the files again to find out what changed and it does not show any of the previous review comments in the code anymore.

https://github.com/lyubomyr-shaydariv/gson/commits/to-number-strategy%2Bhistory

> For a clean commit history, all the commits can then later be squashed when the pull request is merged.

It depends on how the repository maintainers do merges. From what I see in the git log graph, squashes are not a must. and I'd like to keep as clean as possible.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -325,6 +327,30 @@ public GsonBuilder setFieldNamingStrategy(FieldNamingStrategy fieldNamingStrateg
...
+    this.objectToNumberStrategy = objectToNumberStrategy;
+    return this;
+  }
+
+  /**
+   * Configures Gson to apply a specific number strategy during deserialization of {@link Number}.
+   *
+   * @param numberToNumberStrategy the actual number-to-number strategy
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @see ToNumberPolicy#LAZILY_PARSED_NUMBER The default number to number strategy
```
Marcono1234, line 347:
```suggestion
   * @see ToNumberPolicy#LAZILY_PARSED_NUMBER The default number-to-number strategy
```
To be consistent with the parameter description
Or change the parameter description to not use hyphens

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -325,6 +327,30 @@ public GsonBuilder setFieldNamingStrategy(FieldNamingStrategy fieldNamingStrateg
     return this;
   }
 
+  /**
+   * Configures Gson to apply a specific number strategy during deserialization of {@link Object}.
+   *
+   * @param objectToNumberStrategy the actual object-to-number strategy
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @see ToNumberPolicy#DOUBLE The default object to number strategy
```
Marcono1234, line 335:
```suggestion
   * @see ToNumberPolicy#DOUBLE The default object-to-number strategy
```
To be consistent with the parameter description
Or change the parameter description to not use hyphens

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -362,6 +367,39 @@ static void checkValidFloatingPoint(double value) {
     }
   }
 
+  private static TypeAdapterFactory objectAdapterFactory(ToNumberStrategy objectToNumberStrategy) {
+    if (objectToNumberStrategy == ToNumberPolicy.DOUBLE ) {
```
Marcono1234, line 371:
```suggestion
    if (objectToNumberStrategy == ToNumberPolicy.DOUBLE) {
```

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link Long} or {@link Double}
+   * values depending on how JSON numbers are represented: {@link Long} if the JSON number can
+   * fit the {@link Long} range, or {@link Double} if the JSON number can fit the {@link Double}
+   * range and the value cannot be read as {@link Long}. If the parsed double-precision number
+   * results in a positive or negative infinity ({@link Double#isInfinite()}) or a NaN
+   * ({@link Double#isNaN()}) value, {@link MalformedJsonException} is thrown.
```
Marcono1234, line 63:
```suggestion
   * ({@link Double#isNaN()}) value and the {@code JsonReader} is not {@link JsonReader#isLenient() lenient},
   * a {@link MalformedJsonException} is thrown.
```

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -362,6 +367,39 @@ static void checkValidFloatingPoint(double value) {
...
+  private static TypeAdapterFactory objectAdapterFactory(ToNumberStrategy objectToNumberStrategy) {
+    if (objectToNumberStrategy == ToNumberPolicy.DOUBLE) {
+      return ObjectTypeAdapter.FACTORY;
+    }
+    return ObjectTypeAdapter.newFactory(objectToNumberStrategy);
+  }
+
+  private static TypeAdapterFactory numberAdapterFactory(final ToNumberStrategy numberToNumberStrategy) {
+    if (numberToNumberStrategy == ToNumberPolicy.LAZILY_PARSED_NUMBER) {
+      return TypeAdapters.NUMBER_FACTORY;
```
Marcono1234, line 379: Maybe prevent code duplication by removing `TypeAdapters.NUMBER_FACTORY` and instead create a class similar to `ObjectTypeAdapter` which then contains the constant and the logic of this `numberAdapterFactory` method

lyubomyr-shaydariv:
> Sorry, still some more minor things which could possibly be improved (if you like).

No problem, please provide either an inline diff (or a patch as a comment) or provide a patch on top of the "+history" [branch](https://github.com/lyubomyr-shaydariv/gson/tree/to-number-strategy%2Bhistory) as a PR into my repository (so that I could merge/cherry-pick your suggestions on top of the branch and then I could squash all these changes into this branch) addressing the unresolved comments (3 remaining as I could resolve the first two only). Thank you.

Marcono1234:
@googlebot I consent.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (C) 2018 Google Inc.
```
eamonnmcmanus, line 2: Can you use 2021 instead of 2018 here?

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -0,0 +1,99 @@
...
+  LAZILY_PARSED_NUMBER {
+    @Override public Number readNumber(JsonReader in) throws IOException {
+      return new LazilyParsedNumber(in.nextString());
+    }
+  },
+
+  /**
+   * Using this policy will ensure that numbers will be read as {@link Long} or {@link Double}
+   * values depending on how JSON numbers are represented: {@code Long} if the JSON number can
+   * fit the {@code Long} range, or {@code Double} if the JSON number can fit the {@code Double}
```
eamonnmcmanus, line 59: I would reword this to say "can be parsed as a {@code Long}, or otherwise {@code Double} if it can be parsed as a {@code Double}". The distinction being that a value like 3.0 does technically fit the Long range but can not be parsed as a Long.

gson/src/test/java/com/google/gson/ToNumberPolicyTest.java
```diff
@@ -0,0 +1,115 @@
...
+    assertEquals(new LazilyParsedNumber("10.1"), strategy.readNumber(fromString("10.1")));
+    assertEquals(new LazilyParsedNumber("3.141592653589793238462643383279"), strategy.readNumber(fromString("3.141592653589793238462643383279")));
+    assertEquals(new LazilyParsedNumber("1e400"), strategy.readNumber(fromString("1e400")));
+  }
+
+  public void testLongOrDouble() throws IOException {
+    ToNumberStrategy strategy = ToNumberPolicy.LONG_OR_DOUBLE;
+    assertEquals(10L, strategy.readNumber(fromString("10")));
+    assertEquals(10.1, strategy.readNumber(fromString("10.1")));
+    assertEquals(3.141592653589793D, strategy.readNumber(fromString("3.141592653589793238462643383279")));
```
eamonnmcmanus, line 50: You have suffix `D` here but not in `testDouble()`. I think with or without is fine, but it should be consistent.

gson/src/test/java/com/google/gson/ToNumberPolicyTest.java
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (C) 2018 Google Inc.
```
eamonnmcmanus, line 2: Copyright year 2021 here too.

gson/src/test/java/com/google/gson/functional/ToNumberPolicyFunctionalTest.java
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (C) 2018 Google Inc.
```
eamonnmcmanus, line 2: Year. :-)

gson/src/test/java/com/google/gson/functional/ToNumberPolicyFunctionalTest.java
```diff
@@ -0,0 +1,134 @@
...
+    assertEquals(null, gson.fromJson("null", Object.class));
+    assertEquals(10.0, gson.fromJson("10", Object.class));
+    assertEquals(null, gson.fromJson("null", Number.class));
+    assertEquals(10.0, gson.fromJson("10", Number.class));
+  }
+
+  public void testAsLazilyParsedNumbers() {
+    Gson gson = new GsonBuilder()
+        .setObjectToNumberStrategy(ToNumberPolicy.LAZILY_PARSED_NUMBER)
+         .setNumberToNumberStrategy(ToNumberPolicy.LAZILY_PARSED_NUMBER)
```
eamonnmcmanus, line 57: Indentation seems off here.

eamonnmcmanus (changes requested):
Hi, Gson is in maintenance mode and for the most part we are not accepting new features like this. However, given the considerable demand, I think we can make an exception here. Thanks for putting this together!

I've run this code against the tests of all of Google's internal projects and I didn't see any problems. The new API also seems good.

I did have a few minor comments.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
lyubomyr-shaydariv: Done.

gson/src/test/java/com/google/gson/ToNumberPolicyTest.java
lyubomyr-shaydariv: Done.

gson/src/test/java/com/google/gson/functional/ToNumberPolicyFunctionalTest.java
lyubomyr-shaydariv: Done.

gson/src/test/java/com/google/gson/functional/ToNumberPolicyFunctionalTest.java
lyubomyr-shaydariv: Done.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
lyubomyr-shaydariv: Done.

alperentoksoz:
@googlebot I consent.

osmantufekci:
@googlebot I consent.

ogunozdemir:
@googlebot I consent.

rmirabelle:
I don't understand this "feature" of Gson.

Assuming objects of this shape: 
```kotlin
data class MyThing(
    val answer: Any? = null
)
```

And assuming incoming JSON like so:

```json
[
    {
        "answer": "1"
    },
    {
        "answer": 1
    }
    {
        "answer": 1.0
    },
    {
        "answer": null
    }
]
```

The very obvious solution to the above is for Gson to see the target type of `Any?`, then _infer_ the destination type based on the JSON value, setting the answer values to `String`, `Int`, `Double` and `null` respectively.  This doesn't strike me as even marginally difficult.

QUITE unfortunately, Gson still doesn't do this and is happy to corrupt incoming JSON values, requiring that AFTER conversion, we manually inspect the resulting object and manually correct all of Gson's errors.  To compound this issue, I cannot craft my own custom converter either, because that too, would require selecting a single type only.

Marcono1234:
> The very obvious solution to the above is for Gson to see the target type of `Any?`, then _infer_ the destination type based on the JSON value, setting the answer values to `String`, `Int`, `Double` and `null` respectively.

When you tell Gson to deserialize a Java `Object` / Kotlin `Any` and it finds a JSON number it deserializes it by default as `Double` regardless of the format of the number. I assume for historical reasons that could not be changed which is why this PR here was proposed.

The `ToNumberStrategy` allows you to implement the functionality you want. A very simple implementation which just checks if the number contains a `.` could look like this:
```kotlin
val gson = GsonBuilder()
    .setObjectToNumberStrategy { reader ->
        // Read the JSON number as string to inspect its format
        val numberAsString = reader.nextString()

        if (numberAsString.contains('.')) numberAsString.toDouble()
            else numberAsString.toInt()
    }
    .create()
```
You could also have a look at the implementation of Gson's `ToNumberPolicy.LONG_OR_DOUBLE` which implements the same concept except for `Long` and `Double`, and also rejects non-finite values and produces more helpful exception messages.

rmirabelle:
@Marcono1234 Thanks for the very thoughtful response.

PR #1307, opened 2018-05-02
add name derive feature

fantasyRqg:
support different field's name style for deserialization


Marcono1234:
Sorry for this late response, do you think these changes are still relevant? In case you think so, could you please explain your use case a bit more in detail. If I understand the changes correctly, the intention is to allow overwriting the names specified by `@SerializedName`; I think all other functionality could already be achieved with a custom `FieldNamingStrategy`.

PR #1321, opened 2018-05-15
Issue 1242: Printing Gson version when throwing AssertionError and Il…

inder123:
…legalArgumentException

On some versions of Android (probably on some variants of the popular Samsung S4 phone), an older version of Gson is suspected to be bundled in, and gets picked up from the system classpath.
For those versions, the applications that include the latest Gson fail unexpectedly. This debug print will help confirm this issue.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -17,7 +17,7 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
+ 
   <build>
     <plugins>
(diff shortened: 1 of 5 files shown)


inder123:
@swankjesse @JakeWharton @joel-leitch 
I would like to include this PR into the new build 2.8.5. This will help us isolate problems in specific versions of Gson.

Issue #1316, opened 2018-05-10
Version number 11-ea doesn't parse


cayhorstmann:
An easy fix is to add a hyphen to the split regex:

`    String[] parts = System.getProperty("java.version").split("[._-]");`

Marcono1234:
Was probably fixed by #1311. `JavaVersion.getMajorJavaVersion("11-ea")` returns 11.

PR #1322, opened 2018-05-16
Use java.class.version to detect Java version (#1316)

httpdigest:
It has a much more consistent format compared to java.version and
therefore does not require to handle all corner cases of JVMs reporting
the Java version in any arbitrary way they want.

This also makes the JavaVersionTest somewhat obsolete, since no
actual parsing is necessary anymore.


httpdigest:
I signed it!

gson/src/main/java/com/google/gson/internal/JavaVersion.java
```diff
@@ -26,51 +26,19 @@
...
-      }
-      return Integer.parseInt(num.toString());
-    } catch (NumberFormatException e) {
-      return -1;
+    int version = (int) Double.parseDouble(System.getProperty("java.class.version"));
+    switch (version) {
+      case 46: return 2;
+      case 47: return 3;
+      case 48: return 4;
+      case 49: return 5;
```
JakeWharton: Gson doesn't run on these versions so I don't see any reason to include them.

httpdigest:
> We should use multi-release jars to check Runtime.Version on Java 9 or newer which is the only actually supported mechanism for doing this.

Alright. Fine by me. Closing then.

gson/src/main/java/com/google/gson/internal/JavaVersion.java
```diff
@@ -26,51 +26,15 @@
...
-          num.append(c);
-        } else {
-          break;
-        }
-      }
-      return Integer.parseInt(num.toString());
-    } catch (NumberFormatException e) {
-      return -1;
+    int version = (int) Double.parseDouble(System.getProperty("java.class.version"));
+    switch (version) {
```
inder123: Great... Will this work on Android?

httpdigest: Looks like it: https://developer.android.com/reference/java/lang/System.html#getProperties()

httpdigest: Though I would rather use Reflection-based feature detection (depends on what Gson actually is using the Java version for) rather than querying the Java version, because Android is not Java and does not have to adhere to a version of the Java or Virtual Machine Language Specification, so they can write what they want in that property.

Issue #1291, opened 2018-04-17
getAsString with an optional string argument in JsonObject


faissaloo:
So I don't have to use ```.get("tag").getAsString()``` and have to deal with ```.get("tag")``` being null.

lyubomyr-shaydariv:
I would cause API bloating: this would require the entire `get***(name)` and `is***(name)` method family to be replicated in `JsonObject` and `get***(index)` and `is***(index)`  to be replicated in `JsonArray` respectively. Having `.get(name)`/`.get(index)` is simpler to design and guarantees that you can always access whole `JsonElement` API not worrying that some APIs are not replicated to `JsonObject` or `JsonArray`.

philipwhiuk:
I agree it causes API bloating however I disagree that it’s a problem here. The point of the API is to provide easy access to JSON.

Not implementing this just causes implementors to reimplement the null check themselves - which is obvious sub par compared to a tested library implementation.

It would be nice if Java had an ability y where you could mix in the extra methods if you wanted the convenience. But we don’t have that in the language so we are forced to compromise.

My personal use case is providing this as an API for downstream developers to access custom defined structures so this is a common case which I’d like to improve to provide an API that is more accessible. If forking is the only way to implement it that would be annoying.

PR #1332, opened 2018-06-06
Get member value as a string - helper method

philipwhiuk:
This is a proposal for adding helper methods that remove boiler plate when getting values from objects.

Obviously it needs tests and probably other similar helpers for other value types, but I'm interested in feedback on the approach.


NightlyNexus:
This would then open the door for more convenience methods like getAsInt, getAsBoolean, etc.
That's too much overhead for little benefit.

JakeWharton:
Let's keep the discussion in #1291

PR #1337, opened 2018-06-28
Add possibility to maintain type variable in created pojo

lucalas:
Instead of completely remove type variable, give the possibility to decide if maintain type value (if pojo has type variable declaration) or not.


lucalas:
did it!

lucalas:
I signed it!

lucalas:
Used wrong account to commit.
I'll recheck and add new pull request.

PR #1349, opened 2018-07-16
add .gitattributes

jf17:
is not HTML library .

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+gson/docs/javadocs/* linguist-documentation
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-gson/docs/javadocs/* linguist-documentation
+*.html linguist-language = Java
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-*.html linguist-language = Java


jf17:
@googlebot I signed it!

.gitattributes
```diff
@@ -0,0 +1 @@
+*.html linguist-language = Java
```
lyubomyr-shaydariv: I think that mimicking all HTML files as Java source code is not a very good idea. According to the Linguist [documentation](https://github.com/github/linguist#documentation), the following line looks like a more correct way:

```
gson/docs/javadocs/* linguist-documentation
```

jf17:
@lyubomyr-shaydariv look good, thanks!

lyubomyr-shaydariv:
By the way,

* You don't need to have both 4790307eef5be79eff0bc7517754a4b4916134b2 and a6e6c61de10f15ed2272c2bfdcec7a38ddc74f79 commits. Could you please simply combine both commits into a single one (by amending the initial commit or rebasing the two using fixup/squash)?
* I would preserve the commit message from the second commit. I think that its ```Use the linguist-documentation attribute to mark paths as documentation.``` is more accurate.

Issue #566, opened 2015-03-19
Unmarshalling fails when GSON library is used in a managed environment with custom SecurityManager installed


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Try to set the default security manager (without adding any additional 
permissions) 
2. Try to unmarshall a simple JSON string to a Java object which include access 
to private fields

What is the expected output? What do you see instead?
The JSON string should be unmarshalled without any problems but 
AccessControlException is thrown due to missing permissions for using 
reflection and accessing the context classloader within the GSON library. 

What version of the product are you using? On what operating system?
Issue is not dependent currently on the version of the library and OS 
(reproduced with 2.2.5) 

Please provide any additional information below.
The issue occurs the following scenario: The GSON library is used as an OSGi 
bundle installed as a system bundle (meaning it has AllPrivileges). When 
another OSGi bundle tries to use the GSON system bundle it fails if it does not 
have additional permissions (such as ReflectPermissions that allow the GSON 
library to do the unmarshalling) - this is a result of the intersection of the 
protections domains of the application bundle and the GSON bundle (implied by 
the security architecture of the JDK). A patch for the library is also attached 
with unit tests that provide an overview of the issue outside the context of 
OSGi (or another managed environment such as J2EE server, Applet [2, 3], Google 
AppEngine [1] - issue will be reproducible in such a context as well). 
References: 
[1] 
http://stackoverflow.com/questions/3002714/gson-on-google-app-engine-throws-a-se
curity-exception 
[2] 
http://stackoverflow.com/questions/2788017/reflection-permission-problems-when-u
sing-the-gson-library-in-a-applet 
[3] 
http://stackoverflow.com/questions/3850859/java-using-gson-in-an-applet-causes-s
ecurityexception 
[4] https://code.google.com/p/google-gson/issues/detail?id=344 
[5] https://groups.google.com/forum/#!topic/google-gson/2TKlp619p2w 
[6] https://code.google.com/p/google-gson/issues/detail?id=445 

```

Original issue reported on code.google.com by `marto8...@gmail.com` on 18 Mar 2014 at 6:11

Attachments:
- [gson.diff](https://storage.googleapis.com/google-code-attachments/google-gson/issue-566/comment-0/gson.diff)

JochenHiller:
Is there any plan IF and WHEN to integrate this diff into the GSON library?
We are using GSON in an environment with custom security manager installed.

inder123:
[...]

PR #1361, opened 2018-07-30
Proposal fix for #566

hi5san:
Hi @inder123 and @JochenHiller.  I've run into a similar problem to Issue #566 running gson under a default SecurityManager enabled runtime environment and like to propose a fix PR for it.

The fix is simply by surrounding certain portions of code with privileged access where was missing and needed (like the ones proposed in #566 [gson.diff](https://storage.googleapis.com/google-code-attachments/google-gson/issue-566/comment-0/gson.diff), just needed to apply changes to match latest code).  I've also configured unit tests to run under default SecurityManager + Maven surefire + Junit4 which is the current unit test configuration for gson (needed to add java.policy to test resources and surefire test settings in pom.xml) and all 1050 tests pass.

Could you please review and see if it looks good or have any concerns needing some work?  Would be happy to fix as needed.  Either way, it would be great if a fix could be pulled in so that gson would work under security manager without any special hacks (i.e. - giving extra permissions to application code).  Thanks.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -20,6 +20,8 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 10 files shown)


inder123:
Thanks for the PR. However, an easier way would be to just add default TypeAdapter for all java.* types so that we don't need privileged action. What do you think?

hi5san:
@inder123 Thanks for your review comments.  I haven't read in-depth how Gson internally uses pre-builtin TypeAdapters but I imagine making it capable of handling all java.* types may provide additional efficiency for the Gson mapping process and provide a fix for code using Gson to map java.* type classes.  However, would it work as a fix for code with application or user defined classes (non-java.* types)?  If yes, I agree 👍.
If not -  I think the intended issue addressed in #566 is that working codes using Gson all sudden do not work when security manager is enabled with default policy.   Of course, documenting users to define a TypeAdapter especially If running under secmgr is an option, but would be ideal if the same code works in both managed (sec.mgr enabled) and non-managed (same disabled) runtime.  What do you think?

eamonnmcmanus:
Closing since issue #566 has been closed.

PR #1362, opened 2018-07-30
Simplify maintainType logic.

NightlyNexus:
When we maintain the label value, we do not need to make a new JsonObject and copy over the keys and values when writing. The ordering will change, though. Before this change, it always put the label first.


extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
```diff
@@ -241,15 +241,18 @@ private RuntimeTypeAdapterFactory(Class<?> baseType, String typeFieldName, boole
               + "; did you forget to register a subtype?");
         }
         JsonObject jsonObject = delegate.toJsonTree(value).getAsJsonObject();
+
+        if (maintainType) {
+          return jsonObject;
```
JakeWharton: Should this be something like
```java
Streams.write(jsonObject, out);
return;
```

NightlyNexus: oops haha duh

PR #1363, opened 2018-07-30
Deprecate JsonElement.getAsCharacter.

### LINKED ISSUE #1355: JsonElement#getAsCharacter()
JsonElement 
public char getAsCharacter()
“convenience method to get this element as a primitive character value.”

I expected calling getAsCharacter would convert an integer to its character value. Instead, it returns the first character of the toString().

```
JsonPrimitive jp = new JsonPrimitive(100);
System.out.println(jp.getAsCharacter());     Output: 1
System.out.println((char) jp.getAsInt());    Output: d
```

Perhaps the documentation should be more descriptive.

NightlyNexus:
Closes #1355.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonElement.java
+++ b/gson/src/main/java/com/google/gson/JsonElement.java
@@ -259,14 +259,17 @@ public abstract class JsonElement {
   }
 
   /**
-   * convenience method to get this element as a primitive character value.
+   * convenience method to get the first character of this element as a string or the first
+   * character of this array's first element as a string.
    *
-   * @return get this element as a primitive char value.
+   * @return the first character of the string.
    * @throws ClassCastException if the element is of not a {@link JsonPrimitive} and is not a valid
-   * char value.
+   * string value.
    * @throws IllegalStateException if the element is of the type {@link JsonArray} but contains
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/JsonElement.java
```diff
@@ -259,14 +259,17 @@ public byte getAsByte() {
...
    * @throws ClassCastException if the element is of not a {@link JsonPrimitive} and is not a valid
-   * char value.
+   * string value.
    * @throws IllegalStateException if the element is of the type {@link JsonArray} but contains
    * more than a single element.
    * @since 1.3
+   * @deprecated This method is misleading, as it does not get this element as a char but rather as
+   * a string's first character.
    */
   public char getAsCharacter() {
```
JakeWharton: I'm still wary of only using the Javadoc tag. Can you include the annotation as well?

NightlyNexus: yep.

NightlyNexus: i can't add a description in the annotation, so i'll leave the javadoc tag, too, if that's alright.

PR #1375, opened 2018-08-23
Field,  parameter,  and  variable  could  be  final

reudismam:
The final modifier can be used in fields, parameters, and local variablesto indicate they cannot be reassigned. This edit improves clarity and it helps with debugging since it shows what values will change at runtime.

### DIFF
--- a/metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
+++ b/metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
@@ -107,13 +107,13 @@ public final class ParseBenchmark extends SimpleBenchmark {
     parser = api.newParser();
   }
 
-  public void timeParse(int reps) throws Exception {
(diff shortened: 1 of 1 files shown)


metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
```diff
@@ -107,13 +107,13 @@ private Document(TypeToken<?> typeToken, TypeReference<?> typeReference) {
     parser = api.newParser();
   }
 
-  public void timeParse(int reps) throws Exception {
+  public void timeParse(final int reps) throws Exception {
```
JakeWharton: We don't mark parameters as final since it's prohibitively noisey. IntelliJ IDEA 2018.2 will highlight locals and parameters differently depending on whether you overwrite them or not and we think that's good enough.

PR #1401, opened 2018-10-10
Fix JPMS module setup (fixes #1315)

### LINKED ISSUE #1315: Move module-info.java to /src/main/java
To build gson as a JPMS module, the `module-info.java` file needs to be in `src/main/java`, not in `src/main/java/com.google.gson`. 

I checked [https://guides.gradle.org/building-java-9-modules/](https://guides.gradle.org/building-java-9-modules/) for confirmation.

With that change, and with these changes in build.gradle, the JAR builds fine as a modular JAR.

```
sourceCompatibility = 9
targetCompatibility = 9
// sourceSets.main.java.exclude("**/module-info.java")
```

pietvandongen:


pietvandongen:
Aw, right, I'll do a new one that targets 1.6 and up then. Thanks!

PR #1402, opened 2018-10-16
Fix JPMS module setup (fixes #1315)

### LINKED ISSUE #1315: Move module-info.java to /src/main/java

pietvandongen:
I've followed the instructions from the Apache Maven Compiler Plugin team here: https://maven.apache.org/plugins/maven-compiler-plugin/examples/module-info.html

This should ensure compatibility with Java 6+ while providing module info for Java 9+.

Also, in `gson/src/main/java/com/google/gson/Gson.java` I've removed calls to a constructor only introduced in Java 1.7, which broke backwards compatibility.


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -705,7 +705,7 @@ public void toJson(Object src, Type typeOfSrc, JsonWriter writer) throws JsonIOE
     } catch (IOException e) {
       throw new JsonIOException(e);
     } catch (AssertionError e) {
-      throw new AssertionError("AssertionError (GSON " + GsonBuildConfig.VERSION + "): " + e.getMessage(), e);
+      throw new AssertionError("AssertionError (GSON " + GsonBuildConfig.VERSION + "): " + e.getMessage());
```
inder123: why are you changing these?

pietvandongen:
Because that constructor was only introduced in 1.7, which would make it not compatible with 1.6.

gson/src/main/java/com/google/gson/Gson.java
pietvandongen: Because that constructor was only introduced in 1.7, which would make it not compatible with 1.6.

inder123: Ah! Thanks.

JakeWharton: Store the exception in a local and call `initCause(e)` so as not to break the chain of exceptions before throwing.

pietvandongen: Isn't that beyond the scope of this pull request?

JakeWharton: Nope. You're already changing these so I would either leave them as-is to retain the cause or make them correct for Java 6 while retaining the cause.

pietvandongen: OK, agreed. If I revert this change, the build will fail. I do not completely understand what exactly you expect me to do using `initCause`. Could you spell it out for me so I can commit the change?

inder123: I think this is what Jake meant:
```
AssertionError error = new AssertionError("AssertionError (GSON " + GsonBuildConfig.VERSION + "): " + e.getMessage());
error.initCause(e); 
throw error;
```

JakeWharton: Precisely! Just so we don't actually lose the causal chain.

pietvandongen: Ahh, got it, thanks! I've fixed it and pushed those changes.

pom.xml
```diff
@@ -63,13 +63,28 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.7.0</version>
+          <version>3.8.0</version>
+          <executions>
+            <execution>
+              <id>default-compile</id>
+              <configuration>
+                <release>9</release>
```
inder123: Seems like this requires JDK 9 for compilation? 
I can no longer compile with JDK 8. I get this error:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gson: Fatal error compiling: invalid flag: --release -> [Help 1]

How do I fix this?

JakeWharton: Module info require JDK 9 or newer. While 8 isn't EOL yet, it will be in 5 months. You can grab a JDK 11 build here: https://adoptopenjdk.net/

inder123: Thanks Jake. I have not upgraded beyond JDK 8 so far because of my focus on Android. 
I guess time to upgrade!

inder123: Have you used JDK 9 or 10 with Android? And does it work?

JakeWharton: JDK 11 is my default, yes. It mostly just works fine. And wherever it doesn't it's a bug in the tooling.

PR #1404, opened 2018-10-17
Replace explicit synthetic field check with bitmask.

ehenoma:
Lower the cyclomatic complexity by adding the *synthetic* modifier initially to the `modifiers` value. That way, the method can keep its truthfulness while lowering the amount of statements. Currently, the fields bitmask is checked twice, wheras the last check is a call to the `isSynthetic()`method of the Field class. With this little addition, just the first check is required.

```java
static final int SYNTHETIC = 0x00001000;
```

^ *Integer constant in the Modifier class.*

The reason why Javas `Modifier` class doesn't publish its bitmask
for the *synthetic* modifier, is that there is no such keyword in
the language:

```
Bits not (yet) exposed in the public API either because they
have different meanings for fields and methods and there is no
way to distinguish between the two in this class, or because
they are not Java programming language keywords
```

^ *The original comment (Modifier.java Line 333 - 336).*

As you can see, the fact that the original constant is not public should not prevent us from using it.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/Excluder.java
+++ b/gson/src/main/java/com/google/gson/internal/Excluder.java
@@ -50,9 +50,10 @@ import java.util.List;
 public final class Excluder implements TypeAdapterFactory, Cloneable {
   private static final double IGNORE_VERSIONS = -1.0d;
   public static final Excluder DEFAULT = new Excluder();
+  private static final int MODIFIER_SYNTHETIC = 0x00001000;
 
   private double version = IGNORE_VERSIONS;
-  private int modifiers = Modifier.TRANSIENT | Modifier.STATIC;
+  private int modifiers = Modifier.TRANSIENT | Modifier.STATIC | MODIFIER_SYNTHETIC;
   private boolean serializeInnerClasses = true;
   private boolean requireExpose;
   private List<ExclusionStrategy> serializationStrategies = Collections.emptyList();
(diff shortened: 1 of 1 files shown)


lyubomyr-shaydariv:
> the fact that the original constant is not public should not prevent us from using it.

I think, Gson should use public APIs only since non-public things are possibly unstable and may be a subject to change in the future (`0x1000` as of Java 6 though which Gson uses as the minimum required Java version; Dalvik uses the [same value](https://source.android.com/devices/tech/dalvik/dex-format)). `isSynthetic` and `isEnumConstant` are the only modifier-related methods exposed from the `Field` class, and none of `Modifier.SYNTHETIC` and `Modifier.ENUM` are public. I guess there's some design choice behind it, so I would stick to the public API only (unless the constant is made public and Gson drops the older Java versions support (or detects, in runtime, whether the constant is made public)).

merlinosayimwen:
push

Marcono1234:
Is there any reason for this other than lowering the cyclomatic complexity? It appears the disadvantage of this would be that users who use `GsonBuilder.excludeFieldsWithModifiers​` overwrite the default bitmask and therefore would by default include synthetic fields with this change.

eamonnmcmanus:
I'm not concerned that the `SYNTHETIC` bitmask might change, but I agree with @Marcono1234 that this change seems potentially wrong in conjunction with `excludeFieldsWithModifiers`. Or at least this is a behaviour change. While we could probably fix that, I'm not sure it would be worth it.

ehenoma:
This PR is quite old now and it is very small so I don't think it needs that much attention and discussion anymore. 
Of course given that users may change the excluded modifiers this could make synthetic fields included and thus 
produce bugs in certain code.

The idea behind this PR could still be applied by adding the flag to every custom bitmask that is passed in but 
I don't think there is any real benefit since, after optimization, there is probably no performance difference.

PR #1413, opened 2018-10-26
Display null-handling details as sublist

jeremy-w:
Indent the details of null handling.

Comparing against the [old, Google Sites user guide][gsites] shows that the two items following the "handles nulls correctly" item are intended as a sublist under that item.

[gsites]: https://sites.google.com/site/gson/gson-user-guide#TOC-Finer-Points-with-Objects


jeremy-w:
I signed it!

Issue #1417, opened 2018-11-06
Stack Overflow exception when serializing proto map fields


marcoferrer:
The ```ProtoTypeAdapter``` has problems serializing messages with fields of type map. It might have something to do with the way map is represented in the proto field descriptor.

Marcono1234:
In case this is still relevant, could you please test the latest Gson version, with the latest version of the `proto` module? And in case you still experience this problem, could you please provide a small self contained example showing the problem?

PR #1418, opened 2018-11-06
bug fix for serializing messages with fields of type map

marcoferrer:
This pr addresses https://github.com/google/gson/issues/1417

Unit tests still need to be wrapped up

### DIFF
--- a/proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
+++ b/proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
@@ -36,6 +36,7 @@ import com.google.protobuf.Descriptors.EnumValueDescriptor;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.Extension;
+import com.google.protobuf.MapEntry;
 import com.google.protobuf.Message;
 
 import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


proto/src/main/java/com/google/gson/protobuf/ProtoTypeAdapter.java
```diff
@@ -240,7 +241,19 @@ public JsonElement serialize(Message src, Type typeOfSrc,
...
+          JsonObject mapObject = new JsonObject();
+          @SuppressWarnings("unchecked")
+          Collection<MapEntry<Object,Object>> entries =
+                  (Collection<MapEntry<Object, Object>>) fieldPair.getValue();
+          for(MapEntry<Object,Object> entry : entries){
+            String key = context.serialize(entry.getKey()).getAsString();
+            mapObject.add(key, context.serialize(entry.getValue()));
+          }
+          ret.add(name, mapObject);
+        }else{
```
Marcono1234, line 254:
```suggestion
        if (fieldPair.getKey().isMapField() && fieldPair.getValue() instanceof Collection) {
          JsonObject mapObject = new JsonObject();
          @SuppressWarnings("unchecked")
          Collection<MapEntry<Object, Object>> entries =
                  (Collection<MapEntry<Object, Object>>) fieldPair.getValue();
          for (MapEntry<Object, Object> entry : entries) {
            String key = context.serialize(entry.getKey()).getAsString();
            mapObject.add(key, context.serialize(entry.getValue()));
          }
          ret.add(name, mapObject);
        } else {
```

proto/src/test/java/com/google/gson/protobuf/functional/ProtosWithMapTypesTest.java
```diff
@@ -0,0 +1,42 @@
...
+  protected void setUp() throws Exception {
+    super.setUp();
+    gson = new GsonBuilder().registerTypeHierarchyAdapter(
+      GeneratedMessage.class, ProtoTypeAdapter.newBuilder()
+          .setEnumSerialization(EnumSerialization.NUMBER)
+          .build())
+      .create();
+  }
+
+  //TODO
```
Marcono1234, line 41: Do you have a test which shows that this works as expected?

Note that I am not a Maintainer, and it could happen that even if you change this pull request it might not be merged (in the near future).

PR #1419, opened 2018-11-07
Delete LICENSE

Lukety88:


inder123:
Why do we want to delete this? I understand that the license is available in GitHub anyway, but having an explicit LICENSE file is not a bad idea.

Lukety88:
Sorry I was trying to figure out how to delete api put on my phone by my gf
while I was sleeping.

PR #1422, opened 2018-11-10
Infer type in generic instance creation.

reudismam:
This edit replaces type parameters to invoke the constructor of a generic class with an empty set (<>), diamond operator and allow inference of type parameters by the context. This edit ensures the use of generic instead of the deprecated raw types.


JakeWharton:
This projects builds with JDK 6 (or `-release 6`) which does not allow the diamond operator.

PR #1423, opened 2018-11-10
Use the class Path to represent a path instead of a String.

reudismam:
Strings can be used to represent a file system path even though some classes are specifically designed for this task. For instance, java.nio.Path. It is useful to change the type of the variable to Path, since strings can be combined in an undisciplined way, which can lead to invalid paths. Second, different operating systems use different file separators, which can cause bugs.

### DIFF
--- a/metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
+++ b/metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
@@ -31,6 +31,8 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringWriter;
 import java.lang.reflect.Type;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


reudismam:
Paths.get throw an Exception on an invalid path. In this way, we are sure that the path is ok.

JakeWharton:
But if the is invalid either syntactically or semantically we receive a null and immediately throw anyway.

Marcono1234:
Probably obsolete due to the refactoring done by #2019. Also I am not sure if using `Path.toString()` for the `getResourceAsStream` call would have been correct. These two represent two separate concepts; `Path` is a file system path which on Windows uses `\` as separator whereas `getResourceAsStream` reads a resource whose path is always using `/`.

eamonnmcmanus:
Indeed, this change seems both obsolete and incorrect. According to the documentation for [`ClassLoader.getResource`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ClassLoader.html#getResource(java.lang.String)):
> The name of a resource is a '/'-separated path name that identifies the resource.

So, as @Marcono1234 says, using `Path` would be wrong.

PR #1428, opened 2018-11-17
Refactor JsonParser to use static access

Degubi:
Deprecate the default and only constructor of the JsonParser class and make the parser methods statically accessible. Because the constructor is only deprecated, no code change is required. The constructor is marked "forRemoval", although it is not a requirement to keep the "forRemoval" marked as true.
Edit: Removed forRemoval because of Java8


Degubi:
I signed it!

inder123:
While this may be a good idea, we don't accept breaking changes in this project.

Degubi:
Is this a breaking change at the bytecode level? Because at source level it works the old way and the new way too.

PR #1430, opened 2018-11-19
Add static method alternatives to JsonParser

Degubi:
Alternative request to add static methods to JsonParser. This way no bytecode changes. I didn't mark the constructor as deprecated this time to avoid warnings.  The PR adds 3 new static methods to JsonParser and refactors the existing 3 instance methods to use the static ones.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonParser.java
+++ b/gson/src/main/java/com/google/gson/JsonParser.java
@@ -42,7 +42,7 @@ public final class JsonParser {
    * @since 1.3
    */
   public JsonElement parse(String json) throws JsonSyntaxException {
-    return parse(new StringReader(json));
+    return parseString(json);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


asus246:
ok

inder123:
What is the key issue you are trying to solve? 
Is it the creation of the JsonParser object that is bothering you? Or is it a deeper philosophical issue.

Degubi:
I just don't like the look of it. And the need to create a new parser object every time. Also, the method names "parse" vs "parseString" I can see that I'm parsing the Json from a string, not from a Reader.

Marcono1234:
@Degubi why did you close this pull request?

You were definitely not the only one with this opinion, see [this StackOverflow comment](https://stackoverflow.com/questions/4110664/gson-directly-convert-string-to-jsonobject-no-pojo#comment30505372_4527044)

Degubi:
@Marcono1234 I reopened it with better deprecations, see #1513 . I really think it they should be statics, but I'm not sure about others thinking the same... (See the discussion in #1513)

Issue #1440, opened 2018-12-07
RuntimeTypeAdapterFactory isn't serializing the type


rjcarr:
I've happily used gson for years (thanks!) and needed to serialize and deserialize some polymorphic types and came across this *extra* which seems exactly what I need:

https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java

But no matter what I try it isn't serializing the type field, so when it tries to deserialize, naturally, it gives this error:

> cannot deserialize class <...> because it does not define a field named type

Even using the demo code from the file doesn't work for me:

    class Shape {
      int x;
      int y;
    }

    class Diamond extends Shape {
      int width;
      int height;
    }

    ...

    Gson g = (new GsonBuilder()).registerTypeAdapterFactory(
    	RuntimeTypeAdapterFactory
    		.of(Shape.class)
    		.registerSubtype(Diamond.class)
    ).create();

    Diamond d = new Diamond();
    d.x = 10;
    d.y = 20;
    d.width = 25;
    d.height = 35;

    String json = g.toJson(d);

Which produces:

> {"width":25,"height":35,"x":10,"y":20}

Note I've tried specifying custom strings for *of()* and *register()*, adding the type to the *toJson()* call, and even embedding the object it an outer collection, but it makes no difference.

Actually, there is no difference in output between a standard *Gson* object and this one created with the builder and type factory.

I notice if I debug the code and set breakpoints in the *RuntimeTypeAdapterFactory* read and write methods, it will hit the read() (and then get to the failure above), but it never hits the write method.  Just totally skips it.

What is going on here?  The code I'm using is all the latest and directly from that above link (as of today).

lyubomyr-shaydariv:
You must serialize the `d` object with `g.toJson(d, Shape.class)` [because of](https://github.com/google/gson/blob/da5cae371e8b739fe63a6c6d16debf7b297dea0e/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java#L196):

```
if (type.getRawType() != baseType) {
```

rjcarr:
@lyubomyr-shaydariv Thanks so much for the quick response!  And yes, that fixed the problem!  I had tried `g.toJson(d, Diamond.class)` but that made no difference so using `Shape` the `baseType` was the key.  Maybe that info could be put into the notes?  Thanks again!

Marcono1234:
[...]

PR #1441, opened 2018-12-08
The JavaDoc is now more accurate about the type adapter serialization and deserialization

lyubomyr-shaydariv:
This should fix `RuntimeTypeAdapterFactory` JavaDoc missing information for proper serialization and deserialization, originally raised in #1440.

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
@@ -120,6 +120,18 @@ import com.google.gson.stream.JsonWriter;
  *       .registerSubtype(Circle.class)
  *       .registerSubtype(Diamond.class);
  * }</pre>
+ *
+ * <h3>Serialization and deserialization</h3>
+ * In order to serialize and deserialize a polymorphic object,
+ * you must specify the base type explicitly.
+ * <pre>   {@code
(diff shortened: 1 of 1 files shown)


extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
```diff
@@ -120,6 +120,18 @@
  *       .registerSubtype(Circle.class)
  *       .registerSubtype(Diamond.class);
  * }</pre>
+ *
+ * <h3>Serialization and deserialization</h3>
+ * In order to serialize and deserialize a polymorphic object,
+ * you must specify the base type explicitly.
+ * <pre>   {@code
+ *   Diamond diamond = new Diamond();
```
inder123: serialization does not require the type, only deserialization

lyubomyr-shaydariv: @inder123 hm, there must be something wrong then. Please see the original issue #1440 that raised this PR. Additionally all tests in `RuntimeTypeAdapterFactoryTest.java` specify the type explicitly on serialization.

inder123: Ah, ok. This is for RuntimeTypeAdapterFactory not core Gson.

PR #1442, opened 2018-12-11
Add j2objc annotations to fix memory leak in transpiled objective-c 

iousin:
The issue has been discussed on j2objc page here, j2objc team recommended to create a PR to be merged into mainstream gson codebase. 
https://github.com/google/j2objc/issues/921

The guava project already has j2objc annotations. 
These annotations will fix memory leak related to LinkedTreeMap when transpiled to objective-c code using j2objc. 

Thanks.

### DIFF
--- a/gson/build.gradle
+++ b/gson/build.gradle
@@ -9,5 +9,6 @@ targetCompatibility = 1.6
 
 sourceSets.main.java.exclude("**/module-info.java")
 dependencies {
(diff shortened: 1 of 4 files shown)


inder123:
We don't accept any dependency on external jars

catslink:
@inder123  I just wonder if this kind memory leak is fixed in latest Gson version? I also meet the leak in 2.8.0

Issue #1345, opened 2018-07-09
Add hierarchy to the RuntimeTypeAdapterFactory


manfred4:
In the gson extra class **RuntimeTypeAdapterFactory** I want to use a similar behavior to the example above the code, except that I use interfaces with multiple Layers of Hierarchy. This causes in the _create_ method that the raw type does not equal the base type. To fit my need the first line in that method needs to be changed from
`if (type.getRawType() != baseType) {`
to 
`if (!baseType.isAssignableFrom(type.getRawType())) {`.

This can be a new feature of this class, if it is desired to be handled like this.

Marcono1234:
It looks like this was resolved by #2160.

PR #1458, opened 2019-01-18
Allow hierarchical Interfaces to be serialized

manfred4:
If the Shape class from the example above the code would implement an interface named TwoDimensionalObject with the fields x and y and some Shape should be serialized in a class that only knows that its content is a TwoDimensionalObject, this change allows that use case to work. Also, any hierarchical interfaces will work with this improvement.

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
@@ -193,7 +193,7 @@ public final class RuntimeTypeAdapterFactory<T> implements TypeAdapterFactory {
   }
 
   public <R> TypeAdapter<R> create(Gson gson, TypeToken<R> type) {
-    if (type.getRawType() != baseType) {
+    if (!baseType.isAssignableFrom(type.getRawType())) {
       return null;
     }
 


manfred4:
See also my other opened Issue #1345 .

inder123:
Please add a test that indicates the problem being fixed. Thanks.

lyubomyr-shaydariv:
I'm not sure, but the original class check seems to compare against the specific base class and apply the type adapter only and only if the base class is supposed to serialize and deserialize, whereas you can define more specific type adapters for its subclasses. I guess this is a rationale behind the original design of the type adapter factory.

manfred4:
In my example I got an interface as the top level class that i want to serialize with the additional type field with multiple levels of hierarchy, whereas the individual classes are all just implementations for a specific interface for that class.
In the example in the class itself this would represent Shape being an interface with Circle, Rectangle and Diamond as interfaces extending the Shape interface. The classes serialized are then called CircleImpl, RectangleImpl and DiamondImpl and are just implementations of their corresponding interfaces.
In that case, the base class would be the shape interface and the raw type of the class that should be serialized would be the CircleImpl for example. So they just do not match as the original check asks and the adapter is not created.
I might add a test in the future, but if anyone has this repository set up anyway, feel free to check this example.

Marcono1234:
It lookes like this was resolved by #2160, thanks nonetheless for the pull request!

PR #1473, opened 2019-02-20
Added Adapter Layering Feature (Java 6 Compatible)

paulkass:
**Why this is useful:** Suppose you have a class with a lot of fields. Most of them can be easily interpreted by the Gson reader, possibly with a field naming strategy. But some have such weird names that the field naming strategy might not necessarily properly match them to the target properties. Especially if the JSON comes from a third party (like an API), the best course of action as of today is to create your own `TypeAdapter` and manually map *all* of the fields, including the ones that GSON could map on it's own. This is why I added the "Fill-In" feature. The `registerTypeAdapterWithFillIn` method allows you to manually define only the fields that Gson can not map properly, while not having to map trivial mappings.

**Implementation Details:** The way to register a custom adapter with Fill-in is through the `GsonBuilder#registerTypeAdapterWithFillIn` method. This method wraps the adapter into a constructor object that is then passed to an instance of the `ReflectiveTypeAdapterFactory.Adapter` class. In other words, when the `read` method is called on a `ReflectiveTypeAdapterFactory.Adapter` instance, the `ObjectContstructor` object calls the custom adapter and generates an instance of the target class whose fields are written to or overwritten (see the Conditions section).

When calling the `create` method on the `ReflectiveTypeAdapterFactory` responsible for producing a fill-in adapter, it will only return the adapter when the target type is or is a subclass of the desired target class. This is unlike the usual `ReflectiveTypeAdapterFactory` instances which serve to create a catch-all type adapter for non-standard classes.

In order to support the construction of instances while reading from the input `JsonReader`, the `InstanceCreator` and `ObjectConstructor` have methods that accept a `JsonReader` object. If not overridden by the implementation, they default to the normal `construct` method (for backwards compatibility). 

I wrote some tests in the `test/.../gson/internal/bind/LayeredAdapterTest.java` class to demonstrate and test some common use cases that I thought of for this feature.

**Notes:** 
* The outside adapter field will overwrite any field values previously written to in the instance returned by the user-defined Adapter. To avoid this, add the `@Expose` annotation to the fields that you want to be protected from overwriting. 
* ~~Due to the use of `default` methods in interfaces, this only supports java versions of 8+, and **will fail the travis-ci check**. I would really appreciate if someone pointed out a way to achieve the same thing in a backwards-compatible way without relying on Java 8 features. Also, if there is going to be a multi-release jar as suggested in #1469, this can go in the java 8+ jar.~~ I made this compatible with Java 6. See Below.

**Java 6 Compatibility:** I added two wrapper classes, `InstanceCreatorWrapper` and `ObjectConstructorWrapper`,  that wrap the `InstanceCreator` and `ObjectConstructor` interfaces respectively. Therefore, it's backwards compatible since these creators and just implementations of their respective interfaces. However, when the fill-in part of the code tries to use them to create an instance from the contents of the `JsonReader` object, it will case the objects returned by the `ConstructorConstructor` to the proper Wrapper, and use the special methods of those classes to create the object.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -16,6 +16,10 @@
 
 package com.google.gson;
 
+import com.google.gson.internal.ConstructorConstructor;
+import com.google.gson.internal.bind.ReflectiveTypeAdapterFactory;
+import java.io.IOException;
+import java.io.Reader;
 import java.lang.reflect.Type;
 import java.sql.Timestamp;
 import java.text.DateFormat;
@@ -32,6 +36,7 @@ import com.google.gson.internal.bind.TreeTypeAdapter;
 import com.google.gson.internal.bind.TypeAdapters;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
+import java.util.function.Function;
 
 import static com.google.gson.Gson.DEFAULT_COMPLEX_MAP_KEYS;
 import static com.google.gson.Gson.DEFAULT_ESCAPE_HTML;
@@ -509,6 +514,38 @@ public final class GsonBuilder {
       factories.add(TypeAdapters.newFactory(TypeToken.get(type), (TypeAdapter)typeAdapter));
     }
     return this;
+  }
+
+  /**
+   * TODO: Add a meaningful description
+   */
+  public GsonBuilder registerTypeAdapterWithFillIn(Type type, Object objectAdapter) {
+   $Gson$Preconditions.checkArgument(objectAdapter instanceof TypeAdapter<?>);
+   TypeAdapter<?> typeAdapter = (TypeAdapter<?>) objectAdapter;
+
+   factories.add(new TypeAdapterFactory() {
+     @Override
+     public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+       Function<JsonReader, T> function = new Function<JsonReader, T>() {
+         @Override
+         public T apply(JsonReader reader) {
+           try {
+             return (T) typeAdapter.read(reader);
+           } catch (IOException e) {
+             // TODO: wrap this in a reasonable exception
+             return null;
+           }
+         }
+       };
+       ConstructorConstructor constructorConstructor = new ConstructorConstructor(type, function);
+       ReflectiveTypeAdapterFactory reflectiveTypeAdapterFactory =
+           new ReflectiveTypeAdapterFactory(constructorConstructor, gson.fieldNamingStrategy,
+               gson.excluder, gson.jsonAdapterFactory);
+       return reflectiveTypeAdapterFactory.create(gson, type);
+     }
+   });
+
+   return this;
   }
 
   /**
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -16,10 +16,23 @@
 
 package com.google.gson;
 
+import static com.google.gson.Gson.DEFAULT_COMPLEX_MAP_KEYS;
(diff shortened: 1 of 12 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -511,6 +512,36 @@ public GsonBuilder registerTypeAdapter(Type type, Object typeAdapter) {
...
+   * Configures Gson for custom serialization and deserialization with Fill-in. Takes the type
+   * adapter, and wraps the result from the {@link TypeAdapter} into a {@link ReflectiveTypeAdapterFactory.Adapter}. Like
+   * {@link #registerTypeAdapter(Type, Object)}, it only applies to the type specified by
+   * the {@code type} parameter.
+   *
+   * @param baseType the type definition for the type adapter being registered
+   * @param objectAdapter This object must implement the {@link TypeAdapter} class
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   */
+  public GsonBuilder registerTypeAdapterWithFillIn(final Type baseType, Object objectAdapter) {
```
inder123: Instead of this method, why not create a TypeAdapter implementation that users can extend to get fill-in features. 
e.g., 
ReflectiveTypeAdapter that internally will use ReflectiveTypeAdapterFactory

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -16,6 +16,22 @@
...
+import static com.google.gson.Gson.DEFAULT_LENIENT;
+import static com.google.gson.Gson.DEFAULT_PRETTY_PRINT;
+import static com.google.gson.Gson.DEFAULT_SERIALIZE_NULLS;
+import static com.google.gson.Gson.DEFAULT_SPECIALIZE_FLOAT_VALUES;
+
+import com.google.gson.internal.$Gson$Preconditions;
+import com.google.gson.internal.ConstructorConstructor;
+import com.google.gson.internal.Excluder;
+import com.google.gson.internal.bind.ReflectiveTypeAdapterFactory;
+import com.google.gson.internal.bind.TreeTypeAdapter;
```
inder123: Please avoid frivolous changes like adding final to method parameters, imports reordering, or formatting changes.

gson/src/main/java/com/google/gson/InstanceCreatorWrapper.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.reflect.Type;
+
+/**
+ * Acts as a wrapper of the {@link InstanceCreator} interface for fill-in. Fulfills roughly the same
+ * functionality as the {@link com.google.gson.internal.ObjectConstructorWrapper} class. It doesn't
+ * break existing functionality by implementing {@code InstanceCreator} but when casted to this
+ * class, it can expose a method that accepts a {@code JsonReader} as a parameter, allowing fill-in
+ * functionality.
+ */
+public class InstanceCreatorWrapper<T> implements InstanceCreator<T> {
```
inder123: Why is this class part of the public API?

gson/src/main/java/com/google/gson/ReflectiveTypeAdapter.java
```diff
@@ -0,0 +1,30 @@
+package com.google.gson;
+
+import com.google.gson.internal.ConstructorConstructor;
+import com.google.gson.internal.bind.ReflectiveTypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class ReflectiveTypeAdapter<T> extends TypeAdapter<T> {
```
inder123: public API?

inder123:
Overall, I am still not convinced why this needs to be part of Gson, and not be part of extras package.

Overall, we prefer really small changes to the public APIs.

PR #1482, opened 2019-03-06
Fix wrong reference in Versioning Support example

TomyJaya:
Just a minor fix of wrong object being referenced.


TomyJaya:
I signed it!

PR #1488, opened 2019-03-10
Remove object parameterized methods in JsonObject & JsonPrimitive

Degubi:
-Refactor JsonPrimitive.value to be final
-Remove JsonPrimitive object parameterized constructor
-Remove some unnecessary helper methods


JakeWharton:
There was missing test coverage around null which this regresses. Tests added in a817604a417349af15df0daf1589319282565f22 and I've manually merged this as 8bf2949c2934cf9a168df24d0d2a29c702556a91 with a necessary fix to retain behavior.

PR #1490, opened 2019-03-13
Add Json Bodyhandlers for new Http Client (Java 11)

Degubi:
-Add class for creating custom Json parsing BodyHandlers for the new HttpClient added in Java11.
-Only the 5 static methods are public, everything else in the class is implementation detail.
-At the moment everybody needs to create a class like this, or use BodyHandlers.ofString() as a bodyhandler to then parse it.

Example code: https://pastebin.com/ZeHGdkwL

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonBodyHandler.java
+++ b/gson/src/main/java/com/google/gson/JsonBodyHandler.java
@@ -1,0 +1,108 @@
+package com.google.gson;
+
+import java.net.http.HttpResponse.*;
+import java.nio.*;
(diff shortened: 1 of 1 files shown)


Degubi:
Closing this, solvable with a simple method:
`public static BodyHandler<JsonObject> ofJsonObject(){
		return info -> BodySubscribers.mapping(BodySubscribers.ofString(StandardCharsets.ISO_8859_1), data -> parser.parse(data).getAsJsonObject());
	}`

PR #1492, opened 2019-03-15
Remove getter in Gson exposing internal class

Degubi:
-Remove getter for Excluder. Excluder is an internal class, it should not get returned to user code.
-Move DefaultDateTypeAdapter into ...internal/bind package.
-Presize factories list in Gson constructor
-Make the last 2 private TypeAdapter factories static in Gson class
-Refactor Gson toString
-Fix tests for changes


Degubi:
Closing this, might split into smaller requests.

Issue #924, opened 2016-09-06
NoSuchFieldException when serializing objects with an enumeration field on Samsung devices running Android 5.0.x


angelocoiro:
## Stack trace

```
java.lang.AssertionError: java.lang.NoSuchFieldException: ENUM_CONSTANT_NAME
at com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter#<init>(808)
at com.google.gson.internal.bind.TypeAdapters$30#create(834)
at com.google.gson.Gson#getAdapter(423)
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory#createBoundField(115)
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory#getBoundFields(164)
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory#create(100)
at com.google.gson.Gson#getAdapter(423)
at com.google.gson.Gson#toJson(661)
at com.google.gson.Gson#toJson(648)
at com.google.gson.Gson#toJson(603)
at com.google.gson.Gson#toJson(583)
at com.myapp.mypackage.MyClass#myMethod()
```

## Crash Descriprion
I can see this crash on Samsung devices running Android 5.0.x only.
I'm using gson-2.7 and I'm not usin ProGurad.

The crash occurs when I try to serialize an object of a specific type that contains a field of type enumeration.
However, it does not occur sistematically but rather randomly.

As fas as I understand, such an exception cannot be caused by my app's code.
Moreover, by inspecting gson code, I cannot understand why this exception is generated. It seems to me that this should never happen:

```
for (T constant : classOfT.getEnumConstants()) {
          String name = constant.name();
          SerializedName annotation = classOfT.getField(name).getAnnotation(SerializedName.class);
...
}
```

The NoSuchFieldException exception is thrown when the `Class.getField(String name)` method is called (line 797)

Any idea on why this exception is thrown? 
What could be the reason that causes a similar crash?

JakeWharton:
This is a bug in Samsung's version of Android. I don't think there's anything you can do and there's certainly nothing we can do.

angelocoiro:
Thanks for your support @JakeWharton. That is exactly what I suspected.

bpappin:
[...]

PR #1495, opened 2019-03-22
Make EnumTypeAdapter friendly with obfuscation

ganadist:
When enum value was obfuscated by proguard, EnumTypeAdapter raise NoSuchFieldException, even if apply SerializedName annotation.
Because EnumTypeAdapter cannot find obfuscated enum constant field with its name.
But without this workaround, there is no way to obfuscate enum name.


JakeWharton:
Test case?

ganadist:
I tried to integrate proguard obfuscation on testcase like this.

https://github.com/ganadist/gson/commit/251ec476aa49fbb4653b164c67d22e17445625c4

And intent of this mvn rule are this order.
* copy-resources (pre-obfuscate-class)
* proguard (default)
* copy-resources (post-obfuscate-class)

But order of plugin execution does not work as intended.
* copy-resources (pre-obfuscate-class)
* copy-resources (post-obfuscate-class)
* proguard (default)

Can I get any idea?

JakeWharton:
You don't need to integrate ProGuard. Just specify a different name and
ensure it's honored in the JSON output.

ganadist:
> Just specify a different name and ensure it's honored in the JSON output.

Gson has already [com.google.gson.functional.testEnumCaseMapping()](https://github.com/google/gson/blob/master/gson/src/test/java/com/google/gson/functional/EnumTest.java#L147) testcase.

ganadist:
I updated first conversation of this PR. because it could be misleading.

inder123:
I ran into this problem as well. This is a good fix. Let's add a testcase.

ganadist:
To verify this change, enum class in testcase must be obfuscated by proguard.
And I updated testcase on https://github.com/ganadist/gson/commit/1b65e3aae9f549c7a6adaf4dc533263cb7e1be4c , added to check enum class is obfuscated.

But it still has a problem to integrate proguard. :(

ganadist:
Proguard integration on Maven testcase is pretty dirty, but.
```
Running com.google.gson.functional.EnumWithObfuscatedTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec
```

``` 
$ cat gson/target/proguard_map.txt
com.google.gson.functional.EnumWithObfuscatedTest -> com.google.gson.functional.EnumWithObfuscatedTest:
    com.google.gson.Gson gson -> a
    void <init>() -> <init>
    void setUp() -> setUp
    void testEnumClassWithObfuscated() -> testEnumClassWithObfuscated
com.google.gson.functional.EnumWithObfuscatedTest$Gender -> com.google.gson.functional.EnumWithObfuscatedTest$Gender:
    com.google.gson.functional.EnumWithObfuscatedTest$Gender MALE -> a
    com.google.gson.functional.EnumWithObfuscatedTest$Gender FEMALE -> b
    com.google.gson.functional.EnumWithObfuscatedTest$Gender[] $VALUES -> c
    com.google.gson.functional.EnumWithObfuscatedTest$Gender[] values() -> values
    com.google.gson.functional.EnumWithObfuscatedTest$Gender valueOf(java.lang.String) -> valueOf
    void <init>(java.lang.String,int) -> <init>
    void <clinit>() -> <clinit>
```

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -788,7 +794,11 @@ public EnumTypeAdapter(Class<T> classOfT) {
           nameToConstant.put(name, constant);
           constantToName.put(constant, name);
         }
-      } catch (NoSuchFieldException e) {
+      } catch (IllegalAccessException e) {
+        throw new AssertionError(e);
+      } catch (NullPointerException e) {
+        throw new AssertionError(e);
+      } catch (ExceptionInInitializerError e) {
```
Marcono1234, line 801: Why catch this?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -788,7 +794,11 @@ public EnumTypeAdapter(Class<T> classOfT) {
           nameToConstant.put(name, constant);
           constantToName.put(constant, name);
         }
-      } catch (NoSuchFieldException e) {
+      } catch (IllegalAccessException e) {
+        throw new AssertionError(e);
+      } catch (NullPointerException e) {
```
Marcono1234, line 799: Why catch this?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -776,9 +777,14 @@ public void write(JsonWriter out, Locale value) throws IOException {
 
     public EnumTypeAdapter(Class<T> classOfT) {
       try {
-        for (T constant : classOfT.getEnumConstants()) {
+        for (Field field : classOfT.getDeclaredFields()) {
+          if (!field.isEnumConstant()) {
+            continue;
+          }
+          field.setAccessible(true);
```
Marcono1234, line 784: Why make the field accessible? Enum element fields are public so making them accessible should not be necessary.

Marcono1234:
This pull request would fix #924 and #1776

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: For https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Field.html#get(java.lang.Object)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: Without this line, unittests failed with following error.
```
Failed tests:   testCollectionOfEnumsDeserialization(com.google.gson.functional.EnumTest): AssertionError (GSON 2.8.6-SNAPSHOT): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"
  testTopLevelEnumDeserialization(com.google.gson.functional.EnumTest): AssertionError (GSON 2.8.6-SNAPSHOT): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"
  testClassWithEnumFieldSerialization(com.google.gson.functional.EnumTest): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"
  testClassWithEnumFieldDeserialization(com.google.gson.functional.EnumTest): AssertionError (GSON 2.8.6-SNAPSHOT): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"
  testCollectionOfEnumsSerialization(com.google.gson.functional.EnumTest): AssertionError (GSON 2.8.6-SNAPSHOT): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"
  testTopLevelEnumSerialization(com.google.gson.functional.EnumTest): java.lang.IllegalAccessException: class com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter cannot access a member of class com.google.gson.functional.EnumTest$MyEnum with modifiers "public static final"

```

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
Marcono1234: > `NullPointerException` - if the specified object is null and the field is an instance field.

But the enum constants are `static` fields so this should not be possible.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
Marcono1234: Yes, but only because the method can throw the error does not mean that it should be caught here. I assume if that error actually occurred it could also happen in the non-Proguard specific code and there it is not caught either. So in my opinion it should not be caught here.

(Though note that I am not a maintainer of this project)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -776,9 +777,14 @@ public void write(JsonWriter out, Locale value) throws IOException {
...
+          T constant = (T)(field.get(null));
```
Marcono1234, line 785: Please add a `@SuppressWarnings("unchecked")` for this variable.

gson/src/test/java/com/google/gson/functional/EnumWithObfuscatedTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson.functional;
+
+import java.lang.reflect.Field;
```
Marcono1234, line 19: This import is unused

gson/testcases-proguard.conf
```diff
@@ -0,0 +1,20 @@
+# Options from Android Gradle Plugins
```
Marcono1234, line 1: Would it be possible to move this to `gson/src/test/resources`? I think that might be better than having it directly in the `gson` folder.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: Fixed at eba1415f1c20a225e528aac83913c3a09962c282

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: Fixed at eba1415f1c20a225e528aac83913c3a09962c282

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: Fixed at ddee26e283d088fea1ec2807dfb258e9f36b2497

gson/src/test/java/com/google/gson/functional/EnumWithObfuscatedTest.java
ganadist: Fixed at 7dbfce9d9633c65c16f595e9c779c80084ddf894

gson/testcases-proguard.conf
ganadist: Fixed at a25f96b8936625ff4856495e8f8dfc63205bcbd7

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Enum constants are public, but the enum class itself isn't necessarily. If it isn't, then you'll get `IllegalAccessException`.

My concern here is that if there is a SecurityManager, the `field.setAccessible(true)` call might fail, whereas the code using `classOfT.getEnumConstants()` would have worked. So we'll have fixed one problem (getting enum constants in ProGuarded code) but introduced another (getting enum constants when there is a SecurityManager).

I think the code here should read something like this:

```java
for (final Field field : classOfT.getDeclaredFields()) {
  if (!field.isEnumConstant()) {
    continue;
  }
  AccessController.doPrivileged(
      new PrivilegedAction<Void>() {
        @Override Void run() {
          field.setAccessible(true);
          return null;
        }
      });
  T constant = (T) field.get(null);
```

gson/src/test/java/com/google/gson/functional/EnumWithObfuscatedTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @SerializedName("MAIL")
+    MALE,
+
+    @SerializedName("FEMAIL")
+    FEMALE
+  }
+
+  public void testEnumClassWithObfuscated() {
+    for (Gender enumConstant: Gender.class.getEnumConstants()) {
+        try {
```
eamonnmcmanus, line 48: Would you mind fixing the indentation on lines 48–50 so it uses the same two-space indentation as the rest of the code?

eamonnmcmanus (changes requested):
Thanks for putting this change together, especially the tests! Sorry it has taken so long for us to get to it.

There are just a couple of changes that I think would make sense.

I'm going to test this change against Google's many internal uses of Gson and if it comes through clean we should be good to go.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
ganadist: Fixed at 59a8aedb37ae35d4d2a4306c92efd463fffaaf69

gson/src/test/java/com/google/gson/functional/EnumWithObfuscatedTest.java
ganadist: Fixed at d8c5fcf00bdd3170365e92e43880021031c7d005

eamonnmcmanus (approved):
Thanks for the quick update!

Issue #1486, opened 2019-03-07
JsonArray Cannot add/set at Index


jredfox:
the methods of add/set(index, object) are missing from your wrapper class.

PR #1498, opened 2019-03-22
Implement add with index wrapper for JsonElement and primitive types

dacamposol:
- Implementation for the add method in an specifies position of the List to the wrapper.
- Unit tests for the different primitives.

Solves: Issue #1486

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -100,6 +100,61 @@ public final class JsonArray extends JsonElement implements List<JsonElement> {
   }
 
   /**
+   * Adds the specified boolean to self in the specified position.
+   * All the element on the array will be moved to the right (increment index by one)
+   *
+   * @param index the index where the specified parameter will be added.
+   * @param bool the boolean that needs to be added to the array.
+   */
+  public void add(int index, Boolean bool)  {
+    elements.add(index, bool == null ? JsonNull.INSTANCE : new JsonPrimitive(bool));
+  }
+
+  /**
+   * Adds the specified character to self in the specified position.
+   * All the element on the array will be moved to the right (increment index by one)
+   *
+   * @param index the index where the specified parameter will be added.
+   * @param character the character that needs to be added to the array.
+   */
+  public void add(int index, Character character)  {
+    elements.add(index, character == null ? JsonNull.INSTANCE : new JsonPrimitive(character));
+  }
+
+  /**
+   * Adds the specified number to self in the specified position.
+   * All the element on the array will be moved to the right (increment index by one)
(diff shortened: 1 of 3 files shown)


dacamposol:
Someone knows why Travis CI returns this error?

`The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk7" --workspace "/home/travis/.cache/install-jdk" --feature "7" --license "BCL"" failed and exited with 3 during .`

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -99,6 +99,61 @@ public boolean add(JsonElement element) {
     return elements.add(element);
   }
 
+  /**
+   * Adds the specified boolean to self in the specified position.
+   * All the element on the array will be moved to the right (increment index by one)
```
Marcono1234, line 104:
```suggestion
   * All the elements on the array will be moved to the right (increment index by one).
```
Same for documentation of the other methods.

Additionally this is slightly misleading because not "all" elements will be moved to the right, instead something like the following might be more correct:
> The element which is already at that index (if any) and all subsequent elements will be moved to the right, i.e. their index is incremented by one.

Similar to the documentation of [`java.util.List.add(int, E)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#add(int,E))

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -99,6 +99,61 @@ public boolean add(JsonElement element) {
...
+   *
+   * @param index the index where the specified parameter will be added.
```
Marcono1234, line 106: "Parameter" usually describes the named variable in the method signature, e.g. `Bool bool`. It looks like here you are rather writing about an "argument", i.e. a specific value for a parameter. Though using "element" would work as well and might be more specific:
> the index where the specified element will be added.

gson/src/test/java/com/google/gson/functional/JsonArrayTest.java
```diff
@@ -159,4 +160,28 @@ public void testSameAddition() {
...
+    jsonArray.add((Float) null);
+    jsonArray.add((String) null);
+
+    jsonArray.add(1, 'a');
+    jsonArray.add(3, false);
+    jsonArray.add(5, -2.567);
+    jsonArray.add(7, new JsonPrimitive(3));
+    jsonArray.add(9, "test");
+
+    System.out.println(jsonArray);
```
Marcono1234, line 181: Please remove this, it is not needed for the test

Marcono1234:
It might also be good to document the `IndexOutOfBoundsException` these methods can throw.

PR #1499, opened 2019-03-24
Primitives refactor

Degubi:
Refactor & Clean up the Primitives class to use array lookup instead of HashMap lookup. There's no need to use a HashMap for 9 elements.
Benchmark code: https://pastebin.com/cG8w2pyy
Benchmark output: https://pastebin.com/dCKfacCy

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/Primitives.java
+++ b/gson/src/main/java/com/google/gson/internal/Primitives.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.google.gson.internal;
 
 import java.lang.reflect.Type;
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/Primitives.java
```diff
@@ -29,47 +25,34 @@
...
+	return -1;
   }
-
+  
   /**
    * Returns true if this type is a primitive.
    */
   public static boolean isPrimitive(Type type) {
-    return PRIMITIVE_TO_WRAPPER_TYPE.containsKey(type);
+    return indexOf(type, 0) != -1;
```
JakeWharton: Isn't this just `type instanceof Class<?> && ((Class<?>) type).isPrimitive()`?

Degubi:
I was thinking about that, I will update to use normal if-s instead of array lookup.

Degubi:
I think this looks good now.

JakeWharton:
Reformatted, squashed, and merged as 477b3f77e159d15531fd30f41688d4b732a8b042. Thanks!

PR #1501, opened 2019-03-28
Removing unnecessary 'final' access modifier from constructor

manuelkollus:
Hello Dear, Google Developers!

I have removed the 'final' access modifier from the constructor because they are not necessary.

Yours sincerely,
Manuel Kollus


manuelkollus:
I signed it!

PR #1506, opened 2019-04-01
JsonObject Class Testcoverage 100%

Phhofm:
The JsonObject class coverage was at Class: 100%, Method 76%, Line 83%, Branch 62%.
After adding my tests to the JsonObjectTest class, we now achieve a whooping: Class 100%, Method 100%, Line 100% and Branch 100% coverage :)

Two of the added tests, `testGettingAsJasonArray()` and `testGettingAsJasonObject()` fail, but as far as I can figure out for now, the problem is with the explicit conversion inside the JsonObject class methods, and not with the tests. But i might be mistaken.
It seems like tests for the two methods `getAsJsonPrimitive(String memberName)` and `getAsJsonArray(String memberName)` of the JsonObject class actually have been missing in the project before i added mine.

### DIFF
--- a/gson/src/test/java/com/google/gson/JsonObjectTest.java
+++ b/gson/src/test/java/com/google/gson/JsonObjectTest.java
@@ -27,264 +27,275 @@ import junit.framework.TestCase;
  */
 public class JsonObjectTest extends TestCase {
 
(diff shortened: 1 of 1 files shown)


inder123:
Thanks for the tests, but the build failed. Can you take a look please?

PR #1512, opened 2019-04-14
Remove package private helpers in JsonElement & JsonPrimitive

Degubi:
-Remove documentation & code for "getAsBooleanWrapper" in JsonElement.
-Remove uses of "getAsBooleanWrapper" in JsonPrimitive.
-Inline "getAsBooleanWrapper" into methods where it was used.

-Not breaking backwards compatibility, because the methods were package private.


inder123:
👍

PR #1513, opened 2019-04-15
Add static methods to JsonParser

Degubi:
-Add static parser method alternatives to JsonParser.class.
-Mark the default constructor of JsonParser as deprecated.
-Point of this pr is to stop forcing people to create a JsonParser instance per parsing.
-People usually store a JsonParser instance as a global somewhere in their projects, and call the parser methods on that instance. With this pr, all these globals could be removed.
-No deprecations or warnings on parser method calls, single deprecation warning on JsonParser constructor.
-No breaking changes, the old solution works just like before.
-Fixed all tests to use the new static methods.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonParser.java
+++ b/gson/src/main/java/com/google/gson/JsonParser.java
@@ -32,7 +32,13 @@ import com.google.gson.stream.MalformedJsonException;
  * @since 1.3
  */
 public final class JsonParser {
-
+	
+  /**
+   * @deprecated No need to instantiate this class, use the static parser methods instead.
+   */
+  @Deprecated
+  public JsonParser() {}
+  
   /**
    * Parses the specified JSON string into a parse tree
    *
@@ -41,10 +47,10 @@ public final class JsonParser {
    * @throws JsonParseException if the specified text is not valid JSON
    * @since 1.3
    */
-  public JsonElement parse(String json) throws JsonSyntaxException {
-    return parse(new StringReader(json));
+  public static JsonElement parseString(String json) throws JsonSyntaxException {
+    return parseReader(new StringReader(json));
   }
-
+  
   /**
    * Parses the specified JSON string into a parse tree
    *
@@ -53,10 +59,10 @@ public final class JsonParser {
    * @throws JsonParseException if the specified text is not valid JSON
    * @since 1.3
    */
-  public JsonElement parse(Reader json) throws JsonIOException, JsonSyntaxException {
+  public static JsonElement parseReader(Reader reader) throws JsonIOException, JsonSyntaxException {
     try {
-      JsonReader jsonReader = new JsonReader(json);
-      JsonElement element = parse(jsonReader);
+      JsonReader jsonReader = new JsonReader(reader);
+      JsonElement element = parseReader(jsonReader);
       if (!element.isJsonNull() && jsonReader.peek() != JsonToken.END_DOCUMENT) {
         throw new JsonSyntaxException("Did not consume the entire document.");
       }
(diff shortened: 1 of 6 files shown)


JakeWharton:
Merged as c5a3f21fbad3828deba1deb88168e1f76829f807. Thanks!

inder123:
Do we really need new static methods @JakeWharton 
We could instead just provide a singleton instance. 
That would have aligned a little better with the existing design.

JakeWharton:
The existing design was the problem though. These methods are stateless and should never have been instance methods.

inder123:
@JakeWharton Yes, it can be argued that way. A counter-argument will be that making them instance-methods allowed for more testability. 

In general, my philosophy is to NOT try to fix APIs unless they are really bad, or the alternate is substantially better. 

This didn't pass that test. In fact, we have made it worse for the developers as now they will be confused on whether to use static method or the instance method.

Please consider reverting this change. Thanks.

JakeWharton:
More testability how? Static stateless methods are far easier to test than instance methods.

The deprecation makes it clear which to use.

inder123:
Better testability by allowing fake JsonParser, instead of having to mock static methods.

Regarding deprecations: I don't like deprecations much for stylistic changes. If it was a bad bug or broken interface, deprecation is acceptable. 
Gson has had a philosophy of keeping a minimal stable API. Every change has to fight to get in. 
Please help us keep it that way.

JakeWharton:
Except you never need a fake JsonParser because it's stateless so that argument doesn't work.

Sure I'll help keep it stable. Consider me retired from Gson.

inder123:
@JakeWharton It will be a big loss for Gson if you retire. So, I certainly wouldn't want that. Let's talk off-line. 

I don't have a strong use-case. Best I can come up with is a from a Gson user's perspective:
Think of a user who is writing a Java test for code that uses JsonParser. That user may want to install a different JsonParser than may be transforms input before processing.

JakeWharton:
The type was already final so you couldn't replace its functionality in any way. If you need parsing behavior that you can be configured then you would have already needed to build your own abstraction (like `Function<String, JsonElement>`) where one of them can be backed by this type (using `new JsonParser()::parse` or the new `JsonParser::parseString`).

inder123:
@JakeWharton If need be, we can lift the final from the class. That is a much smaller change than new methods. 

Overall, this is a solution in search of a problem.

JakeWharton:
No! That would be so much worse! Gson isn't your abstraction. It's an
implementation.

This change is how good API's are designed. Stateless wherever possible.

inder123:
@JakeWharton I agree with generally designing APIs to be stateless.
We possibly have an inferior design in JsonParser. However, I remain unconvinced that this feature addition is good. I prefer our APIs to be small and solve a burning problem. This doesn't achieve either.

gson/src/main/java/com/google/gson/JsonParser.java
```diff
@@ -69,25 +75,46 @@ public JsonElement parse(Reader json) throws JsonIOException, JsonSyntaxExceptio
...
+  
+  /**
+   * @see JsonReader.parseString
+   */
+  public JsonElement parse(String json) throws JsonSyntaxException {
+    return parseString(json);
+  }
+
+  /**
+   * @see JsonReader.parseReader
```
Marcono1234: Doesn't this and the other `@see JsonReader.parseReader` have to include the parameters since there are two methods with this name?

Also, shouldn't it be a hash (`#`)  sign: `@see #parseReader(Reader)`? ([javadoc](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIEDI))

PR #1516, opened 2019-04-21
Add nextValue() to JsonReader.

jamesrmtwigg:
Also add tests for nextValue().
[google-CLA.txt](https://github.com/google/gson/files/3100733/google-CLA.txt)

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -1221,44 +1221,108 @@ public class JsonReader implements Closeable {
   }
 
   /**
-   * Skips the next value recursively. If it is an object or array, all nested
+   * <p>Reads the next value recursively and returns the JSON value as a String.
+   * This method is intended for use when the JSON token stream contains a 
+   * value for which deserialization must be deferred.</p>
+   * 
+   * <p>Prefer {@link #skipValue()} when the value would not be used as it is
+   * faster.</p>
+   * @return The next value in the JSON input. Not {@code null} but possibly
+   * {@code "null"}. Not {@code ""} but possibly {@code "\"\""}. Any escape 
+   * sequences in {@link com.google.gson.stream.JsonToken.STRING String} values
+   * are converted and whitespace is removed.
+   * @see #skipValue()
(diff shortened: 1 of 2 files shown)


jamesrmtwigg:
An alternative solution to #1453. When values must be deserialized in a particular order, this added method can be used to save portions of the input for later.

inder123:
@swankjesse Can you review this please? 
Seems like an OK addition to me.

jamesrmtwigg:
@swankjesse Thanks for the feedback. I can resolve the first two issues. Would you accept "nextRawJSONValue()" for the method name?

For the third issue, `JsonReader.buffer` is only 1024 chars and it looks like it's filled with portions of the stream as it's read. If the next value is a medium to large array or object it could easily be larger than the buffer so I can't just take a slice. I assume this is why `skipValue()` similarly walks the tree rather than just advancing the stream. I can however keep a local buffer.

In this case I would take the policy of blindly copying the input stream until I find the end of the current value. I would only throw on unbalanced '{}' and '[]' tokens. The intention here is to enable deferring deserialization until later in case the key order matters, so I expect this input would get passed back to the very same Gson instance. The caller could of course do something else, but that's on them. They asked for "raw" JSON to deserialize themselves, so that's what they get.

swankjesse:
Yeah, might not be worth the complexity.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1221,44 +1294,92 @@ public void close() throws IOException {
...
+  }
+  
+  /**
+   * <p>Skips the next value recursively. If it is an object or array, all nested
    * elements are skipped. This method is intended for use when the JSON token
-   * stream contains unrecognized or unhandled values.
+   * stream contains unrecognized or unhandled values.</p>
+   * 
+   * <p>Faster than {@link #nextValue()}.</p>
+   * @see #nextValue()
```
Marcono1234: Should these link to `nextRawJSONValue()`?

jamesrmtwigg: Yes you are correct. I'll fix that, thanks.

PR #1527, opened 2019-05-05
update proguard to fix R8 compatibility issue

faridfor:
[described here](https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md#member-in-a-data-object-is-always)


faridfor:
I signed it! :)

PR #1534, opened 2019-05-16
Refactor CostructorConstructor to be stateless

Degubi:
- Remove ConstructorConstructor.java constructor, make all methods static
- Remove CostructorConstructor field from gson, pass the instanceCreator map to adapters
- Remove ReflectionAccessor fields, no need to store them in fields, avoids loading some classes when not used(Reflection classes)


inder123:
why is this preferred over existing implementation @Degubi ?

Degubi:
> why is this preferred over existing implementation @Degubi ?

-Remove some unnecessary constructors and state.
-Avoid loading some classes.
Nothing special, just to remove some unnecessary code bloat.

PR #1546, opened 2019-07-01
Keep `TypeAdapter`s from being stripped

hqzxzwb:
`TypeAdapter` implementations used in JsonAdapter annotation sometimes are never instantiated explicitly without reflection, causing R8 to strip their instance methods.


hqzxzwb:
I signed it!

hqzxzwb:
Do I need to do anything more for this getting reviewed?

PR #1548, opened 2019-07-11
JsonReader nextFloat support

LiozM:
in order to slightly optimize the performance of GSON JsonReader, I have added nextFloat function to skip converting into a double


LiozM:
I signed the form, is there anything else needed for this to be reviewed?

PR #1574, opened 2019-09-06
Add BUILD files for Bazel

Yannic:


Marcono1234:
I assume similar to the reasons mentioned in https://github.com/google/gson/pull/2150#issuecomment-1194435204 this won't be added because it would make maintenance more difficult.

eamonnmcmanus:
Indeed, the cited comment in #2150 applies here too.

PR #1576, opened 2019-09-16
Prefer writing chars instead of strings of length one

dmap:
This results in a noticeable performance improvement with most writer
implementations (including BufferedWriter).

The scale of the improvement obviously depends on the exact data being written but in my test cases the improvement in the performance critical `JsonWriter.string()` method when writing to a writer constructed as 
```
final Writer writer = new BufferedWriter(new OutputStreamWriter(new GZIPOutputStream(new ByteArrayOutputStream()), "UTF-8"), 8192);
```
was around 12.5%.  Given that the bulk of the time is spent doing the content compression the relative change in the non-compression is much larger... removing the `GZIPOutputStream` from the test makes the improvement around 30%.

The reason for the improvement is that the `BufferedWriter` code for writing a single char is much more efficient than the code for writing a string (when that string is length 1).

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -286,7 +286,7 @@ public class JsonWriter implements Closeable, Flushable {
    */
   public JsonWriter beginArray() throws IOException {
     writeDeferredName();
-    return open(EMPTY_ARRAY, "[");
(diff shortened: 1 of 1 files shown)


dmap:
I'm not sure exactly why the build failed but it appears to be failing in a similar manner on other branches.

inder123:
Thank you for a simple focused change with a great description and performance numbers!

PR #1577, opened 2019-09-17
Deprecated `new Gson()`, added `Gson.DEFAULT_GSON`

Miha-x64:
There is a popular anti-pattern: `new Gson().fromJson(json, type)`. No need to create new Gson instances as all of them have the same default configuration and work equally.


inder123:
We don't like to change Gson api for nice to haves.

PR #1589, opened 2019-10-06
Added ability to serialize down Anon and Local classes

miversen33:
I placed a flag in GsonBuilder that allows the Excluder to serialize Anon and Local classes down during the serialization to JSON. By default, the flag is false, and therefore won't break any existing usages of the Excluder.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -287,6 +287,16 @@ public final class GsonBuilder {
   }
 
   /**
+   * Configures Gson to include local and/or anonymous classes for serialization
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.1
+   */
+  public GsonBuilder enableLocalOrAnonClassSerialization(){
+    excluder = excluder.enableAnonOrLocalClassSerialization();
+    return this;
+  }
+
+  /**
    * Configures Gson to apply a specific serialization policy for {@code Long} and {@code long}
    * objects.
    *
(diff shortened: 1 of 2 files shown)


miversen33:
@googlebot I signed it!

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -286,6 +286,16 @@ public GsonBuilder disableInnerClassSerialization() {
     return this;
   }
 
+  /**
+   * Configures Gson to include local and/or anonymous classes for serialization
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.1
```
Marcono1234, line 292: This is wrong, but has to be adjusted anyways depending on when this is merged.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -286,6 +286,16 @@ public GsonBuilder disableInnerClassSerialization() {
...
+   */
+  public GsonBuilder enableLocalOrAnonClassSerialization(){
```
Marcono1234, line 294:
```suggestion
  public GsonBuilder enableLocalOrAnonClassSerialization() {
```

I think there should be a space to match the style of the rest of the project.

gson/src/main/java/com/google/gson/internal/Excluder.java
```diff
@@ -87,6 +88,12 @@ public Excluder disableInnerClassSerialization() {
     return result;
   }
 
+  public Excluder enableAnonOrLocalClassSerialization(){
```
Marcono1234, line 91:
```suggestion
  public Excluder enableAnonOrLocalClassSerialization() {
```

Issue #1367, opened 2018-08-10
Writing "raw" data using JsonWriter


NicolaSpreafico:
Hi,
I have a flow which read an image from a File, convert it to Base64, put the value inside a JSON and the send it to processing using an HTTP Connection.

Due to image size, I can't read the whole image bytes in memory, then fully convert it to base64, then put it all the base64 value inside the json and create the connection.

So I'm using a JsonWriter in order to read a chunk of data, convert it to base64, write inside the outputstream and the repeat for the whole content of the image. Here is concept code:

```
@Override
public void writeTo(OutputStream out) throws IOException {
	try (JsonWriter jsonWriter = new JsonWriter(new OutputStreamWriter(out, "UTF-8"))) {
	   jsonWriter.beginObject();
	   jsonWriter.name("image_base64");

	   // Read the image file with buffer
	   Encoder encoder = Base64.getEncoder();
	   ByteBuffer bytes = ByteBuffer.allocate(64 * 1024);
	   ReadChannel reader = client.reader(... opening reader for file...);

           // TODO1
	   jsonWriter.beginString();
	   
	   while (reader.read(bytes) > 0) {
		   bytes.flip();
                 // TODO2
		   jsonWriter.appendStringValue(encoder.encode(bytes.array()));
		   bytes.clear();
	   }
           // TODO3
	   jsonWriter.endStringValue();

	   jsonWriter.endObject();
	}
}
```

Because the JsonWriter only provides the method to write an entire String, for the moment the only way I found is reading the whole image, convert it to a full String and the call a single method.
As mentioned in the opening, I need to do this processing with a buffer in order to not having alla the image bytes in memory. Please note the 3 `TODO` parts, which I'm not sure how to implement.

Is there a way to write "raw data" in the OutputStream which simply writes the characters I give to it instead of writing the exact String i give, sorrounded by "?

For the moment I totally ignored the JsonWriter implementation and I'm writing the whole content as raw data:
```
@Override
public void writeTo(OutputStream out) throws IOException {
	String init = "{\"image_bytes\":{\"";
	out.write(init.getBytes(Charset.forName("UTF-8")));

	Encoder encoder = Base64.getEncoder();
	try ( OutputStream outB64 = encoder.wrap(out)){
		ReadChannel reader = client.reader(... opening reader for file...);
		IOUtils.copy(Channels.newInputStream(reader), outB64,64 * 1024);
	}

	String end = "\"}";
[...]

Issue #971, opened 2016-12-05
JsonReader.nextString() producer/consumer alternative


lyubomyr-shaydariv:
Hi,

`JsonReader` is a great class to read JSON data in streaming fashion. However, there are some cases when tokens are too big to fit in JVM memory making the JSON reader streaming idea neutralized. For example, here is a question at StackOverflow asking for managing huge fields somehow: http://stackoverflow.com/questions/39615673/best-way-to-handle-huge-fields-with-gson-jsonreader -- `String JsonReader.nextString()` reads the whole huge literal input into a `StringBuilder` instance that makes the JVM crash on `OutOfMemoryError`.

Do you consider adding a streaming alternative, say `void JsonReader.nextString(StringChunksConsumer)` where the `StringChunksConsumer` interface is expected to be something like the interface below, a valuable idea?

```
interface StringChunksConsumer {

    void accept(char[] buffer, int offset, int length)
            throws IOException;

}
```

The idea of this method would be passing parsed string literal chunks to the consumer rather than aggregating them into a string builder. The `StringChunksConsumer.accept` method signature would perfectly match:

* `StringBuilder.append(char[], int, int)`
* `Writer.write(char[], int, int)`
* etc.

For example, a simple redirecting from a reader to a writer could be just `read.nextString(writer::write);` consuming no more memory than `JsonReader.nextString()` uses internally.

Thanks!

**Update 1**

A similar problem: https://stackoverflow.com/questions/43470927/android-java-lang-outofmemoryerror-failed-to-allocate-with-free-bytes-and-70m

**Update 2**

This would also allow to work with inner string-encoded (for whatever reason) JSON documents more efficiently. Example: http://stackoverflow.com/questions/43929916/gson-parsing-error-for-my-json-string

**Update 3**

This possibly duplicates #415 and #718

swankjesse:
Interesting idea. I think the most likely thing we’d build would be a method like `nextReader()` that returned a `Reader` on the characters of a string.

lyubomyr-shaydariv:
@swankjesse Your suggestion sounds much better!

Macarse:
I kind of need this.
What would happen if the Reader is not consumed completely? JsonReader is in charge of fast forward?

lyubomyr-shaydariv:
@swankjesse Would this also include a `JsonWriter.write(...)` method to return a `Writer` to write very long strings to output streams?

aleripe:
More than one year has passed... any news on this topic?!

[...]

PR #1604, opened 2019-11-02
Add Reader and Writer methods to JsonReader and JsonWriter

Marcono1234:
# Motivation
JSON property names and string values can in same cases be quite large, but JsonReader and JsonWriter currently do not allow reading /  writing them in an efficient way. Instead you have to read / write them as one complete string.

This pull request adds the following methods which return `java.io.Reader` for reading and `java.io.Writer` for writing names and fields.
- `JsonReader`
    - `nextNameReader()`
    - `nextStringReader()`
- `JsonWriter`
    - `nameWriter()`
    - `stringValueWriter()`

@swankjesse, I hope this is how you imagined the API to look like.

Note: This pull request uses a few Java 8 features. It appears the pom.xml specifies that Java 6 should be used, however that might be outdated? Please let me know if it has to be compiled using Java 6.

# Related issues and pull requests
- #415
- #718
- #971
- #1367


lyubomyr-shaydariv:
Wow, cool thing!

> This pull request uses a few Java 8 features. It appears the pom.xml specifies that Java 6 should be used, however that might be outdated? Please let me know if it has to be compiled using Java 6.

I believe it has to use Java 6.

Marcono1234:
Ok, changed it to compile using Java 6. This pull request should now be ready.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -162,6 +167,190 @@
     HTML_SAFE_REPLACEMENT_CHARS['\''] = "\\u0027";
   }
 
+  private class StringValWriter extends Writer {
```
elharo: StringValueWriter  since Google style is to not abbreviate. Perhaps there's a better name since StringWriter is taken?

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -162,6 +167,190 @@
...
+
+    @Override
+    public void write(int c) throws IOException {
+      verifyNotClosed();
+      stringPiece((char) c);
+    }
+
+    /**
+     * @param len
+     *      Length of the data, e.g. array length
```
elharo: one line
length (not capped)

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -162,6 +167,190 @@
...
+     * @param off
+     *      0-based offset where the section begins
+     * @param sectLen
+     *      Length of the section
+     * @throws IndexOutOfBoundsException
+     *      If {@code off} or {@code sectLen} is invalid
+     */
+    private void validateIndices(int len, int off, int sectLen) {
```
elharo: len -> length
off --> offset
sectLen --> sectionLength

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -162,6 +167,190 @@
...
+    @Override
+    public void write(String str, int off, int len) throws IOException {
+      verifyNotClosed();
+      validateIndices(str.length(), off, len);
+      stringPiece(str.substring(off, off + len));
+    }
+
+    @Override
+    public void write(String str) throws IOException {
+      verifyNotClosed();
```
elharo: This doesn't look thread safe but might be OK if these objects never escape one thread.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
Marcono1234: The `write` methods use abbreviated names as well since the `java.io.Writer` methods do this. Do you want all parameter names written out? In my opinion especially here this does not really improve readability because the method name indicates their meaning.

elharo: yes, these are the rules google follows.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
Marcono1234: The doc for JsonWriter says:
> Instances of this class are not thread safe.

I hope that should prevent users from passing the created writers to different threads.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
Marcono1234: I have changed it to `StringValueReader` and `StringValueWriter` now. However, these classes are not publicly visible so their names can be changed at any point.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
Marcono1234: Alright, I have written out the names of fields and parameters I introduced. However, for overridden `Reader`  and `Writer` methods I am still using the JDK names. Guava seems to be doing the same, see for example [AppendableWriter](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/AppendableWriter.java).

Please let me know if you have other change requests.

Marcono1234: Alright, I have written out the names of fields and parameters I introduced. However, for overridden `Reader`  and `Writer` methods I am still using the JDK names. Guava seems to be doing the same, see for example [AppendableWriter](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/AppendableWriter.java).

Please let me know if you have other change requests.

Marcono1234:
I have now pushed a bunch of commits which perform the following changes:

# Changes
## Bug fixes
Fixed bugs which I introduced in my previous commits.
I also removed the following (existing) lines because it appears they don't add any value after these changes and might even unnecessarily loop through all chars again (see the `continue` statement) even though they were already processed:
https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1068-L1075

## JSON path changes
When a JSON object property value is skipped its name will now be kept in the JSON path. Previously it was replaced which in my opinion did not add any value and might have been confusing, e.g.:
```java
JsonReader reader = new JsonReader(new StringReader("{\"a\":1}"));
reader.beingObject();
reader.nextName();
reader.getPath(); // "$.a"
reader.skipValue();
// Previously this returned // "$.null"
reader.getPath();
```

Additionally skipping an object property name now uses the string `"#skippedName"` in the JSON path (previously it used `"null"`) and reading a name using `nextNameReader()` might use (depending on the implementation) `"#streamedName"` instead of the actual name.
All these behaviors are now documented in the javadoc.

## Add public subclass `StringValueReader`
I have thought a lot about how to properly handle closing the name or string value reader and how to detect whether it consumed the complete data. `Reader.close()` should fail fast if necessary, e.g. when it is called from a try-with-resources statement which threw an exception. So having it consume all remaining data of the value is not an option. On the other hand the complete data has to be consumed before the next JSON token can be read from the JsonReader. `Reader.skip(...)` does not provide the necessary guarantees to solve this (you don't know when it reached the end).
Therefore I decided the most user-friendly solution is to add a custom class (`StringValueReader`) which provides the convenience method `skipRemaining()` which the user can call before they process the next JSON token.
Additionally I added two other methods which might be useful and can be implemented efficiently into the current JsonReader implementation.

It is therefore now always necessary to first make sure that a name or string value reader consumed all data (e.g. by calling `skipRemaining()`) and to close the reader before reading the next JSON token from the JsonReader.

## Implement reader and writer methods for JsonTreeReader/-Writer classes
The JsonTreeReader and JsonTreeWriter classes now support the string value reader and writer methods as well.

# State of this pull request
More tests for the reader and writer methods of `JsonTreeReader` and `JsonTreeWriter` might be useful, however I am not going to do this now (or possibly in the near future) since that might require rewriting the test structure to prevent code duplication between the `JsonReader` / `JsonWriter` and `JsonTreeReader` / `JsonTreeWriter` tests.
But besides that I think this pull request might be ready, however there might still be corner cases in which the added / changed code behaves incorrectly. So for now I will leave this pull request as draft and would appreciate any manual testing or review.

Marcono1234:
>Why is it needed? Are there actual cases where the existing code fails?

Please see the "Related issues and pull requests" section of the description of this pull request and the StackOverflow questions  #971 is linking to. It appears it is not that uncommon that people encode binary data using Base64 which therefore can become arbitrarily large.
For what it's worth @swankjesse has expressed interest in such methods twice in the past ([here](https://github.com/google/gson/pull/718#issuecomment-150980208) and [here](https://github.com/google/gson/issues/971#issuecomment-266334881)). The reason for the separate public class is outlined in detail in my comment above.

elharo:
I can see that with values. Does this ever happen in practice with names or only with values?

Marcono1234:
I am not sure if there are existing APIs or programs which store arbitrarily long data in names, but it could be the case. Since the methods for names share in most parts the same implementation with the string value reader/writer I thought it would be good to support it as well.

Issue #992, opened 2017-01-09
JsonPrimitive with LazilyParsedNumber integers breaks contract for hashCode()


einarpehrson:
While most types maintain consistency between `equals()` and `hashCode()` after string serialization and parsing (`JsonElement#toString()` + `JsonParser#parse(String)`, integers do not. `equals()` returns true, but the hash codes are different.

```
  public void testEqualsIntegerAndLazilyParsedInteger() {
    MoreAsserts.assertEqualsAndHashCode(new JsonPrimitive(42), new JsonPrimitive(new LazilyParsedNumber("42")));
  }
```

dwilches:
Just got hit by this 3-year old bug.

jblacker:
Just got hit with this bug as well.  Any chance this issue will get picked up?

PR #1610, opened 2019-11-09
JsonPrimitive: considers LazilyParsedNumbers with integral values as integral when computing hashCode.

pterry26:
This fixes the bug described in https://github.com/google/gson/issues/992 where two JsonPrimitives could be equal but have different hashCodes.


pterry26:
@googlebot I signed it!

pterry26:
@googlebot I fixed it!

PR #1611, opened 2019-11-09
JsonPrimitive: considers integral LazilyParsedNumber as integral when evaluating hashCode

pterry26:
This fixes the bug described in https://github.com/google/gson/issues/992 where two JsonPrimitives could equal each other but have different hashCodes.


pterry26:
@googlebot I fixed it!

pterry26:
pht5 is me; I can't sign the CLA for both accounts

pterry26:
Fixed my issue before that was resulting in the commit being made under the wrong account; closing this and opening a new PR

PR #1612, opened 2019-11-09
JsonPrimitive: considers integral LazilyParsedNumber as integral when evaluating hashCode

pterry26:
This fixes the bug described in google#992 where two JsonPrimitives could equal each other but have different hashCodes.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -288,7 +288,16 @@ public final class JsonPrimitive extends JsonElement {
     if (primitive.value instanceof Number) {
       Number number = (Number) primitive.value;
       return number instanceof BigInteger || number instanceof Long || number instanceof Integer
-          || number instanceof Short || number instanceof Byte;
+          || number instanceof Short || number instanceof Byte || isIntegralLazilyParsedNumber(number);
+    }
+    return false;
+  }
+
+  /** @return true if the specified number is of type LazilyParsedNumber and has an integral value */
+  private static boolean isIntegralLazilyParsedNumber(Number number) {
+    if (number instanceof LazilyParsedNumber) {
+      double value = number.doubleValue();
+      return value == Math.rint(value);
     }
     return false;
   }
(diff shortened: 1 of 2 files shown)


dwilches:
+1 I was just hit by that bug, and it was open 3 years ago!

Marcono1234:
This change does not handle `double`s > `Long.MAX_VALUE` or < `Long.MIN_VAUE` correctly:
```
JsonPrimitive parsedDouble = new Gson().fromJson(Double.toString(Double.MAX_VALUE), JsonPrimitive.class);
// Should fail, but doesn't
assertFalse(parsedDouble.equals(new JsonPrimitive(0)));
```

Additionally it results in poor hash code calculation for integral doubles >= 10<sup>64</sup> (see [`BigDecimal.longValue()`](https://github.com/openjdk/jdk/blob/34c79640e76c001e9f316f9aaba49dbe227bbae7/src/java.base/share/classes/java/math/BigDecimal.java#L3522)), while the current implementation probably returns somewhat good values up to `Double.MAX_VALUE` (10<sup>308</sup>) (because it uses `BigDecimal.doubleValue()`), though that might be acceptable.

Both could be fixed by checking in `isIntegralLazilyParsedNumber` that the number is >= Long.MIN_VALUE && <= Long.MAX_VALUE.

Another problem is that by calling `gson.fromJson("...", Number.class)`  you get a `LazilyParsedNumber` which could store any string, possibly even a malformed one. This could then cause exceptions in `equals` and `hashCode`. However, that is already an issue with the existing code and there is probably no easy way to fix this.
However, for the case where `LazilyParsedNumber` contains `"Infinity"` or `"NaN"` (which might not be that uncommon) it could at least be fixed by also checking in `isIntegralLazilyParsedNumber` that the number is finite, i.e. neither [NaN](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#isNaN(double)) nor [Infinity](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Double.html#isInfinite(double)).

eamonnmcmanus:
I believe this was fixed by #3059.

PR #1616, opened 2019-11-13
support strings empty return null

thinking-github:
support strings empty return null


eamonnmcmanus:
Can you say something more about the problem you are trying to solve here? Right now we just have a bunch of code with a 5-word description. What is it that people will be able to do with this change that they can't do today?

Also, a reminder that Gson is in maintenance mode and we are generally reluctant to accept new features.

PR #1618, opened 2019-11-15
Add asList()

vulkman:
I'd love to have direct access to the element list so I don't have to convert the iterator to something that supports the Java 8 Stream API


vulkman:
@googlebot I signed it!

PR #1619, opened 2019-11-15
Add asList()

vulkman:
I'd love to have direct access to the elements-List so I don't have to fiddle about with the Iterator to use the Java 8 Stream API.


lyubomyr-shaydariv:
Similar: #683
Exactly the same: #1209

Marcono1234:
This has been superseded by #2225. Thank you nonetheless for your effort on this!

Issue #1592, opened 2019-10-16
JsonTreeWriter.put(): IllegalStateException


cqjason:
After calling JsonTreeWriter.name(null), calling JsonTreeWriter.vule() will cause JsonTreeWriter.put(): IllegalStateException because of internal calling JsonTreeWriter.put(). So why not check if the param of name() is null?  The check is done in super class！

fmmmlee:
Can you provide more detail on how to reproduce the issue?

cqjason:
public void testNullStringNameInSuperClass() throws IOException {
    StringWriter stringWriter = new StringWriter();
    JsonWriter jsonWriter = new JsonWriter(stringWriter);
    jsonWriter.beginObject();
    jsonWriter.name(null);     //throw  java.lang.NullPointerException: name == null
                                           //at com.google.gson.stream.JsonWriter.name(JsonWriter.java:386)
    jsonWriter.value((String) null);
    jsonWriter.endObject();
  }

public void testNullStringName() throws IOException {
    JsonWriter jsonWriter = new JsonTreeWriter();
    jsonWriter.beginObject();
    jsonWriter.name(null);
    jsonWriter.value((String) null);//throw java.lang.IllegalStateException
	//at com.google.gson.internal.bind.JsonTreeWriter.put(JsonTreeWriter.java:89)
	//at com.google.gson.internal.bind.JsonTreeWriter.nullValue(JsonTreeWriter.java:153)
	//at com.google.gson.internal.bind.JsonTreeWriter.value(JsonTreeWriter.java:146)
    jsonWriter.endObject();
  }

       I think it is appropriate behavior that throw NullPointerException when name(null). Because It can help client discover problem early. It has done in super class. But it‘s poor that JsonTreeWriter.name() didn't do it when JsonTreeWriter.name() override it in super class.

fmmmlee:
(Note: your reply has the response in code ticks, while the code is in normal text formatting - it's switched around for some reason)

I'll look more at the relevant areas of the code and most likely make a pull request; I agree that it makes sense for the behavior here to be consistent with the superclass unless there is a demonstrable use case in which the inconsistent behavior is necessary.

cqjason:
It's fine to fix it and it‘s very nice to commit it quickly.  I‘ll close the issue for solving.

PR #1620, opened 2019-11-22
Fixing issue #1592

fmmmlee:
As cqjason describes in [issue #1592](https://github.com/google/gson/issues/1592), the behavior of `JsonTreeWriter.name()` is inconsistent with the method it overrides in its parent JsonWriter. When calling `JsonWriter.name(null)`, a NullPointerException is thrown, while calling `JsonTreeWriter.name(null)` does not throw an exception, leading to a potentially misleading IllegalStateException when calling `JsonTreeWriter.value()` later in execution.

To resolve this, I just copied the null check from lines 385-387 of [JsonWriter](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/stream/JsonWriter.java).

Edit: I had made a few style changes to the design document on this fork, but reverted them in order to keep the changes issue-related.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 
 jdk:
-  - oraclejdk9
+  - openjdk11
 
 install: mvn -f gson install -DskipTests=true
 script: mvn -f gson test


Marcono1234:
Would it be possible to force push only 49280df or squash the commits? I assume the maintainers here would like to not include commits which are then reverted in a later commit.

fmmmlee:
I believe maintainers have the option to squash when merging a pull request, so it only shows up as a single commit. I don't think they can be squashed before merging, though; only when performing the merge.

PR #1640, opened 2020-01-14
Add isEmpty method to JsonArray class

GabrielBB:
The method returns true if the array is empty


GabrielBB:
@googlebot I signed it!

Issue #1024, opened 2017-02-24
Gson polymorphism: request similar feature like Json.Net's TypeNameHandling.Auto


zenglian:
( I'm not sure this is a duplicate one as there were issues raised for "polymorphism")

The main idea is to **automatically** add an extra "@type" field if the actual instance's type is not the same as (should be a direct or **indirect** subclass of) the declaring type.

see [**Json.Net's TypeNameHandling.Auto**](http://www.newtonsoft.com/json/help/html/SerializeTypeNameHandling.htm)

[Edit]
 Json.Net's approach is more generic and simpler than [RuntimeTypeAdapterFactory](https://github.com/google/gson/blob/e0f0854ac23c07658cce5877b1fc9bc5995b1688/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java). It holds the full class name so you do not need to register type adapter. Everything is automatic.

It will be good if gson has similar feature. Thanks.

NightlyNexus:
Will this do what you want?
https://github.com/google/gson/blob/e0f0854ac23c07658cce5877b1fc9bc5995b1688/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java

leonsoft:
I think Json.Net's approach is more generic and simpler (more difficult for the author of gson :)   

It holds the full class name so you do not need to register type adapter. Everything is automatic. 
 
E.g., a List<Object> contains any objects can be deseriazlied, including nested collections. Not sure if RuntimeTypeAdapterFactory can do the same thing yet.

carschno:
@NightlyNexus Also, there is no official release (on Maven Central) for gson-extras, is there?

Marcono1234:
Personally I am not that convinced because deserializing based on class names can easily make you vulnerable to deserialization exploits, see [ysoserial](https://github.com/frohoff/ysoserial).
Having this as separate library is fine if you know what you are doing, but having it in Gson might lead to incautious usage just because it is "easier".

Though in the end the maintainers of this project decide of course.

PR #1647, opened 2020-02-03
Enable default inclusion of type information and use this information...

cadaver123:
… during deserialization.

Motivation: 
* It fixes some issues with deserialization of polymorphic types. Without it proper deserialization required from user writing specific Adapters. 
* I've tried to achive generic solution similar to Jackson [enableDefaultTyping()](https://fasterxml.github.io/jackson-databind/javadoc/2.5/com/fasterxml/jackson/databind/ObjectMapper.html#enableDefaultTyping())

It is turned on with GsonBuilder.enableDefaultTyping() method. 

Basically when enabled it changes represantion of user defined types (serialized/deserialized with ReflectiveTypeAdapterFactory) in JSON from:

```
{
 prop1: value1,
 prop2: value2
}
```

...to:

```
{
  _type: "name.of.UserClass",
  _properties: {
    prop1: value1,
    prop2: value2
  }
}
```

For ex., when we're running DefaultTypingTest with this option enabled we will get:

image

Without it by default we will lost all data from/about subtypes:

image

### **Related issue:**
#1024

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -40,6 +40,7 @@ import static com.google.gson.Gson.DEFAULT_LENIENT;
 import static com.google.gson.Gson.DEFAULT_PRETTY_PRINT;
 import static com.google.gson.Gson.DEFAULT_SERIALIZE_NULLS;
 import static com.google.gson.Gson.DEFAULT_SPECIALIZE_FLOAT_VALUES;
+import static com.google.gson.Gson.DEFAULT_TYPING;
 
 /**
  * <p>Use this builder to construct a {@link Gson} instance when you need to set configuration
@@ -94,6 +95,7 @@ public final class GsonBuilder {
   private boolean prettyPrinting = DEFAULT_PRETTY_PRINT;
   private boolean generateNonExecutableJson = DEFAULT_JSON_NON_EXECUTABLE;
   private boolean lenient = DEFAULT_LENIENT;
+  private boolean defaultTyping = DEFAULT_TYPING;
 
   /**
    * Creates a GsonBuilder instance that can be used to build Gson with various configuration
@@ -401,6 +403,18 @@ public final class GsonBuilder {
   }
 
   /**
+   * Enable default inclusion of type information and use this information during deserialization.
+   * Every complex type will be serialized/deserialized using two additional JSON fields - _type and _properties.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   */
+
+  public GsonBuilder enableDefaultTyping() {
+    defaultTyping = true;
+    return this;
+  }
+
+  /**
    * By default, Gson escapes HTML characters such as &lt; &gt; etc. Use this option to configure
    * Gson to pass-through HTML characters as is.
    *
@@ -596,7 +610,7 @@ public final class GsonBuilder {
 
     return new Gson(excluder, fieldNamingPolicy, instanceCreators,
         serializeNulls, complexMapKeySerialization,
-        generateNonExecutableJson, escapeHtmlChars, prettyPrinting, lenient,
+        generateNonExecutableJson, escapeHtmlChars, prettyPrinting, lenient, defaultTyping,
         serializeSpecialFloatingPointValues, longSerializationPolicy,
         datePattern, dateStyle, timeStyle,
         this.factories, this.hierarchyFactories, factories);
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -404,7 +404,11 @@ public final class GsonBuilder {
 
   /**
    * Enable default inclusion of type information and use this information during deserialization.
-   * Every complex type will be serialized/deserialized using two additional JSON fields - _type and _properties.
+   * Every complex type will be serialized/deserialized using two additional JSON fields:
+   * <ul>
+   *   <li>{@code _type}: The class name of the type as JSON string, e.g. {@code "my.CustomClass$Nested"}</li>
+   *   <li>{@code _properties}: The properties of the serialized instance in a JSON object.</li>
+   * </ul>
    *
    * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
    */
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -228,7 +228,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         in.beginObject();
         while (in.hasNext()) {
           String name = in.nextName();
-          if (this.defaultTyping && TYPE_FIELD_NAME.equals(name)) {
+          if (defaultTyping && TYPE_FIELD_NAME.equals(name)) {
             String className = in.nextString();
             if (PROPERTIES_FIELD_NAME.equals(in.nextName())) {
               try {
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -295,8 +295,6 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         out.endObject();
       }
       out.endObject();
-
     }
-
   }
 }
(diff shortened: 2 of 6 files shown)


cadaver123:
@googlebot I signed it!

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -215,6 +227,26 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
         in.beginObject();
         while (in.hasNext()) {
           String name = in.nextName();
+
+          if (this.defaultTyping && TYPE_FIELD_NAME.equals(name)) {
+            String className = in.nextString();
+            if (PROPERTIES_FIELD_NAME.equals(in.nextName())) {
+              try {
+                Class<?> clazz = Class.forName(className);
```
Marcono1234, line 235: This should call <code>Class.forName(className, <b>false</b>, ...)</code>. Otherwise it initializes unrelated classes, which might be exploitable.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -215,6 +227,26 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
         in.beginObject();
         while (in.hasNext()) {
           String name = in.nextName();
+
+          if (this.defaultTyping && TYPE_FIELD_NAME.equals(name)) {
```
Marcono1234, line 231: Maybe this should also check for `instance` not being of type `Object`, otherwise a lazy developer would just deserialize `Object`, which would be exploitable because it allows creating instances of any object.

Marcono1234:
Thanks!
Though for `Object` you misunderstood me. It is fine if the JSON specifies that a `java.lang.Object` is deserialized. What I meant is that a lazy developer would just write:
```
gson.fromJson(json, Object.class)
```
This would allow an adversary to have any type they want deserialized (because `Object` is assignable from any type). The check should therefore be `if (instance.getClass() == Object.class) { throw ... }`.
A lazy developer could of course still use commonly used interfaces as type to deserialize (e.g. `Collection`), however since it allows only a subset of all types, chances that a subclass of it can be abused (see ysoserial "gadget class") are at least lower.

cadaver123:
@Marcono1234 
Greatly thanks for response.
As you said deserialization based on class names can lead to exploits, but after implementing your remarks from code review and due to this check

```java
  Class<?> clazz = Class.forName(className, false, this.getClass().getClassLoader());
  if (!instance.getClass().equals(clazz) && ****instance.getClass().isAssignableFrom(clazz)****) {
    return (T) context.getAdapter(clazz).read(in);
  } else {
    in.beginObject();
    continue;
  }
```

only types/subtypes from user classes' declarations will be deserialized/initialized, so yes... in some cases it can make things worse but only if user has put itself in danger and is already deserializing risky objects.

cadaver123:
> Thanks!
> Though for `Object` you misunderstood me. It is fine if the JSON specifies that a `java.lang.Object` is deserialized. What I meant is that a lazy developer would just write:
> 
> ```
> gson.fromJson(json, Object.class)
> ```

Current state of code blocks this option, because Object is deserialized by ObjectTypeAdapter to LinkedTreeMap/ArrayList/Primitive Types.  `gson.fromJson(json, Object.class)` will omit my changes, so this check is unnecessary.

Ex. based on test (DefaultTypingTest) that I've written (with defaultTyping turned on):
When I had run
`Object complexTestTypeWithObject = gson.fromJson(getJson(), Object.class);`
it returned:
image

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -238,6 +275,12 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
       }
 
       out.beginObject();
+      if(defaultTyping) {
```
Marcono1234, line 278:
```suggestion
      if (defaultTyping) {
```

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -248,7 +291,12 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
       } catch (IllegalAccessException e) {
         throw new AssertionError(e);
       }
+      if (defaultTyping) {
+        out.endObject();
+      }
       out.endObject();
+
     }
+
```
Marcono1234, line 300:
```suggestion
    }
```
These empty lines are probably not needed

gson/src/test/java/com/google/gson/GsonTest.java
```diff
@@ -60,7 +60,7 @@ public void testOverridesDefaultExcluder() {
   public void testClonedTypeAdapterFactoryListsAreIndependent() {
     Gson original = new Gson(CUSTOM_EXCLUDER, CUSTOM_FIELD_NAMING_STRATEGY,
         new HashMap<Type, InstanceCreator<?>>(), true, false, true, false,
-        true, true, false, LongSerializationPolicy.DEFAULT, null, DateFormat.DEFAULT,
+        true, true,false, false, LongSerializationPolicy.DEFAULT, null, DateFormat.DEFAULT,
```
Marcono1234, line 63:
```suggestion
        true, true, false, false, LongSerializationPolicy.DEFAULT, null, DateFormat.DEFAULT,
```

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -215,6 +228,25 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+            if (PROPERTIES_FIELD_NAME.equals(in.nextName())) {
+              try {
+                Class<?> clazz = Class.forName(className, false, this.getClass().getClassLoader());
+                if (!instance.getClass().equals(clazz) && instance.getClass().isAssignableFrom(clazz)) {
+                  return (T) context.getAdapter(clazz).read(in);
+                } else {
+                  in.beginObject();
+                  continue;
+                }
+              } catch (ClassNotFoundException e) {
```
Marcono1234, line 242: Maybe it would be cleaner if this `try-catch` only contained the `Class.forName`, e.g.:
```
Class<?> clazz;
try {
  clazz = Class.forName(...);
} catch (ClassNotFoundException e) {
  throw ...
}
```

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -215,6 +228,25 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
         in.beginObject();
         while (in.hasNext()) {
           String name = in.nextName();
+          if (this.defaultTyping && TYPE_FIELD_NAME.equals(name)) {
```
Marcono1234, line 231:
```suggestion
          if (defaultTyping && TYPE_FIELD_NAME.equals(name)) {
```
Qualifying using `this` is not necessary

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -215,6 +228,25 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+            String className = in.nextString();
+            if (PROPERTIES_FIELD_NAME.equals(in.nextName())) {
+              try {
+                Class<?> clazz = Class.forName(className, false, this.getClass().getClassLoader());
+                if (!instance.getClass().equals(clazz) && instance.getClass().isAssignableFrom(clazz)) {
```
Marcono1234, line 236: Maybe it would be saner to throw an exception (`JsonParseException`) if `instance.getClass().isAssignableFrom(clazz)` is not the case? Currently it would just deserialize as if the type name did not exist which might make noticing this issue more difficult.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -400,6 +402,22 @@ public GsonBuilder setLenient() {
...
+   * Every complex type will be serialized/deserialized using two additional JSON fields:
+   * <ul>
+   *   <li>{@code _type}: The class name of the type as JSON string, e.g. {@code "my.CustomClass$Nested"}</li>
+   *   <li>{@code _properties}: The properties of the serialized instance in a JSON object.</li>
+   * </ul>
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   */
+
+  public GsonBuilder enableDefaultTyping() {
```
Marcono1234, line 416: It would probably be good to have a big warning section here. I can try phrasing one if you like.

cadaver123: Go ahead if you like

cadaver123:
> After thinking some more about it, another way to solve this would be to provide this functionality as `TypeAdapter` which can then be referenced with a [`@JsonAdapter`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/annotations/JsonAdapter.html) annotation. This then only limits this feature to the classes / fields where you really want this functionality and does not apply it to all types deserialized with that Gson instance.

It looks like a good idea - I can try to implement this when I find spare time

Marcono1234:
Note however, when creating a type adapter which is intended for usage with `@JsonAdapter`, it has to check and prevent usage for deserializing `Object.class` because `@JsonAdapter` on fields overwrites all adapters, including ones which can normally not be overwritten (such as `ObjectTypeAdapter` here).

PR #1648, opened 2020-02-04
Official Gradle Wrapper Validation Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


inder123:
Thanks for the PR. Please add some information to indicate that this works.

JLLeitschuh:
> Please add some information to indicate that this works.

That's not how this works, you'll need to merge this first. You should, from then on, see the status check on all PR's. I've tested this exact action workflow on many repositories and have proved it works (including against other Google projects).

JLLeitschuh:
Example:
https://github.com/google/flexbox-layout/pull/531

PR #1651, opened 2020-02-11
Throw UnsupportedOperationException when JsonWriter.jsonValue is not supported

Marcono1234:
Previously `JsonTreeWriter.jsonValue` threw an `AssertionError` because it was not overridden. Now it throws an `UnsupportedOperationException` instead. This is also documented because the `JsonTreeWriter` is passed to user defined type adapters.


Marcono1234:
@inder123, is there something you want me to change (also for #1656), or can I help you in some way? It would be really good if there was more progress with Gson.

Marcono1234:

An alternative to this would be to parse the JSON data to `JsonElement` as implemented in #1819.

eamonnmcmanus:
I agree with the principle of the fix. I don't have an opinion about #1819 except that it's sort of incompatible with our stated Maintenance Mode.

It wasn't obvious to me _why_ the current code throws `AssertionError`. Here's the trace (not sure the line numbers correspond):

```
        at com.google.gson.internal.bind.JsonTreeWriter$1.write(JsonTreeWriter.java:36)
        at java.base/java.io.Writer.write(Writer.java:291)
        at java.base/java.io.Writer.write(Writer.java:250)
        at java.base/java.io.Writer.append(Writer.java:323)
        at com.google.gson.stream.JsonWriter.jsonValue(JsonWriter.java:440)
```

The first line is [here](https://github.com/google/gson/blob/b84b2218f22acdbd141c4ae3fa9e68767d07028d/gson/src/main/java/com/google/gson/internal/bind/JsonTreeWriter.java#L36):

```
  private static final Writer UNWRITABLE_WRITER = new Writer() {
    @Override public void write(char[] buffer, int offset, int counter) {
      throw new AssertionError();
    }
```

Marcono1234:
> It wasn't obvious to me _why_ the current code throws `AssertionError`

But it is now, right? The problem is how `JsonWriter` and `JsonTreeWriter` are designed (same applies for the reader variants). `JsonWriter` is a full implemention which uses a `Writer` to write the JSON. `JsonTreeWriter` however creates a `JsonElement` tree instead of using a `Writer`, therefore it needs to override all of `JsonWriter` methods to prevent any of them calling a method on the `Writer`. This is similar to the issue which was solved by #2132.

Have created #2181 to verify that all methods are overridden.

PR #1654, opened 2020-02-11
Remove unused files

Marcono1234:
It appears `gson/docs/javadocs` has been generated and its content is not used anymore, but please let me know if this is incorrect.


inder123:
Thank you for the change. Yes, we should delete these files. 
However, gson/Gson 2.1 notes.txt shouldn't be deleted.

Marcono1234:
It appears `CHANGELOG.md` already covers this:
https://github.com/google/gson/blob/87b1d26d06d47ae709b7e953392257b7cbef1b62/CHANGELOG.md#L165-L171

If not, I would suggest adding the missing information there instead. However it appears it is partially already covered and changes to a test (`TypeHierarchyAdapterTest`) should not be mentioned anyways.

PR #1665, opened 2020-03-30
Specify timezone for GregorianCalender in testDateFormatString()

LorenzNickel:
The date is formatted using format(Date) which uses UTC-timezone as no other timezone is specified
GregorianCalender instead uses the system-default timezone, which makes this test not pass in all timezones
Therefore the timezone of the calender should be specified before it's used to create a Date-instance using (by default) UTC-timezone

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
@@ -17,7 +17,10 @@ public class ISO8601UtilsTest {
 
     @Test
     public void testDateFormatString() {
-        Date date = new GregorianCalendar(2018, Calendar.JUNE, 25).getTime();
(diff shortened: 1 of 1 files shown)


LorenzNickel:
By the way, I think the reason this "bug" was introduced is, that the test passes if you are western to UTC (which means that your date is somewhere between 2018-08-25 00:00 and 2018-08-25 12:00, but since I'm living in Germany, it's 2018-08-24 22:00 (and in general 24th between 12:00 and 00:00).

So TL;DR currently this test passes only if your computer has a timezone between UTC-12 and UTC

asolntsev:
Why this PR is not still merged? 
`gson` tests are failing on my machine right now. Should merged it!

LorenzNickel:
@asolntsev seems like gson is no longer actively maintained: https://github.com/google/gson/issues/1821

Marcono1234:
This was fixed by #1687; thanks nonetheless for the PR!

PR #1667, opened 2020-03-30
Add clear method to JsonArray

LorenzNickel:
The method will clear the contents of a JsonArray

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -182,6 +182,13 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
   }
 
   /**
+   * Removes all elements from the array.
+   */
+  public void clear() {
+    elements.clear();
+  }
+
+  /**
    * Returns an iterator to navigate the elements of the array. Since the array is an ordered list,
    * the iterator navigates the elements in the order they were inserted.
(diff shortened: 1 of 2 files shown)


Marcono1234:

For consistency a `JsonObject.clear()` method might be useful as well, but that can be done in a separate pull request, if desired.

eamonnmcmanus:
I think if we're able to add an `asList()` method to `JsonArray`, we shouldn't need other new methods like this?

Marcono1234:
Right, though the open question with `asList()` is still `null` handling I think.

Maybe `clear()` is needed often enough to justify having it directly on `JsonArray` either way? But I am not too sure.

eamonnmcmanus:
I think object reuse is poor practice so I don't find that a convincing argument. If people really want to reuse `JsonArray` instances they can do `array.asList().clear()`. But making a new `JsonArray` is generally better. So I'm still unconvinced by `JsonArray.clear()`, and even more so by `JsonObject.clear()`.

dhoard:
> I think object reuse is poor practice so I don't find that a convincing argument. If people really want to reuse `JsonArray` instances they can do `array.asList().clear()`. But making a new `JsonArray` is generally better. So I'm still unconvinced by `JsonArray.clear()`, and even more so by `JsonObject.clear()`.

Fair enough.

Issue #1658, opened 2020-02-24
The rule '-keep class * implements com.google.gson.TypeAdapter' uses implements but actually matches extends.


JasonTian23:
“-keep class * implements com.google.gson.TypeAdapter” in examples/android-proguard-example/proguard.cfg is not valid. TypeAdapter is abstract class, not a interface.

Should use: -keep class * extends com.google.gson.TypeAdapter

Marcono1234:
Fixed by  #1693

Marcono1234:
As mentioned above, this has been fixed.

PR #1668, opened 2020-03-31
TypeAdapter only appears as AbstractClass

LorenzNickel:
Fixes https://github.com/google/gson/issues/1658


Marcono1234:
Has been fixed by  #1693

PR #1671, opened 2020-04-03
Cleanup code & documentation for exceptions

LorenzNickel:
**Add a private generated serialVersionUID**
[According to the java documentation](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/io/Serializable.html) different serializable classes should use unique `serialVersionUID`:
> The serialization runtime associates with each serializable class a version number, called a serialVersionUID, which is used during deserialization to verify that the sender and receiver of a serialized object have loaded classes for that object that are compatible with respect to serialization. If the receiver has loaded a class for the object that has a different serialVersionUID than that of the corresponding sender's class, then deserialization will result in an InvalidClassException.

These should also be private if possible:
> It is also strongly advised that explicit serialVersionUID declarations use the private modifier where possible, since such declarations apply only to the immediately declaring class--serialVersionUID fields are not useful as inherited members.

**Parameters of exceptions can be final**
As these parameters are immutable, they should be marked as final

**Add author-tags if missing**
All files should have the `@author`-tag, therefore it has been added to `MalformedJsonException.java`

**Consistent naming for throwable**
The naming of the throwable in exceptions has been unified to be always `cause`

**Add documentation to all methods**
All methods and constructors in exception-classes have received a unified javadoc-block

**Apply Google Java Style Guide**
The [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html) has been applied

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonIOException.java
+++ b/gson/src/main/java/com/google/gson/JsonIOException.java
@@ -23,7 +23,7 @@ package com.google.gson;
  */
 public final class JsonIOException extends JsonParseException {
 
-  private static final long serialVersionUID = 1818297422735748149L;
+  private static final long serialVersionUID = 1L;
 
   /**
    * Creates exception with the specified message. If you are wrapping another exception, consider
@@ -31,7 +31,7 @@ public final class JsonIOException extends JsonParseException {
    *
    * @param msg error message describing a possible cause of this exception.
    */
-  public JsonIOException(final String msg) {
+  public JsonIOException(String msg) {
     super(msg);
   }
 
@@ -41,7 +41,7 @@ public final class JsonIOException extends JsonParseException {
    * @param msg   error message describing what happened.
    * @param cause root exception that caused this exception to be thrown.
    */
-  public JsonIOException(final String msg, final Throwable cause) {
+  public JsonIOException(String msg, Throwable cause) {
     super(msg, cause);
   }
 
@@ -51,7 +51,7 @@ public final class JsonIOException extends JsonParseException {
    *
    * @param cause root exception that caused this exception to be thrown.
    */
-  public JsonIOException(final Throwable cause) {
+  public JsonIOException(Throwable cause) {
     super(cause);
   }
 }
--- a/gson/src/main/java/com/google/gson/JsonIOException.java
+++ b/gson/src/main/java/com/google/gson/JsonIOException.java
@@ -16,30 +16,42 @@
 package com.google.gson;
 
 /**
- * This exception is raised when Gson was unable to read an input stream
- * or write to one.
- * 
+ * This exception is raised when Gson was unable to read an input stream or write to one.
+ *
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
 public final class JsonIOException extends JsonParseException {
-  private static final long serialVersionUID = 1L;
 
-  public JsonIOException(String msg) {
+  private static final long serialVersionUID = 1818297422735748149L;
+
+  /**
+   * Creates exception with the specified message. If you are wrapping another exception, consider
+   * using {@link #JsonIOException(String, Throwable)} instead.
+   *
+   * @param msg error message describing a possible cause of this exception.
+   */
+  public JsonIOException(final String msg) {
     super(msg);
   }
 
-  public JsonIOException(String msg, Throwable cause) {
+  /**
+   * Creates exception with the specified message and cause.
+   *
+   * @param msg   error message describing what happened.
+   * @param cause root exception that caused this exception to be thrown.
+   */
+  public JsonIOException(final String msg, final Throwable cause) {
     super(msg, cause);
   }
 
   /**
-   * Creates exception with the specified cause. Consider using
-   * {@link #JsonIOException(String, Throwable)} instead if you can describe what happened.
+   * Creates exception with the specified cause. Consider using {@link #JsonIOException(String,
+   * Throwable)} instead if you can describe what happened.
    *
    * @param cause root exception that caused this exception to be thrown.
    */
-  public JsonIOException(Throwable cause) {
+  public JsonIOException(final Throwable cause) {
     super(cause);
   }
 }
--- a/gson/src/main/java/com/google/gson/JsonParseException.java
+++ b/gson/src/main/java/com/google/gson/JsonParseException.java
@@ -17,9 +17,9 @@
 package com.google.gson;
 
 /**
- * This exception is raised if there is a serious issue that occurs during parsing of a Json
- * string. One of the main usages for this class is for the Gson infrastructure. If the incoming
- * Json is bad/malicious, an instance of this exception is raised.
+ * This exception is raised if there is a serious issue that occurs during parsing of a Json string.
+ * One of the main usages for this class is for the Gson infrastructure. If the incoming Json is
+ * bad/malicious, an instance of this exception is raised.
  *
  * <p>This exception is a {@link RuntimeException} because it is exposed to the client. Using a
  * {@link RuntimeException} avoids bad coding practices on the client side where they catch the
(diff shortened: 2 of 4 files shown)


Marcono1234:
Since you are already touching `MalformedJsonException`, would you mind changing the call to `initCause` to simply calling `super(Throwable)`? Gson requires Java 1.6 now and that constructor has been added in Java 1.6.

It looks like you misunderstood `serialVersionUID`. It is used to differentiate between multiple versions of the same class, and not between different classes. It is therefore not an issue if different classes have the same `serialVersionUID`. What it tries to solve is when you serialize an instance of class `A` (v1), then change class `A` (v2) and for example add another field and try to deserialize the previous instance again.
In this case you should have changed the `serialVersionUID` for v2 so the difference is detected and deserialization fails instead of causing issues later at runtime.

Therefore changing the `serialVersionUID` without changing the classes as it is the case for this pull request could even unnecessarily break backward compatibility.

Regarding the `final` parameters: That is mostly a code-style question. It has no effects at runtime (and this information is normally not even present in the class files). The Google style guide apparently does not mention it, but starting to make some parameters `final` which then differs from the majority of all other parameters in the project is probably not a good idea.

Not all classes need to have an `@author` tag in the javadoc. In some cases it might be nice to have and an author might use it to show off their work, however it is definitely not required.

LorenzNickel:
Thanks for your feedback, I tried to address it.

gson/src/main/java/com/google/gson/JsonIOException.java
```diff
@@ -16,26 +16,38 @@
...
- * This exception is raised when Gson was unable to read an input stream
- * or write to one.
- * 
+ * This exception is raised when Gson was unable to read an input stream or write to one.
+ *
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
 public final class JsonIOException extends JsonParseException {
+
```
Marcono1234, line 25: Empty line is probably not necessary

gson/src/main/java/com/google/gson/JsonParseException.java
```diff
@@ -30,7 +30,8 @@
  * @author Joel Leitch
  */
 public class JsonParseException extends RuntimeException {
-  static final long serialVersionUID = -4086729973971783390L;
+
```
Marcono1234, line 33: Empty line is probably not necessary

gson/src/main/java/com/google/gson/JsonIOException.java
```diff
@@ -16,26 +16,38 @@
 package com.google.gson;
 
 /**
- * This exception is raised when Gson was unable to read an input stream
- * or write to one.
- * 
+ * This exception is raised when Gson was unable to read an input stream or write to one.
+ *
```
Marcono1234, line 20: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/JsonIOException.java
```diff
@@ -16,26 +16,38 @@
...
+   */
   public JsonIOException(String msg, Throwable cause) {
     super(msg, cause);
   }
 
   /**
-   * Creates exception with the specified cause. Consider using
-   * {@link #JsonIOException(String, Throwable)} instead if you can describe what happened.
+   * Creates exception with the specified cause. Consider using {@link #JsonIOException(String,
+   * Throwable)} instead if you can describe what happened.
```
Marcono1234, line 50: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/JsonParseException.java
```diff
@@ -17,9 +17,9 @@
 package com.google.gson;
 
 /**
- * This exception is raised if there is a serious issue that occurs during parsing of a Json
- * string. One of the main usages for this class is for the Gson infrastructure. If the incoming
- * Json is bad/malicious, an instance of this exception is raised.
+ * This exception is raised if there is a serious issue that occurs during parsing of a Json string.
+ * One of the main usages for this class is for the Gson infrastructure. If the incoming Json is
+ * bad/malicious, an instance of this exception is raised.
```
Marcono1234, line 22: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/JsonParseException.java
```diff
@@ -45,16 +46,16 @@ public JsonParseException(String msg) {
   /**
    * Creates exception with the specified message and cause.
    *
-   * @param msg error message describing what happened.
+   * @param msg   error message describing what happened.
```
Marcono1234, line 49: Spaces here probably don't matter because there are no other parameters

gson/src/main/java/com/google/gson/JsonParseException.java
```diff
@@ -45,16 +46,16 @@ public JsonParseException(String msg) {
...
    */
   public JsonParseException(String msg, Throwable cause) {
     super(msg, cause);
   }
 
   /**
-   * Creates exception with the specified cause. Consider using
-   * {@link #JsonParseException(String, Throwable)} instead if you can describe what happened.
+   * Creates exception with the specified cause. Consider using {@link #JsonParseException(String,
+   * Throwable)} instead if you can describe what happened.
```
Marcono1234, line 58: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/JsonSyntaxException.java
```diff
@@ -16,8 +16,7 @@
 package com.google.gson;
 
 /**
- * This exception is raised when Gson attempts to read (or write) a malformed
- * JSON element.
+ * This exception is raised when Gson attempts to read (or write) a malformed JSON element.
```
Marcono1234, line 19: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/JsonSyntaxException.java
```diff
@@ -26,18 +25,29 @@
...
   public JsonSyntaxException(String msg, Throwable cause) {
     super(msg, cause);
   }
 
   /**
-   * Creates exception with the specified cause. Consider using
-   * {@link #JsonSyntaxException(String, Throwable)} instead if you can
-   * describe what actually happened.
+   * Creates exception with the specified cause. Consider using {@link #JsonSyntaxException(String,
+   * Throwable)} instead if you can describe what happened.
```
Marcono1234, line 50: Only changes where the line is wrapped (and "actually" has been removed), so change is probably not needed

gson/src/main/java/com/google/gson/stream/MalformedJsonException.java
```diff
@@ -19,26 +19,40 @@
 import java.io.IOException;
 
 /**
- * Thrown when a reader encounters malformed JSON. Some syntax errors can be
- * ignored by calling {@link JsonReader#setLenient(boolean)}.
+ * Thrown when a reader encounters malformed JSON. Some syntax errors can be ignored by calling
+ * {@link JsonReader#setLenient(boolean)}.
```
Marcono1234, line 23: Only changes where the line is wrapped, so change is probably not needed

gson/src/main/java/com/google/gson/stream/MalformedJsonException.java
```diff
@@ -19,26 +19,40 @@
...
  */
 public final class MalformedJsonException extends IOException {
+
```
Marcono1234, line 26: Empty line is probably not necessary

gson/src/main/java/com/google/gson/stream/MalformedJsonException.java
```diff
@@ -19,26 +19,40 @@
...
+  public MalformedJsonException(String msg, Throwable cause) {
+    super(msg, cause);
   }
 
-  public MalformedJsonException(Throwable throwable) {
-    // Using initCause() instead of calling super() because Java 1.5 didn't retrofit IOException
-    // with a constructor with Throwable. This was done in Java 1.6
-    initCause(throwable);
+  /**
+   * Creates exception with the specified cause. Consider using {@link
```
Marcono1234, line 50: Maybe wrap `{@link` to the next line so the complete javadoc tag is in one line

PR #1673, opened 2020-04-04
New getAsStringByForce method added.

akashyadav20111992:
This new method, in addition to what getAsString method has, also checks for null element and returns "" if it is null. It helps when you have large number json attributes and every attribute needs to be checked for null.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -149,6 +149,24 @@ public final class JsonPrimitive extends JsonElement {
       return (String) value;
     }
   }
+	
+  /**
+   * convenience method to get this element as a String
+   * and blank String if element is null
+   * @return get this element as a String.
+   */
+  @Override
+  public String getAsStringByForce() {
+    if(value == null) {
+      return "";
+    } else if (isNumber()) {
+      return getAsNumber().toString();
+    } else if (isBoolean()) {
+      return ((Boolean) value).toString();
+    } else {
+      return (String) value;
+    }
+  }
 
   /**
    * convenience method to get this element as a primitive double.
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -159,13 +159,8 @@ public final class JsonPrimitive extends JsonElement {
   public String getAsStringByForce() {
     if(value == null) {
       return "";
-    } else if (isNumber()) {
-      return getAsNumber().toString();
-    } else if (isBoolean()) {
-      return ((Boolean) value).toString();
-    } else {
-      return (String) value;
     }
+    return getAsString();
   }
 
   /**
--- a/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
+++ b/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
@@ -111,6 +111,9 @@ public class JsonPrimitiveTest extends TestCase {
 
     json = new JsonPrimitive(true);
     assertEquals("true", json.getAsString());
+    
+    json = null;
+    assertEquals("", json.getAsStringByForce());
   }
 
   public void testExponential() throws Exception {


akashyadav20111992:
> @googlebot I signed it!

akashyadav20111992:
> Okay, I'm not talking about the usefulness of this feature or anything, but just from the code perspective, you could and probably respecting the DRY-principle should re-use the `getAsString()` method here:
> 
> ```java
> @Override
> public String getAsStringByForce() {
>   if(value == null) {
>     return "";
>   }
>   return getAsString();
> }
> ```

Didn't came to me. Thanks for the suggestion.

akashyadav20111992:
@LorenzNickel I have made the changes suggested by you. Thanks.

LorenzNickel:
Thanks, additionally I'd encourage you to add test-coverage for this functionality.

akashyadav20111992:
@LorenzNickel test-coverage added for getAsStringByForce().

LorenzNickel:
Okay I'm not sure about the following but:
- regarding your test, shouldn't you use something like `json = new JsonPrimitive(null);`, because you want the value of the json to be null, not the json itself?
- if the point above is true, this shouldn't be allowed, since `$Gson$Preconditions.checkNotNull`, which is part of every constructor, will cause the instantiation of every JsonPrimitve with a null-value to fail? Therefore does you method even make sense, because if I got this right, the value must never be null.

I'm not really into gson, so I might be wrong here, feel free to correct me in this case.

akashyadav20111992:
@LorenzNickel 
{
"id":"123",
"name":"Akash"
}
If for above json, JsonObject.getAsJsonPrimitive("name"), it will result in "Akash", but for below json,
{
"id":"123",
}
now, JsonObject.getAsJsonPrimitive("name") will throw null pointer exception. 
getAsStringByForce() will handle this and will return "". This will spare us the null check for each attribute.
This method will be helpfull where primitive attributes are not fixed and may or may not be present.

Please let know if this clears your concerns or not.

LorenzNickel:
Thanks for the explanation, that was clear enough 👍

akashyadav20111992:
@LorenzNickel 
Sorry, this is my first time. Is there anything pending from side to get it merged? What's the next step?

LorenzNickel:
You need to wait for approval from someone with write access, so as long as you don't get any comments, you just have to wait.

lyubomyr-shaydariv:
I don't see any reason of adding such a method:

1. This merely makes the API bloated without any significant convenience. Many languages have a so-called Elvis operator that can hold such a case right using the language facilities (Groovy, C#, Kotlin). In Java, that does not support the operator, one can use an eager Elvis-like `MoreObjects.firstNonNull` from libraries like Google Guava, or bulild a lazy overload so that the right-hand operand could only be evaluated when necessary, or use `Optional<T>` (which I'm not a fan of), or use `@Nullable`/`@Nonnull` static analysis support.
1. This would only add the String support, but this is not the only "primitive" value: there are booleans and numbers too.
1. Having such a method would encourage bad practices: `""` is not a null and vice versa, `0` is not a null and vice versa, `false` is not a nul too.
1. Some languages (C#, Kotlin) support extension methods that can have the same effect and usually are designed for methods that have such convenience and utility purposes.

akashyadav20111992:
> I don't see any reason of adding such a method:
> 
> 1. This merely makes the API bloated without any significant convenience. Many languages have a so-called Elvis operator that can hold such a case right using the language facilities (Groovy, C#, Kotlin). In Java, that does not support the operator, one can use an eager Elvis-like `MoreObjects.firstNonNull` from libraries like Google Guava, or bulild a lazy overload so that the right-hand operand could only be evaluated when necessary, or use `Optional<T>` (which I'm not a fan of), or use `@Nullable`/`@Nonnull` static analysis support.
> 2. This would only add the String support, but this is not the only "primitive" value: there are booleans and numbers too.
> 3. Having such a method would encourage bad practices: `""` is not a null and vice versa, `0` is not a null and vice versa, `false` is not a nul too.
> 4. Some languages (C#, Kotlin) support extension methods that can have the same effect and usually are designed for methods that have such convenience and utility purposes.

Thanks for such a detailed explanation.

Marcono1234:
> In Java, that does not support the operator, one can use an eager Elvis-like `MoreObjects.firstNonNull` from libraries like Google Guava

Java 9 also added [`Objects.requireNonNullElse`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#requireNonNullElse(T,T)).

PR #1680, opened 2020-04-15
README: +link to StackOverflow

simon04:


Marcono1234:
This is probably obsolete now because the README was updated as part of #1977.

PR #1687, opened 2020-05-02
Fix ISO8601UtilsTest failing on systems with UTC+X

Marcono1234:
Previously `ISO8601UtilsTest.testDateFormatString()` would fail on systems where the time zone is UTC+X because `getTime()` returned "2018-06-24" for them.

Additionally the tests which previously changed the system locale and time zone have been rewritten to create a UTC calendar instead. Setting locale seems to not be necessary because `ISO8601Utils.parse(...)` does not do that either.

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
@@ -15,9 +15,25 @@ public class ISO8601UtilsTest {
     @Rule
     public final ExpectedException exception = ExpectedException.none();
 
+    private static TimeZone utcTimeZone() {
+        return TimeZone.getTimeZone("UTC");
+    }
+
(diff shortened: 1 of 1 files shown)


LorenzNickel:
Hey, I had the same problem while running the test and also created a PR for it (https://github.com/google/gson/pull/1665), where I use another approach to the same problem

Marcono1234:
Thanks for pointing this out, I did not search before creating the pull request, which I should probably have done.

I am biased since I created this pull request, but I prefer this approach because
- It also sets `Locale.US`, so it behaves the same as `ISO8601Utils.format`, not sure if otherwise the test could fail for certain Locales
- It removes changing the system time zone and locale since that is error-prone if a test forgets to reset them and also prevent concurrent use

eamonnmcmanus:
Thanks! I agree that the approach here seems reasonable, even if it is a bit more complicated than the simplest possible fix.

PR #1688, opened 2020-05-03
Fix #1684: Allow explicit JSON null for JsonElement

### LINKED ISSUE #1684: Unable to deserialize a null property into a JsonObject
Repro:

```java
import com.google.gson.Gson;
import com.google.gson.JsonObject;

public class GsonDeserialization {

    public static void main(String... args) {
        final Gson gson = new Gson();
        // succeeds
        final MyObject mo1 = gson.fromJson("{ \"name\": null, \"otherProperties\": {} }", MyObject.class);
        // throws
        final MyObject mo2 = gson.fromJson("{ \"name\": \"Alex\", \"otherProperties\": null }", MyObject.class);
    }

    public static class MyObject {
        public String name;
        public JsonObject otherProperties;
    }
}
```
```
Exception in thread "main" com.google.gson.JsonSyntaxException: Expected a com.google.gson.JsonObject but was com.google.gson.JsonNull
	at com.google.gson.internal.bind.TypeAdapters$35$1.read(TypeAdapters.java:896)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:129)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:220)
	at com.google.gson.Gson.fromJson(Gson.java:888)
	at com.google.gson.Gson.fromJson(Gson.java:853)
	at com.google.gson.Gson.fromJson(Gson.java:802)
	at com.google.gson.Gson.fromJson(Gson.java:774)
	at com.andrew.scratch.GsonDeserialization.main(GsonDeserialization.java:13)
```

I would expect the second example to deserialize into an object with a (java) null `otherProperties` member, but instead it throws a JsonSyntaxException.

Marcono1234:
Adds a new annotation `@ExplicitlyNullableJsonElement` which can be placed on fields of type JsonElement or subclasses. When deserialized using the reflection-based approach, `null` instead of JsonNull will be used as value. This allows explicit `null` JSON values for JsonElement subclasses which would otherwise cause an exception (except for JsonNull).

Note that this does not cover all use cases where one might want `JsonElement` or some of its subclasses to be deserialized as `null`, see discussion in #1684.
Additionally a user might also be able to use `Object` instead of `JsonElement` as type of fields to prevent the issue this pull request tries to solve. However, numbers are treated differently: Deserializing `Object` treats all numbers as  `double` (see `ObjectTypeAdapter`).

E.g. the following currently throws an exception:
```
class MyClass {
  JsonObject f;
}

new Gson().fromJson("{\"f\": null}", MyClass.class);
```

With the changes of this pull request, the following would be possible:
```
class MyClass {
  @ExplicitlyNullableJsonElement
  JsonObject f;
}

new Gson().fromJson("{\"f\": null}", MyClass.class);
```

Edit: It would also be possible to solve this by using the `@JsonAdapter` annotation on the fields and implementing a `TypeAdapterFactory` which creates type adapters which return `null` in case of JSON `null` or otherwise delegate to the default adapter. Edit 2: Probably not possible due to  #1028.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonElement.java
+++ b/gson/src/main/java/com/google/gson/JsonElement.java
@@ -16,6 +16,7 @@
 
 package com.google.gson;
 
+import com.google.gson.annotations.ExplicitlyNullableJsonElement;
 import com.google.gson.internal.Streams;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
@@ -29,6 +30,8 @@ import java.math.BigInteger;
  *
  * @author Inderjeet Singh
  * @author Joel Leitch
+ *
+ * @see ExplicitlyNullableJsonElement
  */
 public abstract class JsonElement {
   /**
--- a/gson/src/main/java/com/google/gson/annotations/ExplicitlyNullableJsonElement.java
+++ b/gson/src/main/java/com/google/gson/annotations/ExplicitlyNullableJsonElement.java
@@ -1,0 +1,66 @@
+package com.google.gson.annotations;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonObject;
+
+/**
+ * An annotation that indicates that upon deserialization {@code null} will be assigned
+ * to an annotated field of declared type {@link JsonElement} or one of its subclasses
+ * in case it has an explicit {@code null} value in JSON.
+ *
+ * <p>This annotation only has an effect if the JSON representation of the declaring
+ * class of the annotated field is deserialized using Gson's reflection-based
+ * deserialization approach (the default if no type adapter is specified for the class).
+ *
+ * <p>Normally {@link JsonElement} and its subclasses are deserialized as instance of
+ * {@link JsonNull} if they have a JSON value of {@code null}. When the expected type
+ * is {@code JsonElement}, this makes it cumbersome because one has to check for both
+ * {@code null} and {@code JsonNull} after deserializing. And for subclasses of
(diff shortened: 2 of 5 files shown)


Marcono1234:
Closing this pull request because the same functionality provided by this pull request can indeed be implemented using `@JsonAdapter`:
```java
/**
 * Factory for deserializing {@link JsonElement} and all of its subclasses as
 * {@code null} instead of {@code JsonNull.INSTANCE}.
 * 
 * <p>Must <b>only</b> be used in combination with {@link JsonAdapter @JsonAdapter}
 * on fields:
 * <pre>
 * class MyClass {
 *     &#x40;JsonAdapter(NullableJsonElementTypeAdapterFactory.class)
 *     JsonElement f;
 * }
 * </pre>
 */
public static class NullableJsonElementTypeAdapterFactory implements TypeAdapterFactory {
    @Override
        if (JsonElement.class.isAssignableFrom(type.getRawType())) {
            // Due to Gson issue #1028 have to call getAdapter(...) instead of getDelegateAdapter(...)
            TypeAdapter<JsonElement> delegateAdapter = gson.getAdapter(JsonElement.class);
            
            // Safe due to type check at beginning
            TypeAdapter<T> adapter = (TypeAdapter<T>) new NullableJsonElementTypeAdapter(delegateAdapter);
            return adapter;
        } else {
            return null;
        }
    }
    
    private static class NullableJsonElementTypeAdapter extends TypeAdapter<JsonElement> {
        private final TypeAdapter<JsonElement> delegate;
        
        public NullableJsonElementTypeAdapter(TypeAdapter<JsonElement> delegate) {
            this.delegate = delegate;
        }
        
        @Override
        public void write(JsonWriter out, JsonElement value) throws IOException {
            delegate.write(out, value);
        }

        @Override
        public JsonElement read(JsonReader in) throws IOException {
            if (in.peek() == JsonToken.NULL) {
                return null;
            } else {
                return delegate.read(in);
            }
        }
    }
}
```

PR #1689, opened 2020-05-03
Prevent custom JsonElement subclasses

Marcono1234:
Adds a package-private constructor to JsonElement to prevent custom subclasses. JsonElement models the JSON data types so there should probably be no need for custom subclasses because the existing JsonElement subclasses cover all of them. Additionally custom subclasses would cause issues during (de-)serialization because the default type adapter does not support them.

However, maybe there are use cases (in unit tests?) where custom subclasses are desired?
Would it be better to only deprecate the public constructor for now?

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonElement.java
+++ b/gson/src/main/java/com/google/gson/JsonElement.java
@@ -31,6 +31,11 @@ import java.math.BigInteger;
  * @author Joel Leitch
  */
 public abstract class JsonElement {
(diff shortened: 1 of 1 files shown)


inder123:
Yeah, we didn't envision JsonElement to be subclassed. However, at this point in time, it will be a backward incompatible change, and not worth making.

Marcono1234:
Should the public constructor be marked as deprecated at least?

Issue #835, opened 2016-04-22
JsonReader not strict about JSON document consumption


NightlyNexus:
Say I have a class:
```final class MyType {}```

and a `TypeAdapter`:
```
final class MyTypeAdapter extends TypeAdapter<MyType> {
  // ...
  @Override public MyType read(JsonReader in) throws IOException {
    in.beginObject();
    return new MyType();
    }
}
```
Note the missing `in.endObject()` call.

My calling code:
```
String json = "{}";
Gson gson = new GsonBuilder().registerTypeAdapter(MyType.class, new MyTypeAdapter()).create();
StringReader stringReader = new StringReader(json);
MyType myType = gson.fromJson(stringReader, MyType.class);
```
The above will correctly fail (`JsonIOException: JSON document was not fully consumed.`).

However, using a `JsonReader` instead (`gson.fromJson(new JsonReader(stringReader), MyType.class)`) will not fail.
Is this working as intended?

**Comment by JakeWharton** [comment]
If you are passing a `JsonReader` instance then it's your responsibility to do the assertion.

The implementation of `fromJson(String, Class<?>)` is:
```java
JsonReader jsonReader = newJsonReader(json);
Object object = fromJson(jsonReader, classOfT);
assertFullConsumption(object, jsonReader);
return object;
```

In the case where you are the one creating the `JsonReader` there's no opportunity to assert completion since you may just be deserializing part of the whole document.

NightlyNexus:
Oh, yes, I should have dug one step deeper into that. Thanks. Makes sense.

Issue #248, opened 2015-03-19
Avoid compiler warnings when deserializing to a parameterized type


GoogleCodeExporter:
```
Foo<Bar> foo = gson.fromJson(json, Foo.class);

currently produces compiler warning for unchecked conversion and needs 
@SuppressWarnings("unchecked")

Explore revising gson.fromJson(String json, Class<? super T> classOfT) to get 
rid of this warning.
```

Original issue reported on code.google.com by `inder123` on 5 Oct 2010 at 1:53

Issue #1247, opened 2018-02-03
Confusing Exceptions


havrikov:
When trying to deserialize the values `[[{},null],false,false]` and `[[null,false],null,true]` with the code `(new Gson()).fromJson(x, Map.class);` two confusing exceptions are thrown:  
`java.lang.ClassCastException: com.google.gson.internal.LinkedTreeMap is not Comparable` and  
`java.lang.NullPointerException: key == null`, respectively.  

Such exceptions should not be thrown to the user. Instead, one would expect these cases to be reported with an `IllegalArgumentException`.

PR #1700, opened 2020-05-21
Add Gson.fromJson(..., TypeToken) overloads

Marcono1234:
Previously only `Gson.fromJson(..., Type)` existed which is however not type-safe since the generic type parameter `T` used for the return type is not bound.
Example:
```
Type type = new TypeToken<List<String>>(){}.getType();
// Compiles fine, but causes issue at runtime
List<Integer> list = gson.fromJson(..., type);
```

Since these methods are often used in the form
```
gson.fromJson(..., new TypeToken<...>(){}.getType())
```
this commit now adds overloads which accept a `TypeToken` and are therefore
more type-safe:
```
TypeToken<List<String>> type = new TypeToken<List<String>>(){};
// This won't compile (as expected)
List<Integer> list = gson.fromJson(..., type);
```

With Java >= 9 another advantage is that you can use the 'diamond syntax' for anonymous classes (see [JDK-8062373](https://bugs.openjdk.java.net/browse/JDK-8062373)). If Gson had `fromJson` overloads with `TypeToken` this would allow quite concise usage:
```java
Map<String, List<String>> map = gson.fromJson(..., new TypeToken<>(){});
```

Additional changes:
- ~~Adds `fromJson(JsonReader, Class<T>)` for symmetry with other methods~~
Edit: Removed again because it breaks source compatibility, see [comment](https://github.com/google/gson/pull/1700#discussion_r973764414)
- Fixed some javadoc grammar mistakes
- Added javadoc `@see` tags; they currently have this order:
  1. Overload which serializes to writer / string; respectively deserializes from reader / string
  2. Overload which supports generic types
- Consistently write "JSON" in uppercase
- Explicitly state in javadoc that `fromJson` methods other than `fromJson(JsonReader, ...)` fully consume JSON (see #835)
- More precise placement of `@SuppressWarnings("unchecked")`

It appears `TypeToken` overloads were considered in the past but were then rejected, see https://github.com/google/gson/issues/248#issuecomment-83703055. However, I have not found the discussion explaining why they were rejected.

Open questions:
- [ ] Maybe it would even be reasonable to deprecate the `fromJson` methods taking a `Type` (but keeping them indefinitely). While this might be disruptive for users initially, switching to the `TypeToken` variant will make their code safer. If they have, for whatever reason, only a `Type` available but no `TypeToken`, then they can construct one with `TypeToken.get(Type)`, but it is their task then to verify that the casts are safe.


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -812,58 +836,95 @@ public void toJson(JsonElement jsonElement, JsonWriter writer) throws JsonIOExce
...
+   * </pre>
+   * @return an object of type T from the string. Returns {@code null} if {@code json} is {@code null}
+   * or if {@code json} is empty.
+   * @throws JsonSyntaxException if json is not a valid representation for an object of the erasure type
+   * of typeOfT
+   *
+   * @see #fromJson(Reader, TypeToken)
+   * @see #fromJson(String, Class)
+   */
+  public <T> T fromJson(String json, TypeToken<T> typeOfT) throws JsonSyntaxException {
```
lyubomyr-shaydariv, line 904: I'm not really sure, but would declaring the `typeOfT` as covariant `TypeToken<? extends T>` have any advantages here?

Marcono1234: I can't currently think of a case where it would matter. Do you have an example?

lyubomyr-shaydariv: @Marcono1234 
no, I don't, sorry. Variance is probably my weakest point in the Java language, and I'd like to know whether declaring the type token parameter might be affected by the choice. I can't think of a realistic example either as the returned value of type `T` would work fine with both `<T>` and `<? extends T>` since `TypeToken` "produces" only, not "consumes". It seems that not making it covariant would work fine even if the type token class and this method might need something to "consume". So leaving it merely `<T>` is most likely the best choice. I still feel I'm missing to know something. Nevermind.

wind57: @lyubomyr-shaydariv strictly speaking one is a bounded wildcard and the other is a type parameter, but that is nit-picking (that I bet you are aware of anyway), for the type system these are pretty distinct things. It seems to me that in this particular case it makes no difference at all, so I'll second your reasonings. As a matter of fact, I like this change, I never understood why it never went in initially.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -76,26 +76,30 @@
  * <pre>
  * Gson gson = new Gson(); // Or use new GsonBuilder().create();
  * MyType target = new MyType();
- * String json = gson.toJson(target); // serializes target to Json
+ * String json = gson.toJson(target); // serializes target to JSON
  * MyType target2 = gson.fromJson(json, MyType.class); // deserializes json into target2
  * </pre>
  *
  * <p>If the object that your are serializing/deserializing is a {@code ParameterizedType}
```
eamonnmcmanus, line 83: You could fix the typo "your" here while you are at it.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -76,26 +76,30 @@
...
  * (i.e. contains at least one type parameter and may be an array) then you must use the
- * {@link #toJson(Object, Type)} or {@link #fromJson(String, Type)} method. Here is an
- * example for serializing and deserializing a {@code ParameterizedType}:
- *
+ * {@link #toJson(Object, Type)} or {@link #fromJson(String, TypeToken)} method. Gson provides
```
eamonnmcmanus, line 85: Since the text says "you _must_ use", I think it should detail all the options, so both `Type` and `TypeToken` overloads. There may be cases where it is still preferable to use the `Type` one. `TypeToken` could come first, though, and the example show just that.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -932,98 +954,165 @@ public void toJson(JsonElement jsonElement, JsonWriter writer) throws JsonIOExce
    * or if {@code json} is empty.
    * @throws JsonSyntaxException if json is not a valid representation for an object of type
    * classOfT
+   *
+   * @see #fromJson(Reader, Class)
+   * @see #fromJson(String, TypeToken)
    */
   public <T> T fromJson(String json, Class<T> classOfT) throws JsonSyntaxException {
-    Object object = fromJson(json, (Type) classOfT);
+    T object = fromJson(json, TypeToken.get(classOfT));
```
eamonnmcmanus, line 962: I was thinking this was introducing an object allocation where there wasn't one before, but actually the code already does `TypeToken.get(...)` later on.

gson/src/main/java/com/google/gson/Gson.java
eamonnmcmanus: I also don't think there's any reason to use `<? extends T>` here.

eamonnmcmanus (changes requested):
This makes sense to me, and I don't know why the idea was summarily dismissed in 2010. The main drawback I see is that `new TypeToken<List<String>>() {}` is an object allocation and an additional anonymous subclass, but you already have that if you're currently getting a `Type` from the same expression with `.getType()` on the end.

I'm running the change against all of Google's internal Gson client code just in case the new overload introduces some compilation problem, though I can't really see what that would be.

eamonnmcmanus (changes requested):
I think there should probably be at least a couple of new test methods covering the new overloads.

gson/src/main/java/com/google/gson/Gson.java
Marcono1234: I have reworded the whole paragraph a bit because it was also incorrect regarding serialization. I hope the new wording is fine; though please let me know if I should adjust something.

Marcono1234:
> The main drawback I see is that `new TypeToken<List<String>>() {}` is an object allocation and an additional anonymous subclass

It appears there is unfortunately no other solution to this until the Java language gets proper compiler support for this (similar to Kotlin's [reified type parameters](https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters)?). Guava with its `TypeToken` and Jackson with its `TypeReference` seem to take the same approach.

However, you can safely store the `TypeToken` in a `static final` field to avoid object allocations.

> I think there should probably be at least a couple of new test methods covering the new overloads.

Fair point; I have added them to `ParameterizedTypesTest`, I hope the way I implemented them is fine for you.

Also, should there be `toJson` overloads with `TypeToken` parameter? Personally I am not sure if the small advantages (if any?) this provides would justify the new methods. For `fromJson` this is a different story because it provides type safety there.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1041,9 +1132,47 @@ private static void assertFullConsumption(Object obj, JsonReader reader) {
...
+   * @param reader the reader whose next JSON value should be deserialized
+   * @param classOfT the class of T
+   * @return an object of type T from the JsonReader. Returns {@code null} if {@code reader} is at EOF.
+   * @throws JsonIOException if there was a problem reading from the JsonReader
+   * @throws JsonSyntaxException if json is not a valid representation for an object of type classOfT
+   *
+   * @see #fromJson(Reader, Class)
+   * @see #fromJson(JsonReader, TypeToken)
+   */
+  public <T> T fromJson(JsonReader reader, Class<T> classOfT) throws JsonIOException, JsonSyntaxException {
```
eamonnmcmanus, line 1161: Perhaps surprisingly, this is not a source-compatible change. It seems completely logical to add this overload to the other `fromJson(JsonReader...)` ones, but there's an issue. When I ran this PR against all of Google's builds, I encountered [this](https://github.com/googleapis/gax-java/blob/3cba6f7d505a9bb69139239b8f9d155a749e4435/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ApiMessageHttpResponseParser.java#L76):
```java
  private static <ResponseT extends ApiMessage> ApiMessageHttpResponseParser<ResponseT> create(
      final ResponseT responseInstance) {
    final Gson baseGson = new GsonBuilder().create();
    TypeAdapter responseTypeAdapter;
    Gson responseMarshaller = null;

    if (responseInstance != null) {
      responseTypeAdapter =
          new TypeAdapter<ResponseT>() {
            @Override
            public void write(JsonWriter out, ResponseT value) {
              baseGson.toJson(value, responseInstance.getClass(), out);
            }

            @Override
            public ResponseT read(JsonReader in) {
              return baseGson.fromJson(in, responseInstance.getClass());
            }
          };
```

This is not type-safe. The type of `responseInstance.getClass()` is `Class<? extends ApiMessage>`, and `? extends ApiMessage` is not assignable to `ResponseT`. It compiles without this PR because there is no `Class<T>` overload so it gets the type-unsafe `Type` overload.

If the code here instead had `return baseGson.fromJson(in, new TypeToken<ResponseT>() {}.getType())` then it _would_ compile, with or without this PR, and then with this PR the `.getType()` can be removed. It also looks as if this class no longer exists in the current version of `gax-java`.

I found a second problem with code that was calling this method with `Map.class`. The unchecked conversion from raw `Map` to `Map<String, ?>` drew a warning, and the code was being compiled with `-Werror`. The same kind of fix applies, of course.

On the whole, I think we should drop this overload. It's probably better to retain the minor inconsistency of not having it than to expose users to new problems.

eamonnmcmanus (changes requested):
New test code looks good! Just one more thing...

eamonnmcmanus:
I forgot to answer the question about `toJson` overloads. I agree that they are probably not worthwhile.

gson/src/main/java/com/google/gson/Gson.java
Marcono1234: Thanks for running it against all Google builds.

> If the code here instead had `return baseGson.fromJson(in, new TypeToken<ResponseT>() {}.getType())` then it _would_ compile

But it would not be type safe either, in fact it would even be less type safe than the original code respectively not work properly at all because it would parse the JSON data as the erasure of the type variable (`ApiMessage`). This is why it would have been good to fix #1219, and why I had this originally in #2072 😕. The previous code would not have been type safe though in case `ResponseT` was a parameterized type.

> I found a second problem with code that was calling this method with `Map.class`. The unchecked conversion from raw `Map` to `Map<String, ?>` drew a warning

For what it is worth this would not have been safe either. Even though one might expect that a raw `Map` can only be deserialized to a `Map<String, ...>` due to JSON only permitting strings as member names, Gson actually supports non string values as seen with #1247.

So yes, this will break source compatibility, but it might get users to write safer code, or at least become aware that their code is not safe and then manually suppress the compiler warnings. Therefore, personally I think it might be acceptable to break source compatibility here; but maybe there are cases where the existing code is actually safe and this would create too much trouble...

If you really want, I can remove it though.

(This also goes into the same direction why I mentioned in the description of this pull request to consider deprecating the `fromJson(..., Type)` overloads. That would be rather extreme, but maybe it could be considered at some point in the future.)

eamonnmcmanus: Well, the title of this PR is _Add Gson.fromJson(..., TypeToken) overloads_, so perhaps we can keep it focused on that. If you want to make a separate PR that adds `fromJson(JsonReader, Class<T>)` then we can debate the source-incompatibility problem there.

The situation is a bit different from #2072, in that already-compiled code will continue to work. It's only new compilations that might fail, and probably only when the code being compiled is doing type-unsafe things. So I think there's a stronger case to be made for the change. Nevertheless, I think it should be separate, if only so I can prepare Google's code-base by fixing the problems I mentioned.

Regarding the type-safety of the first problem, I think the underlying issue is that the method doesn't have enough information to behave correctly. It would need to have something like a `Class<ResponseT>` parameter, and then of course that could be correctly passed to the existing `Type` overload, or to the new `Class<T>` overload. I agree that the new `Class<T>` overload makes it possible to discover this problem, while the new `TypeToken<T>` overload just gives us a _different_ way to sweep the unsoundness under the carpet. I further agree that ideally we would simply fix #1219 to disallow that unsoundness. I haven't changed my mind since #2072, though: I still feel that it's extremely important that people can update their Gson dependency without breaking things that currently work. I'll add that this would be yet _another_ case where a compatibility-level concept might allow us to fix the errors of the past.

Marcono1234: You are right, I have removed the method again and hope the changes are fine like that.

Sorry for being a bit relentlessly. It is just pretty frustrating and annoying knowing that there are bugs / flaws with Gson and seeing that they cannot be fixed or wasting work on trying to fix them. This is of course not your fault, and I think it is quite valuable that you have a different view of maintaining a library as popular as Gson.

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks for your patience, and for all the hard work you've put into this PR!

PR #1701, opened 2020-05-21
Fix #1699: Don't catch OutOfMemoryError

### LINKED ISSUE #1699: Don't catch OutOfMemoryError
Gson currently catches `OutOfMemoryError`s at two places and wraps them inside `JsonParseException`:
https://github.com/google/gson/blob/c5a3f21fbad3828deba1deb88168e1f76829f807/gson/src/main/java/com/google/gson/JsonParser.java#L88-L89
https://github.com/google/gson/blob/b75e1bbc7915f409ba0254fcc92c71ed6798e69b/gson/src/main/java/com/google/gson/JsonStreamParser.java#L90-L91

This is bad practice because Gson is not necessarily the cause for this error and wrapping it inside an `Exception` subclass prevents the caller from noticing the `OutOfMemoryError` until later.

Marcono1234:
Fix #1699

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonParser.java
+++ b/gson/src/main/java/com/google/gson/JsonParser.java
@@ -85,8 +85,6 @@ public final class JsonParser {
       return Streams.parse(reader);
     } catch (StackOverflowError e) {
       throw new JsonParseException("Failed parsing JSON source: " + reader + " to Json", e);
-    } catch (OutOfMemoryError e) {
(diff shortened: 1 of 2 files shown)


Marcono1234:
Withdrawing this PR because catching `OutOfMemoryError` _might_ be desired, see also https://github.com/google/gson/pull/1912#issuecomment-1130767871. However, it appears `OutOfMemoryError` is not caught consistently everywhere.

PR #1704, opened 2020-05-21
Fix  #1697: Mention that GsonBuilder.registerTypeAdapter makes (de-)serialiers null-safe

### LINKED ISSUE #1697: Clarify that GsonBuilder.registerTypeAdapter makes JsonSerializer and JsonDeserializer null-safe
When calling `GsonBuilder.registerTypeAdapter` with a `JsonSerializer` or `JsonDeserializer`, it is made `null`-safe (see `TreeTypeAdapter`), i.e. the (de-)serializer will not be consulted when a `null` value is handled.

It appears this is currently not documented.

It should be documented in `registerTypeAdapter` and it should refer to `TypeAdapter` as alternative which does allow handling `null`-values.

Related:  #1553, #980, #852

Marcono1234:
Fix  #1697


eamonnmcmanus:
LGTM

PR #1706, opened 2020-05-23
Improve AppendableWriter performance

Marcono1234:
Override methods which by default create char arrays or convert CharSequences to Strings.
This is not necessary for `AppendableWriter` because it can directly append these values to the `Appendable` delegate.


Marcono1234:
@eamonnmcmanus do you think this can be merged?

eamonnmcmanus:
LGTM

PR #1656, opened 2020-02-12
Warning fixes

Marcono1234:
Fixes some `javac` warnings and implements `DefaultDateTypeAdapter` (internal class) in a type-safer way.

PR #1707, opened 2020-05-24
Make dependency on java.sql optional

### LINKED ISSUE #1629: Remove support for java.sql types
The original idea #134 was OK in 2015. Since then, Java got modular support and java.sql is no longer available by default. This is causing sporadic issues in random places: 

- https://github.com/google/gson/issues/134#issuecomment-410848727 
- [java.lang.ClassNotFoundException: java.sql.Time](https://travis-ci.org/dernasherbrezon/jsp-openapi/builds/620333411)
- [stackoverflow](https://stackoverflow.com/search?q=gson+java.sql)
- [gson pom.xml](https://github.com/google/gson/blob/master/pom.xml#L84)

Would be great to deprecate support for java.sql and suggest writing type adapters instead.

Marcono1234:
Note: This is based on #1656
Fixes #1629

Makes the dependency on the `java.sql` module optional. If its classes are present at runtime, the respective type adapters will be used, otherwise Gson will not try to load the classes and will therefore not fail with a `ClassNotFoundException` anymore. The entry point for all SQL type adapters is now `SqlTypesSupport`.

I am however, not that familiar with the module system so it would be good if others who are more familiar with it can check if this is a sane implementation.

Especially #1500 (@Degubi) is confusing me because from what I understood about the JPMS `requires transitive` is only so modules depending on Gson would also be "reading" `java.sql`, which however makes no sense because Gson does not expose these types publicly anywhere, so it should not provide them transitively. To my understanding `requires` should have been enough to require `java.sql` being present at runtime, however based on the `ClassNotFoundException` reports this was not the case?
Or are they using tools (as it is the case [here](https://stackoverflow.com/q/57222110) which break the module system)? Though that would be a rather common issue then, see the question ["java.sql.Time exception"](https://stackoverflow.com/q/50877484).

Also pinging @nicolaiparlog, if you have any spare time, any hints regarding the JPMS would be appreciated :)
Sorry to pester you in case you don't have any time.

### DIFF
--- a/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/DefaultDateTypeAdapter.java
@@ -41,11 +41,52 @@ import com.google.gson.stream.JsonWriter;
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
-final class DefaultDateTypeAdapter extends TypeAdapter<Date> {
-
+final class DefaultDateTypeAdapter<T extends Date> extends TypeAdapter<T> {
   private static final String SIMPLE_NAME = "DefaultDateTypeAdapter";
 
-  private final Class<? extends Date> dateType;
+  static abstract class DateType<T extends Date> {
+    private DateType() {
+    }
+
+    public static final DateType<Date> DATE = new DateType<Date>() {
+      @Override
+      protected Date deserialize(Date date) {
+        return date;
+      }
+    };
+    public static final DateType<java.sql.Date> SQL_DATE = new DateType<java.sql.Date>() {
+      @Override
+      protected java.sql.Date deserialize(Date date) {
+        return new java.sql.Date(date.getTime());
+      }
+    };
+    public static final DateType<Timestamp> SQL_TIMESTAMP = new DateType<Timestamp>() {
+      @Override
+      protected Timestamp deserialize(Date date) {
+        return new Timestamp(date.getTime());
+      }
+    };
+
+    protected abstract T deserialize(Date date);
+
+    public DefaultDateTypeAdapter<T> createAdapter(String datePattern) {
+      return new DefaultDateTypeAdapter<T>(this, datePattern);
+    }
+
+    public DefaultDateTypeAdapter<T> createAdapter(int style) {
+      return new DefaultDateTypeAdapter<T>(this, style);
+    }
+
+    public DefaultDateTypeAdapter<T> createAdapter(int dateStyle, int timeStyle) {
+      return new DefaultDateTypeAdapter<T>(this, dateStyle, timeStyle);
+    }
+
+    public DefaultDateTypeAdapter<T> createDefaultsAdapter() {
+      return new DefaultDateTypeAdapter<T>(this, DateFormat.DEFAULT, DateFormat.DEFAULT);
+    }
+  }
+
+  private final DateType<T> dateType;
 
   /**
    * List of 1 or more different date formats used for de-serialization attempts.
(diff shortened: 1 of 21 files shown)


nicolaiparlog:
Don't worry, I don't pester easily. :wink: I'm happy to provide input, but keep in mind that I don't know GSON or the diff in detail (don't have the time to study them right now), so I may be off. Caveat emptor, I guess. Also, I wrote [a blog post about optional dependencies](https://blog.codefx.org/java/module-system-optional-dependencies/) that may help understand the feature.

As I understand the situation, GSON only needs to work with _java.sql_ types if the user already uses them, right? I.e. there's never a situation where a GSON user doesn't need the _java.sql_ module, but GSON uses its types anyway. If that is so, `requires static` is the right way to go.

Note that `requires static` alone does not make the module available at run time. That means, even if GSON is launched on a Java runtime that contains the module, it will only be available at run time if some other module requires it (or the app is launched from the class path).

In case you're interested, the blog post shows a method `isModulePresent(String)` that, well, does what the name suggests. :grin: You can use it (I hereby provide it under the Apache Software License 2.0) if you need to programmatically determine whether _java.sql_ made it into the module graph.

I agree with your confusion about #1500. Turning `requires` into `requires transitive` doesn't make a difference regarding its presence at run time - it is present either way ([another vlog post](https://blog.codefx.org/java/implied-readability/)). I agree with [Alan Bateman's comment](https://stackoverflow.com/questions/50877484/java-sql-time-exception#comment88763419_50877484) - this looks like class loader shenanigans to me.

Marcono1234:
Thanks for your clarification! Yes Gson does not need any `java.sql` types on its own, it only provides (de-)serialization support for them in case they are used.

Your `isModulePresent(String)` is quite verbose for what it tries to solve. Though maybe there is no better / more complete way of doing this (I have created a [StackOverflow question](https://stackoverflow.com/q/61703051) for that). However, Gson cannot use this solution because it wants to support Java 6, so this pull requests checks using `Class.forName` for a class present in `java.sql`, assuming that the other classes must be present then as well:
https://github.com/google/gson/blob/a4a235e14a63b9a4f902805641f1160c6bf5aa93/gson/src/main/java/com/google/gson/internal/sql/SqlTypesSupport.java#L37-L40

Regarding previously reported `ClassNotFoundException` and class loader shenanigans: It looks like Gson itself does not create or uses any special class loaders, so I assume the callers have a broken setup then.

mP1:
Heres another example of the same problem this time with CLosureCompiler, so+++ from me as well.

```java
[INFO]         java.lang.SecurityException: Prohibited package name: java.sql
[INFO]         	at java.base/java.lang.ClassLoader.preDefineClass(ClassLoader.java:889)
[INFO]         	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1005)
[INFO]         	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
[INFO]         	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:545)
[INFO]         	at java.base/java.net.URLClassLoader.access$100(URLClassLoader.java:83)
[INFO]         	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:453)
[INFO]         	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:447)
[INFO]         	at java.base/java.security.AccessController.doPrivileged(Native Method)
[INFO]         	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:446)
[INFO]         	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClassFromSelf(ClassRealm.java:425)
[INFO]         	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:42)
[INFO]         	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
[INFO]         	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
[INFO]         	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
[INFO]         	at com.google.gson.Gson.<init>(Gson.java:265)
[INFO]         	at com.google.gson.Gson.<init>(Gson.java:186)
[INFO]         	at com.google.javascript.jscomp.AbstractCommandLineRunner.<init>(AbstractCommandLineRunner.java:224)
```

neilwightman-intergral:
This pull request worked great for me in a java agent (which cant use java.sql)

When building it I got this test failure though : 

`Failed tests:   testDateFormatString(com.google.gson.internal.bind.util.ISO8601UtilsTest): expected:<2018-06-2[5]> but was:<2018-06-2[4]>`

I'm wondering if this is a timezone issue.  
Tests on master branch passed.

Marcono1234:
Yes the `ISO8601UtilsTest` is faulty and will be fixed by #1665 / #1687

concision:
Thanks for your work on resolving this issue with this pull request, @Marcono1234.

@inder123, Is it possible we could get a review on this PR? 

When using the `jlink` CLI tool to build a custom minimal JRE, `gson`'s explicit module dependency adds another 10MB+ of Java module dependencies. A prominent use case of using `jlink` is building lean Docker images; however, this explicit SQL dependency adds dead weight to the final image (a 20% size increase for my project). Intentionally removing the `java.sql` module results in a runtime error of `java.lang.NoClassDefFoundError: java/sql/Time`, despite it not being necessary for execution.

The indirect/transitive module dependencies that end up being added are listed here: https://docs.oracle.com/javase/9/docs/api/java.sql-summary.html

I would really appreciate seeing the `java.sql` dependency being optional. Thank you for your time.

PR #1711, opened 2020-05-26
Update links and use HTTPS for them

Marcono1234:
- Update links to what they are currently redirecting
- Use HTTPS for links

I have verified that all the links where I changed `http` -> `https` actually support HTTPS.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -25,7 +25,7 @@ import java.io.Reader;
 import java.util.Arrays;
 
 /**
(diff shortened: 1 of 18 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -172,7 +172,7 @@
  *
  * <a id="nonexecuteprefix"/><h3>Non-Execute Prefix</h3>
  * Web servers that serve private data using JSON may be vulnerable to <a
- * href="http://en.wikipedia.org/wiki/JSON#Cross-site_request_forgery">Cross-site
+ * href="https://en.wikipedia.org/wiki/JSON#Cross-site_request_forgery">Cross-site
```
Marcono1234, line 175: Side note: This section does not exist anymore in the Wikipedia article.

PR #1712, opened 2020-05-27
Fix fallback behavior of UnsafeReflectionAllocator when AccessibleObject isn't so accessible

rhernandez35:
I'm the lucky owner of a post-jigsaw Java platform that has restrictive security manager installed. Failing to catch the `SecurityException` thrown by a cross-classloader `getDeclaredField` call in `UnsafeReflectionAllocator::getOverrideField` prevents any Gson instances from being created.

I'm hoping the fallback behavior in this class will be enough for us to use Gson without having to loosen our security policy.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java
+++ b/gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java
@@ -79,7 +79,7 @@ final class UnsafeReflectionAccessor extends ReflectionAccessor {
   private static Field getOverrideField() {
     try {
       return AccessibleObject.class.getDeclaredField("override");
-    } catch (NoSuchFieldException e) {
+    } catch (Exception e) {
       return null;
     }
   }
--- a/gson/src/test/java/com/google/gson/internal/reflect/UnsafeReflectionAccessorTest.java
+++ b/gson/src/test/java/com/google/gson/internal/reflect/UnsafeReflectionAccessorTest.java
@@ -15,10 +15,12 @@
  */
 package com.google.gson.internal.reflect;
 
+import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/internal/reflect/UnsafeReflectionAccessor.java
```diff
@@ -79,7 +79,7 @@ private static Object getUnsafeInstance() {
   private static Field getOverrideField() {
     try {
       return AccessibleObject.class.getDeclaredField("override");
-    } catch (NoSuchFieldException e) {
+    } catch (Exception e) {
```
Marcono1234, line 82: I think it would be better to explicitly catch `NoSuchFieldException` and `SecurityException` to make the intention more clear (though the other methods in that class simply catch `Exception` as well).
Sadly multi-catch was added in Java 7, but since the code in the `catch` block is only a simple return, I think it would be okay to duplicate it.

rhernandez35: hmm.... personally, I'd prefer to leave as is. this access being blocked by my security policy was the specific reason for me to make this change, but I want to try to use the fallback behavior in `makeAccessible` if we should fail to retrieve this field for _any_ reason whatsoever. this increases Gson's chances of continuing to work in strange environments or future JDKs and makes the failure reason and resolution steps more obvious when it can't (compare the nice, readable exception thrown in `makeAccessible` to the arcane stack trace barfed up by ReflectionAccessor's class initializer)

I'm less concerned with signaling intent (which imo is still obvious) than I am with preserving function.

rhernandez35:
@googlebot I signed it!

eamonnmcmanus:
This looks reasonable. Thanks for the carefully-constructed test!

PR #1716, opened 2020-06-04
Improve ArrayTypeAdapter for Object[]

Marcono1234:
Currently `ArrayTypeAdapter` adds the list elements one by one to the created array on deserialization. For primitive arrays this is required, but for `Object[]` this pull request now changes it to call `ArrayList.toArray` instead which should perform much better because it internally performs an `System.arraycopy` from the internal array backing the list to the destination array.


Marcono1234:
@eamonnmcmanus do you think this can be merged?

eamonnmcmanus:
LGTM

PR #1723, opened 2020-06-17
Clarify doc about non-finite numbers for non-lenient JsonReader

Marcono1234:


Marcono1234:

PR #1745, opened 2020-07-31
Rename internal classes whose name contains a `$`

### LINKED ISSUE #1744: Rename internal Gson classes containing $
Please rename Gson classes whose name contains a `$`:
- `com.google.gson.internal.$Gson$Types`
- `com.google.gson.internal.$Gson$Preconditions`

[JLS 14 §3.8](https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.8) says:
>The dollar sign should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

It appears these classes were renamed by 4efb133b4a6cc93708dda782d948b9c88c3b5ff0 to prevent IDE auto-importing. However, nowadays IDEs (at least Eclipse) have ways to prevent certain types or packages from appearing in import suggestions.
Additionally these classes are in the `internal` package and with the Java module system they won't be visible to users anymore.

For now maybe simply removing the `$`s and adding a documentation comment that these classes are internal might already solve the issue.

Currently when developing Gson having to deal with these classes is (at least under Eclipse but I assume IntelliJ has similar issues) pretty annoying:
- Eclipse fails adding the imports
- Eclipse suggests bogus types in the "Open Types" dialog
- Using conditional breakpoints in these classes is not possible at all because Eclipse incorrectly assumes they are nested classes
- ... _likely more_

Marcono1234:
Fixes #1744

### DIFF
--- a/gson/src/main/java/com/google/gson/FieldAttributes.java
+++ b/gson/src/main/java/com/google/gson/FieldAttributes.java
@@ -16,7 +16,7 @@
 
 package com.google.gson;
 
-import com.google.gson.internal.$Gson$Preconditions;
+import com.google.gson.internal.GsonPreconditions;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Type;
@@ -42,7 +42,7 @@ public final class FieldAttributes {
    * @param f the field to pull attributes from
    */
   public FieldAttributes(Field f) {
-    $Gson$Preconditions.checkNotNull(f);
+    GsonPreconditions.checkNotNull(f);
     this.field = f;
   }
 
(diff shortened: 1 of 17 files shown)


inder123:
Is there another character that we can use? Like _

Marcono1234:
Yes using the underscore would be possible as well:
> The "Java letters" include [...] for historical reasons, the ASCII dollar sign (`$`, or `\u0024`) and underscore (`_`, or `\u005f`). [...] The underscore may be used in identifiers formed of two or more characters, but it cannot be used as a one-character identifier due to being a keyword. 

[JLS 14 §3.8](https://docs.oracle.com/javase/specs/jls/se14/html/jls-3.html#jls-3.8)

So `_GsonTypes` and `_GsonPreconditions` should work as well.

PR #1757, opened 2020-08-19
Support force writing null JSON value, ignoring getSerializeNulls()

### LINKED ISSUE #532: Feature request: Force inclusion of specific field when null
```
Gson allows blanket-inclusion of null fields, using serializeNulls().  But this 
produces much bigger JSON output (which is why the default is not to serialize 
nulls).

It would be nice to support a per-field annotation, for just those fields where 
(broken) clients expect a null to be provided.  The alternative is to turn on 
serializeNulls everywhere (I think).

Jackson supports this with @JsonInclude(Include.ALWAYS).
```

Original issue reported on code.google.com by `jus...@fathomdb.com` on 26 Aug 2013 at 12:58

Marcono1234:
Fixes #532

Useful for cases where program is communicating with third-party application which requires explicit `null` for certain fields, but globally enabling `null` serialization is not possible / would be wasteful.

### DIFF
--- a/gson/src/main/java/com/google/gson/annotations/ForceSerializeNull.java
+++ b/gson/src/main/java/com/google/gson/annotations/ForceSerializeNull.java
@@ -1,0 +1,80 @@
+package com.google.gson.annotations;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import com.google.gson.Gson;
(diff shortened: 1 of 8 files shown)


inder123:
Another lighterweight option could be to take a var-args list of classes in GsonBuilder.serializeNulls() method.

Marcono1234:
I have replaced `@ForceSerializeNull` with `@SerializeNull` now which overwrites `JsonWriter.getSerializeNulls()` for that field. This allows additionally to the initial functionality of this pull request to force `null` values, to also omit `null` values regardless of `getSerializeNulls()`.

PR #1759, opened 2020-08-20
Clarify doc about TypeAdapter fieldNamingStrategy default behavior

ravs:
When using TypeAdapter and Gson with setFieldNamingPolicy, if you are not setting TypeAdapter @Gson.TypeAdapters(fieldNamingStrategy = true), then fieldNamingStrategy is ignored.


ravs:
@googlebot I signed it!

Marcono1234:
As mentioned above; please report this to the other repository, it is not directly related to Gson.

PR #1766, opened 2020-08-25
Fix JsonWriter documentation regarding top-level value

Marcono1234:
RFC 7159 allows any top-level value (not only arrays or objects) [0], however when #773 added this functionality it appears the author forgot to update the documentation of these classes.

[0] https://tools.ietf.org/html/rfc7159#appendix-A
> Changed the definition of "JSON text" so that it can be any JSON value, removing the constraint that it be an object or array.


Marcono1234:
@eamonnmcmanus, can you please have a look?

PR #1767, opened 2020-08-25
Fix JsonReader.nextDouble() throwing wrong exception type for NaN and Infinity

Marcono1234:
The documentation says that a NumberFormatException will be thrown in case the value is not finite (i.e. NaN or Infinity). However, the JsonReader implementation actually throws a MalformedJsonException.

Note that JsonTreeReader currently throws a NumberFormatException, as expected.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -907,7 +907,7 @@ public class JsonReader implements Closeable {
     peeked = PEEKED_BUFFERED;
     double result = Double.parseDouble(peekedString); // don't catch this NumberFormatException.
     if (!lenient && (Double.isNaN(result) || Double.isInfinite(result))) {
-      throw new MalformedJsonException(
+      throw new NumberFormatException(
           "JSON forbids NaN and infinities: " + result + locationString());
     }
     peekedString = null;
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -907,7 +907,7 @@ public double nextDouble() throws IOException {
     peeked = PEEKED_BUFFERED;
     double result = Double.parseDouble(peekedString); // don't catch this NumberFormatException.
     if (!lenient && (Double.isNaN(result) || Double.isInfinite(result))) {
-      throw new MalformedJsonException(
```
inder123, line 910: Seems like it is indeed bad JSON since JSON forbids NaN and infinities. 
The number may actually be in correct format.

inder123:
When in doubt, I prefer to preserve the current behavior.

Marcono1234:
So should the documentation then be updated to state that a MalformedJsonException is thrown and JsonTreeReader be adjusted accordingly? Because the current behavior is inconsistent.

gson/src/main/java/com/google/gson/stream/JsonReader.java
Marcono1234: Though then the question is why `NumberFormatException` is thrown in the first place when the number cannot be parsed as double, int or long since it is malformed JSON in these cases as well.

PR #1770, opened 2020-08-27
Update to latest version of bnd-maven-plugin

mbooth101:
This version has a couple of advantages:

1) It is aware of module-info.class files, which means we don't
need to manually override the "Require-Capability" header.

2) External packages used by reflection are now detected, i.e.
the "sun.misc" package is now added to the "Import-Package"
header. This is manually given optional resolution since it
is obviously not present on all JVMs.

3) It also introduces some new useful OSGi headers:
 - Bundle-DocURL (overriden with correct URL for manual)
 - Bundle-License
 - Bundle-SCM

This change also removes unnecessary mention of the felix
bundle-plugin in the pluginManagement section of the parent pom.

Signed-off-by: Mat Booth <mat.booth@redhat.com>

### DIFF
--- a/gson/bnd.bnd
+++ b/gson/bnd.bnd
@@ -3,8 +3,9 @@ Bundle-Name: ${project.name}
 Bundle-Description: ${project.description}
 Bundle-Vendor: Google Gson Project
 Bundle-ContactAddress: ${project.parent.url}
+Bundle-DocURL: ${project.parent.url}/blob/master/UserGuide.md
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6, JavaSE-1.7, JavaSE-1.8
-Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
+Import-Package: sun.misc;resolution:=optional,*
 
 -removeheaders: Private-Package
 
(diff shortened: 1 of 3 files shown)


mbooth101:
@googlebot I signed it!

mbooth101:
Any thoughts on this?

Marcono1234:
Thanks for the PR! Some of this has already been implemented by #1993 and the bnd-maven-plugin has been updated as well. `Bundle-DocURL` seems to be auto-generated to point to this GitHub repository, so maybe that is fine already.

> It is aware of module-info.class files, which means we don't need to manually override the "Require-Capability" header.

Not sure if that would yield the desired results (and work correctly because `module-info.class` generation is a bit complex at the moment, see #2013) because Gson still targets Java 7 and just includes `module-info.class` for convenience for Java 9 or newer users.

Though `Require-Capability` seems to be auto-generated (when omitted from `bnd.bnd`), but maybe it does not hurt to specify it explicitly anyway.

Do you want to update your PR nonetheless in case you think there are still changes needed for Gson?

mbooth101:
> Do you want to update your PR .....?

Not really, looks like this PR was ignored (and I totally forgot about it tbh) and someone ended up duplicating the work in another PR.... Feel free to close.

PR #1782, opened 2020-09-13
Fix JsonTreeReader throwing wrong exception type for non-finite doubles

Marcono1234:
Follow-up for #1767
Documentation for `JsonReader` is adjusted by #1723.

Either `JsonReader` or `JsonTreeReader` has to be adjusted because their behavior is currently inconsistent.


Marcono1234:

Issue #1575, opened 2019-09-11
JsonStreamParser.hasNext() throwing JsonIOException when given empty string


NEUDitao:
If we construct a JsonStreamParser with an empty string in the reader, when hasNext is called, a JsonIOException is thrown, instead of being returned false.

Issue #330, opened 2015-03-19
EOFException in JsonReader on an empty document


GoogleCodeExporter:
```
What steps will reproduce the problem?

  public void test() throws IOException {
    JsonReader reader = new JsonReader(new StringReader(""));
    assertEquals(JsonToken.END_DOCUMENT, reader.peek());
  }

What is the expected output? What do you see instead?

Expected test to pass, but instead it throws an exception:

java.io.EOFException: End of input
    at com.google.gson.stream.JsonReader.nextNonWhitespace(JsonReader.java:887)
    at com.google.gson.stream.JsonReader.nextValue(JsonReader.java:787)
    at com.google.gson.stream.JsonReader.quickPeek(JsonReader.java:365)
    at com.google.gson.stream.JsonReader.peek(JsonReader.java:340)
...

What version of the product are you using? On what operating system?

GSON 1.6

Please provide any additional information below.

Interestingly, we catch EOFException in quickPeek() for the NONEMPTY_DOCUMENT 
scope, but not for the EMPTY_DOCUMENT scope.
```

Original issue reported on code.google.com by `yan...@google.com` on 2 Jun 2011 at 4:10

PR #1785, opened 2020-09-18
Allow usage of JsonReader::peek() on empty stream in lenient mode

StrongerXi:
**Repro**
```java
void processStreamOfJsonValues(Reader stream) throws IOException {
  JsonReader jReader = new JsonReader(stream);
  jReader.setLenient(true)
  while (jReader.peek() != JsonToken.END_DOCUMENT) {
    JsonElement e = JsonParser.parseReader(jReader);
    // handle `e`
  }
}
...
```
**Expected**: Working even if `stream` is empty (i.e. has reached EOF).
**Actual**: If `stream` is empty, `jReader.peek()` throws an `EOFException`

## Summary
In the above function, the input Reader stream could be a socket stream, string stream, or STDIN, and it could be empty.
However, currently `jReader.peek()` will throw an `EOFException`. This issue has been discussed in #330, and the conclusion seems to be "we can allow empty stream in lenient mode", yet the issue is left unresolved.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -925,7 +925,10 @@ public final class Gson {
     boolean oldLenient = reader.isLenient();
     reader.setLenient(true);
     try {
-      reader.peek();
+      JsonToken token = reader.peek();
+      if (token == JsonToken.END_DOCUMENT) {
+        return null; // we can also throw EOFException here but...
+      }
       isEmpty = false;
       TypeToken<T> typeToken = (TypeToken<T>) TypeToken.get(typeOfT);
       TypeAdapter<T> typeAdapter = getAdapter(typeToken);
--- a/gson/src/main/java/com/google/gson/internal/Streams.java
+++ b/gson/src/main/java/com/google/gson/internal/Streams.java
@@ -23,6 +23,7 @@ import com.google.gson.JsonParseException;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.internal.bind.TypeAdapters;
 import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import com.google.gson.stream.MalformedJsonException;
 import java.io.EOFException;
@@ -43,7 +44,10 @@ public final class Streams {
   public static JsonElement parse(JsonReader reader) throws JsonParseException {
     boolean isEmpty = true;
     try {
-      reader.peek();
+      JsonToken token = reader.peek();
+      if (token == JsonToken.END_DOCUMENT) {
+        return JsonNull.INSTANCE; // we can also throw EOFException here but...
+      }
       isEmpty = false;
       return TypeAdapters.JSON_ELEMENT.read(reader);
     } catch (EOFException e) {
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -529,8 +529,8 @@ public class JsonReader implements Closeable {
         throw syntaxError("Expected ':'");
       }
     } else if (peekStack == JsonScope.EMPTY_DOCUMENT) {
-      if (lenient) {
-        consumeNonExecutePrefix();
+      if (lenient && consumeNonExecutePrefix()) {
+        return peeked = PEEKED_EOF;
       }
       stack[stackSize - 1] = JsonScope.NONEMPTY_DOCUMENT;
     } else if (peekStack == JsonScope.NONEMPTY_DOCUMENT) {
@@ -1563,25 +1563,30 @@ public class JsonReader implements Closeable {
   }
 
   /**
-   * Consumes the non-execute prefix if it exists.
+   * Consumes the non-execute prefix if it exists and return whether eof has
+   * been reached.
    */
-  private void consumeNonExecutePrefix() throws IOException {
+  private boolean consumeNonExecutePrefix() throws IOException {
     // fast forward through the leading whitespace
-    nextNonWhitespace(true);
-    pos--;
+    int c = nextNonWhitespace(false);
+    if (c == -1) {
+      return true;
+    } else {
+      pos--;
+    }
 
     int p = pos;
     if (p + 5 > limit && !fillBuffer(5)) {
-      return;
+      return false;
     }
 
     char[] buf = buffer;
-    if(buf[p] != ')' || buf[p + 1] != ']' || buf[p + 2] != '}' || buf[p + 3] != '\'' || buf[p + 4] != '\n') {
-      return; // not a security token!
+    if(! (buf[p] != ')' || buf[p + 1] != ']' || buf[p + 2] != '}' ||
+          buf[p + 3] != '\'' || buf[p + 4] != '\n')) {
+      // we consumed a security token!
+      pos += 5;
     }
-
-    // we consumed a security token!
-    pos += 5;
+    return false;
   }
 
   static {
--- a/gson/src/test/java/com/google/gson/JsonParserTest.java
+++ b/gson/src/test/java/com/google/gson/JsonParserTest.java
@@ -19,12 +19,14 @@ package com.google.gson;
 import java.io.CharArrayReader;
 import java.io.CharArrayWriter;
 import java.io.StringReader;
+import java.io.IOException;
 
 import junit.framework.TestCase;
 
 import com.google.gson.common.TestTypes.BagOfPrimitives;
 import com.google.gson.internal.Streams;
 import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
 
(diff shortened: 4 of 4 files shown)


StrongerXi:
@googlebot I signed it!

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1563,25 +1563,30 @@ private IOException syntaxError(String message) throws IOException {
...
     if (p + 5 > limit && !fillBuffer(5)) {
-      return;
+      return false;
     }
 
     char[] buf = buffer;
-    if(buf[p] != ')' || buf[p + 1] != ']' || buf[p + 2] != '}' || buf[p + 3] != '\'' || buf[p + 4] != '\n') {
-      return; // not a security token!
+    if(! (buf[p] != ')' || buf[p + 1] != ']' || buf[p + 2] != '}' ||
+          buf[p + 3] != '\'' || buf[p + 4] != '\n')) {
```
Marcono1234, line 1585: Might be good to remove the doubled negation:
```suggestion
    if (buf[p] == ')' && buf[p + 1] == ']' && buf[p + 2] == '}' && buf[p + 3] == '\'' && buf[p + 4] == '\n') {
```

gson/src/test/java/com/google/gson/JsonParserTest.java
```diff
@@ -19,12 +19,14 @@
 import java.io.CharArrayReader;
```
Marcono1234, line 19: Might be good to add a test for the non-execute prefix as well.

gson/src/main/java/com/google/gson/internal/Streams.java
```diff
@@ -43,7 +44,10 @@ private Streams() {
   public static JsonElement parse(JsonReader reader) throws JsonParseException {
     boolean isEmpty = true;
     try {
-      reader.peek();
+      JsonToken token = reader.peek();
+      if (token == JsonToken.END_DOCUMENT) {
+        return JsonNull.INSTANCE; // we can also throw EOFException here but...
+      }
```
Marcono1234, line 50: Might be good to add a test for this.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -925,7 +925,10 @@ private static void assertFullConsumption(Object obj, JsonReader reader) {
     boolean oldLenient = reader.isLenient();
     reader.setLenient(true);
     try {
-      reader.peek();
+      JsonToken token = reader.peek();
```
Marcono1234, line 928: Might be good to add a test for this.

gson/src/test/java/com/google/gson/JsonParserTest.java
StrongerXi: It seems that there are already [unit tests](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java#L1283) for the non-executable prefix management. Could you elaborate a little?

StrongerXi:
I just pushed a commit to address your comments.

StrongerXi:
I also wonder whether it's reasonable to remove [this block of code](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/src/main/java/com/google/gson/internal/Streams.java#L54), because based on the comment, my PR handles exactly that case without relying on the EOFException.

The exact same block of code comes up [here](https://github.com/google/gson/blob/b75e1bbc7915f409ba0254fcc92c71ed6798e69b/gson/src/main/java/com/google/gson/Gson.java#L939).

gson/src/test/java/com/google/gson/JsonParserTest.java
Marcono1234: Yes, sorry for not being very clear. I meant a new test for a non-execute prefix followed by EOF (which your code now supports).

Marcono1234:
Good point, the behavior of `Streams.parse(...)` regarding EOF seems to be incorrect indeed. It should probably not treat them in any special way since (as shown by you), the `parse(...)` method is also used for non-empty documents.
Though maybe it would be good to create a separate GitHub issue about that.

----

Marcono1234:
This pull request would likely fix #1575.

gson/src/test/java/com/google/gson/JsonParserTest.java
```diff
@@ -116,4 +118,18 @@ public void testReadWriteTwoObjects() throws Exception {
     BagOfPrimitives actualTwo = gson.fromJson(element2, BagOfPrimitives.class);
     assertEquals("two", actualTwo.stringValue);
   }
+
+  public void testPeekEmptyStream() throws IOException {
+    JsonReader reader = new JsonReader(new StringReader(""));
+    try {
+      assertEquals(JsonToken.END_DOCUMENT, reader.peek());
+      fail();
+    } catch (java.io.EOFException expected) { }
```
Marcono1234, line 127: Might be good to add an import for `java.io.EOFException` instead of using its qualified name here.

StrongerXi:
> This pull request would likely fix #1575.
> 
> Might be good to add a test for this as well then, i.e. `assertFalse(new JsonStreamParser("").hasNext())`.

I agree. I actually ended up using `JsonStreamParser` in my project and found this issue as well.

gson/src/test/java/com/google/gson/JsonParserTest.java
StrongerXi: > to make sure that a non-execute prefix followed by EOF still throws an EOFException

Wouldn't we _not_ want an exception in such case? (I assume you meant from using `JsonReader::peek()`).
In strict mode, non-execute prefix won't be accepted.
In lenient mode, it will be ignored and the together following EOF will be interpreted as an empty document.

gson/src/main/java/com/google/gson/Gson.java
StrongerXi: I agree with your other comment that removing `isEmpty` and `EOFException` block here and in `Stream::parse` should be another issue/PR.

Marcono1234: Ah sorry, you might have misunderstood me. My comments were not related to each other. Here in this method I think you can remove the `isEmpty` variable without introducing any behavior change. An `EOFException` will only be thrown either for empty documents or after `peek()` has succeeded at least once (and therefore the document is non-empty), see the `JsonReader` implementation. However with your change in lenient mode no EOFException will be thrown for empty documents. And because this method (`Gson.fromJson`) always makes the reader lenient, it should therefore be impossible that a EOFException is thrown for an empty document.

gson/src/test/java/com/google/gson/JsonParserTest.java
Marcono1234: It is probably subjective what the _right_ behavior is here. I thought if a document contains a non execute prefix it is non-empty (even though it contains no JSON tokens). Though on the other hand a third party library which always writes the non execute prefix and then checks whether there is actually JSON data to write, might consider a non execute prefix followed by EOF to be an empty document.

StrongerXi: Yeah I agree it does seem subjective.

Based on my understanding, the current behavior is what I described in my comment above. Essentially the current implementation in `Stream::parse` uses `EOFException` to determine EOF, whereas my PR makes it an explicit check on what `JsonReader::peek` returns, without throwing `EOFException`.

I thought `EOFException` should be used when detecting EOF in the middle of parsing a JSON value, but not when the document (excluding non-execute prefix in lenient mode) is empty.

gson/src/main/java/com/google/gson/Gson.java
StrongerXi: I see what you are staying. So we can't remove the similar block in `Stream::parse` because the reader isn't necessarily lenient. I'll remove the block here.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -921,27 +921,19 @@ private static void assertFullConsumption(Object obj, JsonReader reader) {
...
     } catch (EOFException e) {
-      /*
-       * For compatibility with JSON 1.5 and earlier, we return null for empty
-       * documents instead of throwing.
-       */
-      if (isEmpty) {
-        return null;
-      }
       throw new JsonSyntaxException(e);
     } catch (IllegalStateException e) {
```
Marcono1234, line 938: You can even remove this complete `catch` block. `EOFException` is an `IOException` and that is already caught below.

Marcono1234:
> Might be good to add a test for this as well then, i.e. `assertFalse(new JsonStreamParser("").hasNext())`.

Could you please add a test for this in `JsonStreamParserTest` if possible?

gson/src/main/java/com/google/gson/Gson.java
StrongerXi: Good point.

PR #1792, opened 2020-09-30
In user guide show using generic classes for deserialization

josephks:
I've noticed that new users read the documentation and think that they have to write a java class for gson to use to parse a json string.  This change makes it a little more obvious that you can just use `gson.fromJson(<string>, Object.class)` and gson will give you something.


josephks:
I don't know how to fix the cla bot problem.  I signed the CLA but the account I used in the commit isn't associated with my Google account and I'd like to keep it that way.

UserGuide.md
```diff
@@ -140,6 +140,18 @@ Note that you can not serialize objects with circular references since that will
 // Deserialization
 BagOfPrimitives obj2 = gson.fromJson(json, BagOfPrimitives.class);
 // ==> obj2 is just like obj
+
+// If you the type of your json is an Object you and deserialize into a generic Map class
+Map map = gson.fromJson(json, Map.class);
+// ==> map is: {value1=1.0, value2=abc}
+
+// Or you can just use Object.class, and trust gson to return an Object of the correct type
```
Marcono1234, line 148:
```suggestion
// Or you can just use Object.class, and trust Gson to return an Object of the correct type
```

UserGuide.md
```diff
@@ -140,6 +140,18 @@ Note that you can not serialize objects with circular references since that will
 // Deserialization
 BagOfPrimitives obj2 = gson.fromJson(json, BagOfPrimitives.class);
 // ==> obj2 is just like obj
+
+// If you the type of your json is an Object you and deserialize into a generic Map class
+Map map = gson.fromJson(json, Map.class);
```
Marcono1234, line 145: Would be better to use `TypeToken` to get a parameterized `Map<..., ...>`. Using raw types is discouraged.
Though the sections "Collections Examples" and "Serializing and Deserializing Generic Types" might already cover that.

UserGuide.md
```diff
@@ -140,6 +140,18 @@ Note that you can not serialize objects with circular references since that will
...
+// ==> map is: {value1=1.0, value2=abc}
+
+// Or you can just use Object.class, and trust gson to return an Object of the correct type
+Object mapObj = gson.fromJson(json, Object.class);
+// ==> mapObj is a Map
+Object numObj = gson.fromJson("1", Object.class);
+// ==> numObj is a Double
+Object listObj = gson.fromJson("[1, 3, {\"k\":\"v\"}]", Object.class);
+// ==> listObj is a List
```
Marcono1234, line 154: If the user wants a specific type (Map, Double or List) then it is a little bit weird to ask Gson to deserialize an Object. I think this might rather confuse the user.
Additionally it does not match the previous example code where `json` is the JSON representation of `BagOfPrimitives` and therefore **not** a Double or List.

Marcono1234 (commented):
Thanks for this pull request, but I am not sure if this actually makes things clearer. Maybe a separate section would be better.
Regarding the CLA bot, maybe it would work if you associated your commit e-mail address with your GitHub account (see [GitHub Docs](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-your-github-user-account/adding-an-email-address-to-your-github-account)), or if you [changed your e-mail address in your local Git config](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git)  and then force pushed the commit with the different commiter e-mail address.

----

UserGuide.md
josephks: The point is to be simpler than the exiting example code that uses TypeToken.  A beginning user doesn't need to be confused that.  If you are writing throwaway code or if you don't know or have an exact java class using a plain Map is fine.

UserGuide.md
josephks: My point is that you don't have to use a BagOfPrimitives.  If you just call `gson.fromJson("...", Object.class)` Gson gives you a List.

josephks:
@googlebot I consent.

UserGuide.md
Marcono1234: Though a beginner will then see warnings from the IDE and the compiler and might be confused that something shown in the official user guide causes warnings.

UserGuide.md
Marcono1234: Ah nevermind sorry, I overlooked that you are providing a custom JSON string for these cases.

Though I still think that asking Gson to deserialize an `Object` when you are actually interested in a specific type is a little bit weird and might give a bad impression.
If the user is dealing with JSON of unknown structure, then pointing them to `JsonElement` and related methods might be more appropriate.

UserGuide.md
josephks: IntelliJ gives a lot of spurious warnings.  Beginners will have to get used to ignoring them.

UserGuide.md
josephks: JsonElement and related objects are hard to figure out how to use.  If you just have one line of `Object obj = gson.fromJson(jsonStr, Object.class)` you can use standard java if statements with instanceof to handle it.

The motivation for this MR is that I do codepairs with job candidates in HackerRank.  The HackerRank java8 platform only includes a few libraries, one of which is Gson.  Most candidates have not used Gson before so they have to quickly look up some documentation how to use it.  Some try writing custom classes to parse into.  Some try using JsonElement.  The ones that try using JsonElement have a hard time because they don't know exactly how to use the Gson classes.  A lot of the time they end up extracting a string with quotes in it, and have to do `str.replaceAll("\"", "")` which would be unforgivable in real code.  I don't want them wasting time figuring out how to use Gson, I want them to be able to quickly figure out how to get minimal correct functionality out of it.  If they can do  `Object obj = gson.fromJson(jsonStr, Object.class)` that's all they need to know.

UserGuide.md
Marcono1234: I would not call warnings about raw types "spurious". Likely every IDE and `javac` itself warns about it, for a reason. You are basically giving up Java's type safety.

josephks: Yes when you use Object you give up type safety, but type safety can be an illusion.  What happens if you try to use TypeToken but you get the type wrong?

We are kind of straying from the point.  I want to make it clear to the reader what the simplest way to use Gson is.  If they are good Java developers that like type safety they can take the extra step of using TypeToken or doing something else to ensure type safety.

Issue #1812, opened 2020-11-05
ConcurrentModification exception on serialization of a mutable collection


MarcoSignoretto:
In the case where you want to serialize a Mutable collection and this collection get changed by another thread while Gson is serializing it you got the following error:

```
java.util.ArrayList$Itr.next (ArrayList.java:860)
com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write (CollectionTypeAdapterFactory.java:96)
com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write (CollectionTypeAdapterFactory.java:61)
com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write (TypeAdapterRuntimeTypeWrapper.java:69)
com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write (ReflectiveTypeAdapterFactory.java:127)
com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write (ReflectiveTypeAdapterFactory.java:245)
com.google.gson.Gson.toJson (Gson.java:704)
com.google.gson.Gson.toJson (Gson.java:683)
com.google.gson.Gson.toJson (Gson.java:638)
com.google.gson.Gson.toJson (Gson.java:618)
```

Starting from the fact that having a mutable collection shared between threads is bad, I think the library should handle it in the sense that is should take a snapshot of the collection and serialize that one.

Suggested solution: https://github.com/google/gson/pull/1813

MarcoSignoretto:
Duplicate of: https://github.com/google/gson/issues/1159

Issue #1159, opened 2017-09-19
ConcurrentModificationException in CollectionTypeAdapterFactory$Adapter


Loki-Afro:
Hi,
according to https://groups.google.com/forum/#!topic/google-gson/Vju1HuJJUIE gson is thread safe however I am facing the following exception from time to time with gson 2.8.0

```
 Exception in thread "Thread-17" java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:96)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:61)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:69)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write(ReflectiveTypeAdapterFactory.java:125)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write(ReflectiveTypeAdapterFactory.java:243)
	at com.google.gson.Gson.toJson(Gson.java:669)
	at com.google.gson.Gson.toJson(Gson.java:648)
	at com.google.gson.Gson.toJson(Gson.java:603)
```

JakeWharton:
The Gson object is threadsafe, yes, but serialization of user objects is not. You should either make a defensive copy or use a lock when an object needs serialized but can also be manipulated from other threads concurrently.

PR #1813, opened 2020-11-05
Avoid to crash with ConcurrentModificationException on serialization

MarcoSignoretto:
Avoid crashing with ConcurrentModificationException during json serialization when Gson is serializing a Mutable collection 

Fixes: https://github.com/google/gson/issues/1812

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/CollectionTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/CollectionTypeAdapterFactory.java
@@ -96,7 +96,7 @@ public final class CollectionTypeAdapterFactory implements TypeAdapterFactory {
 
       final List<E> collectionSnapshot;
       synchronized (collection){ // Lock collection to do not mutate during shallow copy
-        collectionSnapshot = new ArrayList<>(collection);
+        collectionSnapshot = new ArrayList<E>(collection);
       }
 
       out.beginArray();
--- a/gson/src/main/java/com/google/gson/internal/bind/CollectionTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/CollectionTypeAdapterFactory.java
@@ -28,7 +28,9 @@ import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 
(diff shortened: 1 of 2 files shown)


MarcoSignoretto:
@googlebot I signed it!

Marcono1234:
Could you please edit the description to say "Fixes #XYZ" instead of "Issue: #XYZ", otherwise GitHub does not understand that this pull request is related to the issue.

Most concurrent collection classes of the JDK are "[weakly consistent](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/package-summary.html#Weakly)", i.e. they never throw `ConcurrentModificationException`s. Additionally most, if not all, do not synchronize on `this` when performing modifications, so your `synchronized (collection) { ... }` would have no effect on them.

Do you have a specific class in mind where this is not the case?

Additionally always performing a shallow copy will decrease performance (and increase memory consumption) for applications which are not concurrently modifying collections.

----

MarcoSignoretto:
> Could you please edit the description to say "Fixes #XYZ" instead of "Issue: #XYZ", otherwise GitHub does not understand that this pull request is related to the issue.
> 
> Most concurrent collection classes of the JDK are "[weakly consistent](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/package-summary.html#Weakly)", i.e. they never throw `ConcurrentModificationException`s. Additionally most, if not all, do not synchronize on `this` when performing modifications, so your `synchronized (collection) { ... }` would have no effect on them.
> 
> Do you have a specific class in mind where this is not the case?
> 
> Additionally always performing a shallow copy will decrease performance (and increase memory consumption) for applications which are not concurrently modifying collections.
> 
> Note that I am not a maintainer of this project.

@Marcono1234 

On the PR there is a test that proves the issue with an `ArrayList` as a collection. In particular, what is causing the error is the iterator of the collection, not the collection itself.

The `synchronized(collection)` is used to perform a shallow copy of the collection atomically so no one will update the collection while you are copying and then use this copy on the iteration below where single item serialization is performed as well. I proposed to do `synchronized(collection)` on the shallow copy to minimize the performance overhead since the lock on the collection will be there only for the copy that will be performed by a `System.arrayCopy` under the hood so the performance impact should be low since you are copying references or primitive types but I agree that will add some overhead ( If you have a better proposal please share ).

MarcoSignoretto:
Issue reported: https://github.com/google/gson/issues/1159 where defensive copy should live into the client

PR #1814, opened 2020-11-05
Improve error message when abstract class cannot be constructed

Marcono1234:
Previously Gson tried to invoke the no-arg constructor of abstract classes. This lead to rather unhelpful exception messages, e.g.:
```
java.lang.RuntimeException: Failed to invoke public test.AbstractClass() with no args
    at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:119)
    ...
```

Now Gson will not try to invoke constructors of abstract classes anymore, which generates more helpful exception messages:
```
java.lang.RuntimeException: Unable to invoke no-args constructor for class test.AbstractClass. Registering an InstanceCreator with Gson for this type may fix this problem.
    at com.google.gson.internal.ConstructorConstructor$14.construct(ConstructorConstructor.java:235)
    ...
Caused by: java.lang.UnsupportedOperationException: Abstract class can't be instantiated! Class name: test.AbstractClass
    at com.google.gson.internal.UnsafeAllocator.assertInstantiable(UnsafeAllocator.java:120)
    ...
```


eamonnmcmanus:
Thanks, I actually ran into a user issue with this earlier this week.

PR #1815, opened 2020-11-06
Add tests checking GsonBuilder not affecting existing Gson instances

Marcono1234:
Edit: Copying the mutable collections of `GsonBuilder` has already been implemented by #1905, so this PR is only about adding some tests.

Below is the original description:

----

Previously when a GsonBuilder had created a Gson instance and was afterwards reused and different type adapters were added, new GsonBuilder instances obtained from the previous Gson instance through `Gson.newBuilder()` would have been affected by the GsonBuilder changes, e.g.:
```java
GsonBuilder gsonBuilder = new GsonBuilder();
Gson gson = gsonBuilder.create();
// Modify gsonBuilder
gsonBuilder.registerTypeAdapter(...);

// New Gson instance is erroneously affected by changes to GsonBuilder
Gson gson2 = gson.newBuilder().create();
```

This commit fixes this and additionally adds some more unit tests.


gson/src/test/java/com/google/gson/GsonBuilderTest.java
```diff
@@ -52,20 +143,6 @@ public void testExcludeFieldsWithModifiers() {
     assertEquals("{\"d\":\"d\"}", gson.toJson(new HasModifiers()));
   }
 
-  public void testRegisterTypeAdapterForCoreType() {
```
Marcono1234, line 55: Moved this down to not be between two test method for `GsonBuilder.excludeFieldsWithModifiers`.

eamonnmcmanus:
LGTM

PR #1820, opened 2020-11-16
Fix gradle build

ZacSweers:
Not sure if this is maintained, but this at least updates the gradle version and fixes the project version injection


gson/build.gradle
```diff
@@ -7,7 +7,29 @@ version = '2.8.6-SNAPSHOT'
 sourceCompatibility = 1.6
 targetCompatibility = 1.6
 
+def copyVersionTemplatesProvider = tasks.register("copyVersionTemplates", Copy) {
+    it.inputs.property("gsonversion", version)
+    it.from(layout.projectDirectory.dir("src/main/java-gradle-templates/com/google/gson/internal"))
+    it.into(project.layout.buildDirectory.dir("generated/sources/version-templates/java/main/com/google/gson/internal"))
+    it.expand(["gsonversion": version])
```
Marcono1234, line 14: Would it be possible to change this to:
```suggestion
    it.expand(["{project.version}": version])
```
and then re-use the existing Maven template class? This is a little bit hacky, but might be less error-prone than having two separate source files for this.

Alternatively could you please modify https://github.com/google/gson/blob/master/gson/src/test/java/com/google/gson/internal/GsonBuildConfigTest.java to verify that the version is not `$gsonversion` either (or in general does not start with `$`; that would probably be most reliable)?

---

Note that I am not a member of this project.

ZacSweers: No templates in gradle copies have a different template mechanism. I tried some basic escaping of the dot but without much luck

Marcono1234: Do you think it would make sense to use [`Copy.filter(Closure)`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:filter(groovy.lang.Closure)) instead? And then manually replace `${project.version}` in the lines?

And regardless of whether you decide to do this, could you please adjust the `GsonBuildConfigTest.java` to check for a leading `$` (instead of `"${project.version}"`) so the test becomes more robust?

PR #1825, opened 2020-11-24
Add architecture ppc64le to travis build

ddeka2910:
Add support for architecture ppc64le.  
This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing. For more info tag @gerrith3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
+os: linux
+arch:
+ - amd64
+ - ppc64le
 language: java
 
 jdk:
   - openjdk11
 
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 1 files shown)


ddeka2910:
@googlebot I signed it!

Marcono1234:
Probably obsolete because Gson does not use Travis anymore, but instead uses GitHub Actions (#1974). And it looks like GitHub Actions does not support PowerPC at the moment, see [documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

PR #1826, opened 2020-11-25
fix bug,string json to map,int automatic conversion double

lsm1998:
On the basis of the last submission, a new processing method nextnumber() is defined, and backward compatibility is retained.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
@@ -76,7 +76,7 @@ public final class ObjectTypeAdapter extends TypeAdapter<Object> {
       return in.nextString();
 
     case NUMBER:
-      return in.nextDouble();
+      return in.nextNumber();
 
     case BOOLEAN:
       return in.nextBoolean();
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -893,6 +893,49 @@ public class JsonReader implements Closeable {
       return (double) peekedLong;
     }
 
+    return nextDouble(p);
+  }
+
+  /**
+   * Returns the {@link com.google.gson.stream.JsonToken#NUMBER double} value of the next token,
+   * consuming it. If the next token is a string, this method will attempt to
+   * parse it as a Number using {@link Double#parseDouble(String)}.
+   *
+   * @throws IllegalStateException if the next token is not a literal value.
+   * @throws NumberFormatException if the next literal value cannot be parsed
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -893,6 +893,49 @@ public double nextDouble() throws IOException {
       return (double) peekedLong;
     }
 
+    return nextDouble(p);
+  }
+
+  /**
+   * Returns the {@link com.google.gson.stream.JsonToken#NUMBER double} value of the next token,
```
Marcono1234, line 900:
```suggestion
   * Returns the {@link com.google.gson.stream.JsonToken#NUMBER number} value of the next token,
```

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -893,6 +893,49 @@ public double nextDouble() throws IOException {
...
+   * consuming it. If the next token is a string, this method will attempt to
+   * parse it as a Number using {@link Double#parseDouble(String)}.
+   *
+   * @throws IllegalStateException if the next token is not a literal value.
+   * @throws NumberFormatException if the next literal value cannot be parsed
+   *     as a double, or is non-finite.
```
Marcono1234, line 906:
```suggestion
   *     as a number, or is non-finite double.
```

gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java
```diff
@@ -76,7 +76,7 @@
       return in.nextString();
 
     case NUMBER:
-      return in.nextDouble();
+      return in.nextNumber();
```
Marcono1234, line 79: This might be backward incompatible in case existing code expects a `Double` instead of any `Number`.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -908,7 +951,7 @@ public double nextDouble() throws IOException {
     double result = Double.parseDouble(peekedString); // don't catch this NumberFormatException.
     if (!lenient && (Double.isNaN(result) || Double.isInfinite(result))) {
       throw new MalformedJsonException(
-          "JSON forbids NaN and infinities: " + result + locationString());
+              "JSON forbids NaN and infinities: " + result + locationString());
```
Marcono1234, line 954: Could you please revert this indentation change?

Marcono1234 (commented):
Thanks! Could you please add some tests for `nextNumber()` to make sure it returns the correct `Number` subclasses?

----

mistletoe5215:
@JakeWharton how long can this bug be resloved?😄

Issue #1829, opened 2020-12-01
fromJson intermittently fails with IllegalStateException when deserializing recursive structures in parallel


mwalkerr:
We are seeing an issue where `fromJson` results in a `JsonSyntaxException` with the root cause being an `IllegalStateException`. The stacktrace that comes from within the GSON library is:
```
Caused by: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:226)
at com.google.gson.Gson.fromJson(Gson.java:927)
at com.google.gson.Gson.fromJson(Gson.java:892)
at com.google.gson.Gson.fromJson(Gson.java:841)
at com.google.gson.Gson.fromJson(Gson.java:813)
at com.stripe.net.LiveStripeResponseGetter.request(LiveStripeResponseGetter.java:66)
... 136 common frames omitted
Caused by: java.lang.IllegalStateException: null
at com.google.gson.Gson$FutureTypeAdapter.read(Gson.java:1009)
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
... 141 common frames omitted
```
This seems to be due to a race condition in `Gson.getAdapter` where a `FutureTypeAdapter` is created by one thread and accessed by another before the delegate is set
https://github.com/google/gson/blob/5370b076e329444860b0ea0c7f33b10504281f27/gson/src/main/java/com/google/gson/Gson.java#L433

edit: This looks to be well described by the comment on #625 : https://github.com/google/gson/issues/625#issuecomment-321373437

Marcono1234:
This appears to be the same issue as #625

mwalkerr:
> This appears to be the same issue as #625

@Marcono1234 I believe you're right, thank you for linking that. Wish I would have found that sooner!

Marcono1234:
This should be fixed by #1832, which will be available in the next release

PR #1830, opened 2020-12-01
Avoid returning incomplete TypeAdapters in concurrent calls

mwalkerr:
Resolves google/gson#1829
Resolves google/gson#625 partially

In the current implementation, a `TypeAdapter` may be added to the `typeTokenCache` despite having a `FutureTypeAdapter` with a null delegate as part of its definition. This would occasionally cause us to see an `IllegalStateException` from GSON. This would happen when bringing up an application in a docker container and then running parallelized API tests against it. Some of the objects the application deserializes are recursively defined such as `A` -> `B` -> `A`. I believe what was happening was that one thread would start to deserialize `A` and call to `getAdapter(A)` which would call to `getAdapter(B)` which would call `getAdapter(A)`, which would return the `FutureTypeAdapter` for `A`. Before the first thread is able to update the delegate for the `FutureTypeAdapter` for `A`, another thread tries to deserialize `B`, calls `getAdapter(B)`, and retrieves the entry added by the first thread which contains a `TypeAdapter` for `A` which is a `FutureTypeAdapter` with a null delegate

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -121,8 +121,8 @@ public final class Gson {
    * lookup would stack overflow. We cheat by returning a proxy type adapter.
    * The proxy is wired up once the initial adapter has been created.
    */
-  private final ThreadLocal<Map<TypeToken<?>, FutureTypeAdapter<?>>> calls
-      = new ThreadLocal<Map<TypeToken<?>, FutureTypeAdapter<?>>>();
+  private final ThreadLocal<Map<TypeToken<?>, TypeAdapter<?>>> calls
+      = new ThreadLocal<Map<TypeToken<?>, TypeAdapter<?>>>();
 
   private final Map<TypeToken<?>, TypeAdapter<?>> typeTokenCache = new ConcurrentHashMap<TypeToken<?>, TypeAdapter<?>>();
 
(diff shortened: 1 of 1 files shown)


mwalkerr:
@googlebot

mwalkerr:
@googlebot I signed it!

mwalkerr:
I used the following test to verify the issue and the solution:

```
package com.google.gson.regression;

import com.google.gson.Gson;
import junit.framework.TestCase;

import java.lang.reflect.Field;
import java.util.Map;
import java.util.concurrent.CyclicBarrier;

public class RecursiveTypeRaceConditionTest extends TestCase {
  private final Gson gson = new Gson();

  public void testNullSafeBugSerialize() throws Exception {
    for (int i=0;i<100000;i++){
      Field f = gson.getClass().getDeclaredField("typeTokenCache");
      f.setAccessible(true);
      ((Map) f.get(gson)).clear();
      final CyclicBarrier gate = new CyclicBarrier(3);
      final boolean[] exception = {false};
      Thread t1 = new Thread(new Runnable() {
        @Override
        public void run() {
          try {
            gate.await();
            gson.getAdapter(TypeToken.get(A.class)).fromJson("{\"b\": {\"a\":{}}}");
          } catch (Exception e) {
            exception[0] = true;
            e.printStackTrace();
          }
        }
      });
      Thread t2 = new Thread(new Runnable() {
        @Override
        public void run() {
          try {
            gate.await();
            gson.getAdapter(TypeToken.get(B.class)).fromJson("{\"a\": {\"b\":{}}}");
          } catch (Exception e) {
            exception[0] = true;
            e.printStackTrace();
          }
        }
      });
      t1.start();
      t2.start();
      gate.await();
      t1.join();
      t2.join();
      assertFalse(exception[0]);
    }
  }

  private static final class A {
    B b;
  }
  private static final class B {
    A a;
  }
}
```

Marcono1234:
Thanks a lot for you work on this! This has now be fixed by #1832, but thank you nonetheless!

PR #1838, opened 2020-12-11
Update UserGuide.md

Fiordy:
fixed typo in generic


Marcono1234 (changes requested):
The user guide is actually correct, the signature is [`InstanceCreator<T>`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/InstanceCreator.html).

The type argument here is `A.B`, i.e. the inner class `B` of enclosing class `A`.

eamonnmcmanus:
As noted by @Marcono1234 the existing text is correct.

Issue #1821, opened 2020-11-18
[Question] : is gson still maintained?


radityagumay:
Rationale behind this is, gson looks like not actively maintained. The [last PR](https://github.com/google/gson/commit/ceae88bd6667f4263bbe02e6b3710b8a683906a2) was on may. and looks like there's no roadmap on gson to provide [new features and fix bugs.](https://github.com/google/gson/pulls)

Will appreciated if the maintainer able to put some thoughts on this. 

Thank you!

Marcono1234:
I am not a maintainer / member of this project and had the same question the last months as well and always thought about creating an issue too because activity here has sadly slowed down.

There are several critical issues (with corresponding pull requests) which have not been touched for a rather long time (subjective selection):
- Issues with generic parameter being used multiple times: #1391
- ProGuard issues with enums: #1495
- `ISO8601UtilsTest` being faulty (#1665, #1687, #1781)
- Optional `java.sql` module dependency: #1707
- Problem with SecurityManager: #1712

It appears @joel-leitch, one of the original maintainers, is not very active on GitHub anymore. @swankjesse and @JakeWharton are not active here either. I am not completely familiar with their biography, but if I understand it correctly they (or one of them) quit their job at Google and are now working at Square where they are working on https://github.com/square/moshi, a competitor of Gson. Maybe they do not even have the GitHub project roles and permissions anymore to directly work on Gson.
@inder123 is the one who made the most recent changes, but it looks like they are not that active at the moment either.

It is not my goal to blame them for the lack of activity, they put great efforts in this project in the past and are the reason for the popularity of it. It is completely understandable that after more than 10 years, the motivation of working on the same project decreases. Especially when (as it is likely the case for all projects) design flaws or decisions hindering future development become apparent over time. And as with all open source projects, enough free time might also not always be available.

However, because Gson is still widely used it is important that a clear decision is made about how Gsons future development (if any) will look like. Otherwise users will be forced to switch to other frameworks when bugs remain unfixed. While switching to other frameworks is not necessarily bad, being forced to do this creates extra stress and work for the users; it should ideally happen voluntarily.
Would it at least be possible to keep developing Gson in _maintenance mode_: Only bug fixes are implemented, but no major new features are added?

xileftenurb:
I found this anwser on reddit written by @swankjesse

https://www.reddit.com/r/androiddev/comments/684flw/why_use_moshi_over_gson/dgx3gpm?utm_source=share&utm_medium=web2x&context=3

[...]

PR #1844, opened 2021-01-05
Kotlin support for index access operators

Bixilon:
In Kotlin you can define your own operators. You can use the ::set Method to add a object. See https://kotlinlang.org/docs/reference/keyword-reference.html for more.

I added the set method in JsonObject. It is a simple alias to add or addProperty.

Also I need the current String position in JsonReader. E.g I have a command as a string and the next argument is a JsonObject. I just want to know how many chars were read from the input string to determinate the next argument. Currently this looks like this:
```java

    private static final Field JSON_READER_POS_FIELD;
    private static final Field JSON_READER_LINE_START_FIELD;

        new JsonReader(new StringReader(""));
        Class<?> jsonReadClass = JsonReader.class;
        try {
            JSON_READER_POS_FIELD = jsonReadClass.getDeclaredField("pos");
            JSON_READER_POS_FIELD.setAccessible(true);
            JSON_READER_LINE_START_FIELD = jsonReadClass.getDeclaredField("lineStart");
            JSON_READER_LINE_START_FIELD.setAccessible(true);
        } catch (NoSuchFieldException e) {
            e.printStackTrace();
            throw new RuntimeException(e);
        }

        try {
            return JSON_READER_POS_FIELD.getInt(jsonReader) - JSON_READER_LINE_START_FIELD.getInt(jsonReader) + 1;
        } catch (IllegalAccessException e) {
            throw new IllegalStateException();
        }
```

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -67,7 +67,7 @@ public final class JsonObject extends JsonElement {
    * @param value the member object.
    */
   public void set(String property, JsonElement value) {
-    set(property, value);
+    add(property, value);
   }
 
   /**
@@ -130,7 +130,7 @@ public final class JsonObject extends JsonElement {
    * JsonPrimitive of Boolean.
    *
    * @param property name of the member.
-   * @param value the number value associated with the member.
+   * @param value the boolean value associated with the member.
    */
   public void addProperty(String property, Boolean value) {
     add(property, value == null ? JsonNull.INSTANCE : new JsonPrimitive(value));
@@ -141,7 +141,7 @@ public final class JsonObject extends JsonElement {
    * JsonPrimitive of Boolean.
    *
    * @param property name of the member.
-   * @param value the number value associated with the member.
+   * @param value the boolean value associated with the member.
    */
   public void set(String property, Boolean value) {
     addProperty(property, value);
@@ -152,7 +152,7 @@ public final class JsonObject extends JsonElement {
    * JsonPrimitive of Character.
    *
    * @param property name of the member.
-   * @param value the number value associated with the member.
+   * @param value the char value associated with the member.
    */
   public void addProperty(String property, Character value) {
     add(property, value == null ? JsonNull.INSTANCE : new JsonPrimitive(value));
@@ -163,7 +163,7 @@ public final class JsonObject extends JsonElement {
    * JsonPrimitive of Character.
    *
    * @param property name of the member.
-   * @param value the number value associated with the member.
+   * @param value the char value associated with the member.
    */
   public void set(String property, Character value) {
     addProperty(property, value);
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -59,6 +59,18 @@ public final class JsonObject extends JsonElement {
   }
 
   /**
+   * Adds a member, which is a name-value pair, to self. The name must be a String, but the value
+   * can be an arbitrary JsonElement, thereby allowing you to build a full tree of JsonElements
+   * rooted at this node.
+   *
+   * @param property name of the member.
+   * @param value the member object.
+   */
+  public void set(String property, JsonElement value) {
+    set(property, value);
+  }
+
+  /**
    * Removes the {@code property} from this {@link JsonObject}.
    *
    * @param property name of the member that should be removed.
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -58,6 +58,18 @@ public void add(String property, JsonElement value) {
...
+  /**
+   * Adds a member, which is a name-value pair, to self. The name must be a String, but the value
+   * can be an arbitrary JsonElement, thereby allowing you to build a full tree of JsonElements
+   * rooted at this node.
+   *
+   * @param property name of the member.
+   * @param value the member object.
+   */
+  public void set(String property, JsonElement value) {
+    set(property, value);
```
Marcono1234, line 70: This causes infinite recursion because the method is calling itself?

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -102,6 +136,17 @@ public void addProperty(String property, Boolean value) {
     add(property, value == null ? JsonNull.INSTANCE : new JsonPrimitive(value));
   }
 
+  /**
+   * Convenience method to add a boolean member. The specified value is converted to a
+   * JsonPrimitive of Boolean.
+   *
+   * @param property name of the member.
+   * @param value the number value associated with the member.
```
Marcono1234, line 144:
```suggestion
   * @param value the boolean value associated with the member.
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -113,6 +158,17 @@ public void addProperty(String property, Character value) {
     add(property, value == null ? JsonNull.INSTANCE : new JsonPrimitive(value));
   }
 
+  /**
+   * Convenience method to add a char member. The specified value is converted to a
+   * JsonPrimitive of Character.
+   *
+   * @param property name of the member.
+   * @param value the number value associated with the member.
```
Marcono1234, line 166:
```suggestion
   * @param value the char value associated with the member.
```

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1485,6 +1485,16 @@ public String getPath() {
     return result.toString();
   }
 
+  /**
+   * Return she current position of the input string
+   *
+   * @return Position of the json reader
```
Marcono1234, line 1491:
```suggestion
   * Return the current position of the JSON reader.
   *
   * @return Position of the JSON reader
```

Though this documentation should be improved because it appears the return value is actually the number of chars from the beginning of the current line (if I see that correctly?).

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1485,6 +1485,16 @@ public String getPath() {
...
+   */
+
```
Marcono1234, line 1493: Remove this empty line please.

Marcono1234 (commented):
Personally I am not sure whether adding more methods with different names but the same behavior as existing methods is really worth it, especially since the main target of this project seems to be Java. Have a look at https://github.com/square/moshi which is designed for Kotlin.

I have added some review comments, but it might not be worth it considering them because this project does not seem to be actively maintained anymore, see #1821.

----

gson/src/main/java/com/google/gson/JsonObject.java
Bixilon: Haha, just copied the original comment.

Bixilon:
I don't know if it is still maintained, but still making this PR. Everything is resolved now.

pak9813 (approved):

pak9813 (approved):

eamonnmcmanus:
Reviving this conversation. 🙂 
This project _is_ now maintained. However I agree with @Marcono1234 that syntactic convenience in Kotlin isn't a good enough reason to add a second name to existing methods. You can easily use Kotlin extension functions to achieve the same thing in your project:

```kt
operator fun JsonObject.set(property: String, value: JsonElement) = add(property, value)
operator fun JsonObject.set(property: String, value: String) = addProperty(property, value)
operator fun JsonObject.set(property: String, value: Number) = addProperty(property, value)
operator fun JsonObject.set(property: String, value: Boolean) = addProperty(property, value)
operator fun JsonObject.set(property: String, value: Char) = addProperty(property, value)
```

It would be good if there were a standard artifact to house extension functions like these. I don't think the main Gson artifact (`com.google.code.gson:gson`) would be the right place, though, because it would probably have to acquire a dependency on the Kotlin runtime.

Issue #542, opened 2015-03-19
object graph recursion with polymorphism using GraphAdapterBuilder and RuntimeTypeAdapterFactory


GoogleCodeExporter:
```
When used in conjunction, the classes GraphAdapterBuilder and 
RuntimeTypeAdapterFactory produce an IllegalStateException "Unexpected 
recursive call to read()" on de-serialization.

When used separately, GraphAdapterBuilder can serialize and de-serialize 
circular references, and RuntimeTypeAdapterFactory can tackle classes involving 
polymorphism. When used together, serialization succeeds but de-serialization 
fails.

What steps will reproduce the problem?
Add the attached source files to a project, and execute the JUnit tests.

What is the expected output? What do you see instead?
All tests (testRecursive, testPolymorphic and testPolymorphicRecursive) in the 
attached source should pass. Instead, testPolymorphicRecursive fails with an 
exception.

What version of the product are you using? On what operating system?
Gson 2.2.4
RuntimeTypeAdapterFactory from 
https://code.google.com/p/google-gson/source/browse/trunk/extras/src/main/java/c
om/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
GraphAdapterBuilder from 
https://code.google.com/p/google-gson/source/browse/trunk/extras/src/main/java/c
om/google/gson/graph/GraphAdapterBuilder.java with the changes made as 
suggested in https://code.google.com/p/google-gson/issues/detail?id=137#c12

```

Original issue reported on code.google.com by `huny...@gmail.com` on 17 Nov 2013 at 2:38

Attachments:
- [gson-polymorphic-recursive.zip](https://storage.googleapis.com/google-code-attachments/google-gson/issue-542/comment-0/gson-polymorphic-recursive.zip)

sertoltof:
[PolyRecursiveTest.txt](https://github.com/google/gson/files/1475956/PolyRecursiveTest.txt)

I found that the broken test can work if you call ".addType(SubRoshambo.class)" on the GraphAdapterBuilder before registering to the gsonBuilder. This solves the issue because the constructor for SubRoshambo was defaulting to a default constructor rather than the "createInstance" method of the GraphAdapterBuilder. 

However, the gson doesn't **serialize** properly when you add the SubRoshambo class type. For some reason, it is using the GraphAdapterBuilder write method for SubRoshambo rather than the RunTypeAdapterFactory write method which will add the "type" field to polymorphic objects. 

[...]

PR #1851, opened 2021-01-20
fix error when using GraphAdapterBuilder with RuntimeTypeAdapterFactory

Leo1998:
I made a proper fix for Issue #542 where the object graph fails to deserialize when a RuntimeTypeAdapterFactory is registered.
The problem is that if a custom TypeAdapter is registered for a Type the custom InstanceCreator in GraphAdapterBuilder:215 is not called and thus graph.nextCreate will not have a value set and the nextCreate field will not be cleared. If Element.read is called after that the it throws an Exception "Unexpected recursive call to read() for %". My addition to read just checks if the deserialized result wasn't populated yet (which only happens if T has a custom TypeAdapter registered) and makes sure that 'value' is set and nextCreate is set.

The Test attached in #542 still perfectly show the issue.


Leo1998:
@googlebot I signed it!

PR #1852, opened 2021-01-21
Added deserialization null handling

zcxshare:
Added deserialization null handling


zcxshare:
I feel the need to add such a function

LtLei:
@zcxshare hello, could you try again? It seems you didn't sign the CLA, and didn't reply here with @googlebot I signed it! I feel the need to add such a function, too!

PR #1862, opened 2021-02-17
Remove not needed .gitattributes file

Marcono1234:
The `.gitattributes` file was added to prevent GitHub detecting Gson as HTML repository due to the included generated javadoc files. However, since #1654 has removed the javadoc files, the `.gitattributes` file is no longer needed.


Marcono1234:

PR #1873, opened 2021-03-10
Documentation for collections

hart-james:
Added a brief segment of documentation to the User Guide pertaining to parsing a json response into a usable collection based off of an object tailored to the response.

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -15,6 +15,7 @@
    * [Collections Examples](#TOC-Collections-Examples)
      * [Collections Limitations](#TOC-Collections-Limitations)
    * [Serializing and Deserializing Generic Types](#TOC-Serializing-and-Deserializing-Generic-Types)
+   * [Parsing a Collection with Objects of Non Arbitrary Types](#TOC-Parsing-a-Collection-with-Objects-of-Non-Arbitrary-Types)
    * [Serializing and Deserializing Collection with Objects of Arbitrary Types](#TOC-Serializing-and-Deserializing-Collection-with-Objects-of-Arbitrary-Types)
    * [Built-in Serializers and Deserializers](#TOC-Built-in-Serializers-and-Deserializers)
    * [Custom Serialization and Deserialization](#TOC-Custom-Serialization-and-Deserialization)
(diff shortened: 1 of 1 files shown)


UserGuide.md
```diff
@@ -260,6 +261,36 @@ gson.fromJson(json, fooType);
 ```
 The idiom used to get `fooType` actually defines an anonymous local inner class containing a method `getType()` that returns the fully parameterized type.
 
+### <a name="TOC-Parsing-a-Collection-with-Objects-of-Non-Arbitrary-Types"></a>Parsing a Collection with Objects of Non Arbitrary Types
```
Marcono1234, line 264: Maybe instead of "non arbitrary types" -> "specific types"?
However, this section only covers arrays, so maybe it would be better to add it to the existing section "Array Examples".

UserGuide.md
```diff
@@ -260,6 +261,36 @@ gson.fromJson(json, fooType);
...
+
+At times you will try to parse a collection of a specific object. The `getJson` and `fromJson`
```
Marcono1234, line 266: There is no `getJson` method (neither on `Gson` nor on any other class), which method do you mean?

UserGuide.md
```diff
@@ -260,6 +261,36 @@ gson.fromJson(json, fooType);
...
+  {
+    "name": "Steven",
+    "age": 30
+  }
+]  //Example of json response to parse
+```
+
+`fromJson` provides the ability to add a third party class as an argument for serialization as so:
+```java
+public class Person() {  //example class
```
Marcono1234, line 284:
```suggestion
public class Person {  //example class
```

Marcono1234 (commented):
Thanks! I have added some review comments, I hope they are helpful.
Though please note that I am not a member of this project.

UserGuide.md
```diff
@@ -260,6 +261,36 @@ gson.fromJson(json, fooType);
...
+    String name;
+    int age;
+}
+
+Gson gson = new Gson();
+Person[] people = gson.fromJson(json, Person[].class);
```
eamonnmcmanus, line 290: I think it would be better if this deserialized into a `List<Person>` rather than a `Person[]`. For example as explained [here](https://www.baeldung.com/gson-list). Object arrays are generally better avoided and I don't think there's a strong reason to use one here.

PR #1878, opened 2021-03-31
replaced deprecated constructor call with static factory call

nilswende:


Marcono1234 (approved):

Marcono1234:
Looks like this was already fixed by #1953.

eamonnmcmanus:
Indeed, this is already fixed.

PR #1882, opened 2021-04-11
Changed the name of the method "equal"

RosieKatze:


eamonnmcmanus:
I don't think there's a compelling reason for this change and the CLA has not been signed.

Issue #1409, opened 2018-10-17
Should MapTypeAdapterFactory.Adapter throw an exception while reading objects with duplicate keys?


lyubomyr-shaydariv:
Today I came across an interesting [question](https://stackoverflow.com/questions/52855391/javajson-to-map-using-gson) on StackOverflow. The author asks why the following JSON document cannot be deserialized:

```json
{
	"K1": "V1",
	"K2": "V2",
	"K1": "V3"
}
```

If deserializing the above document with `gson.fromJson(..., Map.class)`, the following exception is thrown:

```
JsonSyntaxException: duplicate key: key0
```

However, if the document is deserialized with `gson.fromJson(..., Object.class)`, as found by another user, the deserialized map is as follows:

```
{K1=V3, K2=V2}
```

This is obviously an inconsistent behavior, but I find the "`Object`" implementation more useful. As far as I know, JSON [does not prohibit](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object) use of duplicate keys simply describing its grammar.

Should `ObjectTypeAdapter` throw an exception on duplicate keys, or should `MapTypeAdapterFactory.Adapter` not check duplicate keys (however, the check was added by a clear intention in a98d6eae472df67317d58d34f13438bb958668b1)?

----

Related: #386, #649

lyubomyr-shaydariv:
Closed as a duplicate of #649.

PR #1884, opened 2021-04-21
allow deserialize duplicate map key

saddays:
in my scenario， the json allow duplicate keys . We can just add a switch to enable the feature。  Pleas review it, thx.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -34,6 +34,7 @@ import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 
 import static com.google.gson.Gson.DEFAULT_COMPLEX_MAP_KEYS;
+import static com.google.gson.Gson.DEFAULT_DUPLICATE_MAP_KEYS;
 import static com.google.gson.Gson.DEFAULT_ESCAPE_HTML;
 import static com.google.gson.Gson.DEFAULT_JSON_NON_EXECUTABLE;
 import static com.google.gson.Gson.DEFAULT_LENIENT;
@@ -89,6 +90,7 @@ public final class GsonBuilder {
   private int dateStyle = DateFormat.DEFAULT;
   private int timeStyle = DateFormat.DEFAULT;
   private boolean complexMapKeySerialization = DEFAULT_COMPLEX_MAP_KEYS;
+  private boolean duplicateMapKeyDeserialization = DEFAULT_DUPLICATE_MAP_KEYS;
   private boolean serializeSpecialFloatingPointValues = DEFAULT_SPECIALIZE_FLOAT_VALUES;
   private boolean escapeHtmlChars = DEFAULT_ESCAPE_HTML;
   private boolean prettyPrinting = DEFAULT_PRETTY_PRINT;
@@ -116,6 +118,7 @@ public final class GsonBuilder {
     this.instanceCreators.putAll(gson.instanceCreators);
     this.serializeNulls = gson.serializeNulls;
     this.complexMapKeySerialization = gson.complexMapKeySerialization;
+    this.duplicateMapKeyDeserialization = gson.duplicateMapKeyDeserialization;
     this.generateNonExecutableJson = gson.generateNonExecutableJson;
     this.escapeHtmlChars = gson.htmlSafe;
     this.prettyPrinting = gson.prettyPrinting;
@@ -272,6 +275,18 @@ public final class GsonBuilder {
    */
   public GsonBuilder enableComplexMapKeySerialization() {
     complexMapKeySerialization = true;
+    return this;
+  }
+
+  /**
+   * Configure Gson to deserialize duplicate map key. By default, Gson throw a {@code JsonSyntaxException} when there are more than
+   * one same key.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.8
+   */
+  public GsonBuilder enableDuplicateMapKeyDeserialization() {
+    duplicateMapKeyDeserialization = true;
     return this;
   }
 
@@ -595,7 +610,7 @@ public final class GsonBuilder {
     addTypeAdaptersForDate(datePattern, dateStyle, timeStyle, factories);
 
     return new Gson(excluder, fieldNamingPolicy, instanceCreators,
-        serializeNulls, complexMapKeySerialization,
+        serializeNulls, complexMapKeySerialization, duplicateMapKeyDeserialization,
         generateNonExecutableJson, escapeHtmlChars, prettyPrinting, lenient,
         serializeSpecialFloatingPointValues, longSerializationPolicy,
         datePattern, dateStyle, timeStyle,
--- a/gson/src/test/java/com/google/gson/GsonTest.java
+++ b/gson/src/test/java/com/google/gson/GsonTest.java
@@ -46,7 +46,7 @@ public final class GsonTest extends TestCase {
 
   public void testOverridesDefaultExcluder() {
     Gson gson = new Gson(CUSTOM_EXCLUDER, CUSTOM_FIELD_NAMING_STRATEGY,
-        new HashMap<Type, InstanceCreator<?>>(), true, false, true, false,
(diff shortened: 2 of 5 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -275,6 +278,18 @@ public GsonBuilder enableComplexMapKeySerialization() {
     return this;
   }
 
+  /**
+   * Configure Gson to deserialize duplicate map key. By default, Gson throw a {@code JsonSyntaxException} when there are more than
+   * one same key.
```
Marcono1234, line 283: Maybe describe in more detail how duplicate keys will be treated:
```suggestion
   * Configures Gson to deserialize duplicate map keys. Only the value of last entry with the same key will be used, previous values
    * will be discarded. By default, Gson throws a {@link JsonSyntaxException} when a key occurs more than once.
```

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -275,6 +278,18 @@ public GsonBuilder enableComplexMapKeySerialization() {
...
+   *
```
Marcono1234, line 284: Might be good to add a security note:
```suggestion
   *
   * <p>Note that enabling support for duplicate maps keys is discouraged because it can make an application less secure.
   * When an application interacts with other components using different JSON libraries, they might treat duplicate keys
   * differently, allowing an attacker to circumvent security checks.
   *
```

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -275,6 +278,18 @@ public GsonBuilder enableComplexMapKeySerialization() {
...
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.8
+   */
+  public GsonBuilder enableDuplicateMapKeyDeserialization() {
```
Marcono1234, line 288: Maybe it would be good to rename it from "...Key" to "...Key<ins>s</ins>" since it applies to key**s** in general.
```suggestion
  public GsonBuilder enableDuplicateMapKeysDeserialization() {
```
What do you think?

gson/src/test/java/com/google/gson/GsonTest.java
```diff
@@ -46,7 +46,7 @@
 
   public void testOverridesDefaultExcluder() {
     Gson gson = new Gson(CUSTOM_EXCLUDER, CUSTOM_FIELD_NAMING_STRATEGY,
-        new HashMap<Type, InstanceCreator<?>>(), true, false, true, false,
+        new HashMap<Type, InstanceCreator<?>>(), true, false,false, true, false,
```
Marcono1234, line 49:
```suggestion
        new HashMap<Type, InstanceCreator<?>>(), true, false, false, true, false,
```

gson/src/test/java/com/google/gson/GsonTest.java
```diff
@@ -59,7 +59,7 @@ public void testOverridesDefaultExcluder() {
 
   public void testClonedTypeAdapterFactoryListsAreIndependent() {
     Gson original = new Gson(CUSTOM_EXCLUDER, CUSTOM_FIELD_NAMING_STRATEGY,
-        new HashMap<Type, InstanceCreator<?>>(), true, false, true, false,
+        new HashMap<Type, InstanceCreator<?>>(), true, false,false, true, false,
```
Marcono1234, line 62:
```suggestion
        new HashMap<Type, InstanceCreator<?>>(), true, false, false, true, false,
```

Marcono1234 (commented):
Supersedes #649
Relates to #1409

Personally I think this is a reasonable change which does not break security of existing applications relying on duplicate detection, but supports it for applications which explicitly enable this feature; however I am not a member of this project.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -279,13 +279,17 @@ public GsonBuilder enableComplexMapKeySerialization() {
   }
 
   /**
-   * Configure Gson to deserialize duplicate map key. By default, Gson throw a {@code JsonSyntaxException} when there are more than
-   * one same key.
+   * Configures Gson to deserialize duplicate map keys. Only the value of last entry with the same key will be used, previous values
+    * will be discarded. By default, Gson throws a {@link JsonSyntaxException} when a key occurs more than once.
```
Marcono1234, line 283: Sorry, looks like I got the indentation wrong.
```suggestion
   * will be discarded. By default, Gson throws a {@link JsonSyntaxException} when a key occurs more than once.
```

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -275,6 +278,22 @@ public GsonBuilder enableComplexMapKeySerialization() {
...
+   *
+   * <p>Note that enabling support for duplicate maps keys is discouraged because it can make an application less secure.
+   * When an application interacts with other components using different JSON libraries, they might treat duplicate keys
+   * differently, allowing an attacker to circumvent security checks.
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   * @since 2.8
+   */
+  public GsonBuilder enableDuplicateMapKeysDeserialization() {
```
Marcono1234, line 292: Sorry, I missed that `GsonBuilder` already has `enableComplexMapKeySerialization` ("...Key<del>s</del>"). So maybe for consistency it would be good to use "...Key<del>s</del>" here as well.
```suggestion
  public GsonBuilder enableDuplicateMapKeyDeserialization() {
```

Marcono1234 (commented):
Thanks for considering the review!

Maybe it would be good to keep the original "enableDuplicateMapKeyDeserialization" name ("Key" instead of "Keys") you proposed (see review comment).
Also note that CI is apparently not running, so the method renaming would have caused a compilation error anyways.

saddays:
> Thanks for considering the review!
> 
> Maybe it would be good to keep the original "enableDuplicateMapKeyDeserialization" name ("Key" instead of "Keys") you proposed (see review comment).
> Also note that CI is apparently not running, so the method renaming would have caused a compilation error anyways.

You are really dedicated to this project  ,respect!

saddays:
@eamonnmcmanus   could you review this pr?

eamonnmcmanus:
Gson is in maintenance mode and we are reluctant to accept new features. More features means more maintenance. I think there are already ways to handle maps with duplicate keys, for example as discussed in [this StackOverflow thread](https://stackoverflow.com/questions/47876328/ignore-duplicate-keys-in-jsonobject). Is there a reason those aren't sufficient?

Marcono1234:
> as discussed in [this StackOverflow thread](https://stackoverflow.com/questions/47876328/ignore-duplicate-keys-in-jsonobject)

That question appears to be for org.json (based on the exception), with the answer with the most votes being for Jackson. A [question for Gson](https://stackoverflow.com/q/37202584) only has a rather complex answer. Other workarounds might be parsing as `Object`, but that won't work when the map values are complex types.

eamonnmcmanus:
Nevertheless, I would prefer to document a way of achieving this with the existing API, if there is a way that is not too onerous. I agree that the StackOverflow discussion I linked to isn't very satisfactory. But in a previous issue, @swankjesse [wrote](https://github.com/google/gson/pull/649#issuecomment-248290958):
> you can install your own TypeAdapter to workaround this if you like. It’s a few lines of code to write the adapter and install it.

Marcono1234:
> It’s a few lines of code to write the adapter

I am rather skeptical whether that is actually true. If you have a homogenous map with simple value types (e.g. strings or numbers), then such an adapter might indeed not be that complex. However, as soon as you are dealing with heterogenous maps with custom classes as values or when you want to have a generic implementation you most likely end up copying parts of the `MapTypeAdapterFactory` code.

Though you could also wait with merging this pull request to gather information about how many users are interested in this feature. Or @loyili maybe you could describe whether your use case would have also allowed writing a custom Map adapter without too much overhead?

loyili:
@eamonnmcmanus  @Marcono1234   Thanks for your attention. Currently we fork a private branch for this feature instead of writing the compatibility code. We don't want to copy the duplicate code from one project to another independent project, maybe it is more suitable as a feature.

Issue #61, opened 2015-03-19
Provide an @Required annotation to indicate that a field must be present


GoogleCodeExporter:
```
See discussion at http://groups.google.com/group/google-
gson/browse_thread/thread/6cfdcc56742357da

During deserialization, Gson should throw a JsonParseException if it finds 
that a field marked with @Required is not present. 

During serialization, Gson should always write out a field marked @Required 
even if it is null. 
```

Original issue reported on code.google.com by `inder123` on 17 Oct 2008 at 10:31

dangrasso:
+1

mattak:
+1

karupanerura:
+1

RahulRvR:
Does Gson have this feature yet ?

aldoborrero:
+1

chenDoInG:
+1

preetamp:
I am able to achieve this using this approach. This is for serialization (fail Java to Json if some @required attribute is null)
http://stackoverflow.com/questions/35587252/java-to-json-validation-using-gson/

jobinar:
+1

pine613:
:+1:

BrianOn99:
+1

Ryanmt2:
+1

tir38:
Please use this thumbs up button to show your support. Every time someone replies with +1 we *all* get a (useless) email.
<img width="825" alt="screen shot 2016-07-28 at 12 03 59 pm" src="https://cloud.githubusercontent.com/assets/1980553/17219997/91b7dee4-54bb-11e6-98ad-502171a3d7b5.png">

nzackoya:
Seems that no one cares this library. This issue should be marked as critical.

mnouh:
+1

naturalwarren:
There's a library that's can help with this:

https://github.com/uber-common/rave

arbitur:
Still waiting for this feature... I want my Gson deserializing to throw an exception when a JSON doesnt have a parameter which my object do have and is either marked as @Required or not marked as @Optional. This is a big dealbreaker that Gson doesn't have this feature...

Darkhax:
I would love to see this feature implemented as well.

charlesganza:
To anyone still still frustrated with the lack of required fields option in Gson, you can use [https://github.com/uber-archive/rave](Rave). It solves this problem beautifully if you want to avoid NPEs from unreliable APIs.

charlesganza:
[...]

PR #1900, opened 2021-05-30
Add a @Required annotation for fields

grishka:
This is a long-requested feature (#61).

I'm not sure if this is the best way to do this, especially the `Set<String>` part to keep track of required fields that were found in the input, but I haven't come up with anything better. I'm also not sure if this annotation should affect serialization — maybe it should also throw an exception when attempting to serialize a required field that is null?

### DIFF
--- a/gson/src/main/java/com/google/gson/annotations/Required.java
+++ b/gson/src/main/java/com/google/gson/annotations/Required.java
@@ -1,0 +1,18 @@
+package com.google.gson.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation causes a {@link com.google.gson.JsonParseException} to be thrown
+ * during deserialization if this field is not present in the input JSON, is null,
+ * or is not of a type assignable to this field.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface Required {
+}
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -18,10 +18,12 @@ package com.google.gson.internal.bind;
 
 import com.google.gson.FieldNamingStrategy;
 import com.google.gson.Gson;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
 import com.google.gson.annotations.JsonAdapter;
+import com.google.gson.annotations.Required;
 import com.google.gson.annotations.SerializedName;
 import com.google.gson.internal.$Gson$Types;
 import com.google.gson.internal.ConstructorConstructor;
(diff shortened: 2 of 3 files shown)


grishka:
@googlebot I signed it!

Marcono1234:
Looks reasonable, but maybe two things which have to be considered:
1. In some use cases an explicit JSON `null` is considered a value, so maybe the annotation should have an `allowNull` (or similar) annotation element with default value `false` to support that use case.
2. Instead of tracking which field names have been deserialized you should probably track the `Field` (needs some refactoring). Otherwise this won't work for [`@SerializedName` with `alternate`](https://www.javadoc.io/static/com.google.code.gson/gson/2.8.7/com.google.gson/com/google/gson/annotations/SerializedName.html#alternate()), e.g.:
    ```java
    class Test {
        @SerializedName(value = "a", alternate = "b")
        String s;
    }
    ```

Note that I am not a member of this project, so these are only suggestions.

gson/src/main/java/com/google/gson/annotations/Required.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation causes a {@link com.google.gson.JsonParseException} to be thrown
+ * during deserialization if this field is not present in the input JSON, is null,
+ * or is not of a type assignable to this field.
```
Marcono1234, line 12: > or is not of a type assignable to this field.

That is the default behavior for fields anyways, isn't it? I don't think this is worth mentioning here then, otherwise it might cause confusion whether this does / does not apply to optional fields.

grishka: Well, if you try to parse something that has an object where you expect a string, the string will be null. The idea behind this annotation is to guarantee that a required field will never, under any circumstances, be null.

Marcono1234: > Well, if you try to parse something that has an object where you expect a string, the string will be null.

Are you sure? The following throws an exception:
```java
public class GsonParsingTest {
    static class Test {
        String s;
    }
    
        new Gson().fromJson("{\"s\":{}}", Test.class);
    }
}
```

Maybe there are custom type adapters which return `null` for non-`null` JSON in case of type mismatch, but we cannot generally assume that. Not every type adapter which returns `null` must have encountered wrong JSON data.

lyubomyr-shaydariv:
`@Required` sounds like a twin of JSR-303 `@NotNull` and does not seem to make it extensible. I would not create a doubling annotation, but rather would think of implementing a generic method to apply _any kind_ of post-processing at least at DTO class level (probably by extending `ReflectiveTypeAdapterFactory` that makes a lot of things hard to accompilsh being not extensible). Validation is only a particular case of such a generic thing. This all seems to be pretty achievable using post-processing mechanisms like `RuntimeTypeAdapterFactory` from the Gson Extras demo:

- it's possible to do anything with a deserialized object, including validation (entire JSON document must be consumed though);
- it's possible to incorporate any strategy (including JSR-303 validation as a Jackson-like module that can be built on top of type adapter factories) right into type adapter factories, so that any `ReflectiveTypeAdapterFactory.Adapter`-derived type adapter could invoke the strategy and quick-validate the just-deserialized object (subobjects revalidated every time a super object is completed to be serialized, so it may be time expensive -- but it sounds like it can be optimized greatly).

Any other thoughts on making Gson deserialization-only, but extensible on user demand only, tool?

grishka:
> but rather would think of implementing a generic method to apply any kind of post-processing

The problem with the post-processing approach is that it, in case of a field of a primitive type, it won't let you tell the difference between the lack of that field in the json and its presence with its default value (0/false). Yes, boxed types can kinda awkwardly solve this by allowing those fields to be null. But you won't want to use boxed types in all your objects just so you could validate them in a post-processing step.

> JSR-303 `@NotNull`

This is the first time I've seen this after more than 10 years writing Java. I do know about `@NotNull`/`@Nullable` commonly used in many places these days, but these usually either come from a JetBrains library for Java SE or an AndroidX library for Android. And — I just looked — their retention is "class", so not visible at runtime.

Issue #1540, opened 2019-06-05
com.google.gson.internal.reflect.UnsafeReflectionAccessor  illegal reflective access operation on jdk12


annarozin:
Hi,

Since upgrading to JDK 12 (Gson v2.8.5), we keep getting warning messages like the following when doing JSON-serialization using Gson:

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.gson.internal.reflect.UnsafeReflectionAccessor (file:/C:/UA/wss-unified-agent-19.5.1.1.jar) to field java.util.concurrent.atomic.AtomicReference.value
WARNING: Please consider reporting this to the maintainers of com.google.gson.internal.reflect.UnsafeReflectionAccessor
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

It doesn't happen on JDK 8 or JDK 11 it only happens on JDK 12

Thanks :)

aaron-suarez:
+1 for this issue, JDK 12 and Gson v2.8.5 attempts illegal reflective access. Here's the stack trace of the error in case that helps:

```
java.lang.reflect.InaccessibleObjectException: Unable to make jdk.internal.ref.PhantomCleanable() accessible: module java.base does not "opens jdk.internal.ref" to unnamed module @2eea88a1
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:341)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:281)
        at java.base/java.lang.reflect.Constructor.checkCanSetAccessible(Constructor.java:189)
        at java.base/java.lang.reflect.Constructor.setAccessible(Constructor.java:182)
        at com.google.gson.internal.reflect.UnsafeReflectionAccessor.makeAccessible(UnsafeReflectionAccessor.java:44)
        at com.google.gson.internal.ConstructorConstructor.newDefaultConstructor(ConstructorConstructor.java:103)
        at com.google.gson.internal.ConstructorConstructor.get(ConstructorConstructor.java:85)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:101)
        at com.google.gson.Gson.getAdapter(Gson.java:458)
[...]

PR #1902, opened 2021-06-04
Fix failing to serialize Collection or Map with inaccessible constructor

### LINKED ISSUE #1875: InaccessibleObjectException with JDK 16
GSon (2.8.6) has trouble with JDK 16, seemingly due to https://bugs.openjdk.java.net/browse/JDK-8256358 and https://bugs.openjdk.java.net/browse/JDK-8260600 which prevent access through reflection to JDK internals.

This leads to these errors (for example when trying to serialize an empty list constructed with `Collections.emptyList()`) : 

```
java.lang.reflect.InaccessibleObjectException: Unable to make private java.util.Collections$EmptyList() accessible: module java.base does not "opens java.util" to unnamed module @ffaa6af
	at com.google.gson.internal.reflect.UnsafeReflectionAccessor.makeAccessible(UnsafeReflectionAccessor.java:44)
	at com.google.gson.internal.ConstructorConstructor.newDefaultConstructor(ConstructorConstructor.java:103)
	at com.google.gson.internal.ConstructorConstructor.get(ConstructorConstructor.java:85)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory.create(CollectionTypeAdapterFactory.java:54)
	at com.google.gson.Gson.getDelegateAdapter(Gson.java:541)
	at io.gsonfire.gson.WrapTypeAdapterFactory.create(WrapTypeAdapterFactory.java:26)
	at com.google.gson.Gson.getAdapter(Gson.java:458)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:56)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.write(MapTypeAdapterFactory.java:208)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.write(MapTypeAdapterFactory.java:145)
	at com.google.gson.Gson.toJson(Gson.java:704)
	at com.google.gson.Gson.toJson(Gson.java:683)
...
```

I would be happy to contribute but I do not even know how to fix this in Gson. Any ideas ?

Marcono1234:
Fixes #1875

Note that this introduces an incompatibility:
Projects which previously relied on `UnsafeReflectionAccessor` making internal members accessible will now encounter warnings or exceptions when the Java Platform Module System does not allow access. These projects should write their own `TypeAdapter`s, or temporarily use `java --add-opens ...`.
This "resolves" #1540, in the sense that that issue is not a problem with Gson, and that this pull request further strengthens this standpoint.
(In theory this change could be omitted from this pull request while still solving #1875, but it would render the added unit test ineffective.)


gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -97,33 +96,53 @@ public ConstructorConstructor(Map<Type, InstanceCreator<?>> instanceCreators) {
...
-                e.getTargetException());
-          } catch (IllegalAccessException e) {
-            throw new AssertionError(e);
-          }
+        @Override
+        public T construct() {
+          // New exception is created every time to avoid keeping reference
+          // to exception with potentially long stack trace, causing a
+          // memory leak
+          throw new JsonIOException(exceptionMessage);
```
Marcono1234, line 122: `JsonIOException` thrown here and in `ReflectionHelper` is not fitting very well, but Gson has currently no better fitting exception type.
(maybe the whole Gson exception hierarchy should be refactored in the future)

Marcono1234:
@eamonnmcmanus could you please have a look at this pull request? The fixed issue (#1875) causes problems for some legit usages of certain JDK types.

If you want I can also split off the removal of `UnsafeReflectionAccessor`; though ultimately it should be removed either way because it circumvents JDK module system restrictions and is ineffective in the latest JDK versions anyway (see pull request description).

gson/src/test/java/com/google/gson/functional/ThrowableFunctionalTest.java
```diff
@@ -1,65 +0,0 @@
...
-
-import java.io.IOException;
-
-import junit.framework.TestCase;
-
-import com.google.gson.Gson;
-import com.google.gson.annotations.SerializedName;
-
-@SuppressWarnings("serial")
-public final class ThrowableFunctionalTest extends TestCase {
```
Marcono1234, line 12: Has been removed because it relies on JDK internals.

gson/src/test/java/com/google/gson/internal/bind/RecursiveTypesResolveTest.java
```diff
@@ -53,21 +53,6 @@ public void testRecursiveResolveSimple() {
     assertNotNull(adapter);
   }
 
-  /**
```
Marcono1234, line 56: Have been removed because they rely on JDK internals.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -0,0 +1,71 @@
+package com.google.gson.internal.reflect;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+
+import com.google.gson.JsonIOException;
```
eamonnmcmanus, line 6: Can you sort these imports lexicographically and omit blank lines?

eamonnmcmanus (changes requested):
This looks very helpful, thanks! I have mostly nits. Also a merge is needed.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -0,0 +1,71 @@
...
+   * Creates a string representation for a constructor.
+   * E.g.: {@code java.lang.String#String(char[], int, int)}
+   */
+  private static String constructorToString(Constructor<?> constructor) {
+    StringBuilder stringBuilder = new StringBuilder(constructor.getDeclaringClass().getName())
+      .append('#')
+      .append(constructor.getDeclaringClass().getSimpleName())
+      .append('(');
+    Class<?>[] parameters = constructor.getParameterTypes();
+    for (int i = 0; i < parameters.length; i++) {
```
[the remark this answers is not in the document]
Marcono1234: Both probably have their advantages and disadvantages. If you don't mind I would prefer keeping the existing code here.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.gson.JsonIOException;
+
+public class ReflectionHelper {
+  private ReflectionHelper() { }
+
+  /**
+   * Tries making the field accessible, wrapping any thrown exception in a
+   * {@link JsonIOException} with descriptive message.
+   *
+   * @param field
```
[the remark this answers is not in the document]
Marcono1234: Have changed it. It probably depends on the IDE you are using, but I kind of like this wrapped documentation style because it makes the distinction between parameter name and description clearer.
Though it does not follow the style guide of many projects.

eamonnmcmanus (approved):

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -759,22 +760,31 @@ public void write(JsonWriter out, Locale value) throws IOException {
...
-          AccessController.doPrivileged(new PrivilegedAction<Void>() {
-            @Override public Void run() {
-              field.setAccessible(true);
-              return null;
+        // Uses reflection to find enum constants to work around name mismatches for obfuscated classes
+        // Reflection access might throw SecurityException, therefore run this in privileged context;
+        // should be acceptable because this only retrieves enum constants, but does not expose anything else
+        Field[] constantFields = AccessController.doPrivileged(new PrivilegedAction<Field[]>() {
+          @Override public Field[] run() {
+            Field[] fields = classOfT.getDeclaredFields();
```
Marcono1234, line 770: Have changed this to also run `getDeclaredFields()` in privileged context because it can throw a `SecurityException` as well.

Marcono1234:
Sorry for these changes after your approval; I tried adding back the Maven toolchain usage, but unfortunately even with version range it apparently requires to explicitly configure toolchains. Therefore I have removed that commit again and force-pushed; you can see the changes by clicking on the "force-pushed" text here:
force-pushed text screenshot

Also just to be sure there is no misunderstanding; these changes might cause backward compability issues for Gson users running < JDK 12 who rely on Gson's `Unsafe` usage for making fields and constructors accessible to which they should not have any access, because this pull request removes that code.
However, on the other hand these versions < JDK 12 will by default permit illegal access for the module system (was changed in Java 16 to `deny` by default, see [JDK-8255363](https://bugs.openjdk.java.net/browse/JDK-8255363)), so unless the users explicitly changed this behavior, the worse thing they will experience is warnings about illegal access on the console.

eamonnmcmanus:
This looks fine now.

PR #1905, opened 2021-06-04
Add support for reflection access filter

Marcono1234:
Adds a `ReflectionAccessFilter` which allows preventing to use the reflection based type adapter by accident, instead throwing an exception. This is quite useful to avoid depending on Java platform implementation details.

A use case where this would help is https://github.com/stripe/stripe-java/pull/1197, that pull request adds a check whether a returned adapter is `com.google.gson.internal.bind.ReflectiveTypeAdapterFactory`; the addition of a `ReflectionAccessFilter` would allow implementing this without having to rely on Gson internal implementation details.

In the future it would probably be good to apply some of these access filters by default since it appears users are often by accident depending on implementation details, or are not aware of the consequences of this. However, the way `ReflectionAccessFilter` is designed would still allow overwriting this by registering a filter which returns `FilterResult.ALLOW` for all classes, in case users for some reason want to serialize or deserialize Java platform classes using reflection.
Moshi by default prevents reflective access for platform classes, see internal [`Util.isPlatformType(Class)`](https://github.com/square/moshi/blob/3c108919ee1cce88a433ffda04eeeddc0341eae7/moshi/src/main/java/com/squareup/moshi/internal/Util.java#L141) and its callers.


Marcono1234:
Alternatively `FilterResult` could be a regular class with a nullable `String filterName` and a `FilterResultType type`, where `FilterResultType` would be an enum with `ALLOW`, `INDECISIVE` and `BLOCK_INACCESSIBLE` and `BLOCK_ALL`.
Then a filter could optionally specify its name (for example "platform type filter") for the filter result. This could help with debugging and would allow including it in the message of the exception thrown when not allowed reflective access occurs.

But I am not sure if this is worth it. What do you think?

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Blocks reflection access if a member of the class is not accessible
+     * by default and would have to be made accessible. This if unaffected
+     * by any {@code java} command line arguments being used to make packages
+     * accessible.
+     *
+     * <p>Note that this <b>only works for Java 9 and higher</b>, for older
+     * Java versions its functionality will be limited and it might behave like
+     * {@link #ALLOW}. Access checks are only performed as defined by the Java
+     * Language Platform System, restrictions imposed by a {@link SecurityManager}
```
Marcono1234, line 60: Should this be?
```suggestion
     * Platform Module System, restrictions imposed by a {@link SecurityManager}
```

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * <p>This result type is mainly intended to help enforcing the access checks of
+     * the Java Platform Module System. It allows detecting illegal access, even if
+     * the used Java version would only log a warning, or is configured to open
+     * packages for reflection using command line arguments.
+     *
+     * @see AccessibleObject#setAccessible(boolean)
+     * @see AccessibleObject#canAccess(Object)
+     */
+    BLOCK_INACCESSIBLE,
```
Marcono1234, line 71: It appears the used method `AccessibleObject.canAccess` completely ignores whether a module is open for reflective access. Maybe this should be made clearer here. But the question is then how useful this filter result type actually is.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -577,6 +580,26 @@ public GsonBuilder serializeSpecialFloatingPointValues() {
     return this;
   }
 
+  /**
+   * Adds a reflection access filter. A reflection access filter allows restricting for
+   * which classes Gson is allowed to use reflection for serialization and deserialization.
```
eamonnmcmanus, line 585: I think this sentence could be reworded for clarity. What about this?

```
A reflection access filter prevents Gson from using reflection for the serialization and deserialization of certain classes. The logic in the filter specifies which classes those are.
```

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
eamonnmcmanus: I think that is the official name, yes.

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+     * be used.
+     */
+    BLOCK_ALL
+  }
+
+  /**
+   * Blocks all reflection access to members of standard Java classes which are
+   * not accessible by default. However, reflection access is still allowed for
+   * classes for which all fields are accessible and which have an accessible
+   * no-args constructor (or for which an {@link InstanceCreator} has been registered).
```
eamonnmcmanus, line 84: I think it would be good to spell out here exactly what a "standard Java class" is. Maybe just, at the end of the paragraph:

"A standard Java class is any class in a {@code java.*} or {@code javax.*} package."

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+  ReflectionAccessFilter BLOCK_INACCESSIBLE_JAVA = new ReflectionAccessFilter() {
+    @Override public FilterResult check(Class<?> rawClass) {
+      return ReflectionAccessFilterHelper.isJavaType(rawClass)
+        ? FilterResult.BLOCK_INACCESSIBLE
+        : FilterResult.INDECISIVE;
+    }
+  };
+
+  /**
+   * Blocks all reflection access to members of standard Java classes.
```
eamonnmcmanus, line 110: Say what a standard Java class is here too.

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+  ReflectionAccessFilter BLOCK_ALL_JAVA = new ReflectionAccessFilter() {
+    @Override public FilterResult check(Class<?> rawClass) {
+      return ReflectionAccessFilterHelper.isJavaType(rawClass)
+        ? FilterResult.BLOCK_ALL
+        : FilterResult.INDECISIVE;
+    }
+  };
+
+  /**
+   * Blocks all reflection access to members of standard Android classes.
```
eamonnmcmanus, line 134: "A standard Android class is any class in a {@code android.*} or {@code androidx.*} package, or any standard Java class ({@code java.*} or {@code javax.*})."

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override public FilterResult check(Class<?> rawClass) {
+      return ReflectionAccessFilterHelper.isAndroidType(rawClass)
+        ? FilterResult.BLOCK_ALL
+        : FilterResult.INDECISIVE;
+    }
+  };
+
+  /**
+   * Blocks all reflection access to members of classes belonging to programming
+   * language platforms, such as Java, Android, Kotlin or Scala.
```
eamonnmcmanus, line 157: "This includes all classes that would be blocked by {@link #BLOCK_ALL_JAVA} or {@link #BLOCK_ALL_ANDROID}, as well as language-specific platform classes like {@code kotlin.*}." I think we can avoid spelling out exactly what those language-specific classes are.

eamonnmcmanus (changes requested):
Sorry for taking so long to get to this. I think this is an excellent addition. Accidental reflective access seems to be a major problem with Gson, and this will go some way towards mitigating that.

If you merge to HEAD I will try running this against Google's internal tests. I don't really expect any failures but we are sometimes surprised. Feel free to do just the merge initially and hold off on the documentation changes I suggested, until we know the results. Or go ahead and make them, as you prefer.

Marcono1234:
Have improved the documentation and fixed some issues with the previous implementation. I chose a slightly different wording for the sentences describing the standard platform packages. Please let me know if it is alright like this.

As mentioned in https://github.com/google/gson/pull/1905#issuecomment-897945293, should this be extended to support describing which filter returned the final result? Or do you think that is not worth it?

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -632,6 +635,28 @@ public GsonBuilder disableJdkUnsafe() {
     return this;
   }
 
+  /**
+   * Adds a reflection access filter. A reflection access filter prevents Gson from using
+   * reflection for the serialization and deserialization of certain classes. The logic in
+   * the filter specifies which classes those are.
+   *
+   * <p>Filters will be invoked in reverse registration order, that is, the least recently
```
eamonnmcmanus, line 643: Isn't this just registration order? I would expect "registration order" to mean that the earliest one is invoked first.

eamonnmcmanus:
Google's internal tests pass with this change, so we can merge it once the conflicts have been resolved and the remaining minor comments addressed.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -101,6 +102,7 @@
   private boolean useJdkUnsafe = DEFAULT_USE_JDK_UNSAFE;
   private ToNumberStrategy objectToNumberStrategy = DEFAULT_OBJECT_TO_NUMBER_STRATEGY;
   private ToNumberStrategy numberToNumberStrategy = DEFAULT_NUMBER_TO_NUMBER_STRATEGY;
+  private final LinkedList<ReflectionAccessFilter> reflectionFilters = new LinkedList<ReflectionAccessFilter>();
```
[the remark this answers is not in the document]
Marcono1234: Good point! I completely forgot that Java 7 added that feature already. Would you mind if I create a separate PR which replaces all redundant type arguments with the diamond operator?

gson/src/main/java/com/google/gson/GsonBuilder.java
Marcono1234: Yes, you are right, thanks for pointing that out!

Marcono1234:
Thanks for running this against Google's internal tests! Just to verify: You do not think that supporting specifying which filter returned a result (see https://github.com/google/gson/pull/1905#issuecomment-1030339509) is worth it, correct?

I have added one commit improving `IllegalAccessException` handling. Currently when Gson users use Java < 9, but use `FilterResult.BLOCK_INACCESSIBLE`, they might run into `IllegalAccessException` because `ReflectionAccessFilterHelper.canAccess(AccessibleObject, Object)` won't validate whether the object is accessible. Do you think that is acceptable, or should it try to emulate the access check (but in a simplified way because package-private and `protected` checks are irrelevant because Gson is in a different package)?

Marcono1234:

gson/src/main/java/com/google/gson/GsonBuilder.java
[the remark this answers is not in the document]
eamonnmcmanus: Seems fair.

eamonnmcmanus (approved):
Sorry, I lost track of this.

gson/src/main/java/com/google/gson/GsonBuilder.java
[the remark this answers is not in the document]
Marcono1234: Thanks! Have created #2104 for that.

PR #1909, opened 2021-06-11
Test cases for testing the exceptional behavior of JsonArray get... methods

HiFromAjay:
Test cases for testing the exceptional behavior of `get`, `getAsBoolean`, `getAsDouble`, `getAsInt`, `getAsJsonArray`, `getAsJsonObject`, `getAsLong`, and `getAsString` methods of `JsonArray` class. These test cases, which we wrote according to the specified behavior of each method, that helped us in identifying the documentation bugs in `JsonArray` and `JsonElement` classes, which we submitted issues for (Issue #1908). Note that we have adapted these test cases based on similar tests from the [JSON-java](https://github.com/stleary/JSON-java) project.

### DIFF
--- a/gson/src/test/java/com/google/gson/JsonArrayTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayTest.java
@@ -16,12 +16,8 @@
 
 package com.google.gson;
 
-import junit.framework.TestCase;
-
 import com.google.gson.common.MoreAsserts;
-
-import static org.junit.Assert.*;
-import static org.junit.Assert.assertEquals;
+import junit.framework.TestCase;
 
 /**
  * @author Jesse Wilson
--- a/gson/src/test/java/com/google/gson/JsonArrayTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayTest.java
@@ -20,6 +20,9 @@ import junit.framework.TestCase;
 
 import com.google.gson.common.MoreAsserts;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
 /**
  * @author Jesse Wilson
  */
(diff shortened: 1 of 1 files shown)


HiFromAjay:
@googlebot I signed it!

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -99,4 +102,65 @@ public void testDeepCopy() {
     assertEquals(1, original.get(0).getAsJsonArray().size());
     assertEquals(0, copy.get(0).getAsJsonArray().size());
   }
+
+  public void testFailedGetArrayValues() {
+    String arrayStr = "[" + "true," + "false," + "\"true\"," + "\"false\"," + "\"hello\"," + "23.45e-4," + "\"23.45\"," + "42," + "\"43\"," + "[" + "\"world\"" + "]," + "{" + "\"key1\":\"value1\"," + "\"key2\":\"value2\"," + "\"key3\":\"value3\"," + "\"key4\":\"value4\"" + "}," + "0," + "\"-1\"" + "]";
+    JsonArray jsonArray = (JsonArray) JsonParser.parseString(arrayStr);
+    try {
+      jsonArray.get(10).getAsBoolean();
+      assertTrue("expected getBoolean to fail", false);
```
Marcono1234, line 111: Instead of `assertTrue(..., false)` it might be better to use `fail(...)` instead; applies to other tests as well.

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -99,4 +102,65 @@ public void testDeepCopy() {
...
+    } catch (UnsupportedOperationException e) {
+      assertEquals("Expected an exception message",
+              "JsonObject",e.getMessage());
+    }
+    try {
+      jsonArray.get(-1);
```
Marcono1234, line 117: Except for this test, all other ones are not specific to JsonArray, it might be better to put them to [`JsonPrimitiveTest`](https://github.com/google/gson/blob/master/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java) (or maybe there exists another more fitting test class).

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -99,4 +102,65 @@ public void testDeepCopy() {
     assertEquals(1, original.get(0).getAsJsonArray().size());
     assertEquals(0, copy.get(0).getAsJsonArray().size());
   }
+
+  public void testFailedGetArrayValues() {
+    String arrayStr = "[" + "true," + "false," + "\"true\"," + "\"false\"," + "\"hello\"," + "23.45e-4," + "\"23.45\"," + "42," + "\"43\"," + "[" + "\"world\"" + "]," + "{" + "\"key1\":\"value1\"," + "\"key2\":\"value2\"," + "\"key3\":\"value3\"," + "\"key4\":\"value4\"" + "}," + "0," + "\"-1\"" + "]";
```
Marcono1234, line 107: Not all values of this array are used; might be good to remove unused values to make this test easier to understand.

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -99,4 +102,65 @@ public void testDeepCopy() {
...
+    JsonArray jsonArray = (JsonArray) JsonParser.parseString(arrayStr);
+    try {
+      jsonArray.get(10).getAsBoolean();
+      assertTrue("expected getBoolean to fail", false);
+    } catch (UnsupportedOperationException e) {
+      assertEquals("Expected an exception message",
+              "JsonObject",e.getMessage());
```
Marcono1234, line 114: Really minor, but for consistent code style there should probably be a space:
```suggestion
              "JsonObject", e.getMessage());
```
(applies to other tests as well)

Marcono1234 (commented):
I hope the review comments are useful; note that I am not a member of this project so feel free to consider these only as suggestions.

gson/src/test/java/com/google/gson/JsonArrayTest.java
HiFromAjay: Changed to fail(...)

gson/src/test/java/com/google/gson/JsonArrayTest.java
HiFromAjay: Changed how the methods are used.

gson/src/test/java/com/google/gson/JsonArrayTest.java
HiFromAjay: Removed the unused values.

gson/src/test/java/com/google/gson/JsonArrayTest.java
HiFromAjay: Resolved the formatting issues.

gson/src/test/java/com/google/gson/JsonArrayTest.java
Marcono1234: Thanks! Also my comment was incorrect, I missed that the `getAs...` methods of `JsonArray` actually try to convert the array element in case the array contains only a single element.

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -20,6 +20,9 @@
 
 import com.google.gson.common.MoreAsserts;
 
+import static org.junit.Assert.*;
```
eamonnmcmanus, line 23: It's not clear to me why these imports are needed. Aren't `assertEquals` etc inherited from `TestCase`?

eamonnmcmanus (changes requested):
Thanks for this! Just one small comment.

gson/src/test/java/com/google/gson/JsonArrayTest.java
HiFromAjay: I have removed those imports.

eamonnmcmanus (approved):

Issue #1896, opened 2021-05-25
`build.gradle` project version is outdated


Marcono1234:
The version specified by `build.gradle` is outdated:
https://github.com/google/gson/blob/f5fcb0f4cbe554a237459db96f2aebd41ca3f394/gson/build.gradle#L5

Currently the correct version would be `2.8.8-SNAPSHOT`.

However in general it might be good to drop Gradle as build tool, or only use Gradle (instead of Maven) in the future. Having to maintain two build tools increases maintenance work. For example Gradle builds currently do not seem to correctly include the project version in `GsonBuildConfig.java`, see:
- #1554 (not sure if that pull request actually tries to fix that issue)
- #1820

In case support for Gradle is dropped, the corresponding GitHub workflow (see #1648) can be removed again.

PR #1913, opened 2021-06-24
Inline declaration of Java compatibility

Ab20zipYT:
There is no need to write two lines to declare the source and the target Java compatibility.

### DIFF
--- a/gson/build.gradle
+++ b/gson/build.gradle
@@ -4,8 +4,7 @@ apply plugin: 'maven'
 group = 'com.google.code.gson'
 version = '2.8.6-SNAPSHOT'
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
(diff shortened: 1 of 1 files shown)


Ab20zipYT:
@googlebot I signed it!

eamonnmcmanus:
Is there a particular advantage to doing it one way or the other? I see that the combined assignment is slightly shorter, but it might be fractionally less easy to read and doesn't easily allow for the (unlikely) possibility of specifying different source and target versions.

Marcono1234:
Is the Gradle build config even actively maintained? See #1896

Ab20zipYT:
I just want to contribute to Google Gson please merge it. There are so many great developers out there but I'm just a lonely little kid who joined Github like 3 days ago. Please have some mercy.

PR #1915, opened 2021-07-01
Refactored three test methods in CustomSerializerTest, and two test methods in ThrowableFunctionalTest

AliIman:
As part of a research project at the University of Waterloo, we are exploring automatic test refactoring tools and submitting vetted proposed refactorings upstream. These refactorings make it easier to add additional related test methods and should help with the maintainability of the tests. In this pull request, we have two refactoring cases that refactors five test methods. We've checked both cases and ensured that all the test runs pass successfully.

If it helps, we can also propose a test case that uses these five refactored test methods. We explained each of the two refactoring cases below.

1. In CustomSerializerTest, after declaration of gson, the only difference between the three test methods "testBaseClassSerializerInvokedForBaseClassFields() ", "testSubClassSerializerInvokedForBaseClassFieldsHoldingSubClassInstances", and "testSubClassSerializerInvokedForBaseClassFieldsHoldingArrayOfSubClassInstances" is in a class type (new Base() vs new Sub()) and a string (BaseSerializer.NAME vs SubSerializer.NAME). We create "testSerializerInvokedForBaseClassFieldsTemplate(Gson gson, Class<TBase> clazzTBase, String string1)" parametrized method, which allows you to add similar test cases with different arguments gson, clazzTBase and string1.

2. In ThrowableFunctionalTest, the only difference between the two test methods "testExceptionWithoutCause()" and "testErrorWithoutCause()" is in a class type (RuntimeException.class vs OutOfMemoryError.class). We create "testThrowableWithoutCauseTemplate(Class<TThrowable> clazzTThrowable, String msg, String jsonString)" parametrized method, which allows you to add similar test cases with different arguments clazzTThrowable, msg, and jsonString.

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/CustomSerializerTest.java
+++ b/gson/src/test/java/com/google/gson/functional/CustomSerializerTest.java
@@ -16,6 +16,7 @@
 
 package com.google.gson.functional;
 
+import com.google.gson.common.TestTypes;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonArray;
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Thanks for your contribution!

A few remarks.

1. This project is in maintenance mode, and we're generally going to be reluctant to accept PRs that are essentially cosmetic, especially if it is not trivially obvious that they don't change anything.
2. Anyway, I'm not really convinced that the change would be an improvement. The new helper methods seem more complicated than I would expect. I don't see why they catch and rethrow exceptions, for example. They also have type parameters that don't seem useful.
3. As a more general remark, refactoring test methods to remove duplication isn't as obviously beneficial as with production code. It's an advantage to be able to understand a test method in isolation, without having to look elsewhere in the test class.

So I think we're not going to accept this one. Thanks for thinking of us, though, and best wishes for your research!

PR #1929, opened 2021-08-05
Provide proguard/r8 rules

ganadist:
Currently, developer needs to add proguard rules manually.
But rules can be provided as resource and proguard/r8 will read and apply automatically.

Rules are copied from https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg


eamonnmcmanus:
If these resources are present, does that mean that `android-proguard-example` can be made to work correctly even if its own copy of the proguard rules is deleted?

How do we test that these rules are working correctly?

ganadist:
Note: This repo needs `git-lfs`
```
$ git clone https://github.com/ganadist/VersionCodeDemo -b gson_proguard_rule_demo 
$ cd VersionCodeDemo

### use AGP 4.2
$ git checkout HEAD~1
$ ./gradlew :app:pPRUA
...
> Task :app:packageProductionReleaseUniversalApk

$ cat app/build/outputs/mapping/productionRelease/configuration.txt

### check whether META-INF/com.android.tools/r8/gson.pro is included

### use AGP 7.0
$ git checkout origin/gson_proguard_rule_demo
$ ./gradlew :app:pPRUA
...
> Task :app:packageProductionReleaseUniversalApk

$ cat app/build/outputs/mapping/productionRelease/configuration.txt

### check whether META-INF/com.android.tools/r8/gson.pro and META-INF/com.android.tools/r8-from-3.0.0/gson.pro are included 

```

See also 
* https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/dependency/FilterShrinkerRulesTransform.kt
* https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-core/src/test/java/com/android/build/gradle/internal/dependency/FilterShrinkerRulesTransformTest.kt

eamonnmcmanus:
Sorry, I should have been a bit more explicit about my concerns. This PR is adding 4 different ProGuard configuration files. Is there a way to add an automated test that will check that each of those files is correct? I think it will be hard to maintain the files otherwise.

thekingrenz23:
Any update to this one? r8 rules should be included its 2022 now

Marcono1234:
Default rules have been added by #2397 and will be available in the next version, so this pull request here might be obsolete.

@ganadist and everyone else who wanted this feature, could you please try building Gson locally and see if the snapshot version of Gson correctly applies the rules.

Any feedback regarding these rules, such as if something is missing or is too coarse, is highly appreciated!

Also keep in mind that these default library rules cannot cover everything; if you have specific classes or fields which should be preserved and you are not using `@SerializedName` you might have to add additional rules for your build.

ganadist:
Closed by #2397

Issue #690, opened 2015-08-21
publish proto to Maven central?


borud:
While waiting for protobuf version 3 to get support for JSON I've started using Gson to transform protobuffers into JSON and I found `ProtoTypeAdapter` written by Inderjeet Singh under the `proto` directory.  However this does not appear to be part of Gson proper (wouldn't hurt that much if it was) and it does not appear to be pushed to Maven central.

In my project I've just included the code in the source, but I'm not really comfortable with that.  I'd like to get it from somewhere that is being properly maintained.

Would you consider moving this code into the `gson` proper or pushing this artifact to Maven central?

inder123:
Hi. Proto is not a very popular technology and will not make into Gson proper. We may publish an additional maven sub-project. However, I haven't maintained this version after writing it.
I think you did the right thing by including the source code.

borud:
OK, thanks for the quick response @inder123!

tobegit3hub:
com.google.code.gson.proto is needed in our project. Now we need to copy the source code to use and it would be much better if we can rely to the maven repo with compatible version. @borud @inder123

PR #1949, opened 2021-09-01
Update dependencies in proto/pom.xml.

### LINKED ISSUE #1797: There is a vulnerability in Guava: Google Core Libraries for Java(18.0) ,upgrade recommended
https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/proto/pom.xml#L69-L74

CVE-2018-10237

Recommended upgrade version：1.0.1
latest version：29.0-jre

eamonnmcmanus:
Also use `GeneratedMessageV3` rather than `GeneratedMessage`, consistent with recent versions of `protoc`.

Closes #1797.

### DIFF
--- a/proto/pom.xml
+++ b/proto/pom.xml
@@ -55,35 +55,35 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.4</version>
+      <version>2.8.8</version>
       <scope>compile</scope>
     </dependency>
 
     <dependency>
       <groupId>com.google.protobuf</groupId>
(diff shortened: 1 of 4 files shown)


SNDST00M:
When will this release?

eamonnmcmanus:
According to `proto/pom.xml`, the artifact id is `com.google.code.gson:proto`. I don't think that has ever been released. https://github.com/google/gson/issues/690 suggests that was mostly intentional. Given that Gson is in maintenance mode we are probably not going to be releasing new subprojects.

SNDST00M:
Right, struggled to understand what "maintenance mode" would mean but #1821 explained, thanks

Issue #501, opened 2015-03-19
Gson fails to parse simple map


GoogleCodeExporter:
```
When parsing the following json with gson (using 2.2.2 from maven-central), 
gson (fromJson) throws an:

JsonSyntaxException: "duplicate key: null"

whereby the json is created by gson (toJson) itself.

Json:
{"appProperties":{"server.configuration.reset":"false","server.db.jndi":"testDB"
,"xml.basedata.list":"basedata.xml","server.configuration.cache":"false","xml.la
yout.list":"","server.rolemapper":"InsertionRoleMapper","xml.content.list":"data
.xml"}}

http://json.parser.online.fr/ validates the string as correct...
```

Original issue reported on code.google.com by `phil...@meisen.net` on 15 Feb 2013 at 12:04

busylee999:
I faced the same trouble. But I figured out it was because of we have map<enum, T> and ones  server started returns new enum type, we had not had yet, so it leaded to this error, two nulls as keys.

tylerjameshurst:
I'm serializing a map.. the json object only has one key in it, and it's throwing this exception. Not sure why.
Every call to the map's put is using putIfAbsent.
`"map": {
      "string_key": 0
    }`
`com.google.gson.JsonSyntaxException: duplicate key: string_key`

TheDoctorOne:
Since issue is still open, i'm guessing its still here.  Here is how i solved my case: 
I had an issue with Map<enum, list> like everyone else. After seeing this issue and debugging realized it was because of the enum. 
So I cast my Map<enum, list> to Map<Integer, list>. Then did my serialization / de-serialization.

Here is an example code, similar to mine. Changed variable names etc.
```java
public static Map<Enum, List<String>> fromOrdinal(Map<Integer, List<String>> ordMap) {
    Map<Enum, List<String>> org = new HashMap<>();
    for(int ord : ordMap.keySet()) {
	    org.put(Enum.values()[ord], ordMap.get(ord));
    }
    return org;
}

public static Map<Integer, List<String>> toOrdinal(Map<Enum, List<String>> map) {
    Map<Integer, List<String>> ord = new HashMap<>();
    for(Enum key : map.keySet()) {
	    ord.put(key.ordinal(), map.get(key));
    }
    return ord;
}
```

eamonnmcmanus:
I don't think there's enough information in this issue to be actionable. There may be more than one underlying issue, or there may be none and people are just not using the API right. (So there might be a documentation issue.)

[...]

Issue #1920, opened 2021-07-15
Gson Serializetion Bug in Enum Map


phooeny:
When the enum type is used as the key of one map,  if there is `@SerializedName("name")` at the declaration of the enum items, the Gson serialization and deserialization do NOT work well.

Following is the bad case

```java
import com.google.gson.annotations.SerializedName;

public enum RoomIdentifier {
    @SerializedName("MARKER_NAME")  // Here is the bad case
    ROOM_NAME;
}
```

Test code is here

```java
    public void testConvert() {
        RoomIdentifier id = RoomIdentifier.ROOM_NAME;
        
        Map<RoomIdentifier, String> slots = new HashMap<SlotIdentifier, String>();
        slots.put(RoomIdentifier.ROOM_NAME,"ROOM_NAME_TEST");
        
        String strSerialized = gson.toJson(slots);

        Type type = new TypeToken<Map<RoomIdentifier, String>>() {
        }.getType();
        Map<RoomIdentifier, String> slotsDeserialized = gson.fromJson(strSerialized, type);
        
        System.out.println(gson.toJson(id));   // "MARKER_NAME"   //As expect, SerializedName is used
        System.out.println(strSerialized);                // {"ROOM_NAME":"ROOM_NAME_TEST"}   // the key name is not SerializedName
        System.out.println(slotsDeserialized);            // {null=ROOM_NAME_TEST}    //The key is null
    }
```

**Root cause**

- In the serialization function `GSON::toJson`  ,  the **MapTypeAdapterFactory::Adapter** uses `String.valueOf(entry.getKey())` to generate the key.  `@SerializedName("name")` ("MARKER_NAME" in the example) is NOT used at all.  Here is Gson [code](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java#L207)
- But in the deserialization function `GSON::fromJson`  , **EnumTypeAdapter** is used to deserialize the key(here is  [code](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java)).  If one field is marked by  `@SerializedName("name")` ,  the original name string ("ROOM_NAME" in the example ) could NOT be recognized, `null` is returned.

Marcono1234:
Relates to #127, #1722

It is unfortunate that Gson uses `String.valueOf` for map keys, since this behavior is not obvious to the user and most likely not symmetric with deserialization, as shown here.

[...]

PR #1950, opened 2021-09-04
Parsing Map<Enum,Obj> - 501

TheDoctorOne:
Related to issue #501. 

Parsing "Map<Enum,Obj>" where enum's toString method is overridden and not returning the original name of the constant. 
Test code and backtrace of the issue: https://github.com/google/gson/issues/501#issuecomment-912810928

With the added implementation, it first checks for name, as it was doing before. And if the result does not match with any constant that Enum has, it also checks for the output of the "toString()" method and returns the result of it.


Marcono1234:
Relates to #1920, which is (partially) about `toString()` being used for map keys during serialization.

Chaostical:
## 
### 
# ****

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -803,7 +807,11 @@ public EnumTypeAdapter(final Class<T> classOfT) {
         in.nextNull();
         return null;
       }
-      return nameToConstant.get(in.nextString());
+      String key = in.nextString();
+      T constant = nameToConstant.get(key);
+      if (constant == null)
```
eamonnmcmanus, line 812: Google style requires a space between `if` and `(` (which I added) and also braces around the body. Alternatively, you could do this:

```java
return (constant == null) ? stringToConstant.get(key) : constant;
```

eamonnmcmanus (changes requested):
This seems reasonable. Just a small quibble about style.

Marcono1234:
Might be good to add a test for this, for example similar to what you have shown in https://github.com/google/gson/issues/501#issuecomment-912810928.
Possibly also verifying that constant names have higher precedence, e.g.:
```java
enum SwitchedToString {
  A("B"),
  B("A");

  private final String toString;

  SwitchedToString(String toString) {
    this.toString = toString;
  }

  @Override
  public String toString() {
    return toString;
  }
}
```

---

Note that I am not a member of this project so feel free to consider this only as suggestion.

TheDoctorOne (commented):
Didn't know that, thx for the heads up

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Sorry to quibble about this, but can you add braces around the body of the `if`, or rewrite using `?:`?

TheDoctorOne: Changed to `?:`

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks! I've run this change against the tests for all of Google's internal code and I didn't find any problems.

PR #1958, opened 2021-09-11
Add `GsonBuilder.disallowDuplicateProperties()`

### LINKED ISSUE #386: Fail if a JSON object has a duplicated property name
```
1. Instantiate a JSONReader(Reader in) where the specified input stream 
includes an object where a given object in the structure has a repeated 
property name.
  - This can be done directly, or indirectly via, say, Gson.fromJson()
2. Examine the object as parsed from the JSON.

The observed result is that of the multiple specifications of the value for a 
given property name, only the last actually ends up applying.

Per http://www.ietf.org/rfc/rfc4627.txt, there's some implementation 
flexibility on what the expected result should be:

2.2.  Objects

   An object structure is represented as a pair of curly brackets
   surrounding zero or more name/value pairs (or members).  A name is a
   string.  A single colon comes after each name, separating the name
   from the value.  A single comma separates a value from a following
   name.  The names within an object SHOULD be unique.

--

So the implementation doesn't seem to be _incorrect_ with respect to the RFC in 
how it handles this situation, which SHOULD not arise, when it happens. 
However, it would be nice if an issue with repeated names in the object caused 
an exception to be thrown or otherwise brought the issue to the user's 
attention. At least optionally. Silently letting the latest-specified version 
of a name win is a bit surprising.

The issue is particularly acute if the JSON data comes from an outside source 
not directly under the programmer's control.

v2_0_RC1, Linux
```

Original issue reported on code.google.com by `hil...@google.com` on 6 Dec 2011 at 4:55

Marcono1234:
Resolves  #386
Relates to #1884

Adds the method `GsonBuilder.disallowDuplicateProperties()` to disallow multiple properties with the same name. Security sensitive applications can use this to protect against malicious JSON data which tries to circumvent security checks.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -35,6 +35,7 @@ import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 
 import static com.google.gson.Gson.DEFAULT_COMPLEX_MAP_KEYS;
+import static com.google.gson.Gson.DEFAULT_DISALLOW_DUPLICATE_PROPERTIES;
 import static com.google.gson.Gson.DEFAULT_ESCAPE_HTML;
 import static com.google.gson.Gson.DEFAULT_JSON_NON_EXECUTABLE;
 import static com.google.gson.Gson.DEFAULT_LENIENT;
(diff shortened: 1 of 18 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -326,6 +329,26 @@ public GsonBuilder setFieldNamingStrategy(FieldNamingStrategy fieldNamingStrateg
     return this;
   }
 
+  /**
+   * Disallows duplicate JSON object properties during deserialization. When the JSON
+   * data contains multiple properties with the same name a {@link JsonSyntaxException}
+   * will be thrown. This affects the built-in adapters for {@code Object}, {@code Map}
```
Marcono1234, line 335: The `Calendar` adapter is not mentioned here (despite checking for duplicates as well) because it is most likely not that commonly used and I am not sure if its JSON format is specified anywhere in the first place.

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -127,11 +127,6 @@ public void testWritePropertyWithEmptyStringName() {
 
   }
 
-  public void testReadPropertyWithEmptyStringName() {
```
Marcono1234, line 130: Have moved this to `JsonParserTest`

Issue #1768, opened 2020-08-25
JsonReader / JsonTreeReader.promoteNameToValue() causes incorrect JSON path


Marcono1234:
The internal method `JsonReaderInternalAccess.promoteNameToValue(JsonReader)` which is used before the key of a `Map<K,V>` is read causes an incorrect JSON path because after the name has been read as value, the JSON path is not updated.

For example:
```java
String malformedJson = "{\"name\":\"value";
try {
    new Gson().fromJson(malformedJson, new TypeToken<Map<String, String>>() {}.getType());
} catch (JsonSyntaxException exception) {
    // Incorrect: Path is missing name
    // Unterminated string at line 1 column 15 path $.
    System.out.println(exception.getCause().getMessage());
}

JsonReader jsonReader = new JsonReader(new StringReader(malformedJson));
jsonReader.beginObject();
jsonReader.nextName();
try {
    jsonReader.nextString();
} catch (MalformedJsonException exception) {
    // Correct: Path includes name
    // Unterminated string at line 1 column 15 path $.name
    System.out.println(exception.getMessage());
}
```

Issue #1385, opened 2018-09-13
Deserisation fails with null numeric primitive keys


AudriusButkevicius:
**What did you do:**

Ran:
```
import com.google.gson.Gson;
import com.google.gson.GsonBuilder;

import java.util.HashMap;
import java.util.Map;

public class Test {
    static class MyType {
        public Map<Integer, Integer> map = new HashMap<>();
    }

    public static void main(String[] args) {
        MyType instance = new MyType();
        instance.map.put(null, 1);

        Gson serialiser = new GsonBuilder()
                .enableComplexMapKeySerialization()
                .serializeNulls()
                .setPrettyPrinting()
                .create();

        String data = serialiser.toJson(instance);
        System.out.println(data);

        try {
            MyType newInstance = serialiser.fromJson(data, MyType.class);
            System.out.println("Success");
        } catch (Exception e) {
            e.printStackTrace(System.out);
        }
    }
}
```

**What happened**

Deserialisation fails:

```
{
  "map": {
    "null": 1
  }
}
com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: For input string: "null"
	at com.google.gson.internal.bind.TypeAdapters$7.read(TypeAdapters.java:228)
	at com.google.gson.internal.bind.TypeAdapters$7.read(TypeAdapters.java:218)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:41)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:186)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:145)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
	at com.google.gson.Gson.fromJson(Gson.java:927)
	at com.google.gson.Gson.fromJson(Gson.java:892)
	at com.google.gson.Gson.fromJson(Gson.java:841)
	at com.google.gson.Gson.fromJson(Gson.java:813)
	at Test.main(Test.java:27)
Caused by: java.lang.NumberFormatException: For input string: "null"
[...]

PR #1961, opened 2021-09-12
Add TypeAdapter methods for custom property name conversion

### LINKED ISSUE #127: Should support custom serializers for map keys
```
This is actually a RFE for supporting custom serializers for map keys, the 
current implementation 
relies on result of String.valueOf for map keys, but if I've registered a 
custom serialization type 
adapter for the type of map key it (obviously) doesn't get called

create any class with toString method and register a JsonSerializer for it, try 
to serialize to JSON, 
you'll see the toString method beeing called instead of the registered 
serializer

a call to JsonSerializer instead to toString method, and result from the 
serializer in the JSON 
output

gson 1.4 on OSX Leopard 10.5.7

patch is attached, with a possible solution that calls into serialization 
context while honoring 
current implementation's behavior for keys without registered serializers, and 
a unit test for the 
specific situation

```

Original issue reported on code.google.com by `zregv...@gmail.com` on 31 May 2009 at 6:30

Attachments:
- [map_key.patch](https://storage.googleapis.com/google-code-attachments/google-gson/issue-127/comment-0/map_key.patch)

### LINKED ISSUE #1722: TypeAdapters don't apply to map keys
When converting a Map to Json, `TypeAdapters` are not being used to generate the keys. Here is a minimal repro:

```java
import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import com.google.gson.TypeAdapter;
import com.google.gson.stream.JsonReader;

import java.io.IOException;
import java.util.HashMap;
import java.util.Map;
import java.util.Objects;

class Main {

    /** An example class that just wraps a String. */
    static class Thinger {
        public final String key;

        Thinger(String key) {
            this.key = key;
        }

        @Override
        public String toString() {
            return "Thinger(" + key + ")";
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;
            Thinger that = (Thinger) o;
            return Objects.equals(key, that.key);
        }

        @Override
        public int hashCode() {
            return Objects.hash(key);
        }
    }

    /** A TypeAdapter that can encode and decode a Thinger */
    static class ThingerAdapter extends TypeAdapter<Thinger> {

        @Override
        public void write(JsonWriter out, Thinger value) throws IOException {
            out.value(value.key);
        }

        @Override
        public Thinger read(JsonReader in) throws IOException {
            return new Thinger(in.nextString());
        }
    }

        // Sample data with custom keys and values
        final Map<Thinger, Thinger> map = new HashMap<>();
        map.put(new Thinger("Foo"), new Thinger("Bar"));
        map.put(new Thinger("Baz"), new Thinger( "Qux"));

        // Register the adapter we defined above
        final Gson gson = new GsonBuilder().registerTypeAdapter(Thinger.class, new ThingerAdapter()).create();

        // Print the encoded JSON
        System.out.println(gson.toJson(map));
    }
}
```

I expect this to print:

```
{"Foo":"Bar","Baz":"Qux"}
```

But it actually calls the `toString` method and prints:

```
{"Thinger(Foo)":"Bar","Thinger(Baz)":"Qux"}
```

I'm not familiar with your code base, but it looks like the problem is related to [this line of code](https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java#L207)

Marcono1234:
Resolves #127
Relates to / resolves  #1722

Currently Gson uses `String.valueOf` for Map key serialization and the regular `TypeAdapter.read` method for deserialization. This is often not symmetric, preventing serialized data from being deserialized again, and is also not easily customizable by the user.

This pull request adds the `TypeAdapter` methods `createPropertyName(T)` and `readFromPropertyName(String)`. These allow the user to define how values should be converted to a property name and how it should be converted back. The default implementation of using `String.valueOf` was kept for backward compatibility.
These methods have the following advantages:
- Users can customize how conversion to property names should happen; and for classes where this makes no sense can also prevent it by throwing an exception
- Custom type adapter factories for `Map`-like types can uses these methods as well, having a well defined way to convert Map keys
- For these cases where Gson is currently not behaving well for Map keys (e.g. #1385 or #1920) users can at least provide their own type adapter which properly performs the conversion delegating to the existing adapter. An example for this can be seen in the added test method `MapTest.testCustomEnumMapKeyConversion()`.

It might have been ideal if Gson added such methods in the past already so only types for which conversion to property names makes sense would have opted-in into the conversion. The current opt-out situation (especially with `String.valueOf` usage) is error-prone because users convert values to property names for which this was never intended or considered.

The implementation of this pull request allowed removing the error-prone `promoteNameToValue` method (see also #1768). This comes at the cost that the default implementation of `readFromPropertyName` creates a new JsonTreeReader wrapping the name as JsonPrimitive. This is not as performant, loses information about whether the original JsonReader was lenient and about the JSON path. The lenient setting is probably not that important because it would have only been used when the name was read as non-finite double. The missing JSON path can be worked around by having the caller catch the exceptions and add additional context, such as the JSON path. In general it is recommended that adapters which do want to support property name conversion should override `readFromPropertyName` for better performance.
An alternative would have been to use the signature `readFromPropertyName(JsonReader)`; this would solve these issues and would also allow streaming reading of the name (as proposed by #1604; though most likely rarely / never needed), but would require  keeping `promoteNameToValue` and I think it would reduce usability because `readFromPropertyName` will only be called for names, so only `nextName()` (and a few other methods) can be called; all other ones would always throw an IllegalStateException due to token mismatch. The approach of only passing a `String` to the method seems less error-prone and easier to use to me.


gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -287,4 +292,105 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
+   *
+   * @param name the JSON property name; not {@code null}
+   * @return Java object (possibly {@code null}) parsed from the property name
+   * @throws UnsupportedOperationException If deserialization from a property
+   *        name is not supported by this adapter
+   * @throws JsonParseException If parsing the property name fails
+   */
+  // TODO: Throwing UnsupportedOperationException is not ideal because it 'bubbles up'
+  // too far the call stack, but Gson has no better fitting exception which also
+  // fits for createPropertyName
```
Marcono1234, line 355: Not sure about this one

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -287,4 +292,105 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
+   * <p>For serialization using Gson's built-in {@code Map} adapter, an alternative
+   * to overriding this method can be to use {@link GsonBuilder#enableComplexMapKeySerialization()}.
+   *
+   * @param value the Java object to serialize. May be null.
+   * @return The JSON property name. Must not be {@code null}.
+   * @throws UnsupportedOperationException If serialization to a property
+   *        name is not supported by this adapter
+   */
+  // TODO: Throwing UnsupportedOperationException is not ideal because it 'bubbles up'
+  // too far the call stack, but Gson has no better fitting exception
```
Marcono1234, line 322: Not sure about this one

gson/src/main/java/com/google/gson/TypeAdapter.java
Marcono1234, line 355: Not sure about this one

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -287,4 +292,105 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
+
+  static class PropertyNameJsonReader extends JsonTreeReader {
+    public PropertyNameJsonReader(String propertyName) {
+      super(new JsonPrimitive(propertyName));
+    }
+
+    @Override public String getPath() {
+      // Does not have access to actual path, so at least make it obvious that
+      // path is unknown
+      return "#fake-property-name-path";
```
Marcono1234, line 389: This path and especially the "fake" might be a bit irritating for users; maybe just use `<property-name>` or similar as path? That hopefully should make it obvious enough that this is not the actual name.

PR #1962, opened 2021-09-13
JsonWriter make indent accessible

Speiger:
Motivation for this change.
When using indents in the JsonWriter usually spam newlines for arrays.
I wrote a custom JsonWriter on top of this one which just implements a nicer Pretty Printing style.
Said Style disables newlines & indents when a Array is started, but enables them if a object is started within the array.
It uses a List<Boolean> to keep track of the blockage state.

The only thing that I couldn't override without adding a extra Method is the "indent". Which is set to final for reasons I assume,
but a getter should be present to aid people who want to improve their "Human Readable" JsonFile.

The idea is that you can seamlessly override the JsonWriter without the need of requiring extra functions to make original behavior work.

Thanks for reading and your time.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -219,7 +219,15 @@ public class JsonWriter implements Closeable, Flushable {
       this.separator = ": ";
     }
   }
-
+  
+  /**
+   * Returns the indentation string to be repeated for each level of indentation
+   * in the encoded document.
+   */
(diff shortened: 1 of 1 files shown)


Speiger:
Eh yeah not going to sign it:
Because google lets me not decide which information may or may not be shared publicly.
The main problem being, I would like to select which of my email addresses it uses. (Already tried alternative email address and it says both instead of select one)

If someone wants to take over this patch and submit it instead of me. Go ahead I would be happy enough.
(Technically these lines of code are not copyright-able anyways)

PR #1964, opened 2021-09-14
Improve Maven build

Marcono1234:
- Specify missing plugin versions
- Fix or suppress ProGuard notes and warnings

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -10,27 +10,25 @@
   <artifactId>gson</artifactId>
   <name>Gson</name>
 
+  <properties>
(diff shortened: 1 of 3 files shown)


gson/pom.xml
```diff
@@ -131,9 +132,15 @@
         </configuration>
         <dependencies>
           <dependency>
-            <groupId>net.sf.proguard</groupId>
+            <groupId>com.guardsquare</groupId>
```
Marcono1234, line 135: ProGuard seems to have changed its group ID

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks!

Issue #1794, opened 2020-10-08
Java 14/15 records can not set final field


IsmailMarmoush:
Hi, as the following error describes, java records (preview feature) deserialization isn't working in gson, Jackson is adding support in its very soon release 2.12, is there going to be same support/fix in gson ? 

`java.lang.AssertionError: AssertionError (GSON 2.8.6): java.lang.IllegalAccessException: Can not set final java.lang.String field JsonGsonRecordTest$Emp.name to java.lang.String`

Here's a sample test

```
  record Emp(String name) {}

  @Test
  void deserializeEngineer() {
    Gson j = new GsonBuilder().setPrettyPrinting().create();
    var empJson = """
            {
              "name": "bob"
            }""";
    var empObj = j.fromJson(empJson, Emp.class);
  }
```

Baizley:
The problem seems to exist only in Java 15 with preview enabled. Running your test under Java 14 with preview enabled and printing the `empObj` gives `Emp[name=bob]`.

JDK used 
1. openjdk-14.0.2_linux-x64_bin
2. openjdk-15_linux-x64_bin

GandalfTheWhite80:
The following Deserializer works fine with JDK16 (workaround)

```java
package de.sph.gsonrecordtest;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertNotSame;
import static org.junit.Assert.assertTrue;

import java.lang.reflect.Type;

import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import com.google.gson.JsonDeserializationContext;
import com.google.gson.JsonDeserializer;
import com.google.gson.JsonElement;
import com.google.gson.JsonObject;
import com.google.gson.JsonParseException;

import org.junit.jupiter.api.Test;


class RecordTest {
	
	public RecordTest()
	{
		//
	}

	@Test
	void testSerAndDeser() {
		// creating a new record
		final TestMe testMeRecord = new TestMe(42L, "Question");

		// Use a GsonBuilder
		final GsonBuilder gb = new GsonBuilder();
		// register our deserializer
		gb.registerTypeAdapter(TestMe.class, new TestMeDeserializer());

		// creating a gson object.
		final Gson gson = gb.create();

		// Test serialize
		final String str = gson.toJson(testMeRecord);
		assertEquals("{\"longValue\":42,\"stringValue\":\"Question\"}", str);

		// Test deserialize
[...]

PR #1968, opened 2021-09-17
Add Gson method for creating new instances (using `InstanceCreator`s)

### LINKED ISSUE #537: Adds Gson.buildFrom(Gson): exposes instance creator
```
Adds getter for ConstructorConstructor's internal instance creator Map.
```

Original issue reported on code.google.com by `dandre.a...@gmail.com` on 22 Oct 2013 at 11:44

Attachments:
- [ConstructorConstructor.java.patch](https://storage.googleapis.com/google-code-attachments/google-gson/issue-537/comment-0/ConstructorConstructor.java.patch)

Marcono1234:
Adds the method `Gson.createInstance(TypeToken)` exposing `ConstructorConstructor`'s functionality.
This allows custom TypeAdapterFactories to easily create new instances which can be useful for example for custom collection-like types or for custom reflection based adapters which want to support `@JsonAdapter`, see https://github.com/google/gson/issues/1794#issuecomment-919890214.

Might relate to or resolve #537, though from its description it is not completely clear what that issue was proposing and the attached patch does not seem to be available anymore.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1006,6 +1006,25 @@ public final class Gson {
     return (T) fromJson(new JsonTreeReader(json), typeOfT);
   }
 
+  /**
+   * Creates a new instance of the type using either one of the {@link InstanceCreator}s
+   * {@linkplain GsonBuilder#registerTypeAdapter(Type, Object) registered with the <code>GsonBuilder</code>}
+   * or a built-in creator.
+   *
+   * <p>This method is mainly intended for {@link TypeAdapterFactory} implementations of
+   * collection-like types which want to create an empty collection instance and then add
+   * the values read from JSON to it.
+   *
+   * @param <T> type to create
+   * @param type token representing the type to create
+   * @return the created instance
+   */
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1006,6 +1006,25 @@ private static void assertFullConsumption(Object obj, JsonReader reader) {
...
+   * <p>This method is mainly intended for {@link TypeAdapterFactory} implementations of
+   * collection-like types which want to create an empty collection instance and then add
+   * the values read from JSON to it.
+   *
+   * @param <T> type to create
+   * @param type token representing the type to create
+   * @return the created instance
+   */
+  // TODO: Specify thrown exception type; but ConstructorConstructor currently uses
+  // RuntimeException instead of specific exception
```
Marcono1234, line 1023: Not sure if this can be resolved in a reasonable way yet.
Ideally in the future Gson's exception hierarchy should be refactored. There are exception types missing, such as for this case or when serialization fails (Gson's exceptions all seem to be related to deserialization), which currently leads to usage of `RuntimeException`.

Issue #298, opened 2015-03-19
Allow serialization of anonymous and local classes


GoogleCodeExporter:
```
What steps will reproduce the problem?

        Gson gson = new Gson();
        String json = gson.toJson(new Object() {
            String url = "http://localhost:10080/live/list";
            String[] names = new String[] {"dsc0001","dsc0002","dsc0003"};
        });
        System.out.println(json);

What is the expected output?
{"url":"http://localhost:10080/live/list","name":[{"dsc0001","dsc0002","dsc0003"
]}

What do you see instead?
blank line

What version of the product are you using? On what operating system?
1.6

Please provide any additional information below.
I understand that there are default exclusion strategies that prevent the 
serialization of objects belonging to an anonymous or local class, but it would 
be great if these could be bypassed to allow the above notation.


```

Original issue reported on code.google.com by `robby...@gmail.com` on 17 Mar 2011 at 2:14

codeblooded:
I see this closed, but I don't see a reason.  Can we reopen or provide clarification?  It sounds like this thread seemed in overwhelming support for anonymous class serialization.

mintern:
This is especially problematic with Java 8. I've tried to create a TypeAdapterFactory for serializing Streams, but my `create` method never even gets called because `ReferencePipeline$2` is anonymous, so the Excluder just returns null.

lyubomyr-shaydariv:
@mintern

> This is especially problematic with Java 8. I've tried to create a TypeAdapterFactory for serializing Streams, but my `create` method never even gets called because `ReferencePipeline$2` is anonymous, so the Excluder just returns null.

It should work if you provide a type to the `toJson()` method:

```java
... = gson.toJson(stringStream, new TypeToken<Stream<String>>() {}.getType());
```

----

I think that this issue should be reopened, because it's possible to serialize and deserialize an anonymous/local class object, if the class does not have its outer class reference `this$0`, for example, if declared in a static method (don't think it has much practical use, though).

PR #1969, opened 2021-09-17
Don't exclude static local classes

Marcono1234:
Java 16 added Records, which can be declared locally but are always `static`. It also allows interfaces to be declared locally.

This pull request makes makes `Excluder` allow such static local classes. Despite Gson not having a built-in adapter for Record types yet, it at least makes it easier for users to write their own, see https://github.com/google/gson/issues/1794#issuecomment-803611447.
The reason why local classes were previously not permitted seems to be that they can capture enclosing references which cannot be serialized and deserialized properly, see #298. However for `static` classes this won't be an issue.

I have not added a test for this because Gson currently uses Java 6 to compile, so it would require adjustments to the Maven build to support compiling and running one test with Java 16. With Maven this is slightly more cumbersome than with Gradle because users need to manually set up a toolchain to specify a Java 16 JDK, whereas Gradle can handle that on its own.
But you can test this manually by running:
```java
public void test() {
  record R(int i) {}
  record C(R r) {}
  // Without the change this prints "null"
  System.out.println(new Gson().toJson(new C(new R(1))));
}
```

Relates to #1510

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/Excluder.java
+++ b/gson/src/main/java/com/google/gson/internal/Excluder.java
@@ -173,7 +173,7 @@ public final class Excluder implements TypeAdapterFactory, Cloneable {
       return true;
     }
 
-    if (isAnonymousOrLocal(field.getType())) {
+    if (isAnonymousOrNonStaticLocal(field.getType())) {
       return true;
     }
 
@@ -199,7 +199,7 @@ public final class Excluder implements TypeAdapterFactory, Cloneable {
           return true;
       }
 
-      if (isAnonymousOrLocal(clazz)) {
+      if (isAnonymousOrNonStaticLocal(clazz)) {
           return true;
       }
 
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/Excluder.java
```diff
@@ -199,7 +199,8 @@ private boolean excludeClassChecks(Class<?> clazz) {
           return true;
       }
 
-      if (isAnonymousOrLocal(clazz)) {
+      // Permit static local classes, such as Records (Java 16 feature)
+      if (!isStatic(clazz) && isAnonymousOrLocal(clazz)) {
```
eamonnmcmanus, line 203: I am wondering if it makes sense for `isAnonymousOrLocal` to become `isAnonymousOrLocalAndNotStatic`. In the other case where we call the method (`excludeField`), would we want to accept static local classes as well?

Marcono1234: Ah, you are right `excludeField` should not exclude fields with static local classes as type either.
Moving it into `isAnonymousOrLocal` sounds good. What about `isAnonymousOrNonStaticLocal` as name instead?

eamonnmcmanus: Sure, sounds good.

Marcono1234: Have force-pushed the change. It now in theory also allows static anonymous classes (to not make the boolean expression too nested), but I don't think such a thing is possible in Java at the moment.

PR #1974, opened 2021-09-24
Switch to GitHub Actions

### LINKED ISSUE #1971: Travis CI is not running / switch to GitHub Actions
It appears Travis CI has not been running for the latest commits, the last run was on September 1st, see [build history](https://travis-ci.org/github/google/gson/builds). Maybe this is related to the migration from travis-ci.org to travis-ci.com?

What do you think about switching to GitHub Actions instead? Its syntax is also pretty simple and the build commands for this project are not complex.
Should I create a pull request for that?

The only question is why Travis CI currently runs `mvn install` without tests and then `mvn test`:
https://github.com/google/gson/blob/aa5554e69a2d5ee7e557b3cc1f5d68275fd4c1ab/.travis.yml#L6-L7

We could keep that, but it might make sense to simplify this to just run `mvn install` in the root directory of this repository.

Marcono1234:
Resolves #1971


eamonnmcmanus:
Thanks!

Marcono1234:
I am not familiar with Travis, so maybe there is some work left to be done by you as maintainer to disable the Travis CI app for this repository and maybe also disable builds on travis-ci.com?

PR #1976, opened 2021-09-24
Improve TreeTypeAdapter thread-safety

Marcono1234:
Gson [claims to be thread-safe](https://github.com/google/gson/blob/3b7835a18b7ce999fe79ad12cd85014642f30968/gson/src/main/java/com/google/gson/Gson.java#L63) so `TreeTypeAdapter.delegate()` might be called by multiple threads. To guarantee that each thread sees a fully constructed `delegate`, the field has to be `volatile`.

Otherwise when a user-defined type adapter with non-`final` fields (even though it is stateless) is being used as delegate a thread might see an incomplete constructed instance. See [Safe Publication and Safe Initialization in Java](https://shipilev.net/blog/2014/safe-public-construction/#_correctness); the current case of `TreeTypeAdapter.delegate()` would be similar to the "Unsafe Local DCL & Unsafe Singleton" case shown in that blog post.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
@@ -47,7 +47,7 @@ public final class TreeTypeAdapter<T> extends TypeAdapter<T> {
   private final GsonContextImpl context = new GsonContextImpl();
 
   /** The delegate is lazily created because it may not be needed, and creating it may fail. */
-  private TypeAdapter<T> delegate;
+  private volatile TypeAdapter<T> delegate;
 
   public TreeTypeAdapter(JsonSerializer<T> serializer, JsonDeserializer<T> deserializer,
       Gson gson, TypeToken<T> typeToken, TypeAdapterFactory skipPast) {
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
```diff
@@ -83,6 +83,7 @@ public TreeTypeAdapter(JsonSerializer<T> serializer, JsonDeserializer<T> deseria
   }
 
   private TypeAdapter<T> delegate() {
+    // Allows racy initialization of `delegate` by multiple threads
```
eamonnmcmanus, line 86: More than one thread can observe `d == null` and therefore call `gson.getDelegateAdapter(...)`. That's _probably_ not a problem? Presumably multiple calls will return the same or at least an equivalent result, and one of them will end up remaining in `delegate`.

Marcono1234: Yes exactly, that is what I meant with the comment, to clarify that this case has been considered.
Maybe for some rare corner cases it might make a difference, but I assume for the majority of the use cases it does not make a difference.
However, if you want I can also replace this with double checked locking to be on the safe side.

eamonnmcmanus: I think it's fine as it stands but the comment could be a bit clearer. For example:

```java
// A race might lead to `delegate` being assigned by multiple threads but the last assignment will stick.
```

PR #1977, opened 2021-09-24
Add GitHub issue templates

Marcono1234:
Adds [GitHub issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). This helps making sure that bug reports contain all necessary information and automatically adds labels for bug reports and feature requests, making it easy to tell them apart.

This is a proof of concept implementation; please let me know what you think and if you want anything changed. You can also try them out here: https://github.com/Marcono1234/gson/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -16,9 +16,11 @@ assignees: ''
 
 
 # Used tools
-<!-- List relevant build tools here which might affect Gson -->
-- [ ] ProGuard
-- ...
+<!-- List relevant build tools and plugins with version number here which might affect Gson -->
+- [ ] Maven; version: 
+- [ ] Gradle; version: 
+- [ ] ProGuard (attach the configuration file please); version: 
+- [ ] ...
 
 # Description
 <!-- Describe the bug you experienced -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,47 @@
+---
+name: Bug report
+about: Report a Gson bug.
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+# Gson version
+<!-- Gson version you are using, for example '2.8.8' -->
+
+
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,47 @@
...
+<!-- Gson version you are using, for example '2.8.8' -->
+
+
+# Java / Android version
+<!-- Version of the Java or Android platform on which the bug occurred -->
+
+
+# Used tools
+<!-- List relevant build tools here which might affect Gson -->
+- [ ] ProGuard
```
eamonnmcmanus, line 20: Perhaps Maven and Gradle too? I wonder if it is worth asking for versions as well.

eamonnmcmanus (commented):
Thanks, this looks very helpful!

.github/ISSUE_TEMPLATE/bug_report.md
Marcono1234: Sounds like a good idea; Maven and Gradle themselves probably in most cases 'only' fetch the dependencies and build the project, but there could be cases where the build tool itself makes a difference.

What do you think about this change?
```markdown
# Used tools
<!-- List relevant build tools and plugins with version number here which might affect Gson -->
- [ ] Maven; version: 
- [ ] Gradle; version: 
- [ ] ...
```

eamonnmcmanus: That looks good to me.

Marcono1234:
@eamonnmcmanus, I just noticed that this project does not have a `feature-request` label (used by the issue template) yet.
Do you want to add one or should the issue template be changed to use the [`enhancement`](https://github.com/google/gson/labels/enhancement) label instead?

eamonnmcmanus:
I think it would make sense to use `enhancement` instead. Sorry I didn't notice that.

Marcono1234:
Have created #1982 for that now. It was also my fault for not noticing / considering this earlier, sorry.

PR #1986, opened 2021-10-11
Remove `Gson.excluder()` exposing internal `Excluder` class

### LINKED ISSUE #1903: `Gson.excluder()` exposes internal `Excluder`
The method `Gson.excluder()` exposes the internal `Excluder` class, see also Javadoc:
Javadoc screenshot

The method should be package-private.

Marcono1234:
Resolves #1903

Additionally adds documentation to some Gson methods which did not have documentation yet.


eamonnmcmanus:
Seems like [some people are using this](https://sourcegraph.com/search?q=context:global+Gson.excluder%28%29&patternType=literal) so I'm not sure we can remove it.

kaqqao:
Hi everyone. I'm chiming in as the maintainer of one of the projects that uses `excluder()`.

> It looks like most projects are interested in whether Gson is configured to ignore a certain class (`excludeClass`) or field (`excludeField`). 
This is exactly our situation. We're generating a GraphQL schema from Java types and, in order to decide what fields to expose, we're effectively interrogating Gson's configuration to make that decision.

> Maybe Gson should then expose (parts of) the Excluder class?
For what it's worth, this would suffice and sounds like a very fair approach to me.

Marcono1234:
@kaqqao, would you mind creating a new feature request issue for this (I don't think one exists yet)? That would probably make it easier to track this request.

PR #1987, opened 2021-10-11
Improve number strategy implementation

Marcono1234:
Follow-up for #1290 for some things I overlooked when I reviewed the changes a while ago.


gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -71,11 +71,11 @@
         try {
           Double d = Double.valueOf(value);
           if ((d.isInfinite() || d.isNaN()) && !in.isLenient()) {
-            throw new MalformedJsonException("JSON forbids NaN and infinities: " + d + in);
+            throw new MalformedJsonException("JSON forbids NaN and infinities: " + d + "; at path " + in.getPath());
```
Marcono1234, line 74: Uses `getPath()` instead of (implicit) `in.toString()` because the line and column number included by `toString()` would point behind the value (since it has already been consumed).

PR #1988, opened 2021-10-11
Some code suggestion from CodeGuru

XinyuLiu5566:
- Some code uses '%s' to format int: 'size' expression. Use %d, not %s, for integers. This ensures locale-sensitive formatting.
- Some lines of code contain a resource that might not have closed properly. closing the _write_ resource in a try-finally block.


eamonnmcmanus:
These changes make sense but we do need the CLA before we can merge.

XinyuLiu5566:
Hi sorry for the late reply, I just signed the CLA. Could you check that again?

PR #1990, opened 2021-10-11
Fix LongSerializationPolicy null handling being inconsistent with Gson

Marcono1234:
Gson does not actually use the specified `LongSerializationPolicy` but instead uses type adapters which emulate the behavior. However, previously Gson's implementation did not match LongSerializationPolicy regarding `null` handling.

Because it is rather unlikely that LongSerializationPolicy has been used on its own, this pull request adjusts its implementation to match Gson's behavior (instead of the other way around).


eamonnmcmanus:
Thanks! I checked that the tests of all of Google's internal projects that use Gson pass with this change.

PR #1991, opened 2021-10-11
Prevent Java deserialization of internal classes

Marcono1234:
Adversaries might be able to forge data which can be abused for DoS attacks.
These classes are already writing a replacement JDK object during serialization for a long time, so this change should not cause any issues.

I have not added tests for the prevented deserialization because it would require hardcoding the binary deserialization data which would decrease transparency of this project.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
+++ b/gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
@@ -15,6 +15,9 @@
  */
 package com.google.gson.internal;
 
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.io.ObjectStreamException;
 import java.math.BigDecimal;
 
@@ -75,6 +78,11 @@ public final class LazilyParsedNumber extends Number {
    */
   private Object writeReplace() throws ObjectStreamException {
     return new BigDecimal(value);
+  }
+
+  private void readObject(ObjectInputStream in) throws IOException {
+    // Don't permit directly deserializing this class; writeReplace() should have written a replacement
+    throw new InvalidObjectException("Deserialization is unsupported");
   }
 
   @Override
--- a/gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
+++ b/gson/src/main/java/com/google/gson/internal/LinkedHashTreeMap.java
@@ -17,6 +17,9 @@
 
 package com.google.gson.internal;
 
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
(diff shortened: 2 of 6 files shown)


Simon-Good:
> Adversaries might be able to forge data which can be abused for DoS attacks. These classes are already writing a replacement JDK object through `writeReplace()` during serialization for a long time (multiple years), so this change should not cause any issues.
> 
> I have not added tests for the prevented deserialization because it would require hardcoding the binary deserialization data which would decrease transparency of this project.

This commit is now related to CVE-2022-25647.There's no clue to analyse how this vulnerable   affect our products.How could adversaries forge data to make DoS attacks,is there any example? Why throw an exception could prevent these Dos attacks?

Marcono1234:
Disclaimer: The following is not an official statement; I am not a member of this project.

---

This only affects [Java Serialization](https://docs.oracle.com/en/java/javase/17/docs/specs/serialization/index.html), if neither your code nor your dependencies are not using it, then you are not affected by this.

I was not the one who filed / requested the CVE or the Snyk vulnerability entry and was not even aware of its existence until I saw your comment. I created this pull request as preventive action, but _not_ because I was aware of any specific existing exploit, nor have I investigated whether a denial of service attack is really possible. The description of this pull request was meant as hypothetical worst case consequence of the current code. (Though arguably I did not make that clear enough.)

However, to put this into perspective: Java Serialization is in general dangerous and should be avoided by applications. It is rather likely that other libraries have not paid close attention to sanitization during deserialization, and that even the JDK classes might be vulnerable by chaining multiple objects with recursive references together, allowing to cause a denial of service in some situations. But unless you (or your dependencies) use Java Serialization such serializable classes without proper prevention of denial of service attacks is not a big issue (remote code execution vulnerabilities caused by serializable classes is of course a different topic).

Therefore, in my personal opinion the current CVSS score of 7.3 with claimed high impact on availability and integrity is pretty exaggerated. Similarly the "Details" section of the [Snyk vulnerability entry](https://security.snyk.io/vuln/SNYK-JAVA-COMGOOGLECODEGSON-1730327) seems to contain a generic Java Serialization vulnerabilities text, without being specific to the situation of Gson. To my knowledge the old Gson code could for example not directly have been exploited for a remote code execution attack. (It could however be the case that Snyk has private knowledge about a possible exploit.)

---

@eamonnmcmanus, would be good if you could comment on this, and maybe provide an official statement here. Might also be worth to [contact Snyk](https://support.snyk.io/hc/en-us/requests/new) to clarify the situation and to find out how they determined the CVSS score.

eamonnmcmanus:
Thanks @Marcono1234 for this detailed analysis. I agree with it. For me the risk is pretty theoretical, but it's _conceivable_ that versions without this fix might be combined with other vulnerabilities elsewhere to produce a Denial of Service. Therefore I would recommend that people use version Gson 2.8.9 or any later version, _if_ serialization is or might be in use. I also agree that in general serialization should be avoided.

PR #1992, opened 2021-10-11
Delete unused LinkedHashTreeMap

Marcono1234:
Class seems to be unused since commit f29d5bc37b52c4b8d2ad15a10bb0c7f684c1d45d.
Gson currently only uses LinkedTreeMap.


eamonnmcmanus:
OK, given that the need for this class is obviated by DoS protection in the JDK and given that the class is `.internal.` I think we can probably delete it. However, as I mentioned in #1989, I'd like to make a Gson release before merging these potentially-breaking changes.

eamonnmcmanus:
2.8.9 has now been released, so as agreed I am merging this.

Issue #1981, opened 2021-09-30
Missing Constraint: Import-Package: sun.misc


JaroslavTulach:
# Gson version

The Apache NetBeans project has just tried to migrate from version `2.8.5` to version `2.8.8` in the pull request [PR-3200](https://github.com/apache/netbeans/pull/3200/files) and the update fails with unsatisfied OSGi missing constraint: `Import-Package: sun.misc`. 

# Java / Android version

NetBeans project builds & runs primarily on JDK8.

# Used tools

NetBeans is using home-made Ant based build system. I don't think it matters for this bug report.

# Description

As [PR-3200](https://github.com/apache/netbeans/pull/3200/files) states: The start of NetBeans fails with exception coming from an OSGi container (Apache NetBeans project is using OSGi behind the scene):
```
SEVERE [org.netbeans.modules.netbinox]: framework event com.google.gson type 2
org.osgi.framework.BundleException: The bundle "com.google.gson_2.8.8 [111]" could not be resolved. Reason: Missing Constraint: Import-Package: sun.misc; version="0.0.0"
	at org.eclipse.osgi.framework.internal.core.AbstractBundle.getResolverError(AbstractBundle.java:1332)
	at org.eclipse.osgi.framework.internal.core.AbstractBundle.getResolutionFailureException(AbstractBundle.java:1316)
	at org.eclipse.osgi.framework.internal.core.BundleHost.startWorker(BundleHost.java:323)
	at org.eclipse.osgi.framework.internal.core.AbstractBundle.resume(AbstractBundle.java:390)
	at org.eclipse.osgi.framework.internal.core.Framework.resumeBundle(Framework.java:1184)
	at org.eclipse.osgi.framework.internal.core.StartLevelManager.resumeBundles(StartLevelManager.java:559)
	at org.eclipse.osgi.framework.internal.core.StartLevelManager.resumeBundles(StartLevelManager.java:544)
	at org.eclipse.osgi.framework.internal.core.StartLevelManager.incFWSL(StartLevelManager.java:457)
	at org.eclipse.osgi.framework.internal.core.StartLevelManager.doSetStartLevel(StartLevelManager.java:243)
	at org.eclipse.osgi.framework.internal.core.EquinoxLauncher.internalStart(EquinoxLauncher.java:271)
	at org.eclipse.osgi.framework.internal.core.EquinoxLauncher$2.run(EquinoxLauncher.java:246)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1993, opened 2021-10-14
#1981: Avoid OSGi bundle's dependency on sun.misc package

JaroslavTulach:
As #1981 wrotes: The Apache NetBeans project has just tried to migrate from version `2.8.5` to version `2.8.8` in the pull request https://github.com/apache/netbeans/pull/3200 and the update fails with unsatisfied OSGi missing constraint: `Import-Package: sun.misc`. The best solution is to avoid such dependency.

The PR introduces a test that verifies `Import-Package` statement in GSON JAR's manifest no longer contains reference to `sun.misc`. Then this PR modifies `bnd` configuration file to explicitly avoid `sun.misc` package from list of imported packages.

### DIFF
--- a/gson/bnd.bnd
+++ b/gson/bnd.bnd
@@ -5,6 +5,7 @@ Bundle-Vendor: Google Gson Project
 Bundle-ContactAddress: ${project.parent.url}
 Bundle-RequiredExecutionEnvironment: JavaSE-1.6, JavaSE-1.7, JavaSE-1.8
 Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
+Import-Package: com.google.gson.*,!sun.misc
 
 -removeheaders: Private-Package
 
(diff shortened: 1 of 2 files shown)


JaroslavTulach:
@googlebot I signed it!

Marcono1234:
Have done some basic testing based on the setup described in https://www.baeldung.com/osgi with Apache Karaf by editing its `etc/config.properties` to exclude `sun.*` / `sun.misc` from `org.osgi.framework.bootdelegation` and `org.osgi.framework.system.packages.extra`.

| | Gson 2.8.8 | Pull request |
| - | :-: | :-: |
| bootdelegation + extra-package | ✔️ | ✔️ |
| extra-package | ✔️ | ✔️ |
| bootdelegation | ❌ (fails due to missing requirement for `sun.misc`)  | ✔️ |
| extension-bundle | ✔️ | ✔️ |
| - | ❌ (fails due to missing requirement for `sun.misc`) | ⚠️ (bundle starts, but fails during runtime because no custom Gson `InstanceCreator` is registered and `Unsafe` is not available; as expected) |

Key:
- `bootdelegation`: `sun.*` set as `org.osgi.framework.bootdelegation` value
- `extra-package`: `sun.misc` set as `org.osgi.framework.system.packages.extra` value
- `extension-bundle`: extension bundle from https://github.com/diffplug/osgiX being installed

Note that I have not tested how Karaf interacts with the Java Platform Module System.

PR #2006, opened 2021-10-31
Update project version in README

Marcono1234:


eamonnmcmanus:
Thanks, I should have remembered that. There are a couple of other places to update too, which I'll take care of.

Issue #2007, opened 2021-11-02
`codegen` project is incomplete / non-functional


Marcono1234:
It looks like the `codegen` project is incomplete. The only thing it does is write an empty class, see:
https://github.com/google/gson/blob/a92bbf849c99b1fc708eb3dd209af9c72110ed4d/codegen/src/main/java/com/google/gson/codegen/GeneratedTypeAdapterProcessor.java#L56-L58

For example, the output for the `Order` test class is just:
```java
package com.google.gson.codegen.functional;
final class Order$TypeAdapter {
}
```

Should the project be removed, or should its functionality be implemented eventually?

eamonnmcmanus:
I think I would be in favour of deleting it. The code is 9 years old. It seems very half-baked, has no documentation, and has what appears to be a copy of a very old version of JavaPoet. There _might_ be interest in some sort of code-generating solution, but if so I think it would be better built with modern tools, including the latest JavaPoet.

Perhaps @inder123 wants to weigh in, though?

inder123:
proto has users. May be best to leave it as is?

eamonnmcmanus:
Yes, sorry, the question was about the codegen project, not the proto project.

NBXXF:
you can use https://github.com/NBXXF/gson_plugin

PR #2008, opened 2021-11-02
Convert codegen, metrics and proto to Maven submodules

Marcono1234:
Converts the `codegen`, `metrics` and `proto` projects to Maven submodules.
This does not necessary add much value for these projects (see #2007, and note that `metrics` has to be run manually), but it at least cleans up these projects and makes sure they compile / build.

I have marked this pull request as draft for now; please let me know what you think.


pom.xml
```diff
@@ -65,35 +68,6 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.8.1</version>
-          <executions>
```
Marcono1234, line 68: Have moved this to the `gson` submodule.

metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
```diff
@@ -16,14 +16,6 @@
 
 package com.google.gson.metrics;
 
-import com.google.caliper.Param;
```
eamonnmcmanus, line 19: Google's style rules concerning imports have changed over time, but I think this file follows the current guidelines now and doesn't after these changes. The imports should be in lexicographical order.

eamonnmcmanus (changes requested):
I think this makes sense. As discussed in #2007, we may just delete the `codegen` project.

metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
Marcono1234: Sorry, have fixed this now. I was using Eclipse's save actions and had imported the Google Style Guide formatter settings, but had not adjusted the additional Eclipse settings as described in https://github.com/google/styleguide/issues/273#issuecomment-457843194.

Would it make sense to add the [spotless Maven plugin](https://github.com/diffplug/spotless/tree/main/plugin-maven#google-java-format) to enforce the style guide, or was your concern here mainly that the formatting of an existing file was changed?

eamonnmcmanus: My main concern about doing that would be that we would have to reformat everything to conform, which would invalidate most outstanding PRs. (I have the same concern about updating the tests to JUnit4 + Truth.)

eamonnmcmanus (approved):

metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
Marcono1234: Ok, might be best to not apply such a plugin then at the moment.

Issue #1859, opened 2021-02-10
Version 2.8.6 JDK 7 jar export error


wansleynery:
Good morning,

The version 2.8.6 compiles the jar with module-info with JDK 9, at least. It looks wrong when compared with the compatibility of another versions.

![image](https://user-images.githubusercontent.com/42624198/107504403-004a8d00-6b7a-11eb-8c39-ad2093c895d0.png)
![image](https://user-images.githubusercontent.com/42624198/107503954-5ff46880-6b79-11eb-894e-37b98066a85d.png)
![image](https://user-images.githubusercontent.com/42624198/107504814-9a123a00-6b7a-11eb-9f3a-ae11dc799031.png)
![image](https://user-images.githubusercontent.com/42624198/107504598-54ee0800-6b7a-11eb-8262-4867f0c0b69a.png)

Marcono1234:
Similar to  #1850 and #1627

Marcono1234:
Can you please check if this issue still occurs with the latest Gson version? The way the `module-info.class` is included in the Gson JAR was changed and should hopefully now cause fewer issues. Also make sure to use the latest build tools and plugins (e.g. Gradle).

Marcono1234:
I am going to close this because no answer has been provided. If you still experience this with the latest Gson version, feel free to comment here again with more details.

Issue #1850, opened 2021-01-19
Can't process class [module-info.class] (Unsupported class version number [53.0] (maximum 52.0, Java 1.8))


raminhbbAU:
Hey There,

I'm facing with a trouble while compiling my android project with gson-2.8.6.jar. The following error has appeared:

Warning:Exception while processing task java.io.IOException: Can't read [D:\Parsic Source\ParsiLab\ParsiLab\app\libs\gson-2.8.6.jar(;;;;;;**.class)] (Can't process class [module-info.class] (Unsupported class version number [53.0] (maximum 52.0, Java 1.8)))

Any help would be appreciated
RaminHbb

Marcono1234:
Probably related to #1627

Marcono1234:
Can you please check if this issue still occurs with the latest Gson version? The way the `module-info.class` is included in the Gson JAR was changed and should hopefully now cause fewer issues. Also make sure to use the latest build tools and plugins (e.g. Gradle).

Marcono1234:
I am going to close this because no answer has been provided. If you still experience this with the latest Gson version, feel free to comment here again with more details.

Issue #1827, opened 2020-11-25
Android desugaring error with 2.8.6 


ShiremanSun:
when use android.enableD8.desugaring=false

errors Caused by: java.lang.IllegalArgumentException: Type without superclass: module-info

Error while running desugar 
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.android.build.gradle.internal.transforms.DesugarWorkerItem$DesugarAction.run(DesugarWorkerItem.java:94)
	at org.gradle.workers.internal.AdapterWorkAction.execute(AdapterWorkAction.java:50)
	at org.gradle.workers.internal.DefaultWorkerServer.execute(DefaultWorkerServer.java:50)
	at org.gradle.workers.internal.AbstractClassLoaderWorker$1.create(AbstractClassLoaderWorker.java:47)
	at org.gradle.workers.internal.AbstractClassLoaderWorker$1.create(AbstractClassLoaderWorker.java:37)
	at org.gradle.internal.classloader.ClassLoaderUtils.executeInClassloader(ClassLoaderUtils.java:98)
	at org.gradle.workers.internal.AbstractClassLoaderWorker.executeInClassLoader(AbstractClassLoaderWorker.java:37)
	at org.gradle.workers.internal.IsolatedClassloaderWorker.execute(IsolatedClassloaderWorker.java:49)
	at org.gradle.workers.internal.WorkerDaemonServer.execute(WorkerDaemonServer.java:84)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.process.internal.worker.request.WorkerAction$1.call(WorkerAction.java:129)
	at org.gradle.process.internal.worker.child.WorkerLogEventListener.withWorkerLoggingProtocol(WorkerLogEventListener.java:41)
[...]

Issue #1641, opened 2020-01-16
Android R8 issue with version 2.8.6


shubhambansal:
I started getting the following error after updating to version 2.8.6.  Until 2.8.5 everything works fine.

```
> Task :app:transformClassesAndResourcesWithR8ForBetaRelease FAILED
R8 is the new Android code shrinker. If you experience any issues, please file a bug at
https://issuetracker.google.com, using 'Shrinker (R8)' as component name. You can
disable R8 by updating gradle.properties with 'android.enableR8=false'.

.../transforms/FirebasePerformancePlugin/beta/release/66/module-info.class:
D8: Unsupported source file type 
```

`gradle.properties configuration`
android.enableR8.fullMode=true

elirex:
I have the same issue when used Android Gradle plugin v3.4.0 and Gradle v5.1.1 to build Apk. My workaround is Android Gradle plugin upgrades to 3.5.3 and Gradle upgrades to v5.5 then succeed to build Apk. But in build log still seen

```
Can't instrument module-info.class
java.lang.RuntimeException
        at org.objectweb.asm.ClassVisitor.visitModule(ClassVisitor.java:148)
        at org.objectweb.asm.ClassReader.readModule(ClassReader.java:731)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:632)
        at com.google.firebase.perf.plugin.instrumentation.Instrument.instrument(Instrument.java:151)
        at com.google.firebase.perf.plugin.instrumentation.Instrument.instrumentClassesInJar(Instrument.java:100)
...
```

Marcono1234:
Can you please check if this issue still occurs with the latest Gson version? The way the `module-info.class` is included in the Gson JAR was changed and should hopefully now cause fewer issues. Also make sure to use the latest build tools and plugins (e.g. Gradle).

Marcono1234:
I am going to close this because no answer has been provided. If you still experience this with the latest Gson version, feel free to comment here again with more details.

PR #2013, opened 2021-11-06
Put `module-info.class` into Multi-Release JAR folder

Marcono1234:
Places the `module-info.class` into the version specific folder `META-INF/versions/9`. This should hopefully fix issues with build tools for older JDK versions (or Android) which expect only Java 6 compatible classes in the JAR:
- #1627
- Maybe #1630
- Maybe #1641
- #1827
- #1850
- #1859

If users still encounter issues afterwards, they probably have to update their build tool, see for example https://github.com/google/gson/issues/1627#issuecomment-688921218.

When I tested the newly created JAR manually, Java was able to find the module descriptor file in the Multi-Release JAR without issues.


gson/bnd.bnd
```diff
@@ -17,3 +17,9 @@ Import-Package: sun.misc;resolution:=optional, *
     com.google.gson.annotations,\
     com.google.gson.reflect,\
     com.google.gson.stream
+
+# Ignore error about module-info.class being in wrong directory
+# See related https://issues.apache.org/jira/browse/FELIX-5592 (but that is for different plugin)
+# and https://github.com/bndtools/bnd/issues/2227
+-fixupmessages:\
+    "Classes found in the wrong directory: \\{META-INF/versions/9/module-info\\.class=module-info\\}";is:=warning
```
Marcono1234, line 25: If I understand https://github.com/bndtools/bnd/issues/2227 correctly, OSGi won't be able to load classes from `META-INF/versions`. But for `module-info.class` this might not matter because the OSGi manifest specifies the required packages.
The only issue might be packages which are since Java 9 part of a different optional module (e.g. `java.sql`), however for them maybe we should adjust `bnd.bnd` and specify these packages (in case that ever comes up as issue).

pom.xml
```diff
@@ -70,30 +70,32 @@
...
-                <excludes>
-                  <exclude>module-info.java</exclude>
-                </excludes>
+                <jdkToolchain>
+                  <version>9</version>
+                </jdkToolchain>
+                <release>9</release>
+                <includes>
+                  <include>module-info.java</include>
+                </includes>
```
Marcono1234, line 94: Surprisingly this works exactly as desired:
- It sees the other source files (this is necessary for Java to check if packages referenced in `module-info` exist)
- It only compiles `module-info`

pom.xml
```diff
@@ -104,12 +106,15 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-jar-plugin</artifactId>
           <version>3.2.0</version>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.felix</groupId>
-          <artifactId>maven-bundle-plugin</artifactId>
-          <version>5.1.2</version>
-          <inherited>true</inherited>
```
Marcono1234, line 112: Have removed this because it looks like this is not used (unless I am overlooking something). The `gson` Maven submodule is using the `bnd-maven-plugin`.

pom.xml
```diff
@@ -68,23 +68,45 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.8.1</version>
+          <configuration>
+            <release>${javaVersion}</release>
```
Marcono1234, line 72: This change caused all Maven modules and test classes to be build using Java 1.6, which is why I removed `@SafeVarargs` from two test classes and adjusted `ProtoTypeAdapter` because `Map.putIfAbsent` was added in Java 8 while `ConcurrentMap.putIfAbsent` already existed in Java 1.6.

pom.xml
```diff
@@ -68,23 +68,45 @@
...
+            <jdkToolchain>
+              <!-- Require at most JDK 11 because it is the last version 
+                supporting compiling Java 1.6, see https://bugs.openjdk.java.net/browse/JDK-8028563 -->
+              <!-- When Gson increases lowest supported Java version, can 
+                change this to an open version range (e.g. `[11,)`) -->
+              <version>[9,11]</version>
```
Marcono1234, line 78: Not completely sure why this works for the GitHub CI workflow (or rather why it is ignored?). For Surefire plugin it caused build failures, see https://github.com/Marcono1234/gson/runs/4146407481.

eamonnmcmanus:
This is great.

Concerning Java 6, I think we could reasonably consider dropping support for that platform now. It has been 10 years since Java 7. So maybe we can drop Java 6, separately, and undo the changes in this PR that were required to support it.

Marcono1234:
Have resolved the merge conflict. Is there anything else you want to have changed?

pom.xml
```diff
@@ -28,7 +28,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <java.version>1.6</java.version>
+    <javaVersion>6</javaVersion>
```
eamonnmcmanus, line 31: Would it work if you set this to 7? If so then I think it is probably time for a separate PR to fix #2018 by changing the target version to 7 everywhere.

Marcono1234: Do you mean to work around the fuzzer exception? That can probably rather be solved by adjusting the fuzzer script to use `-DjavaVersion=8` instead (which should hopefully overwrite this property here then).

Changing this property to 7 would effectively change the complete build to use Java 7.

eamonnmcmanus: Sorry, I was unclear. I was indeed suggesting that if changing the complete build to use Java 7 fixes this problem, then that's another reason to do it now. So if the CIFuzz check passes with that change, we can make a separate PR for just the Java version change, close #2018, and proceed with this PR.

Marcono1234: The fuzzer seems to run successfully now.

PR #2014, opened 2021-11-06
Fix warnings

Marcono1234:
Note: The visibility changes in the test classes are not strictly necessary. Eclipse IDE complained that the exposed types had lower visibility than the exposing field.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -35,7 +35,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
-          <includePackageNames>com.google.gson</includePackageNames>
           <excludePackageNames>com.google.gson.internal:com.google.gson.internal.bind</excludePackageNames>
           <links>
             <link>https://docs.oracle.com/javase/6/docs/api/</link>
--- a/gson/src/test/java/com/google/gson/InnerClassExclusionStrategyTest.java
+++ b/gson/src/test/java/com/google/gson/InnerClassExclusionStrategyTest.java
@@ -26,8 +26,8 @@ import junit.framework.TestCase;
  * @author Joel Leitch
  */
 public class InnerClassExclusionStrategyTest extends TestCase {
-  public InnerClass innerClass = new InnerClass();
-  public StaticNestedClass staticNestedClass = new StaticNestedClass();
+  private InnerClass innerClass = new InnerClass();
+  private StaticNestedClass staticNestedClass = new StaticNestedClass();
   private Excluder excluder = Excluder.DEFAULT.disableInnerClassSerialization();
 
   public void testExcludeInnerClassObject() throws Exception {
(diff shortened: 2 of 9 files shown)


gson/pom.xml
```diff
@@ -35,7 +35,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
-          <includePackageNames>com.google.gson</includePackageNames>
```
Marcono1234, line 38: Was detected by Eclipse; the [`maven-javadoc-plugin`](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html) does not have a parameter with this name (anymore?).

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -303,6 +303,7 @@ public GsonBuilder newBuilder() {
    * @deprecated This method by accident exposes an internal Gson class; it might be removed in a
    * future version.
    */
+  @SuppressWarnings("exports")
```
Marcono1234, line 306: Eclipse showed a warning here because `Excluder` is not exported by the module descriptor (which is why we deprecated this method).
`"exports"` seems to be an Eclipse specific warning name.

gson/src/test/java/com/google/gson/InnerClassExclusionStrategyTest.java
```diff
@@ -26,8 +26,8 @@
  * @author Joel Leitch
  */
 public class InnerClassExclusionStrategyTest extends TestCase {
-  public InnerClass innerClass = new InnerClass();
```
eamonnmcmanus, line 29: I'm not sure this change is great. Previously the test showed that this field was excluded, _despite_ being public. Now it's still excluded, but maybe that's because the field is private. It might be better to add comments saying why the fields are public.

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -92,6 +92,7 @@ public void testAddingStringProperties() throws Exception {
     assertEquals(value, jsonElement.getAsString());
   }
 
+  @SuppressWarnings("deprecation")
```
eamonnmcmanus, line 95: This is for `jsonElement.getAsCharacter()`, right? Perhaps we could do the same thing of putting it in a local and putting the `@SuppressWarnings` on the declaration of the local?

eamonnmcmanus (changes requested):
Thanks! This is good for hygiene.

gson/src/test/java/com/google/gson/InnerClassExclusionStrategyTest.java
Marcono1234: Good point! I have reverted the changes for this test class.

eamonnmcmanus (approved):

PR #2019, opened 2021-11-10
Update Caliper dependency; disable automatic result upload

Marcono1234:
Supersedes #2015

Note that I am not very familiar with Caliper, so please let me know if there is anything I should change.


metrics/src/main/java/com/google/gson/metrics/ParseBenchmark.java
```diff
@@ -116,25 +120,39 @@ public void timeParse(int reps) throws Exception {
...
     }
+    File file = new File(url.toURI());
+    if (!file.isFile()) {
+      throw new IllegalArgumentException("Resource " + path + " is not a file");
+    }
+    return file;
+  }
+
+  private static String resourceToString(String fileName) throws Exception {
+    ZipFile zipFile = new ZipFile(getResourceFile("/ParseBenchmarkData.zip"));
```
Marcono1234, line 136: Directly reads from the ZIP file; previously it appears user would need to have unzipped it manually.

eamonnmcmanus:
Thanks! I think Caliper support is pretty much best-effort at this stage but it's definitely better to have a more recent version.

Issue #1102, opened 2017-06-14
Support for `Optional`


prdoyle:
Usually, when I use Gson, I need to add special logic so that a field of `Optional<X>` gets serialized as `X` if it's present, and gets omitted if it's empty; and conversely on deserialization.  This seems like the only reasonable way to handle `Optional`.  Would you consider offering this feature (or accepting a PR that implements it)?

JakeWharton:
Gson can't depend on Java 8 types as it supports down to Java 6.

sspies8684:
I would really like to see this.

kaqqao:
@JakeWharton So? It could always be in a separate module.

stevenschlansker:
It looks like such a module exists: https://github.com/hsiafan/gson-java8-datatype
The errors emitted if you forget to include it are very confusing and do not obviously lead you to determining that you must install this module, but once you do, `Optional` types work fine.

marcblanchet:
really would like to see this. Coming from Swift, the use of Optional with JSON is very useful.

Marcono1234:
> Gson can't depend on Java 8 types as it supports down to Java 6.

With [Multi-Release JARs](https://openjdk.java.net/jeps/238) this would now be possible, at least for JRE >= 9.

mleonhard:
I need this.

chakpongchung:
I need this

kaqqao:
@mleonhard @chakpongchung Did you guys read the thread? What you want is already there, use it: https://github.com/hsiafan/gson-java8-datatype

piegamesde:
@kaqqao gson-java8-datatype worked great, but sadly it broke with Java 16 and isn't maintained anymore (the linked repository is archived).

> Gson can't depend on Java 8 types as it supports down to Java 6.

Java 6 is now end of life for 8(!) years, and Java 8 (and thus, Optional) exist for 7 years now. I think it's about time to move on.

alx741:
I agree that keeping that kind of retro-compatibility over including new useful features for this long might be unreasonable. Java 8 has been around for a long time now

prdoyle:
What is the value of supporting Java versions past their end of life?

eamonnmcmanus:
[...]

Issue #1059, opened 2017-04-07
GSON Serialization/Deserialization of Java 8 Data API


reinaldomjr:
When serializing / deserializing an object with the Date and Time classes of the new java.time API, the results are diferent from expected.

For excample, here's a java.util.Date serialized with gson (using version 2.8.0):

**"created": "2017-04-07T18:07:00",**

and here's a java.time.LocalDate:

**"expiration": {
    "year": 2017,
    "month": 4,
    "day": 7
  }**
the expected result for javatime api was a ISO-8601.

JakeWharton:
This is due to some historical baggage where Gson will happily serialize any object inside the platform package (`java.*`). Ideally Gson would throw an error when you did this, but alas, we can't make that change. Also Gson's built-in support of `java.util.Date` is also historical baggage we cannot remove or change.

The correct solution here would be to register your own type adapter for the `java.time.*` types so that they show up as ISO 8601 when serialized and deserialization parses them appropriately.

A quick search yields https://github.com/gkopff/gson-javatime-serialisers which you may just be able to drop in, but otherwise building a `TypeAdapter` yourself for these shouldn't be too bad.

kkolman:
a) still it would be great to have an "official" gson plugin/package with Java 8 types support
b) a note in https://github.com/google/gson/blob/master/UserGuide.md with the explanation about targetting Java 6 would helpe otherwise

jugimaster:
Dates come in from JSON as string values, right?

I keep seeing examples of registering adapters for specific Java classes like ZonedDateTime etc, but how would Gson know that you want to convert a particular incoming _string_ value into a ZonedDateTime specifically?

joffrey-bion:
@jugimaster Because you tell Gson the expected class when deserializing:

```java
SomeClass obj = gson.fromJson(jsonString, SomeClass.class);
```

jugimaster:
> Because you tell Gson the expected class when deserializing

But I'm not using JavaBeans/"domain model classes" - I just have Maps, and I was calling Gson from Clojure too.

Can strings be converted into Dates when deserializing into a Map?

joffrey-bion:
[...]

PR #2022, opened 2021-11-18
Optional data type

nadiawoodninja:
Resolves the enhancement request to add Support for java.util.Optional.  [issue 1102](https://github.com/google/gson/issues/1102)

### DIFF
--- a/codegen/pom.xml
+++ b/codegen/pom.xml
@@ -82,8 +82,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.5.1</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.8</source>
+          <target>1.8</target>
           <compilerArgument>-proc:none</compilerArgument>
         </configuration>
       </plugin>
--- a/gson/build.gradle
+++ b/gson/build.gradle
@@ -4,8 +4,8 @@ apply plugin: 'maven'
 group = 'com.google.code.gson'
 version = '2.8.6-SNAPSHOT'
 
-sourceCompatibility = 1.6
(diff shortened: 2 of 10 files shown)


eamonnmcmanus (changes requested):
Thanks!

This PR does two things:
1. Update the build to Java 8. I think we are indeed ready to do that, but it should be in a PR of its own. Do you want to make that?
2. Add support for serializing `Optional`. I'm not 100% convinced that this is useful. I note that the `Optional` class is not serializable with Java serialization (does not implement `java.io.Serializable`) for example. Do you have a need for this?

Also, the PR includes generated javadoc files, which it should not.

Marcono1234:
> I'm not 100% convinced that this is useful. I note that the Optional class is not serializable with Java serialization (does not implement java.io.Serializable) for example. Do you have a need for this?

If I may jump in here; the number of 'thumbs up' on #1102 and the number of upvotes on the [corresponding StackOverflow question](https://stackoverflow.com/questions/12161366/how-to-serialize-optionalt-classes-with-gson) suggest that there is some need for this.

GeenDutchman:
With Java7 dropped as of Gson 2.12.0, are there still blockers on this?

Marcono1234:
> are there still blockers on this?

Probably that Gson currently supports Android API level 21 as minimum (see README), but `Optional` was added in API level 24. And maybe backward compatibility concerns where currently users (accidentally?) rely on the reflection-based serialization format for `Optional` (which depends on Android / JDK implementation details).

But as mentioned in #1059, maybe some opt-in way or a separate "java-8-adapters" (or similar) Maven module could work?

(These are just my personal thoughts on this, the direct project members might think differently about it.)

PR #2027, opened 2021-11-26
cifuzz: add integration

DavidKorczynski:
Add cifuzz integration so oss-fuzz fuzzers are run 600 secs each PR

This adds CI integration with CIFuzz. CIFuzz will run the gson OSS-Fuzz fuzzers for a short period of time for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,25 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Thanks for doing this! I approved running the checks, and it appears that the new one doesn't work. Hopefully it will not be too hard to fix now that you are not working blind.

DavidKorczynski:
Thanks @eamonnmcmanus - It looks like it's a CIFuzz issue that happens across java projects. I have filed an issue here: https://github.com/google/oss-fuzz/issues/6913
Will ping you here when it has been sorted!

DavidKorczynski:
I think this should be working now, could we try to retrigger the CI?

eamonnmcmanus:
Yup, looks good. Thanks again!

Vichugo1995 (commented):
Positive

Issue #1994, opened 2021-10-15
Integer overflow when calling `fromJson()` on `JsonObject`


mroloux:
# Gson version
2.8.8

# Java / Android version
11

# Description
I'm using `fromJson()` to convert between a `JsonObject` and a POJO. When the `JsonObject` has a `long` primitive property, and I convert to an `int` on the POJO, an overflow occurs.

On the other hand, when I convert from string to the POJO, an exception gets thrown (as expected).

## Expected behavior
Exception when putting a `long` primitive value in an `int` property in `fromJson()`.

## Actual behavior
No exception, but incorrectly set `int` property.

# Reproduction steps

```
package domain.drawing;

import com.google.gson.Gson;
import com.google.gson.JsonObject;

public class MyClass {

    public int foo;

    public static void main(String[] args) {
        JsonObject jsonObject = new JsonObject();
        jsonObject.addProperty("foo", 10000000000L);
        System.out.println("-- fromJson on JsonObject: ");
        System.out.println(new Gson().fromJson(jsonObject, MyClass.class));

        System.out.println("");
        System.out.println("-- fromJson on string: ");
        System.out.println(new Gson().fromJson("{foo: 10000000000}", MyClass.class));
    }

    @Override
    public String toString() {
        return Integer.toString(foo);
    }
}
```

Running this main method outputs:

```
-- fromJson on JsonObject: 
1410065408

-- fromJson on string: 
Exception in thread "main" com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: Expected an int but was 10000000000 at line 1 column 18 path $.foo
	at com.google.gson.internal.bind.TypeAdapters$7.read(TypeAdapters.java:231)
	at com.google.gson.internal.bind.TypeAdapters$7.read(TypeAdapters.java:221)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
	at com.google.gson.Gson.fromJson(Gson.java:932)
	at com.google.gson.Gson.fromJson(Gson.java:897)
	at com.google.gson.Gson.fromJson(Gson.java:846)
	at com.google.gson.Gson.fromJson(Gson.java:817)
	at domain.drawing.MyClass.main(MyClass.java:18)
[...]

PR #2031, opened 2021-12-04
Issue 1994

dctrue2:
Added validation checks for parsing integers to ensure they fit within each data type's defined range; added corresponding tests.


NatalieSty:
@eamonnmcmanus I just signed the CLA, can you re-run please? Thanks.

Marcono1234:
Your changes look good, thanks! Could you please:
1. Add `@throws` to the documentation of the `JsonElement` methods to mention that they throw an exception for lossy conversion
2. Maybe consider throwing an exception other than `NumberFormatException`. That class is [described as](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/NumberFormatException.html):
    > Thrown to indicate that the application has attempted to convert a string to one of the numeric types

    That does not seem very fitting because there is no conversion from string when the overflow check is done. Maybe throwing `java.lang.ArithmeticException` would be better. That one is also used by JDK methods such as [`Math.addExact`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Math.html#addExact(int,int)) when overflow occurs.
    Though on the other hand the `JsonReader` methods say that they throw `NumberFormatException`, so that would require catching and wrapping the `ArithmeticException`. Maybe we should wait here for the comment of a maintainer.
3. It might be good to retrieve the `short`, `int`, ... value from the parsed `BigInteger` you are using for the range check (and also use it for the exception messages) instead of parsing the value again in the `return` statement at the end.
4. It could also be worth trying to avoid the overhead of always converting to String then parsing as BigInteger. When `value` is a boxed primitive, `BigInteger` or `BigDecimal` its value could be tested directly (without parsing). E.g. `getAsInt()` could check if `value` is `Short`, `Integer` (for those `intValue()` can be returned without overflow check) or `Long` (here an overflow check would be needed). But this might be to some extent premature optimization, so we should wait for a comment from a maintainer as well.
5. Add a test which tests the use case shown in #1994? All your tests directly interact with the `JsonPrimitive`, but it would be good to verify that `JsonTreeReader` also uses your changed methods. E.g. by calling `gson.fromJson(new JsonPrimitive(...), int.class)` and verifying that it detects the overflow.
6. Maybe move the new test methods you added to the class `JsonPrimitiveTest` instead because they are not really parsing anything.

One issue with this approach might be lossless conversion from floating point values. Because your code uses `BigInteger` any decimal numbers will result in a `NumberFormatException`. Maybe it would therefore be better to parse as `BigDecimal`. Or there might also be better solutions to this. What do you think?

Sorry for this long comment; I hope some of the points are helpful though.

Also as side note: GitHub understands certain [keywords](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue). Could you please edit the description of your pull request and add the following, so GitHub understands that this pull request closes that issue:
> Fixes #1994

---

Note that I am not a member of this project. Feel free to consider my comment only as suggestion.

eamonnmcmanus (changes requested):
Thanks for taking this on!

Though this is potentially a breaking change, it seems like the right thing to do, both because it fixes potential silent data loss and because it is consistent with what we do elsewhere (as detailed in the issue).

The new tests are great, but I had a few suggestions for the implementation changes.

eamonnmcmanus:
I've run this change against Google's internal tests and unfortunately there were quite a few failures. It looks as if there might be about 10 distinct root causes, which fall into two main categories:
1. Reading floating-point values into `int` fields. In the failures I saw, the floating-point values were things like `45.0`, so these actually do the right thing currently and will fail with this change.
2. Genuine overflows, sometimes because the value is big enough to fit into an unsigned `int` or `long` but not a signed one.

Although we could imagine tweaks to allow these cases to continue to work, or perhaps to update the tests so they don't trigger the problems, my feeling is that it is probably best to leave well enough alone. So I am probably going to close this PR and the associated issue.

dctrue2:
Sounds good @eamonnmcmanus and @Marcono1234! We greatly appreciate your feedback. Clearly, building in the necessary checks to make all possible test cases resolve as expected is a bit more laborious and expensive than at first glance from the initial issue posting.

Marcono1234:
> 2\. Genuine overflows, sometimes because the value is big enough to fit into an unsigned `int` or `long` but not a signed one.

Just curious, are those tests running directly against the `JsonPrimitive` methods modified here, or are they using `Gson.fromJson` with a `JsonElement` (or subtype) as input? In case of the latter, then at least personally I think these tests are somewhat questionable because they explicitly depend on the discrepancy between parsing from a `Reader` (respectively `JsonReader`, where unsigned values are not supported) vs. parsing from a `JsonElement` (respectively `JsonTreeReader`).
In case of the former, then this could be solved by adding new <code>getAs<i>X</i>Exact</code> methods, which could even be internal at first and only be used by `JsonTreeReader`. However, the question is whether that would be desired.

PR #2039, opened 2021-12-26
Add READMEs to Maven modules

Marcono1234:
Adds `README` files to the Maven modules to make it easier for new contributors to understand the project structure.

### DIFF
--- a/examples/android-proguard-example/README.md
+++ b/examples/android-proguard-example/README.md
@@ -1,8 +1,8 @@
 # android-proguard-example
 
 Example Android project showing how to properly configure [ProGuard](https://www.guardsquare.com/proguard).
-ProGuard is a tool for 'shrinking' and obfuscating compiled classes, it removes methods and fields if
-they appear to be unused or renames them. This can cause issues for Gson which uses Java reflection to
+ProGuard is a tool for 'shrinking' and obfuscating compiled classes. It can rename methods and fields,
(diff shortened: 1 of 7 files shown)


examples/android-proguard-example/README.md
```diff
@@ -0,0 +1,9 @@
+# android-proguard-example
+
+Example Android project showing how to properly configure [ProGuard](https://www.guardsquare.com/proguard).
+ProGuard is a tool for 'shrinking' and obfuscating compiled classes, it removes methods and fields if
```
eamonnmcmanus, line 4: This sentence could perhaps be improved. What about this?

> ProGuard is a tool for 'shrinking' and obfuscating compiled classes.
> It can rename methods and fields, or remove them if they appear to be unused.

gson/README.md
```diff
@@ -0,0 +1,4 @@
+# gson
```
eamonnmcmanus, line 1: I agree that the original `gson/README` file is not particularly helpful, since it is essentially a truncated form of the `README.md` file from the parent directory.

eamonnmcmanus (commented):
This is very helpful, thanks! Just one small comment.

examples/android-proguard-example/README.md
Marcono1234: That sounds better, thanks!

PR #2040, opened 2021-12-26
Fix Javadoc warnings and errors

Marcono1234:
Fixes warnings and errors reported by the `javadoc` tool.

### DIFF
--- a/extras/src/main/java/com/google/gson/interceptors/Intercept.java
+++ b/extras/src/main/java/com/google/gson/interceptors/Intercept.java
@@ -28,8 +28,8 @@ import java.lang.annotation.Target;
  * after it has been deserialized from Json.
  * Here is an example of how this annotation is used:
  * <p>Here is an example of how this annotation is used:
- * <p><pre>
- * &#64Intercept(postDeserialize=UserValidator.class)
+ * <pre>
+ * &#64;Intercept(postDeserialize=UserValidator.class)
  * public class User {
  *   String name;
  *   String password;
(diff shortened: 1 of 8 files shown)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -548,7 +548,7 @@ static void checkValidFloatingPoint(double value) {
    *  read or written.
    * @param skipPast The type adapter factory that needs to be skipped while searching for
    *   a matching type adapter. In most cases, you should just pass <i>this</i> (the type adapter
-   *   factory from where {@link #getDelegateAdapter} method is being invoked).
+   *   factory from where {@code getDelegateAdapter} method is being invoked).
```
Marcono1234, line 551: Changed this to `code` because it refers to the current method itself, so having a `link` is redundant and might confuse readers (making them think the link refers to a different method).

extras/src/main/java/com/google/gson/interceptors/Intercept.java
```diff
@@ -28,8 +28,8 @@
  * after it has been deserialized from Json.
  * Here is an example of how this annotation is used:
  * <p>Here is an example of how this annotation is used:
- * <p><pre>
```
Marcono1234, line 31: Changed these because `<pre>` is already a block tag and therefore `javadoc` complains that it is nested inside the `<p>` block tag.

eamonnmcmanus:
Thanks, this is great!

PR #2043, opened 2021-12-30
Change target Java version to 7

### LINKED ISSUE #2018: Drop support for Java 6
Continuing to support Java 6 causes considerable headaches. Building with `-source 6 -target 6` is not supported in JDK versions after 11. I don't think we need to support further back than Java 7 even for the most ancient Android devices still in use.

Marcono1234:
Resolves #2018

Also tries to add some documentation to the README regarding Java version and JPMS module requirements. I am not sure if the formatting and wording is ideal; feedback is appreciated.

Follow-up task:
Remove the Java version overwrite (and comment) in https://github.com/google/oss-fuzz/blob/master/projects/gson/build.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,25 @@ Maven:
 
 ![Build Status](https://github.com/google/gson/actions/workflows/build.yml/badge.svg)
 
+### Requirements
+#### Java version
+- Gson 2.9.0 and newer: Java 7
+- Gson 2.8.9 and older: Java 6
+
+Despite supporting older Java versions, Gson also provides a JPMS module descriptor (module name `com.google.gson`) for users of Java 9 or newer.
+
+#### JPMS dependencies (Java 9+)
+These are the optional Java Platform Module System (JPMS) JDK modules which Gson depends on.
+This only applies when running Java 9 or newer.
+
+- `java.sql` (optional since Gson 2.8.9)  
+When this module is present, Gson provides default adapters for some SQL date and time classes.
+
+- `jdk.unsupported`, respectively class `sun.misc.Unsafe` (optional)  
+When this module is present, Gson can use the `Unsafe` class to create instances of classes without no-args constructor.
+However, care should be taken when relying on this. `Unsafe` is not available in all environments and its usage has some pitfalls,
+see [`GsonBuilder.disableJdkUnsafe()`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/GsonBuilder.html#disableJdkUnsafe()).
+
 ### Documentation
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -34,6 +34,24 @@ Maven:
...
+#### JPMS dependencies (Java 9+)
+These are the optional Java Platform Module System (JPMS) JDK modules which Gson depends on.
+This only applies when running Java 9 or newer.
+
+- `java.sql` (optional since Gson 2.8.9)  
+When this module is present, Gson provides default adapters for some SQL date and time classes.
+
+- `jdk.unsupported`, respectively class `sun.misc.Unsafe` (optional)  
+When this module is present, Gson can use the `Unsafe` class to create instances of classes without no-args constructor.
+However, care should be taken when relying on this. `Unsafe` is not available in all environments and its usage has some pitfalls.
```
Marcono1234, line 53: Once #1904 is merged, this can also be adjusted to refer to the new `GsonBuilder.disableJdkUnsafe()` method.

codegen/README.md
```diff
@@ -1,4 +1,4 @@
-# gson-codegen
```
Marcono1234, line 1: Was an oversight by me; all other modules are using the folder name (instead of the artifact ID).

gson/build.gradle
```diff
@@ -4,8 +4,8 @@ apply plugin: 'maven'
 group = 'com.google.code.gson'
 version = '2.8.6-SNAPSHOT'
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
```
Marcono1234, line 8: Probably does not matter because the Gradle build is outdated anyway, see #1896.

gson/src/test/java/com/google/gson/internal/LinkedTreeMapTest.java
```diff
@@ -161,7 +161,7 @@ public void testJavaSerialization() throws IOException, ClassNotFoundException {
   }
 
   @SafeVarargs
-  private <T> void assertIterationOrder(Iterable<T> actual, T... expected) {
+  private final <T> void assertIterationOrder(Iterable<T> actual, T... expected) {
```
Marcono1234, line 164: Interestingly when changing the version to Java 7 this caused a build failure (due to [private methods not supporting it in Java 7](https://bugs.openjdk.java.net/browse/JDK-8047159)), but when the version was Java 6 this compiled fine. Most likely because `@SafeVarargs` was added in Java 7, but ([unlike `--release`](https://stackoverflow.com/a/43103038)) the `-source` and `-target` compiler flags do not prevent using newer API.

eamonnmcmanus:
Thanks! Alea jacta est.

PR #2048, opened 2022-01-05
Switch to JUnit 5

### LINKED ISSUE #1896: `build.gradle` project version is outdated

Marcono1234:
Resolves #1896 (see below)

Changes:
- Remove Gradle build support
The Gradle build setup is outdated, and trying to keep two setups (Maven and Gradle) up to date is error-prone. See also #1896.
Personally I think it would be useful to switch completely to Gradle at some point because its features such as [toolchains](https://docs.gradle.org/current/userguide/toolchains.html) make it easier to create a reliable build setup.
- Switch to JUnit 5
  - Reduce test class and method visibility to package-private (because JUnit 5 supports that). This is less verbose and also reduces unrelated test classes showing up in code completion while writing tests.
  - Consistently use `@Disabled` to disable tests (though maybe I missed some). Previously prefixes such as `disabled_` were used to disable test methods of `TestCase` subclasses.
  - Use `@BeforeEach` and `@AfterEach` for setup and teardown methods
  - Switch some method arguments because JUnit 4 has the assertion _message_ as first argument while JUnit 5 has it as last
  - Use Java 11 for compiling and running tests; IDE support for that is limited, see below

This pull request is 'complete', but I have marked it as Draft for now because there are some open issues:
- The `proto` module uses Google Truth which has a transitive dependency on JUnit 4; we therefore need to be careful to not accidentially mix JUnit 4 and JUnit 5 usage in the `proto` module
- Some IDEs (at least Eclipse IDE) only seem to support one Java version for the complete project. They do not understand that main and test sources should be compiled with different Java versions. JUnit 5 requires at least Java 8. This causes spurious IDE compiler errors, but running the tests might work nonetheless. However, it makes using newer Java features, such as lambdas for JUnit 5's `assertThrows` more annoying. And these spurious IDE compiler errors might be irritating for new contributors. Due to this I have also not replaced usage of `try { ... fail(); } catch (...) { }` constructs with `assertThrows` yet.
Note that Maven itself does not seem to have a problem with this.

Because this pull request removes a few test methods (see review comments), marks some as test but disables them with `@Disabled` and because some test methods are converted to parameterized tests, the number of executed tests changed slightly. However, it looks like all tests are still executed:
```diff
-[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.functional.ConcurrencyTest
+[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.functional.ConcurrencyTest

-[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.functional.CustomTypeAdaptersTest
+[WARNING] Tests run: 21, Failures: 0, Errors: 0, Skipped: 2 in com.google.gson.functional.CustomTypeAdaptersTest

-[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.functional.MapAsArrayTypeAdapterTest
+[WARNING] Tests run: 6, Failures: 0, Errors: 0, Skipped: 1 in com.google.gson.functional.MapAsArrayTypeAdapterTest

-[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.GsonBuilderTest
+[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.GsonBuilderTest

-[INFO] Running com.google.gson.metrics.PerformanceTest
-[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.metrics.PerformanceTest

-[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.MixedStreamTest
+[WARNING] Tests run: 13, Failures: 0, Errors: 0, Skipped: 1 in com.google.gson.MixedStreamTest

-[WARNING] Tests run: 20, Failures: 0, Errors: 0, Skipped: 1 in com.google.gson.stream.JsonReaderPathTest
+[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.stream.JsonReaderPathTest

-[INFO] Tests run: 121, Failures: 0, Errors: 0, Skipped: 0 in com.google.gson.stream.JsonReaderTest
+[WARNING] Tests run: 125, Failures: 0, Errors: 0, Skipped: 4 in com.google.gson.stream.JsonReaderTest
 ```
 (difference to 4ec67c00a08348bff6abe1a4a285f8a78483ee37)

If this pull request is merged, it is important to pay attention that new or existing PRs properly use `@Test` because merely naming a method `test...` will not suffice anymore.

No worries if you don't want to merge this pull request or if you want to wait until (eventually) the target Java version of this project is Java 8 or newer.

### CODE REVIEW

File: gson/pom.xml [hunk 1]
```diff
@@ -50,7 +50,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>3.0.0-M5</version>
```
Marcono1234, line 53: Is managed by parent now.

gson/src/main/java/com/google/gson/internal/JavaVersion.java
```diff
@@ -37,7 +37,7 @@ static int getMajorJavaVersion(String javaVersion) {
       version = extractBeginningInt(javaVersion);
     }
     if (version == -1) {
-      return 6;  // Choose minimum supported JDK version as default
+      return 7;  // Choose minimum supported JDK version as default
```
Marcono1234, line 40: Missed that for #2043, but this change currently has no effect anyway because Gson only uses `JavaVersion.isJava9OrLater()`.

gson/src/test/java/com/google/gson/GsonBuilderTest.java
```diff
@@ -16,53 +16,68 @@
...
-  public void testExcludeFieldsWithModifiers() {
+  @Test
+  void testExcludeFieldsWithModifiers() {
     Gson gson = new GsonBuilder()
         .excludeFieldsWithModifiers(Modifier.VOLATILE, Modifier.PRIVATE)
         .create();
     assertEquals("{\"d\":\"d\"}", gson.toJson(new HasModifiers()));
   }
 
-  public void testRegisterTypeAdapterForCoreType() {
```
Marcono1234, line 55: Previously this test was not performing any assertion (and was only testing that no exception is thrown?).
Have refactored it; see below.

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -123,27 +136,30 @@ public void testPropertyWithQuotes() {
...
   }
 
-  public void testEqualsNonEmptyObject() {
+  @Test
+  void testEqualsNonEmptyObject() {
     JsonObject a = new JsonObject();
     JsonObject b = new JsonObject();
 
-    assertEquals(a, a);
+    assertTrue(a.equals(a));
```
Marcono1234, line 162: Changed this because `assertEquals` should not be used to test `equals` implementation (it might return fast without calling `equals` at all).

gson/src/test/java/com/google/gson/functional/ConcurrencyTest.java
```diff
@@ -15,62 +15,43 @@
...
+  @BeforeEach
+  void setUp() throws Exception {
     gson = new Gson();
   }
 
   /**
    * Source-code based on
    * http://groups.google.com/group/google-gson/browse_thread/thread/563bb51ee2495081
    */
-  public void testSingleThreadSerialization() {
```
Marcono1234, line 46: Removed these two test methods because they were not performing any assertion. And there are (probably) enough tests already performing basic reflection-based serialization and deserialization.

gson/src/test/java/com/google/gson/functional/MapTest.java
```diff
@@ -443,8 +485,8 @@ public void testSerializeMaps() {
     assertEquals("{\"a\":12,\"b\":null,\"c\":{\"test\":1,\"TestStringArray\":[\"one\",\"two\"]}}",
         new GsonBuilder().serializeNulls().create().toJson(map));
     assertEquals("{\n  \"a\": 12,\n  \"b\": null,\n  \"c\": "
-  		+ "{\n    \"test\": 1,\n    \"TestStringArray\": "
-  		+ "[\n      \"one\",\n      \"two\"\n    ]\n  }\n}",
+        + "{\n    \"test\": 1,\n    \"TestStringArray\": "
+        + "[\n      \"one\",\n      \"two\"\n    ]\n  }\n}",
```
Marcono1234, line 489: Was mixing tabs and spaces.

gson/src/test/java/com/google/gson/internal/bind/RecursiveTypesResolveTest.java
```diff
@@ -46,38 +46,42 @@
...
-
-  public void testRecursiveResolveSimple() {
+  @Test
+  void testRecursiveResolveSimple() {
     @SuppressWarnings("rawtypes")
     TypeAdapter<Foo1> adapter = new Gson().getAdapter(Foo1.class);
     assertNotNull(adapter);
   }
 
-  /**
```
Marcono1234, line 56: Changed this here and a few lines below because they are not Javadoc comments for a specific method.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -561,24 +603,27 @@ public void testLongLargerThanMinLongThatWrapsAround() throws IOException {
...
   /**
    * Issue 1053, negative zero.
    * @throws Exception
    */
-  public void testNegativeZero() throws Exception {
-	  	JsonReader reader = new JsonReader(reader("[-0]"));
-	    reader.setLenient(false);
-	    reader.beginArray();
-	    assertEquals(NUMBER, reader.peek());
-	    assertEquals("-0", reader.nextString());
```
Marcono1234, line 574: Was using tabs and spaces for indentation.

pom.xml
```diff
@@ -68,13 +70,39 @@
...
+          </executions>
         </plugin>
+
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>3.0.0-M5</version>
+          <!-- Note: Ideally this would specify `jdkToolchain` to make sure 
+            a JDK >= ${testJavaVersion} is used, but unlike other plugins maven-surefire-plugin 
+            seems to fail when no Maven toolchains are configured locally -->
```
Marcono1234, line 103: The reason why this works for the compiler plugin is most likely a "quirk" (or bug?) because it simply [ignores any exception](https://github.com/apache/maven-compiler-plugin/blob/6335382f97aad2e69b9f9a0a4af3cf801cd29273/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java#L1472-L1476) when the toolchain lookup fails.
Maybe we run into the same problem there as well once they fix that.

eamonnmcmanus:
I think it would make sense to switch the Gradle removal to a separate PR. I agree with the idea; we clearly haven't been maintaining the Gradle build and it's better to have nothing than something that misleads people into wasting time trying to make it work. It's conceivable that we would switch the build to Gradle at some point but given that Gson is in maintenance mode that might be hard to justify.

Concerning the switch to JUnit 5, internally Google is still on JUnit 4 with no concrete plans to migrate to JUnit 5. So I don't think we can take that change. We'd certainly be interested in a change upgrading the JUnit 3 tests to JUnit 4. Though that would lead to a lot of churn, the improvement is enough to justify it.

Marcono1234:
Ok, I am closing this pull request then for now. Maybe it becomes relevant in the future again, or it might serve as basis for converting JUnit 3 tests to JUnit 4 tests, in case someone wants to do that.

PR #2051, opened 2022-01-11
Add more `Gson` default constants to be used by `GsonBuilder`

Marcono1234:
This hopefully makes it a bit easier to verify that the default values are in sync between `new GsonBuilder().create()` and `new Gson()`.


eamonnmcmanus:
LGTM

PR #2062, opened 2022-01-23
Improve `JsonReader.skipValue()`

### LINKED ISSUE #605: JsonReader skipValue() can cause an infinite loop
```
2. run reader.beginObject() to start reading the stream

3. run reader.skipValue() repeatedly until you try and skip over the "}" to end 
the json stream.  the last skip will cause an infinite loop.


I would expect an exception to be thrown indicating EOF was hit, or that you 
cannot skip over an object/array closing.  Even silently ignoring would be 
better then an infinite loop.



2.3 w/ Java 1.7.0_67



JsonReader line 1249 decrements the count to -1 because PEEKED_END_OBJECT is 
encountered on the last skip, but since the next item is EOF, it forever 
infinite loops waiting for the counter to get incremented back to 0, which 
never occurs.

I put in a quick hack for my problem by adding an extra if condition

} else if (p == PEEKED_EOF) {
    count = 0;
}

Probably a better fix would be to check for count>0 instead of !=0 (assuming 
that doesn't cause a problem somewhere else, but I am not familiar enough with 
the code base to try this).

This code snippet can reproduce the problem:

   public void testEarlyEOF() throws Exception {
        String json = "{ \"errors\":[ { \"field\":null,\"code\":\"loanId-not-found\" } ] }";

        final InputStream is = new ByteArrayInputStream(json.getBytes());

        Thread t = new Thread() {
            @Override
            public void run() {
                JsonReader reader = new JsonReader(new InputStreamReader(is));

                try {
                    reader.beginObject();     // start of overall json object
                    reader.skipValue();       // skip "errors" label
                    reader.skipValue();       // skip array value for "errors"
                    reader.skipValue();       // try and skip the closing "}"
                } catch (Exception e) {
                    e.printStackTrace();
                }
            }
        };

        t.setDaemon(true);
        t.start();

        t.join(10000);
        boolean alive = t.isAlive();
        if (alive) t.interrupt();
        assertFalse("should have finished by now, we must be hung", alive);
    }



```

Original issue reported on code.google.com by `rocco.ga...@gmail.com` on 3 Nov 2014 at 2:15

Marcono1234:
Fixes #605

See commit messages.


gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -269,14 +270,18 @@ JsonElement nextJsonElement() throws IOException {
...
+    JsonToken peeked = peek();
+    if (peeked == JsonToken.NAME) {
       nextName();
-      pathNames[stackSize - 2] = "null";
+      pathNames[stackSize - 2] = "<skipped>";
+    } else if (peeked == JsonToken.END_ARRAY) {
+      endArray();
+    } else if (peeked == JsonToken.END_OBJECT) {
+      endObject();
+    } else if (peeked == JsonToken.END_DOCUMENT) {
```
eamonnmcmanus, line 281: I'm going to suggest doing nothing here rather than throwing an exception.

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -269,14 +270,18 @@ JsonElement nextJsonElement() throws IOException {
   }
 
   @Override public void skipValue() throws IOException {
-    if (peek() == JsonToken.NAME) {
+    JsonToken peeked = peek();
+    if (peeked == JsonToken.NAME) {
```
eamonnmcmanus, line 274: Can we use a `switch` statement here?

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1255,22 +1262,45 @@ public void skipValue() throws IOException {
...
         count--;
       } else if (p == PEEKED_END_OBJECT) {
+        // Only update when object end is explicitly skipped, otherwise stack is not updated anyways
+        if (count == 0) {
+          pathNames[stackSize - 1] = null; // Free the last path name so that it can be garbage collected
+        }
         stackSize--;
         count--;
-      } else if (p == PEEKED_UNQUOTED_NAME || p == PEEKED_UNQUOTED) {
+      } else if (p == PEEKED_UNQUOTED) {
```
eamonnmcmanus, line 1271: While you are here, could you change this to a `switch` statement too? Maybe the JIT compiler is smart enough to avoid having to do a sequence of up to 12 comparisons, but we could avoid relying on that.

eamonnmcmanus:
This is great. Thanks for these improvements!

I ran this past all of Google's tests and found one that failed. Because of what is arguably a logic bug, it is calling `JsonReader.skipValue()` when `peek()` reports `END_DOCUMENT`. In outline:

```java
    switch (reader.peek()) {
      case BEGIN_ARRAY:
        return handleArray(reader);
      case BEGIN_OBJECT:
        return handleObject();
      case END_ARRAY:
        reader.endArray();
        break;
      case END_OBJECT:
        reader.endObject();
        break;
      default: // All other values can be skipped.
        reader.skipValue();
    }
```

This now fails with `IllegalStateException`. I think that means that we should have skipValue() do nothing in this case rather than throwing an exception, even if an exception is what we would probably have chosen if we were designing this from scratch. And of course we should document this.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1255,22 +1262,45 @@ public void skipValue() throws IOException {
...
+      } else if (p == PEEKED_DOUBLE_QUOTED_NAME) {
+        skipQuotedValue('"');
+        // Only update when name is explicitly skipped, otherwise stack is not updated anyways
+        if (count == 0) {
+          pathNames[stackSize - 1] = "<skipped>";
+        }
       } else if (p == PEEKED_NUMBER) {
         pos += peekedNumberLength;
+      } else if (p == PEEKED_EOF) {
+        throw new IllegalStateException("Cannot skip at end of document");
```
eamonnmcmanus, line 1298: Do nothing here too. (Except maybe `return`?)

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1233,9 +1232,19 @@ public int nextInt() throws IOException {
   }
 
   /**
-   * Skips the next value recursively. If it is an object or array, all nested
-   * elements are skipped. This method is intended for use when the JSON token
-   * stream contains unrecognized or unhandled values.
+   * Skips the next value recursively. This method is intended for use when
+   * the JSON token stream contains unrecognized or unhandled values.
+   *
+   * <p>The behavior depends on the type of the next JSON token:
```
Marcono1234, line 1238: I hope the wording here and below is fine; please let me know if you think it could be improved.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1505,7 +1551,7 @@ private String getPath(boolean usePreviousPath) {
    *   <li>For JSON arrays the path points to the index of the previous element.<br>
    *   If no element has been consumed yet it uses the index 0 (even if there are no elements).</li>
    *   <li>For JSON objects the path points to the last property, or to the current
-   *   property if its value has not been consumed yet.</li>
+   *   property if its name has already been consumed.</li>
```
Marcono1234, line 1554: The previous documentation (introduced in my PR #2000) might have been a bit misleading because whether or not the value has been consumed does not make a difference. (It might only affect which property you consider the "last" and the "previous".)

PR #1554, opened 2019-07-17
Fixed gradle build.

bufistov:
Signed-off-by: Dmitry Bufistov <dmitry@midokura.com>

PR #2063, opened 2022-01-23
Remove Gradle build support

### LINKED ISSUE #1896: `build.gradle` project version is outdated

Marcono1234:
Resolves #1896
Renders #1554 obsolete

Build script was outdated and not actively maintained anymore.
This was split off from #2048 (see also comments there).


eamonnmcmanus:
Thanks! This is a bit sad, but in the current maintenance-mode state of things I don't think we can afford to support two build systems and require maintainers to keep both working.

PR #2065, opened 2022-01-25
Add visitor support for traversing json tree.

jaiswalsk:
This PR adds the visitor support for traversing the JSON tree. See below example of custom visitor.

**CustomJsonElementVisitor**
```
public class CustomJsonElementVisitor implements JsonElementVisitor<String> {
    /**
     * Handles the processing of JsonObject.
     *
     * @param jsonObject holds json object.
     * @param data       Buffer for storing processed data.
     * @return Data of processed json object.
     */
    @Override
    public String visitJsonObject(JsonObject jsonObject, String data) {
        return jsonObject.accept(this,"");
    }

    /**
     * Handles the processing of JsonArray.
     *
     * @param jsonArray holds json array.
     * @param data      Buffer for storing processed data.
     * @return Data of processed json array.
     */
    @Override
    public String visitJsonArray(JsonArray jsonArray, String data) {
        jsonArray.iterator().forEachRemaining(element -> {
            element.accept(this, data);
        });
        return data;
    }

    /**
     * Handles the processing of JsonPrimitive.
     *
     * @param jsonPrimitive holds json primitive.
     * @param data          Buffer for storing processed data.
     * @return Data of processed json primitive.
     */
    @Override
    public String visitJsonPrimitive(JsonPrimitive jsonPrimitive, String data) {
        return data + jsonPrimitive.getAsString();
    }

    /**
     * Handles the processing of JsonNull.
     *
     * @param jsonNull holds json null.
     * @param data     Buffer for storing processed data.
     * @return Data of processed json null.
     */
    @Override
    public String visitJsonNull(JsonNull jsonNull, String data) {
        return data;
    }
}
```

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -16,6 +16,8 @@
 
 package com.google.gson;
 
+import com.google.gson.visitor.JsonElementVisitor;
+
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


eamonnmcmanus:
Thanks for this PR. Gson is in maintenance mode and for the most part we are not adding new features. I feel like the case for this one isn't all that compelling. It's easy enough for somebody to make their own code to do this without changing Gson. Is there a reason it has to be built in to Gson?

I also think that if we did add this, we would probably want `JsonElementVisitor` to have type parameters for both the parameter and the return type, like the JDK's [`TypeVisitor`](https://docs.oracle.com/en/java/javase/17/docs/api/java.compiler/javax/lang/model/type/TypeVisitor.html).

Issue #1564, opened 2019-08-12
JsonReader throws NumberFormatException which does not contain location info


jdlib:
The message of exceptions thrown by `JsonReader` usually contains location information including the current path. This is really helpful to locate the error in the input JSON.

Example: `java.lang.IllegalStateException: Expected an int but was BOOLEAN at line 7 column 10 path $.x.y`

Now `JsonReader.nextInt(), .nextLong(), .nextDouble()` may throw exceptions that contain no such location information. The exception is thrown in the lines
`double asDouble = Double.parseDouble(peekedString); // don't catch this NumberFormatException.`

How to provoke that specific exception:
Model: `public class Model { int x; }`
and then call `new GsonBuilder().create().fromJson("{ x: ''}", Model.class}`

The exception message will be:  `java.lang.NumberFormatException: empty String`, containg no location information.
The exception itself was thrown by `Double` and `JsonReader` just passed it along.

Imho JsonReader should catch the exception from `Double.parseDouble` and add location information to a rethrown copy.

PR #2076, opened 2022-02-08
Add CodeQL GitHub code scanning workflow

Marcono1234:
Adds a GitHub code scanning workflow running CodeQL, see GitHub documentation "[About code scanning with CodeQL](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql)".

The workflow is currently configured to run all queries in the "security-and-quality" suite; this is more than the default which only runs [certain security queries](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs). I have fixed some of the findings; it looks like all other ones are false positives, or can be ignored.

~~It appears by default generated and test code is scanned as well (and cannot be excluded?), but results for it can be [filtered in the security view](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#filtering-code-scanning-alerts).~~
Edit: Have changed the build command to `mvn compile`; therefore test code is not analyzed anymore.

I have marked this pull request as draft for now to get some feedback. No worries in case you don't want to include this.


eamonnmcmanus:
Were you planning to fix the remaining CodeQL failures?

Marcono1234:
I think all remaining findings are false positives, or can be ignored (for now):
- "Array index out of bounds" for `JavaWriter.java`
This is part of the codegen module, which is non-functional (#2007); additionally as mentioned by you on that issue, that class is most likely a third party class coming from JavaPoet
- "Container contents are never accessed" for `CollectionsDeserializationBenchmark.java`
For this benchmark accessing the collection elements is not necessary, assuming that the Gson unit tests would detect any incorrect functional Gson behavior. I assume the `List` is used in this benchmark to make it more realistic?
- "Dereferenced variable may be null" for `UtcDateTypeAdapter.java` and `ISO8601Utils.java`
These findings seem to be reported due to the `date == null` check below the `try` statement. However, to me that looks reasonable; in general these methods do not seem to support `null` as `date`, but validation of the other arguments might have caused an exception before any null check. Additionally in both cases Gson always seems to call the methods with non-`null` arguments. (And these classes originally come from Jackson, as indicated by the comments)
- "Dereferenced variable may be null" for `LinkedTreeMap.java`
I am not very familiar with the code of that class, but to me it looks like the dereferencing expressions are guarded by the check on the value of the `delta` variable, and that only seems to be -2 respectively 2 when `rightHeight` respectively `leftHeight` are > 0, which won't be the case when `right` or `left` are `null`.
- "Reference equality test on strings" for `LazilyParsedNumber.java`
Check happens in `equals` and is followed by `equals` call for the String value. Maybe using `==` here is premature optimization, but it is functionally correct.
- "Missing catch of NumberFormatException"
Maybe some of them could be revisited in the future, one of them is #1564.
- ... all other notes can probably be ignored (for now).

You as maintainer can [dismiss](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#dismissing-or-deleting-alerts) the false positive alerts later on, if you like.

PR #2079, opened 2022-02-16
Improve Maven build and GitHub Maven workflow

Marcono1234:
Two minor changes to improve the Maven build:

- Fix consecutive Maven builds failing without performing `clean`
By default moditect-maven-plugin refuses to overwrite the JAR file it generated in a previous run. See also https://github.com/moditect/moditect/issues/161.
- Make GitHub Maven build workflow detect Javadoc issues


eamonnmcmanus:
This is great, thanks!

PR #2080, opened 2022-02-21
Add tests for enum constant toString() reading

Marcono1234:
Follow-up for #1950, adding tests for the new behavior.


eamonnmcmanus:
Thanks!

PR #2082, opened 2022-02-25
Added more tests for skipUnquotedValues()

oskaredv:
Five new tests that increase the branch coverage of skipUnquotedValues() by covering five branches of the method that was not covered yet

### DIFF
--- a/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
+++ b/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
@@ -1782,6 +1782,46 @@ public final class JsonReaderTest extends TestCase {
     assertEquals(JsonToken.NUMBER, token);
   }
 
+  public void testSkipUnquotedSlash() throws IOException {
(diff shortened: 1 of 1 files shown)


Marcono1234:
Might be good to add some kind of assertion after the `skipValue();` calls to make sure it actually had an effect. However, since these cases all seem to represent invalid lenient JSON, I am not sure if there is a way to perform any reasonable assertion; maybe checking `JsonReader.getPath()` would work at least.

Feel free to only consider this a suggestion; I am not a member of this project.

oskaredv:
Thanks for the suggestion! I initially tried to add assertions after `skipValue()`, but did not manage to do it since the cases represent invalid lenient JSON. With `JsonReader.getPath()`and `JsonReader.getPreviousPath()` I managed to add assertions after `skipValue()` in two of the tests. For the other three `JsonReader.getPath()` returned null values after `skipValue() and I did not know what do to with that.

PR #2083, opened 2022-02-25
Refactoring and reducing code duplication in JsonReader

oskaredv:
Refactored code that was duplicated in several methods into its own method. Five of those methods(peek, nextDouble, nextInt, nextLong and nextString) have high cyclomatic complexity and this will help reducing that a bit.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -338,10 +338,7 @@ public class JsonReader implements Closeable {
    * beginning of a new array.
    */
   public void beginArray() throws IOException {
-    int p = peeked;
-    if (p == PEEKED_NONE) {
-      p = doPeek();
-    }
+    int p = peeked();
     if (p == PEEKED_BEGIN_ARRAY) {
       push(JsonScope.EMPTY_ARRAY);
       pathIndices[stackSize - 1] = 0;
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -406,24 +394,29 @@ public void endObject() throws IOException {
...
   /**
-   * Returns true if the current array or object has another element.
+   * Returns peeked if not PEEKED_NONE, if PEEKED_NONE then doPeek() is returned instead
    */
-  public boolean hasNext() throws IOException {
+  private int peeked() {
     int p = peeked;
     if (p == PEEKED_NONE) {
-      p = doPeek();
+         p = doPeek();
```
Marcono1234, line 402: It looks like the indentation was changed here by accident:
```suggestion
      p = doPeek();
```

Marcono1234 (commented):
There is one more pattern of this in the `static` block at the end of the file; do you want to replace this as well? Might be good then to make `peeked()` package-private to avoid a synthetic accessor method (I assume).
Maybe at that point it might also make sense to merge `peeked()` and `doPeek()` into one method?

(I am not a member of this project; the maintainers will probably give you more in-depth feedback)

Marcono1234:
> There is one more pattern of this in the `static` block at the end of the file; do you want to replace this as well? Might be good then to make `peeked()` package-private to avoid a synthetic accessor method (I assume).
> Maybe at that point it might also make sense to merge `peeked()` and `doPeek()` into one method?

@oskaredv, do you want to address that as well? Or do you think that this is not worth it?

PR #2085, opened 2022-02-28
Mention in README that Gson is in maintenance mode

Marcono1234:
Mention in README that Gson is in maintenance mode. Currently this is only mentioned in the [GitHub issue template](https://github.com/google/gson/blob/91256532788aeee39d67087936be9bb458022c7a/.github/ISSUE_TEMPLATE/feature_request.md?plain=1#L3) which might not be seen by users when they directly create a pull request.

I have prefixed the sentence with ":information_source:" to draw some attention to it. Please let me know if that should be removed or changed. Feedback regarding the wording is welcome as well.


eamonnmcmanus:
Thanks, this is very helpful!

PR #2091, opened 2022-03-15
Removal of magic numbers in CollectionTest

Lhandi29:
Sync Forked Repo


eamonnmcmanus:
Can you explain what this PR is about? The description is not very helpful.

Lhandi29:
> 

I wanted to remove magic numbers of this class. Magic numbers are not explicit to the project and subsequently one can forget what they correspond to. In my opinion, deleting them will allow future contributors to work more efficiently on the project.

eamonnmcmanus:
While I agree that magic numbers are often a hindrance, here I don't think there's much advantage to the rewrite. It's pretty obvious that the 3 in `assertEquals(3, queue.size())` corresponds to the 3 elements in `[10, 20, 31]`, and 3 explicit assertions are if anything easier to understand than a loop.

One change that I think _would_ be worthwhile would be to use [Truth](https://truth.dev/) for the assertions. Since `PriorityQueue`, `Vector`, and `Stack` all implement `Iterable`, you can use `IterableSubject` to write:

```
assertThat(queue).containsExactly(10, 20, 31);
```

instead of checking the size and removing the elements one by one. This doesn't test that the elements would be retrieved in the expected order from the queue, but they must be if `PriorityQueue` is behaving correctly, and it's not our job to test that there. For the other types, you could tack `.inOrder()` on the end of the assertion, since you do in fact expect the elements to be present in the given order.

I'm a bit surprised to see that Truth is not yet a dependency of the main `gson/pom.xml` project, though it is in `proto/pom.xml`. I think it would be worth adding it now.

Please do sign the CLA if you do decide to go ahead with this.

eamonnmcmanus:
Sorry to have neglected this. I believe the Truth migration changes have already happened in other PRs.

PR #2098, opened 2022-04-03
Delete CodeGenFunctionalTest.java

kentrachmat:
[DEAD CODE] CodeGenFunctionalTest.java does not do anything and therefore the test is not used.


eamonnmcmanus:
I think it would make more sense to delete the whole `codegen/` hierarchy. It seems to be an unfinished experiment that was abandoned ten years ago. Feel free to send another PR to that effect.

PR #2099, opened 2022-04-03
codegen removal

kentrachmat:
Thank you for your reply, in this PR i have deleted the whole codegen file


pom.xml
```diff
@@ -13,15 +13,14 @@
   <artifactId>gson-parent</artifactId>
   <version>2.9.1-SNAPSHOT</version>
   <packaging>pom</packaging>
-
+
```
eamonnmcmanus, line 16: This doesn't seem necessary?

eamonnmcmanus (commented):
Thanks! We do need a signed CLA before we can merge, though.

pom.xml
kentrachmat: Yes sorry, i was trying to solve the CLA problem, this is actually my first ever PR and i thought by changing my email in the next commit would solve this problem

kentrachmat:
Hai I have signed the CLA but i still can't merge, can you help ? thank you in advance

PR #2102, opened 2022-04-04
Code Smell (GraphAdapterBuilder.java, RawCollectionsExample.java, PostConstructAdapterFactory.java)

kentrachmat:
- Standard outputs should not be used directly to log anything (System.out -> logger)
- Modifiers should be declared in the correct order
- A field should not duplicate the name of its containing class
- Cognitive Complexity of methods should not be too high


extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
```diff
@@ -98,10 +98,10 @@ public void registerOn(GsonBuilder gsonBuilder) {
             out.nullValue();
             return;
           }
+          writeTypeAdapter( out, value);
+        }
 
-          Graph graph = graphThreadLocal.get();
-          boolean writeEntireGraph = false;
-
+        public void writeTypeAdapter(JsonWriter out, T value) throws IOException {
```
Marcono1234, line 104: Not sure if these few lines really have any effect on the complexity (same for the `readTypeAdapter` method below). Maybe this change reduces a theoretical complexity value; but in reality I would say it does not make a (big) difference.

Marcono1234 (commented):
Not sure if all of these changes are really needed. However, I am not a member of this project, and the maintainers might think differently about this.

extras/src/main/java/com/google/gson/extras/examples/rawcollections/RawCollectionsExample.java
```diff
@@ -17,6 +17,9 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.logging.Level;
```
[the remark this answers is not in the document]
kentrachmat: i analysed the code with sonarqube and it recommends me to use logger instead of System.out (better visibility and security)

extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
eamonnmcmanus: I typed up a longer comment which seems to have got lost. I was suggesting pulling original lines 122-124 into a method and getting rid of the `writeEntireGraph` variable. I think that would be a better way of simplifying the code than just separating out the null case.

extras/src/main/java/com/google/gson/extras/examples/rawcollections/RawCollectionsExample.java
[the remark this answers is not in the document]
kentrachmat: I see, thank you for the input

kentrachmat:
Hello,

What do you thing regarding  my recent modification ?

extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
```diff
@@ -111,20 +110,17 @@ public void registerOn(GsonBuilder gsonBuilder) {
...
 
-          @SuppressWarnings("unchecked") // graph.map guarantees consistency between value and T
-          Element<T> element = (Element<T>) graph.map.get(value);
-          if (element == null) {
-            element = new Element<T>(value, graph.nextName(), typeAdapter, null);
-            graph.map.put(value, element);
-            graph.queue.add(element);
-          }
+            @SuppressWarnings("unchecked") // graph.map guarantees consistency between value and T
+            Element<T> element = (Element<T>) graph.map.get(value);
```
eamonnmcmanus, line 117: This will always be null since nothing has been written to the IdentityHashMap from the previous statement. So you can simplify here.

eamonnmcmanus:
As a reminder, the Gson project is in maintenance mode. We can only accept cosmetic changes like these if they are handled efficiently: ideally just uncontroversial changes that we can merge as-is, but certainly if changes are requested they should be applied. Please revert the changes to `RawCollectionsExample` and the unnecessary space changes.

extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
kentrachmat: You're right, i'll do that

PR #2121, opened 2022-05-22
Rename ReflectiveTypeAdapterFactory field inclusion check method

### LINKED ISSUE #2066: Exclude field condition
https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L68

Is it ok? If follow that logic, the field going to be excluded when
A. **NOT** exclude a class
**AND**
B. **NOT** exclude a field

I guess it should be `return excluder.excludeClass(f.getType(), serialize) || excluder.excludeField(f, serialize)`

that means, that, the field going to be excluded when
A. exclude a class
OR
B. exclude a field

Correct me if I wrong

Marcono1234:
Fixes #2066


Marcono1234:
@eamonnmcmanus, could you please have a look? This only changes method names and removes one internal method.

eamonnmcmanus:
LGTM

Issue #1208, opened 2017-12-15
Why is the JsonParser always lenient?


xfxyjwf:
See this line:
https://github.com/google/gson/blob/6b20e3aaec0b3280f01c7785b5dd391ab6de9cc7/gson/src/main/java/com/google/gson/JsonParser.java#L82

It basically makes the JsonReader.setLenient() API useless. 

How can we make the parser strict? For reference this is for http://b/34734814

JakeWharton:
You can call `getAdapter(type).fromJson(gson.newJsonReader(input))` instead of just `fromJson(input)` to get strict parsing. We should really deprecate all of the `fromJson` methods and add new versions that are strict by default.

The reason is bad decisions long ago that we can no longer change 😢

xfxyjwf:
Thanks for the quick reply! I'm not sure how to apply this to our code though. Our code currently looks like this:

```java
JsonParser jsonParser = new JsonParser();
JsonReader reader = new JsonReader(json);
reader.setLenient(false);
JsonElement element = jsonParser.parse(reader);
... use element ...
```

See: https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L1170

How do I apply `getAdapter(type).fromJson(gson.newJsonReader(input))`?

NightlyNexus:
unfortunate there's no `public JsonElement JsonReader.readJsonValue()` to expose this.

you might be have to mix in the object binding APIs. something like:
```java
private static TypeAdapter<JsonElement> JSON_ELEMENT_ADAPTER = new Gson().getAdapter(JsonElement.class);

JsonElement jsonElement(JsonReader reader) {
  JSON_ELEMENT_ADAPTER.read(reader);
}
```

There may be subtle behavior difference in there. I don't recall!

vkopichenko:
Here is an example of Gson strict parsing workaround proposed here with extensive test case: https://stackoverflow.com/questions/43233898/how-to-check-if-json-is-valid-in-java-using-gson/47890960#47890960

Issue #763, opened 2015-12-31
why gson can parse some string not stand json such as "[a,b,c]" ?


somewon:
``` java
        JsonParser parser = new JsonParser();
        JsonElement el = parser.parse("[a,b,c]");
        JsonArray array = new JsonArray();
        array = el.getAsJsonArray();
```

? why?

JakeWharton:
It's not valid JSON. Try `"[\"a\", \"b\", \"c\"]"` instead.

somewon:
Yes. I know that.  It's not valid JSON. But gson could parse this. 
I'm curious about why gson have this feature.

JakeWharton:
Oh you're asking why it _can_ parse it? For historical reasons Gson is lenient by default and unfortunately there's no sane API to disable it for `JsonParser` that I can see. Even if you create your own `JsonReader` and call `setLenient(false)` Gson will ignore that setting.

Disabling the lenient behavior can be done when using the object binding API (via the `Gson` type), but there's hoops to jump through. Are you able to use the binding API instead of the DOM API?

pretizy:
I haven't reviewed the code, but can you explain why the lenient behavior
affects the json standard format

Marcono1234:
Probably resolved (but not really fixed) by the documentation improvements of #2122.

eamonnmcmanus:
I think those documentation improvements are probably the best we can do given compatibility requirements and the maintenance-mode status of this project.

PR #2122, opened 2022-05-22
Improve lenient mode documentation

### LINKED ISSUE #1720: Gson.assertFullConsumption should throw JsonSyntaxException if JSON was not fully consumed
Currently `Gson.assertFullConsumption` throws a `JsonIOException` if the JSON was not fully consumed:
https://github.com/google/gson/blob/ceae88bd6667f4263bbe02e6b3710b8a683906a2/gson/src/main/java/com/google/gson/Gson.java#L902-L906

Maybe it would be better to throw a `JsonSyntaxException` instead, because:
- The problem is not really IO related because reading worked without issues (no `IOException` was thrown)
- This code is also called by `fromJson(String, ...)` which does not declare that it can throw a `JsonIOException`

For the case where a type adapter is not properly consuming all input (see also #835) neither would match perfectly though that might be acceptable.

### LINKED ISSUE #1923: Mistake on Javadoc method description (?)
**It appears to me that you've made a mistake on the description of the `parseReader​(Reader reader)` method of the JsonObject class.**

image

You can check said method's doc yourself [here](https://javadoc.io/static/com.google.code.gson/gson/2.8.7/com.google.gson/com/google/gson/JsonParser.html)

**I believe it should be _"Parses the specified stream/reader into a parse tree"._**

I hope that my lack of English/Java knowledge isn't the problem here and I'm not wasting your time.

Marcono1234:
Fixes #1720
Resolves #1923

Relates to: #372, #763, #1208, #1609

Improves the documentation regarding lenient mode and trailing data handling. The current Gson behavior is still not ideal, but now at least the documentation (hopefully) matches the actual behavior.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -16,25 +16,6 @@
 
 package com.google.gson;
 
-import java.io.EOFException;
-import java.io.IOException;
-import java.io.Reader;
-import java.io.StringReader;
-import java.io.StringWriter;
-import java.io.Writer;
-import java.lang.reflect.Type;
-import java.math.BigDecimal;
-import java.math.BigInteger;
-import java.text.DateFormat;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicLongArray;
-
 import com.google.gson.internal.ConstructorConstructor;
 import com.google.gson.internal.Excluder;
 import com.google.gson.internal.GsonBuildConfig;
@@ -58,6 +39,24 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import com.google.gson.stream.MalformedJsonException;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.io.Writer;
+import java.lang.reflect.Type;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.text.DateFormat;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongArray;
 
 /**
  * This is the main class for using Gson. Gson is typically used by first constructing a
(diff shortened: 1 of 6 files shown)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -965,7 +1021,7 @@ public <T> T fromJson(Reader json, Type typeOfT) throws JsonIOException, JsonSyn
   private static void assertFullConsumption(Object obj, JsonReader reader) {
     try {
       if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {
-        throw new JsonIOException("JSON document was not fully consumed.");
+        throw new JsonSyntaxException("JSON document was not fully consumed.");
```
Marcono1234, line 1024: This fixes #1720

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -252,6 +252,9 @@ public final JsonElement toJsonTree(T value) {
    * read is strict. Create a {@link JsonReader#setLenient(boolean) lenient}
    * {@code JsonReader} and call {@link #read(JsonReader)} for lenient reading.
    *
+   * <p>No exception is thrown if the JSON data has multiple top-level JSON elements,
```
Marcono1234, line 255: This behavior is probably rather questionable though...

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -304,8 +304,6 @@ public JsonReader(Reader in) {
    *       prefix</a>, <code>")]}'\n"</code>.
    *   <li>Streams that include multiple top-level values. With strict parsing,
    *       each stream must contain exactly one top-level value.
-   *   <li>Top-level values of any type. With strict parsing, the top-level
```
Marcono1234, line 307: This and the changes below are from #1609 and the comments there.

This specific sentence was outdated because Gson already supports top-level values of other types, even in strict mode.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -321,6 +319,18 @@ public JsonReader(Reader in) {
    *       {@code :}.
    *   <li>Name/value pairs separated by {@code ;} instead of {@code ,}.
    * </ul>
+   *
+   * <p>Note: Even in strict mode there are slight derivations from the JSON
```
Marcono1234, line 323: Maybe this should clarify that these derivations are not guaranteed behavior and might change in the future?

gson/src/main/java/com/google/gson/Gson.java
eamonnmcmanus: This is a behaviour change, if minor, so I'll want to run it past all of Google's tests first.

eamonnmcmanus: OK, no issues found.

eamonnmcmanus:
Thanks for doing this! I'm merging as-is, but feel free to convert some of the open questions into other PRs or new issues.

Marcono1234:
Thanks for merging this!

> feel free to convert some of the open questions into other PRs or new issues

The main open problem is probably the missing end of document check for the `TypeAdapter` methods, but changing that would be backward incompatible. At least this behavior is now documented, so I don't think any further action is necessary at the moment.

PR #2132, opened 2022-06-10
Add explicit support for floats in JsonTreeWriter.

Capstan:
Follow-up to comments on #2130, which introduced a new override which was not overridden by `JsonTreeWriter`. Also tweaks the doccomments for the `float`, `double`, and `Number` variants of `JsonWriter.value`.

Supplement to the fix for #1127.


Marcono1234 (approved):
> I'll give @Marcono1234 a chance to comment this time before merging. :-)

Looks good to me. It is also great that this adjusts the tests to cover the newly added method!

Capstan:
Does this need further review?

Issue #712, opened 2015-10-15
Serialization issue with RuntimeTypeAdapterFactory


ultraon:
Hello, i found after serialization root object 'type' field is not injected, for inner object 'type' fields are added. Example:

```
public class Test {
  public class Base {
    String baseField = "base";
  }
  public class Child extends Base {
    String childField = "child";
    Base baseInChild = new Base();
  }

  public static void main(String[] args) {
        Gson mGson = new GsonBuilder()
                .setPrettyPrinting()
                .registerTypeAdapterFactory(
                        RuntimeTypeAdapterFactory.
                                of(Base.class).
                                registerSubtype(Base.class).
                                registerSubtype(Child.class)
                )
                .create();

        final String jsonStr = mGson.toJson(new Child());
        final Base child = mGson.fromJson(jsonStr, Base.class); // ERROR HERE: "Caused by: com.google.gson.JsonParseException: cannot deserialize class mypackage.Test$Base because it does not define a field named type"
  }
}
```

and json output after serialization:

```
{
  "baseInChild": {
    "type": "Base",
    "baseField": "base"
  },
  "childField": "child",
  "baseField": "base"
//in this place should be generated field 'type' with value 'Child'
}
```

ultraon:
Hello, i've found bug in RuntimeTypeAdapterFactory (line185, method create(), and resolved this issue, need replace
```
if (type.getRawType() != baseType)
```
with
```
if (null == type || !baseType.isAssignableFrom(type.getRawType()))
```

Then RuntimeTypeAdapterFactory works well as expected.

renanferrari:
@ultraon Thanks, just ran into this issue today and your answer solved it.

calrsom:
Thank you.  This solved an issue I had with serializing an subtype with a custom serializer and type field not being included.  This fix should really get pulled into the codebase.

swankjesse:
Try replacing this:

```
        final String jsonStr = mGson.toJson(new Child());
```

With this:

```
        final String jsonStr = mGson.toJson(new Child(), Base.class);
```

nsarathk:
@swankjesse i guess thats where the problem is. i tried using the same it didnt work. Am about to try with the solution which @ultraon has given.

ultraon:
Hello, i put fixed version of the RuntimeTypeAdapterFactory here: [https://gist.github.com/ultraon/e7b4de17ad8ec6c0a255](url)

headsvk:
[...]

PR #2139, opened 2022-06-27
Fix RuntimeTypeAdapterFactory

t-oster:
Trying to use this class as is results in the type-property not being serialized into the JSON, thus it is not present on deserialization.
The fix from https://github.com/google/gson/issues/712#issuecomment-148955110 works. No idea why this is not merged yet.

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
@@ -16,10 +16,6 @@
 
 package com.google.gson.typeadapters;
 
-import java.io.IOException;
-import java.util.LinkedHashMap;
-import java.util.Map;
-
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
@@ -30,6 +26,10 @@ import com.google.gson.TypeAdapterFactory;
 import com.google.gson.reflect.TypeToken;
 import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
 
 /**
  * Adapts values whose runtime type may differ from their declaration type. This
@@ -205,7 +205,7 @@ public final class RuntimeTypeAdapterFactory<T> implements TypeAdapterFactory {
 
   @Override
   public <R> TypeAdapter<R> create(Gson gson, TypeToken<R> type) {
-    if (type.getRawType() != baseType) {
+    if (type == null || !baseType.isAssignableFrom(type.getRawType())) {
       return null;
     }
 
--- a/extras/src/test/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactoryTest.java
+++ b/extras/src/test/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactoryTest.java
@@ -34,7 +34,9 @@ public final class RuntimeTypeAdapterFactoryTest extends TestCase {
 
     CreditCard original = new CreditCard("Jesse", 234);
     assertEquals("{\"type\":\"CreditCard\",\"cvv\":234,\"ownerName\":\"Jesse\"}",
-        gson.toJson(original, BillingInstrument.class));
+        //do not give the explicit typeOfSrc, because if this would be in a list
+        //or an attribute, there would also be no hint. See #712
+        gson.toJson(original));
     BillingInstrument deserialized = gson.fromJson(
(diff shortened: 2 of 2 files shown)


eamonnmcmanus:
Thanks for the proposal! Would it be possible to add a test that fails without this fix and passes with it?

t-oster:
@eamonnmcmanus like this? Or in a single commit? The thing is in the existing test, the base class was explicitely given, so the bug did not occur. I just removed the explicit typeOfSrc parameter. In a real world example, the object could be a collection element or an attribute of a class being serialized, so there would be no typeOfSrc either. If it is not given, the RuntimeTypeAdapterFactory.create method does not work, because it just compares if the type matches the base-type excaclty, instead of if it is supertype.

t-oster:
...btw it would be great, if the gson-extras would be released on maven somehow, because currently many people just copy the class to their source-code and won't profit from future updates.

extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
```diff
@@ -205,7 +205,7 @@ public RuntimeTypeAdapterFactory<T> registerSubtype(Class<? extends T> type) {
 
   @Override
   public <R> TypeAdapter<R> create(Gson gson, TypeToken<R> type) {
-    if (type.getRawType() != baseType) {
+    if (null == type || !baseType.isAssignableFrom(type.getRawType())) {
```
eamonnmcmanus, line 208: Would you mind fixing the Yoda condition? i.e. `type == null` instead.

Marcono1234:
> In a real world example, the object could be a collection element or an attribute of a class being serialized, so there would be no typeOfSrc either.

It does work if the field type or collection element type is the base type. Consider the following class:
```java
static class Test {
  Base base;
  Sub sub;

  List<Base> baseList;
  List<Sub> subList;
}
```
When creating an instance of `Sub` and assigning it to `base` and `sub`, and a List containing a `Sub` object and assigning it to `baseList` and `subList`, the JSON will currently contain the `type` for the `Base` fields, but not for the `Child`. There hasn't been any official statement about the behavior (and I am not a member of this project), but I could imagine that this behavior might have been intended. When your model class explicitly uses a specific subclass, then including the `type` might be redundant because deserialization would work without the `type`.
An issue only occurs when you serialize an object as a different type than you deserialize it, as shown in the example in #712. A similar issue would also occur when there is an intermediate class, e.g. `Base` > `Sub1` > `Sub2`. Then a field with type `Sub1` but runtime value of `Sub2` will not have the `type` member either.

So even if the behavior was originally intended to be like this, it might be good nonetheless to change it.
However, @t-oster, maybe it would be good to not modify the original test method but instead add a separate one to verify that it still behaves as expected when the class is explicitly specified for `toJson`. What do you think?

t-oster:
@eamonnmcmanus I inverted the null check. @Marcono1234 I am not sure if I understand, but I had the error when trying to expliciteley serialize a list of the base-class, so this was definitely a bug IMHO. Could you maybe create an example? If the serialization does not add the type field, the deserialization will fail. And if it adds the field, the deserialization will remove it, so I do not see the redundancy.

@eamonnmcmanus is only the extras on maintenance mode or the whole GSON project? Why? Are the receommended alternatives?

Marcono1234:
> so I do not see the redundancy

What I meant was when serializing as `Sub` and also deserializing as `Sub`, then the `type` field would be redundant, and I assumed that might have been the intention for the current behavior. But as you say, when deserializing as `Base` this will fail because `type` is missing, so the fix introduced by this PR is needed. Sorry if my comment caused confusion.

Regarding the tests, what I meant was that it should ideally cover both:
- `toJson(subInstance)`
- `toJson(subInstance, BaseClass.class)`

Your changes modify the tests to only cover `toJson(subInstance)`, but not `toJson(subInstance, BaseClass.class)` anymore (if I see that correctly).

However, I am not a member of this project so unlike eamonnmcmanus' comments feel free to consider mine only a suggestion.

eamonnmcmanus:
> is only the extras on maintenance mode or the whole GSON project? Why? Are there recommended alternatives?

The whole project. It is fine for people to continue using Gson but for new projects I would probably recommend Moshi.

eamonnmcmanus:
FYI this change caused problems at Google, where we do in fact use the `extras` project as-is. The problems are similar to the `List` scenario described by @Marcono1234 . I think the change here should be an optional behaviour of `RuntimeTypeAdapterFactory` that is triggered by calling a new `recognizeSubclasses()` method, rather than always happening. I'm planning a PR to that effect. It means that other people who have copies of the class can update their copy without fear of being broken.

t-oster:
Ok. Thanks for the heads up. I'm sorry this introduced errors, but with your `recognizeSubclasses()` method this should be OK.

PR #2142, opened 2022-07-07
Update README.md

kvv79:
Change the Gradle dependency in README.md to be more consistent with Guava

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 Gradle:
 ```gradle
 dependencies {
-  implementation 'com.google.code.gson:gson:2.9.0'
+  implementation("com.google.code.gson:gson:2.9.0")
 }
 ```
 


README.md
```diff
@@ -19,7 +19,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 Gradle:
 ```gradle
 dependencies {
-  implementation 'com.google.code.gson:gson:2.9.0'
+  implementation("com.google.code.gson:gson:2.9.0")
```
eamonnmcmanus, line 22: My impression is that single quotes are preferred in Gradle scripts. So possibly it would make more sense to propose the opposite change for the Guava documentation.

Marcono1234: `implementation("...")` is probably what you see more commonly for the Kotlin build script DSL, but apparently it also valid for the Groovy DSL, see https://github.com/google/guava/pull/3673 which proposed that for Guava.
See also the [Gradle documentation](https://docs.gradle.org/current/userguide/dependency_management_for_java_projects.html#sec:setting_up_a_standard_build_script_java_tutorial) for an example and click the "Groovy" and "Kotlin" tabs of the code block.

eamonnmcmanus: Thanks for the background! In that case, I think this change makes sense. But, we can't accept it until the CLA has been signed.

Issue #2144, opened 2022-07-11
JsonPrimitive#equals Method behaves incorrect when used with BigInteger


noeppi-noeppi:
# Gson version
2.9.0

# Java / Android version
Eclipse Adoptium OpenJDK 64-Bit Server VM 17.0.2+8 on Linux

# Used tools
- [ ] Maven; version: 
- [x] Gradle; version: 7.3.3
- [ ] ProGuard (attach the configuration file please); version: 
- [ ] ...

# Description
Two `JsonPrimitive`s with `BigInteger` values return `true` from their `equals` method if their long value obtained from `Number#longValue` is equal. This for example causes the numbers `23648748948963264236478263434` and `808933803388026932016705674` to be considered equal.

This is because both numbers are greater than `Long.MAX_VALUE`, so only the least significant 64 bits of these numbers are considered in the `longValue()` method. These happen to be `6484271001536295050` for both of the numbers, so `JsonPrimitive` considers them equal.

## Expected behaviour
The expected behaviour is that `JsonPrimitive#equals` returns `false` for values which don't actually equal each other.

## Actual behaviour
The actual behaviour is that `JsonPrimitive#equals` returns `true` for `BigIntegers` whose 64 least significant bits are the same.

# Reproduction steps
```java
JsonPrimitive p1 = new JsonPrimitive(new BigInteger("23648748948963264236478263434"));
JsonPrimitive p2 = new JsonPrimitive(new BigInteger("808933803388026932016705674"));
System.out.println(Objects.equals(p1, p2)); // is true, should be false
```

# Exception stack trace
No exception occurred.

Marcono1234:
Seems to be related to #904

MaicolAntali:
This issue can be closed. Related PR: #2311

Marcono1234:
@MaicolAntali, thanks! The reproduction steps provided above print `false` now, as expected.

PR #2145, opened 2022-07-14
Fix for issues #904 and #2144

OmarAlJarrah:
As a BigDecimal is essentially a wrapper around a BigInteger that "remembers where the decimal point is", casting BigInteger to BigDecimal using the provided constructor (or the method that uses the constructor in the conversion process), and do the comparing process makes more sense, avoiding the bug of max long/double values comparison.


gson/src/main/java/com/google/gson/JsonPrimitive.java
```diff
@@ -292,4 +298,16 @@ private static boolean isIntegral(JsonPrimitive primitive) {
     }
     return false;
   }
+
+  private static boolean isBigInteger(JsonPrimitive primitive) {
+    return primitive.value instanceof BigDecimal;
```
eamonnmcmanus, line 303: I think this should be `BigInteger`? Which in turn makes me think that there should be a test that would have caught this.

Speaking of tests, in general if you are proposing a fix for an issue, you should include a test that fails without the fix and passes with it.

eamonnmcmanus (changes requested):
Thanks for looking into this!

gson/src/main/java/com/google/gson/JsonPrimitive.java
omar2682000: @eamonnmcmanus 
Yes that should be a BigInteger, as for the testing part, thank you for the tip, I will make sure to provide tests with any future commits.

PR #2149, opened 2022-07-20
Upgrade to oss-parent 9

liutikas:
org.sonatype.oss:oss-parent:7 is unsigned (missing .asc files), which forces users of gson library
that enforce signature verification to explicitly allowlist oss-parent. Luckily oss-parent 9 is signed,
thus fixing this issue.

See:
https://repo1.maven.org/maven2/org/sonatype/oss/oss-parent/7/
vs
https://repo1.maven.org/maven2/org/sonatype/oss/oss-parent/9/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.sonatype.oss</groupId>
     <artifactId>oss-parent</artifactId>
-    <version>7</version>
+    <version>9</version>
   </parent>
 
   <groupId>com.google.code.gson</groupId>


jvirtanen:
This looks like a good change, but for the long term, it would be best to move away from oss-parent as [it is deprecated and might not work correctly anymore](https://central.sonatype.org/publish/publish-maven/#deprecated-oss-parent).

liutikas:
> This looks like a good change, but for the long term, it would be best to move away from oss-parent as [it is deprecated and might not work correctly anymore](https://central.sonatype.org/publish/publish-maven/#deprecated-oss-parent).

I think that seems to be a fine thing to do, but I would prefer it is done in a separate step :)

eamonnmcmanus:
Thanks! I agree that ultimately we should get rid of oss-parent. I may try that before the next release, since I think the main impact is on releasing.

Marcono1234:
Getting rid of oss-parent would be really good. It appears at the moment it apparently effectively [downgrades certain plugins during release](https://github.com/sonatype/oss-parents/blob/5cc24e62cae0fc8a1a6a917c581574b71dec476f/oss-parent/pom.xml#L110-L149), which seems to be the reason why the published Javadoc has no external links to the JDK classes. The following is logged:
> ```
> [INFO] --- maven-javadoc-plugin:2.7:jar (attach-javadocs) @ gson ---
> [ERROR] Error fetching link: https://docs.oracle.com/en/java/javase/11/docs/api/package-list. Ignored it.
> ...
> ```
Notice the `2.7` as version even though Gson specifies 3.4.0 as version for that plugin (and a regular `mvn package javadoc:javadoc` works without issues).

PR #2150, opened 2022-07-24
BUILD files to build Gson with Bazel and easily depend on it in Bazel projects

pawelz:
Gson builds cleanly without any external dependencies, so Bazel BUILD files are very simple. I believe this won't add much maintanance burden, but will make it very easy to use Gson in Bazel projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ build
 .DS_Store
 
 examples/android-proguard-example/gen
+
+bazel-*
--- a/README.md
+++ b/README.md
@@ -32,6 +32,17 @@ Maven:
 </dependency>
 ```
 
+<!-- TODO(pawelz): replace branch="master" with tag="gson-parent-X.Y.Z" once the bazel BUILD files make it to a release. -->
+Bazel:
+```bzl
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+git_repository(
+    name="gson",
+    remote="https://github.com/google/gson.git",
+    branch="master",
+)
+```
+
 [Gson jar downloads](https://maven-badges.herokuapp.com/maven-central/com.google.code.gson/gson) are available from Maven Central.
 
 ![Build Status](https://github.com/google/gson/actions/workflows/build.yml/badge.svg)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,1 @@
+workspace(name = "com_google_gson")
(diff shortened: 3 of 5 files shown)


Marcono1234:
> but will make it very easy to use Gson in Bazel projects

Is this about building Gson from source using Bazel? Because to me it looks like using Gson (respectively any Maven artifact?) is already possibly through [`rules_jvm_external`](https://github.com/bazelbuild/rules_jvm_external#usage).

I am not really familiar with Bazel, but shouldn't the `BUILD` file be at the repository root, or at least in the `gson` subfolder, and not in the sources package? (See [Migrating from Maven to Bazel](https://bazel.build/migrate/maven#2-build))

Unfortunately it looks like the Bazel build is not complete:
- `module-info.class` seems to be missing
(might not be so important, but certainly a minor annoyance)
- `GsonBuildConfig` does not seem to be built properly (or I am using Bazel wrongly)
The compiled classes using `GsonBuildConfig.VERSION` just seem to contain the placeholder value:
    ```java
    throw new IllegalArgumentException("GSON (${project.version}) cannot handle " + type);
    ```
    (decompiled from `Gson.class` in final JAR)

Note that this project originally also had a Gradle build script but it was removed (#2063) because it became outdated, and maintaining two separate build scripts was likely not worth the effort.

I am a little bit skeptical whether Bazel build support really adds much value. However, I am not a member of this project and they might think differently about this.

pawelz:
Thank you for taking a look and commenting.

After thinking this over, you are probably right. I can work with my own fork and this is probably not worth merging in.

I will leave it open for now, in case the project owners want this change. But if they say they are also skeptical, I'll withdraw the PR.

Marcono1234:
Apparently there is also another PR adding a Bazel build configuration:  #1574
Though that PR looks more complex and is probably outdated by now.

eamonnmcmanus:
We recently removed the Gradle build because of the maintenance burden it imposed. So I think we'd be reluctant to add another build system. It happens that within Google we do build Gson with Blaze (the Google precursor to Bazel) but I'm still not really seeing the utility of having this. As @Marcono1234 notes, Bazel clients can get Gson from Maven Central.

Feel free to reopen this PR if there is some subtle reason why this _is_ actually important for Bazel clients. In that case we would also need to update the [`build.yml`](https://github.com/google/gson/blob/master/.github/workflows/build.yml) workflow so that it validates the Bazel build as well as the Maven one.

PR #2151, opened 2022-07-24
Improve settings propagation for temporary `JsonTreeReader` and `JsonTreeWriter`

### LINKED ISSUE #676: `RuntimeTypeAdapterFactory` overrides explicit control over `serializeNulls` behavior of contained objects

Marcono1234:
Fixes #676
Supersedes #677 (this PR here does not contain the same test code, but have verified that the test code there passed with these changes too)

### Description of the issue
In some situations adapters need to:
- first parse the JSON as `JsonElement`, perform modifications to it and then parse the `JsonElement` as desired target type
- first write the JSON to a `JsonElement`, perform modifications to it and then write the `JsonElement` to the final `JsonWriter`

In these situations a temporary `JsonTreeReader` respectively `JsonTreeWriter` is used. The issue was that these readers and writers were not created with the same settings (lenientness, serialize null) as the original reader or writer. This could cause behavior mismatch or unexpected exceptions during deserialization or serialization.

### Implementation notes for this PR
This PR introduces two new `TypeAdapter` methods: `fromJsonTreeWithSettingsFrom` and `toJsonTreeWithSettingsFrom`. Both apply the settings from an existing reader / writer to the temporary `JsonTreeReader` / `JsonTreeWriter`.
This might not be the cleanest solution, so any feedback is appreciated. However, it appears a solution similar to this is necessary because `JsonTreeReader` and `JsonTreeWriter` are internal classes and therefore cannot be used from outside of Gson (not even from the `extras` module at the moment), which was the proposed solution in #677.

### DIFF
--- a/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
+++ b/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
@@ -194,7 +194,7 @@ public final class GraphAdapterBuilder {
             // now that we know the typeAdapter for this name, go from JsonElement to 'T'
             if (element.value == null) {
               element.typeAdapter = typeAdapter;
-              element.read(graph);
+              element.read(graph, in);
             }
             return element.value;
           } finally {
@@ -299,12 +299,12 @@ public final class GraphAdapterBuilder {
       typeAdapter.write(out, value);
     }
 
-    void read(Graph graph) throws IOException {
+    void read(Graph graph, JsonReader originalReader) throws IOException {
       if (graph.nextCreate != null) {
         throw new IllegalStateException("Unexpected recursive call to read() for " + id);
       }
       graph.nextCreate = this;
-      value = typeAdapter.fromJsonTree(element);
+      value = typeAdapter.fromJsonTreeWithSettingsFrom(element, originalReader);
       if (value == null) {
         throw new IllegalStateException("non-null value deserialized to null: " + element);
       }
--- a/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java
@@ -257,7 +257,7 @@ public final class RuntimeTypeAdapterFactory<T> implements TypeAdapterFactory {
         JsonObject jsonObject = delegate.toJsonTreeWithSettingsFrom(value, out).getAsJsonObject();
 
         if (maintainType) {
-          jsonElementAdapter.write(out, jsonObject);
+          writeObjectPermissively(out, jsonObject);
           return;
         }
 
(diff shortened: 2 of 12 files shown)


Marcono1234:
Needs additional `@throws` tags once #2172 is merged

Issue #1708, opened 2020-05-25
ConstructorConstructor.newDefaultImplementationConstructor(Type, Class<? super T>) breaks symmetry


Marcono1234:
The method `com.google.gson.internal.ConstructorConstructor.newDefaultImplementationConstructor(Type, Class<? super T>)` breaks symmetry.

Normally it is desired that if you can serialize `obj` as `A`, then you should be able to deserialize it as `A` again. This is however not possible if `A` extends / implements one of the types checked in `newDefaultImplementationConstructor` and does not have a no-arg constructor.
Example:
```
static class MyList extends ArrayList<Integer>{
  // Any constructor with parameters to prevent default constructor
  public MyList(int i) {}
}
public void testCustomList() {
  Gson gson = new Gson();
  MyList list = new MyList(1);
  list.add(1);
  String json = gson.toJson(list);
  // java.lang.ClassCastException: Cannot cast java.util.ArrayList to MyList
  gson.fromJson(json, MyList.class);
}
```

The proper fix would be to have `newDefaultImplementationConstructor` only create instances if the requested type is the type it creates (e.g. it creates `ArrayList` and `ArrayList` is requested) or if the requested type is a supertype of the created one (e.g. it creates `ArrayList` and `List` is requested).
Similarly `CollectionTypeAdapterFactory` would have to be adjusted to only support these types and types for which a constructor has been registered or types which have a default (= no-arg) constructor.

This would in theory not cause any issues if people already serialized **and** deserialized their objects and therefore the types they are using are matching the above stated requirements.

However, there might be cases where people only serialize objects (such as `MyList` in the example above), which would break with such a change.

Would it make sense to instead introduce an option to `GsonBuilder` to prevent serialization which is not symmetric? So people can catch such issues early?

Marcono1234:
Maybe a duplicate of #419

PR #2152, opened 2022-07-25
Fix LinkedTreeMap being used for non-Comparable keys

### LINKED ISSUE #1247: Confusing Exceptions

Marcono1234:
Fixes #1247

As shown in #1247 there are some corner cases where Gson chooses a `LinkedTreeMap` as `Map` implementation, even though the map keys are `null` or do not implement `Comparable`. This leads to runtime exceptions.

(I have also included a small documentation enhancement for `JsonObject` since it internally uses `LinkedTreeMap` and therefore does not support `null` keys either. That is already covered by existing tests.)


gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -257,6 +257,21 @@ public T construct() {
...
+    if (!(mapType instanceof ParameterizedType)) {
+      return false;
+    }
+
+    Type[] typeArguments = ((ParameterizedType) mapType).getActualTypeArguments();
+    if (typeArguments.length == 0) {
+      return false;
+    }
+    // Consider String and supertypes of it
+    return TypeToken.get(typeArguments[0]).getRawType().isAssignableFrom(String.class);
```
Marcono1234, line 272: Maybe this should also be changed to check the resolved type arguments (in case of subtypes with a different count of type parameters), though the underlying issue is #1708, and `LinkedTreeMap` should only be used if `Map` is requested and not for any subtypes. And in that case there is no need to resolve type arguments.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -257,6 +257,20 @@ public T construct() {
     };
   }
 
+  private static boolean hasStringKeyType(Type mapType) {
+    // If mapType is not parameterized, cannot assume that key is String, because if it
+    // is not String and does not implement Comparable it would lead to a ClassCastException
+    if (!(mapType instanceof ParameterizedType)) {
+      return false;
+    }
```
Marcono1234, line 265: This would prevent `LinkedTreeMap` from being used for raw map types, which should be invisible to the user (since `LnkedTreeMap` is an internal class) but would disable the DoS protection in that case for Java 7. Not sure if that would be acceptable, but on the other hand it prevents potential `ClassCastException`s.

eamonnmcmanus: This is kind of a big change, isn't it? It seems to me that `gson.fromJson(s, Map.class)` is a pretty common idiom and we're changing what it does. Agreed that people shouldn't be relying on getting a `LinkedTreeMap`, but is it something they could observe indirectly? For example because of memory usage. Though it looks to me as if `LinkedTreeMap` is fairly expensive in memory terms because of all those fields in its `Node` class.

I see some other options. I think the whole situation is a bit of a mess to be honest, but here are some ways we *might* be able to clean it up a bit:

1. Keep the approach here: raw `Map` never deserializes into a `LinkedTreeMap`, but some parameterized `Type` instances will. (We can work on adjusting exactly which ones.)
2. Never deserialize into `LinkedTreeMap` here. We would still use the class for `JsonObject` but here we would always use `LinkedHashMap`.
3. Give `LinkedTreeMap` a fallback mode, where if a key is added that is not `Comparable`, or that can't be compared to existing keys, it copies the existing entries into a private `LinkedHashMap` and uses that for all subsequent operations.

I suppose my underlying question is why `LinkedTreeMap` exists in the first place. What problem does it solve? It seems to me that it is usually going to be both bigger and less performant than `LinkedHashMap`, though I haven't measured that.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -257,6 +257,20 @@ public T construct() {
     };
   }
 
+  private static boolean hasStringKeyType(Type mapType) {
```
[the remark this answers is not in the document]
Marcono1234: That is what I mean with https://github.com/google/gson/pull/2152#discussion_r928335419 (unfortunately GitHub showed it as "outdated").
`hasStringKeyType` is really only intended for `Map` and not any subclasses, to determine whether a `LinkedTreeMap` should be used. The issue that it is also called for `Map` subtypes (as shown in your examples) is basically #1708, therefore I am not so keen to add any special logic here.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
Marcono1234: Below where this method `hasStringKeyType` is called, I switched the `if` statement branches, which is why it might be a bit confusing.

I think the intended logic here is:
If it is a `Map<String, ...>` use Gson's `LinkedTreeMap` to protect against potential DoS for JDK's `LinkedHashMap` in Java < 8 ([JDK-8046170](https://bugs.openjdk.org/browse/JDK-8046170)). Otherwise use JDK's `LinkedHashMap`.

The check below was previously (roughly):
```java
if (type instanceof ParameterizedType && !(String.class.isAssignableFrom(
          TypeToken.get(((ParameterizedType) type).getActualTypeArguments()[0]).getRawType()))) {
    ...
} else {
    return new LinkedTreeMap<>();
}
```
For raw `Map.class` and for any parameterized type `Map<String, ...>` it would have used `LinkedTreeMap`.
However, using `LinkedTreeMap` for raw `Map.class` can lead to `ClassCastException`s, as shown by the examples in #1247.

Therefore the changes by this PR actually reduce the cases where `LinkedTreeMap` is used (which was the concern I wanted to express with my original comment here).
But if you think that is fine, then I guess we can leave it like this.

Marcono1234:
@eamonnmcmanus do you think this can be merged (see my responses to your comments above)?

gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
```diff
@@ -164,13 +166,23 @@ public Adapter(Gson context, Type keyType, TypeAdapter<K> keyTypeAdapter,
...
           K key = keyTypeAdapter.read(in);
           V value = valueTypeAdapter.read(in);
+
+          // LinkedTreeMap requires that keys are Comparable, if they are not must switch to
+          // a different map type. LinkedTreeMap is an internal class so switching map type
+          // should not be noticeable for user code.
+          if (isLinkedTreeMap && !(key instanceof Comparable)) {
+            map = new LinkedHashMap<>(map);
+            isLinkedTreeMap = false;
+          }
```
Marcono1234, line 184: If the deserialization code just switches from `LinkedTreeMap` to `LinkedHashMap` automatically, then a malicious user could force this (e.g. by including `[null,false]` as first map entry).

But then the question is if there is really any advantage at all in using `LinkedTreeMap` or if it would make more sense to just always use `LinkedHashMap` (#2737).

Marcono1234:
Obsolete due to #3057

For the other changes in this PR here have created new dedicated PRs:
- #3068
- #3069

PR #2155, opened 2022-07-25
addAll in JsonArray

xtay2:
The JsonArray now supports two addAll methods. These add common functionality as arrays and Collections can now get added in one go.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -18,9 +18,7 @@ package com.google.gson;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 
 /**
  * A class representing an array type in Json. An array is a list of {@link JsonElement}s each of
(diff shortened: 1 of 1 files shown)


Marcono1234:
My main concern with this change is that it "blurs" the API of the project. Currently the API is fairly well grouped: `Gson` and `TypeAdapter` provide the API for converting from POJO to JSON and the other way around; `JsonElement` and subclasses model a JSON document in memory. With this PR `JsonArray` would now suddenly also be responsible for converting POJOs to JSON.

As seen by the changes done by this PR it would also not be that complicated if users wrote the code for this themselves. It might also be possible to shorten it a bit by not having the `instanceof` checks but always calling `toJsonTree`, or possibly by directly calling `toJsonTree(values)` (that is with the `Collection` as argument). That would then also allow the user to use their own `Gson` instance which they can customize.

However, I am not a member of this project and they might think differently about this.

xtay2:
> My main concern with this change is that it "blurs" the API of the project. Currently the API is fairly well grouped: Gson and TypeAdapter provide the API for converting from POJO to JSON and the other way around; JsonElement and subclasses model a JSON document in memory. With this PR JsonArray would now suddenly also be responsible for converting POJOs to JSON.

I dont think that this would be a problem. The method/s is/are very straight-forward and the use-cases would outweight the "confusion".

> As seen by the changes done by this PR it would also not be that complicated if users wrote the code for this themselves.

For me this was an instant issue when first looking at the implementation. This isn't code that every user should have to write. There also exists another unfinished pull request that lets JsonArray implement the List interface. (Which I'm also strongly in favour of.)

> It might also be possible to shorten it a bit by not having the instanceof checks but always calling toJsonTree, or possibly by directly calling toJsonTree(values) (that is with the Collection as argument).

In newer versions of Java this could also be done with a pattern switch-statement, but for me it was important to use the fastest implementation for java 1.5.

>That would then also allow the user to use their own Gson instance which they can customize.
Fair point.

eamonnmcmanus:
It seems odd to me that we would have an `addAll` method that allows any compatible type and uses `instanceof` to figure out what kind of `JsonPrimitive` to add for each element, but we don't have an `add(Object)` method. I agree with @Marcono1234 that this API change doesn't seem very consistent.

I don't think we can accept this PR.

(For future PRs, please note that we don' t use wildcard imports, and that new APIs should come with new tests.)

PR #2159, opened 2022-07-26
Remove usage of oss-parent Maven parent and update `ReleaseProcess.md`

### LINKED ISSUE #1605: Remove oss-parent pom
It's [deprecated](https://github.com/sonatype/oss-parents) and a source of flakes:

[FATAL] Non-resolvable parent POM for com.google.code.gson:gson-parent:2.8.7-SNAPSHOT: Could not transfer artifact org.sonatype.oss:oss-parent:pom:7 from/to google-maven-central (https://maven-central.storage-download.googleapis.com/repos/central/data/): sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target and 'parent.relativePath' points at wrong local POM @ com.google.code.gson:gson-parent:2.8.7-SNAPSHOT, /home/travis/build/google/gson/pom.xml, line 6, column 11
 @

Marcono1234:
Resolves #1605 (even though the bug described there was most likely not casued by oss-parent)
Related: https://github.com/google/gson/pull/2149#issuecomment-1191746895

The oss-parent configuration has been applied manually to the root `pom.xml`, however there are some differences:
- Newer plugin versions are used for source and javadoc JAR creation, and for GPG signing
Note that the GPG plugin currently does not handle a missing GPG binary very well, [throwing a `NullPointerException`](https://issues.apache.org/jira/browse/MGPG-86)
- No snapshot repository for distribution is specified since Gson does not deploy snapshots (as far as I know)
- Have not included Maven enforcer rule to prevent old Maven 2 versions (most plugins probably already require Maven 3 anyways)
- Uses `<releaseProfiles>` option of the Maven Release Plugin. This  apparently only enables the release profile for `release:perform` whereas the previous approach of using the `<arguments>` option with `-P...` seems to have also affected `release:prepare`. This means that Javadoc or signing issues would not be noticed until after `release:perform`. Should that be changed?

See also `oss-parent` POM here: https://repo1.maven.org/maven2/org/sonatype/oss/oss-parent/9/oss-parent-9.pom

This required some reordering of the plugins for the `gson` module to make sure they are executed in the correct order. Otherwise this would cause failures for `javadoc:jar` executed as part of the release.

I have tested this locally a bit by setting the `scm` information and the repository URL to a `file:///` URI and it seems to work as desired (have not tested GPG). Though would be good if you could nonetheless check if these changes are reasonable.

Sonatype also seems to provide their own convenience plugin for publishing to Maven Central, see [guide](https://central.sonatype.org/publish/publish-maven/), maybe that would be useful in the future, but I think at the moment it would complicate migrating away from `oss-parent` a bit.


ReleaseProcess.md
```diff
@@ -8,17 +8,18 @@ The following is a step-by-step procedure for releasing a new version of Google-
...
-1. Update the release version in [Wikipedia](https://en.wikipedia.org/wiki/GSON) and update the current "stable" release.
+1. Download and sanity check all downloads. Do not skip this step! Once you release the staging repository, there is no going back. It will get synced with Maven Central and you will not be able to update or delete anything. Your only recourse will be to release a new version of Gson and hope that no one uses the old one.
+1. Release the staging repository for Gson. Gson will now get synced to Maven Central with-in the next hour. For issues consult [Sonatype Guide](https://central.sonatype.org/publish/release/).
+1. Update [Gson Changelog](CHANGELOG.md). Also, look at all bugs that were fixed and add a few lines describing what changed in the release.
+1. Update version references in (version might be referenced multiple times):
+    - [`README.md`](README.md)
+    - [`UserGuide.md`](UserGuide.md)
+1. Optional: Create a post on the [Gson Discussion Forum](https://groups.google.com/group/google-gson).
+1. Optional: Update the release version in [Wikipedia](https://en.wikipedia.org/wiki/Gson) and update the current "stable" release.
```
Marcono1234, line 22: Have prefixed them with "Optional" because the Google Group does not seem to be that actively used anymore, so updating both of them might not be as important?

eamonnmcmanus:
Sorry I didn't get to this earlier. Would you mind resolving the conflicts? Then I think we can try this out at the next release.

PR #2162, opened 2022-07-29
Remove unused package-private `FieldAttributes` methods

Marcono1234:


eamonnmcmanus:
Hard to imagine that this could cause any problems.

PR #2164, opened 2022-08-03
Add Gradle.kts implementation statement

wilinz:


README.md
```diff
@@ -16,6 +16,13 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 
 ### Download
 
+Gradle.kts:
```
eamonnmcmanus, line 19: I would put this after the existing Gradle one since I think Groovy-Gradle is still commoner.

eamonnmcmanus (changes requested):
Thanks! We do need you to sign the CLA before we can merge this.

ix0rai:
alternatively, this could replace the existing groovy snippet as it works in both versions

PR #2166, opened 2022-08-03
Validate `TypeToken.getParameterized` arguments

Marcono1234:
Adds validation for the `TypeToken.getParameterized` arguments. This prevents creating malformed parameterized types such as `Map<String>` (1 type argument instead of 2) which could later cause confusing issues for type adapters, e.g. `MapTypeAdapterFactory` would fail with `ArrayIndexOutOfBoundsException`.

I hope I have implemented the validation in a correct way, though maybe I am overlooking some valid corner cases which would be rejected with these changes.

### DIFF
--- a/gson/src/main/java/com/google/gson/reflect/TypeToken.java
+++ b/gson/src/main/java/com/google/gson/reflect/TypeToken.java
@@ -16,8 +16,8 @@
 
 package com.google.gson.reflect;
 
-import com.google.gson.internal.$Gson$Types;
 import com.google.gson.internal.$Gson$Preconditions;
+import com.google.gson.internal.$Gson$Types;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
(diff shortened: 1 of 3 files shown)


gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -319,8 +319,46 @@ public static <T> TypeToken<T> get(Class<T> type) {
...
+   *   the raw type
    */
   public static TypeToken<?> getParameterized(Type rawType, Type... typeArguments) {
+    $Gson$Preconditions.checkNotNull(rawType);
+    $Gson$Preconditions.checkNotNull(typeArguments);
+
+    // Perform basic validation here because this is the only public API where users
+    // can create malformed parameterized types
+    if (!(rawType instanceof Class)) {
+      // See also https://bugs.openjdk.org/browse/JDK-8250659
```
eamonnmcmanus, line 334: Grumble. I wish we hadn't duplicated this mistake.

gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -319,8 +319,46 @@ public static <T> TypeToken<T> get(Class<T> type) {
...
+      throw new IllegalArgumentException(rawClass.getName() + " requires " + expectedArgsCount +
+          " type arguments, but got " + actualArgsCount);
+    }
+
+    for (int i = 0; i < expectedArgsCount; i++) {
+      Type typeArgument = typeArguments[i];
+      Class<?> rawTypeArgument = $Gson$Types.getRawType(typeArgument);
+      TypeVariable<?> typeVariable = typeVariables[i];
+
+      for (Type bound : typeVariable.getBounds()) {
```
eamonnmcmanus, line 352: I'm wondering about this. Suppose you have `interface NumberSet<E extends Number> extends Set<E>`. Then Java will allow you to write `NumberSet<?>`, which effectively means `NumberSet<? extends Number>`. Would people expect to be able to write `TypeToken.getParameterized(NumberSet.class, wildcard)` where `wildcard` is a `Type` representing `?`? I may well be overthinking this.

eamonnmcmanus (commented):
LGTM, assuming it passes all of Google's tests, which I am checking.

eamonnmcmanus:
Google tests all pass so I think we are good here.

gson/src/main/java/com/google/gson/reflect/TypeToken.java
Marcono1234: Good point, I did not consider that. Though users probably cannot easily obtain a `WildcardType`, and additionally Gson's current type resolution logic does not seem to handle wildcards that well. It seems to only use their bounds but ignores the bounds of the type variable. So users would most likely run into a `ClassCastException`.

PR #2168, opened 2022-08-04
Clarify that `GsonBuilder.setExclusionStrategies` does not replace existing ones

Marcono1234:
Normally `set...` methods on builders make it sound like the previously specified value is overwritten, especially when there also exist `add...` methods. However, this is not the case here so the documentation should explicitly mention this.

Feedback regarding the wording is appreciated.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -368,6 +368,8 @@ public final class GsonBuilder {
    * deserialization. Each of the {@code strategies} will be applied as a disjunction rule.
    * This means that if one of the {@code strategies} suggests that a field (or class) should be
    * skipped then that field (or object) is skipped during serialization/deserialization.
+   * The strategies are added to the existing strategies (if any), the existing strategies
+   * are not replaced.
    *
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
That's annoying. Obviously the method should have been called `addExclusionStrategies`. I guess this isn't important enough to justify deprecating the `set...` method in favour of a new `add...` one, but certainly this documentation fix is needed.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -368,6 +368,8 @@ public GsonBuilder setNumberToNumberStrategy(ToNumberStrategy numberToNumberStra
    * deserialization. Each of the {@code strategies} will be applied as a disjunction rule.
    * This means that if one of the {@code strategies} suggests that a field (or class) should be
    * skipped then that field (or object) is skipped during serialization/deserialization.
+   * The strategies are added to the existing strategies (if any), the existing strategies
```
eamonnmcmanus, line 371: Punctuation nit: the comma should be a semicolon. :-)

eamonnmcmanus (changes requested):

eamonnmcmanus (approved):

PR #2170, opened 2022-08-05
Convert null to JsonNull for `JsonArray.set`

Marcono1234:
All other methods perform the same implicit conversion.

This is a backward incompatible change, though I hope it does not affect any users, and this current `null` handling is pretty inconsistent.


eamonnmcmanus:
I agree that consistency trumps hypothetical incompatibility here.

PR #2172, opened 2022-08-06
Fix `TypeAdapter.toJson` throwing AssertionError for custom IOException

Marcono1234:
Currently `TypeAdapter.toJson` assumes that no `IOException` can occur because a `StringWriter` is used. However, this assumption is incorrect because the implementation of `write` may manually throw an `IOException`, unrelated to usage of the writer.

### DIFF
--- a/gson/src/main/java/com/google/gson/TypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/TypeAdapter.java
@@ -216,7 +216,7 @@ public abstract class TypeAdapter<T> {
     try {
       toJson(stringWriter, value);
     } catch (IOException e) {
-      throw new AssertionError(e); // No I/O writing to a StringWriter.
+      throw new JsonIOException(e);
     }
     return stringWriter.toString();
   }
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -216,7 +216,7 @@ public final String toJson(T value) {
     try {
       toJson(stringWriter, value);
     } catch (IOException e) {
-      throw new AssertionError(e); // No I/O writing to a StringWriter.
+      throw new JsonIOException(e);
```
eamonnmcmanus, line 219: I guess this means there should be `@throws JsonIOException` in the javadoc?

eamonnmcmanus (changes requested):
Thanks for fixing this subtle problem. Just one minor thing.

gson/src/main/java/com/google/gson/TypeAdapter.java
Marcono1234: Have also added it for the other `TypeAdapter` methods wrapping `IOException`s. I hope the wording is alright.

eamonnmcmanus (approved):

Issue #1578, opened 2019-09-18
2.8.5；NullPointerException；TypeAdapterFactory


geolo123:
java.lang.NullPointerException: Attempt to invoke interface method 'com.google.gson.TypeAdapter com.google.gson.TypeAdapterFactory.create(com.google.gson.Gson, com.google.gson.reflect.TypeToken)' on a null object reference
	at com.google.gson.Gson.getAdapter(Gson.java:458)
	at com.google.gson.Gson.fromJson(Gson.java:926)
	at com.google.gson.Gson.fromJson(Gson.java:892)
	at com.google.gson.Gson.fromJson(Gson.java:841)
	at com.google.gson.Gson.fromJson(Gson.java:813)

geolo123:
help

Omega-Ariston:
Please provide context and paste the code.

aalaprawat:
Iam not able to replicate this issue on my device. But various devices are facing this issue across various android api levels.

`Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'com.google.gson.TypeAdapter com.google.gson.TypeAdapterFactory.create(com.google.gson.Gson, com.google.gson.reflect.TypeToken)' on a null object reference
       at com.google.gson.Gson.getAdapter(Gson.java:458)
       at com.google.gson.Gson.toJson(Gson.java:696)
       at com.google.gson.Gson.toJson(Gson.java:683)
       at com.google.gson.Gson.toJson(Gson.java:638)
       at com.google.gson.Gson.toJson(Gson.java:618)
       at in.workindia.nileshdungarwal.helpers.JsonHelper.convertJobContentsToJson(JsonHelper.java:322)
       at in.workindia.nileshdungarwal.data_parser.ContentValuesParser.parseJobs(ContentValuesParser.java:335)
       at in.workindia.nileshdungarwal.data_parser.ContentValuesParser.parseNewJobs(ContentValuesParser.java:70)
       at in.workindia.nileshdungarwal.retrofit.RetrofitSyncAll.finalProcessOfJobs(RetrofitSyncAll.java:844)
       at in.workindia.nileshdungarwal.retrofit.RetrofitSyncAll$101.onResponse(RetrofitSyncAll.java:6591)
       at retrofit2.DefaultCallAdapterFactory$ExecutorCallbackCall$1.lambda$onResponse$0(DefaultCallAdapterFactory.java:81)
[...]

PR #2179, opened 2022-08-20
Add null checks for GsonBuilder methods

### LINKED ISSUE #1528: Describe that GsonBuilder setFieldNamingPolicy and setFieldNamingStrategy are the same
The methods `GsonBuilder#setFieldNamingPolicy` and `GsonBuilder#setFieldNamingStrategy` both do the same, they assign a value to the `fieldNamingPolicy` field. This is not obvious from the documentation.

It would be good if the documentation clarified that the methods overwrite each other's changes. Possibly `setFieldNamingPolicy` could also be deprecated, because it is redundant, but this might not be a strong enough reason for a deprecation.

Marcono1234:
Currently most `GsonBuilder` methods do not check for `null`. This can lead to `NullPointerException`s being thrown at completely unrelated places when the created `Gson` instance is used. (Might be the reason for #1578.)

Have not added a `null` check to `setDateFormat(String pattern)` because it looks like the pattern can be `null` (at least that is the default value).

Also resolves #1528

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -166,6 +166,7 @@ public final class GsonBuilder {
    * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
    */
   public GsonBuilder excludeFieldsWithModifiers(int... modifiers) {
+    $Gson$Preconditions.checkNotNull(modifiers);
     excluder = excluder.withModifiers(modifiers);
     return this;
   }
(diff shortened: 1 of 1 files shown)


eamonnmcmanus (approved):
Thanks! This looks great.

Separately, perhaps we should get rid of `$Gson$Preconditions.checkNotNull` entirely and replace it with `Objects.requireNonNull` now that we are require Java 7.

Marcono1234:
> Separately, perhaps we should get rid of `$Gson$Preconditions.checkNotNull` entirely and replace it with `Objects.requireNonNull` now that we are require Java 7.

Good point! Have created #2180 for that. (I hope you weren't working on this as well already.)

PR #2180, opened 2022-08-21
Replace `$Gson$Preconditions.checkNotNull` with `Objects.requireNonNull`

Marcono1234:


gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
```diff
@@ -34,13 +34,6 @@
     throw new UnsupportedOperationException();
   }
 
-  public static <T> T checkNotNull(T obj) {
```
eamonnmcmanus, line 37: To my distress I found 3 places in Google's code base that are calling this method, even though it's hard to imagine how it could be more obviously internal. I can fix these, but on the assumption that code elsewhere is doing the same thing, maybe leave the method intact.

eamonnmcmanus (changes requested):
Thanks!

gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
eamonnmcmanus: I'm guessing BTW that these happened because an IDE auto-imported the wrong `checkNotNull`. Which can easily happen to other projects.

eamonnmcmanus (approved):

PR #2181, opened 2022-08-22
Verify that JsonTreeReader and JsonTreeWriter override all methods

Marcono1234:
If those classes do not override one of the `JsonReader` or `JsonWriter` methods the user might encounter an `AssertionError`.

This was originally suggested in https://github.com/google/gson/pull/2132#issuecomment-1154194193.

Related issues:
- #1651
- #2132


eamonnmcmanus (changes requested):
Very helpful, thanks! Just one small suggestion to simplify slightly.

eamonnmcmanus (approved):

PR #2182, opened 2022-08-22
Add GitHub workflow for checking API compatibility

### LINKED ISSUE #2171: Check for binary compatibility during Maven (or CI) build
As seen in https://github.com/google/gson/pull/2169#issuecomment-1207197608, one can easily by accident break binary backward compatibility, for example by changing the return type of a method.

# Feature description
Either as part of a regular Maven build, or at least triggered by the CI, a check for binary compatibility should be performed.
Breaking changes should either cause an error or at least a warning, ideally with GitHub annotation for the code causing the issue. Optionally it might be useful to also generate a report which can be obtained as GitHub workflow artifact, showing the binary API differences. This could be useful to detect accidentially exposed methods or fields.

It appears there are multiple tools and Maven plugins for this; some of them are:
- https://github.com/siom79/japicmp
- https://github.com/revapi/revapi

# Alternatives / workarounds
Do nothing and review pull requests thoroughly to not break binary compatibility by accident?

Marcono1234:
Resolves #2171

This currently only runs as GitHub workflow but not as part of a regular Maven build. It first checks out the pull request base, builds the project and installs the artifacts into the local Maven repository. Then it checks out the pull request head builds it and compares the artifacts with the previous ones.

The API differences report is uploaded as GitHub artifact, an example can be seen here: https://github.com/google/gson/actions/runs/2905297668
Note that the artifact is produced even when there are non-breaking changes, or no changes. Though in these cases the workflow won't fail.

This is mainly a proof-of-concept, I don't know how well this approach will work for real pull requests.

This uses https://github.com/siom79/japicmp. I don't know whether it covers all cases, so it would be good to not rely on it exclusively.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
           java-version: '11'
--- a/.github/workflows/check-api-compatibility.yml
+++ b/.github/workflows/check-api-compatibility.yml
@@ -1,0 +1,50 @@
+name: Check API compatibility
+
+on: pull_request
+
+jobs:
+  check-api-compatibility:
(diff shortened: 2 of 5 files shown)


eamonnmcmanus:
Thanks! This may well protect us from future missteps.

PR #1374, opened 2018-08-23
Remove a raw type.

reudismam:

PR #2183, opened 2022-08-22
Fail Maven build on compiler warnings; remove some warning suppressions

Marcono1234:
Currently this enables all compiler warnings (`-Xlint:all`); I am not sure if that is a good idea. If you want we can also restrict it to [specific warnings](https://docs.oracle.com/en/java/javase/11/tools/javac.html#GUID-AEEC9F07-CB49-4E96-8BC7-BCC2C7F725C9__GUID-C2076F46-1200-4082-9634-06E7F4851EEA) instead. For example warnings for `serial` seem to be redundant here and `varargs` seems to not be very helpful. I am also a bit afraid that the supported and detected warnings depend on the JDK version being used (instead of the `--release` version being specified).

It also looks like `javac` and Eclipse IDE disagree regarding when a warning needs to be suppressed. For some of the required suppressions here Eclipse is telling me "Unnecessary @ SuppressWarnings".

Supersedes #1374

Please let me know what you think.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -136,7 +136,8 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     final boolean jsonAdapterPresent = mapped != null;
     if (mapped == null) mapped = context.getAdapter(fieldType);
 
-    final TypeAdapter<?> typeAdapter = mapped;
+    @SuppressWarnings("unchecked")
+    final TypeAdapter<Object> typeAdapter = (TypeAdapter<Object>) mapped;
     return new ReflectiveTypeAdapterFactory.BoundField(name, serialize, deserialize) {
       @Override void write(JsonWriter writer, Object value)
           throws IOException, IllegalAccessException {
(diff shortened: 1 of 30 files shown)


Marcono1234:
> I am also a bit afraid that the supported and detected warnings depend on the JDK version being used (instead of the `--release` version being specified).

Yes, that can be seen with the failed CIFuzz run:
```
Warning:  COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
Warning:  source value 7 is obsolete and will be removed in a future release
Warning:  target value 7 is obsolete and will be removed in a future release
Warning: [WARNING] To suppress warnings about obsolete options, use -Xlint:-options.
[INFO] 3 warnings 
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  warnings found and -Werror specified
```

eamonnmcmanus (changes requested):
I'm OK with this if you can fix the CIFuzz problem. I *suspect* we may find that `failOnWarning` is more trouble than it's worth but it won't be hard to back just that part out if so.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -126,10 +126,10 @@ private ReflectiveTypeAdapterFactory.BoundField createBoundField(
       final TypeToken<?> fieldType, boolean serialize, boolean deserialize,
       final boolean blockInaccessible) {
     final boolean isPrimitive = Primitives.isPrimitive(fieldType.getRawType());
-    // special casing primitives here saves ~5% on Android...
```
Marcono1234, line 129: Have removed this comment because it is a bit misleading and even when it was [added in 2011](https://github.com/google/gson/commit/4b042671af018f308c2a261145dc6712e39b96e7#diff-98964cf2f1728356669a5e22fba88f24d52a20c6fc974f36c798976cb3fc0ee4R131) it is not really clear to me what exactly this is about, or which special casing was used to check this performance impact.

Marcono1234:
Have excluded the `options` warnings which were causing the failure for CIFuzz, and have also tried to improve the placement for some `@SuppressWarnings` annotations. I hope that is alright.

Though I agree, if `failOnWarning` is causing too much issues we could remove it again.

eamonnmcmanus (approved):

eamonnmcmanus:
This is great, thanks!

PR #2188, opened 2022-08-31
Return self at add() for method-chaining

Osiris-Team:
Enables stuff like this:
```java
new JsonObject()
    .addProperty("key", "value")
    .addProperty("key", "value");
```


Osiris-Team:
So it would break stuff? How?

eamonnmcmanus:
Any application that calls these methods and that has been compiled against a version of Gson before this change would fail when run with a version of Gson after the change. The reason is that the return type of a method is part of its signature at the JVM level, even though that is not true for the Java language.

PR #2192, opened 2022-09-06
updated JsonDeserializer example

### LINKED ISSUE #1425: JsonDeserializer API doc example may be incomplete-cannot resolve symbol 'id'
Hello,
The example provided in [JsonDeserializer API documentation](http://static.javadoc.io/com.google.code.gson/gson/2.8.5/com/google/gson/JsonDeserializer.html) may be incomplete.

Executing the same program giving out compilation error "cannot resolve symbol 'id'".

Please tell me how to get "Id" class's object and it's value.

Thank you.

deeptimarrivada1:
resolves #1425

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonDeserializer.java
+++ b/gson/src/main/java/com/google/gson/JsonDeserializer.java
@@ -51,6 +51,8 @@ import java.lang.reflect.Type;
  * class IdDeserializer implements JsonDeserializer&lt;Id&gt;() {
  *   public Id deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
  *       throws JsonParseException {
+ *         JsonObject jsonObject = jsonElement.getAsJsonObject();
+ *         long idValue = jsonObject.get("value").getAsLong();
  *     return new Id((Class)typeOfT, id.getValue());
  *   }
  * </pre>


deeptimarrivada1:
Sample test method:  @Test
    public void test() {
        String inputString
                = "{\"clazz\":com.foo.MyObject,\"value\":20}";
        IdDeserializer deserializer = new IdDeserializer();
        Gson gson = new GsonBuilder()
                .registerTypeAdapter(Id.class, deserializer)
                .create();

        Id outList = gson.fromJson(inputString, new TypeToken<Id>(){}.getType());
        System.out.println(outList.getValue());
    }

gson/src/main/java/com/google/gson/JsonDeserializer.java
```diff
@@ -51,6 +51,8 @@
  * class IdDeserializer implements JsonDeserializer&lt;Id&gt;() {
  *   public Id deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
  *       throws JsonParseException {
+ *         JsonObject jsonObject = jsonElement.getAsJsonObject();
```
Marcono1234, line 54: Could you please adjust the indentation so that it is the same as for the `return` statement below?

It looks like this example is still not functional; you added a variable `long idValue` but the line below still references an unknown variable `id` and calls a method `getValue()` on it.

deeptimarrivada1: updated

gson/src/main/java/com/google/gson/JsonDeserializer.java
```diff
@@ -51,7 +51,9 @@
  * class IdDeserializer implements JsonDeserializer&lt;Id&gt;() {
```
Marcono1234, line 51: Could you please also adjust this here to fix #626 (the description of that issue is not very great but it seems to be about this)?
```suggestion
 * class IdDeserializer implements JsonDeserializer&lt;Id&gt; {
```

Marcono1234 (commented):
Are you able to sign the CLA? I think it will be necessary for this PR to be merged.

gson/src/main/java/com/google/gson/JsonDeserializer.java
```diff
@@ -51,7 +51,9 @@
...
  *   public Id deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
  *       throws JsonParseException {
- *     return new Id((Class)typeOfT, id.getValue());
+ *     JsonObject jsonObject = json.getAsJsonObject();
+ *     long idValue = jsonObject.get("value").getAsLong();
```
Marcono1234, line 55: Maybe this should use `getAsJsonPrimitive​` to make the logic here a bit stricter, otherwise it would for example also accept a JSON array (see [`JsonArray.getAsLong()`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonArray.html#getAsLong())) which is probably not desired.
```suggestion
 *     long idValue = jsonObject.getAsJsonPrimitive​("value").getAsLong();
```

gson/src/main/java/com/google/gson/JsonDeserializer.java
```diff
@@ -51,7 +51,9 @@
...
+ *     return new Id((Class) typeOfT, idValue);
```
Marcono1234, line 56: Or actually maybe this logic should be:
```suggestion
 *     long idValue = json.getAsJsonPrimitive().getAsLong();
 *     return new Id((Class) typeOfT, idValue);
```
This would match the logic in [`JsonSerializer`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonSerializer.html), and also what the sentences a few line above this code block in the documentation say.

gson/src/main/java/com/google/gson/JsonDeserializer.java
deeptimarrivada1: updated

deeptimarrivada1:
> Are you able to sign the CLA? I think it will be necessary for this PR to be merged.

yes

gson/src/main/java/com/google/gson/JsonDeserializer.java
deeptimarrivada1: Thanks for the suggestion. updated

gson/src/main/java/com/google/gson/JsonDeserializer.java
deeptimarrivada1: we cannot use getAsJsonPrimitive on a Json Object. 
In the example mentioned above, json is ID object and will read as JsonObject and can get variables from that object.

Refer the screenshoot: 
Screen Shot 2022-09-09 at 10 58 33 AM

gson/src/main/java/com/google/gson/JsonDeserializer.java
Marcono1234: Thanks, but maybe it would be better to replace it with what I mentioned in https://github.com/google/gson/pull/2192#discussion_r967005649 (sorry for the confusion).

gson/src/main/java/com/google/gson/JsonDeserializer.java
Marcono1234: Yes that is how the `Id` class would have been serialized without custom serializer, however the sample serializer in the documentation of `JsonSerializer` serializes it as `20`.

And the `JsonDeserializer` documentation a few lines above the code sample you have edited says (emphasis mine):
> The default deserialization of `Id(com.foo.MyObject.class, 20L)` will require the JSON string to be `{"clazz":"com.foo.MyObject","value":20}`. Suppose, you already know the type of the field that the `Id` will be deserialized into, and hence just **want to deserialize it from a JSON string `20`**. You can achieve that by writing a custom deserializer:

deeptimarrivada1:
Sample test: 

 @Test
    public void test() {
        Id id = new Id(Foo.class, 20);
        Gson gson = new GsonBuilder().registerTypeAdapter(Id.class, new IdSerializer()).create();
        Gson gson2 = new GsonBuilder()
                .registerTypeAdapter(Id.class, new IdDeserializer())
                .create();
        Id outList = gson2.fromJson(gson.toJson(id), new TypeToken<Id>(){}.getType());
        System.out.println(outList.getValue());
    }

gson/src/main/java/com/google/gson/JsonDeserializer.java
deeptimarrivada1: updaed

gson/src/main/java/com/google/gson/JsonDeserializer.java
```diff
@@ -48,10 +48,10 @@
...
  *
  * <pre>
- * class IdDeserializer implements JsonDeserializer&lt;Id&gt;() {
+ * class IdDeserializer implements JsonDeserializer&lt;Id&gt; {
  *   public Id deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
  *       throws JsonParseException {
- *     return new Id((Class)typeOfT, id.getValue());
+ *     return new Id((Class) typeOfT, json.getAsJsonPrimitive().getAsLong());
  *   }
  * </pre>
```
Marcono1234, line 56: Sorry, I just noticed the original code is also missing a closing `}`:
```suggestion
 * }
 * </pre>
```

gson/src/main/java/com/google/gson/JsonDeserializer.java
Marcono1234: Thanks! Personally I would have added a separate `long idValue` variable to increase readability, but it does not matter that much. If you like you can change it though.

gson/src/main/java/com/google/gson/JsonDeserializer.java
deeptimarrivada1: updated

Marcono1234 (approved):
Thanks a lot for your work on this!

eamonnmcmanus:
Thanks for doing this!

PR #2193, opened 2022-09-09
Improve documentation

Marcono1234:
Improves Javadoc of multiple classes and adjusts the User Guide; some changes are based on the documentation changes in #2189.

### DIFF
--- a/gson/src/main/java/com/google/gson/ExclusionStrategy.java
+++ b/gson/src/main/java/com/google/gson/ExclusionStrategy.java
@@ -17,11 +17,8 @@
 package com.google.gson;
 
 /**
- * A strategy (or policy) definition that is used to decide whether or not a field or top-level
- * class should be serialized or deserialized as part of the JSON output/input. For serialization,
- * if the {@link #shouldSkipClass(Class)} method returns true then that class or field type
- * will not be part of the JSON output. For deserialization, if {@link #shouldSkipClass(Class)}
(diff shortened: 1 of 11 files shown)


gson/src/main/java/com/google/gson/ExclusionStrategy.java
```diff
@@ -17,11 +17,8 @@
 package com.google.gson;
 
 /**
- * A strategy (or policy) definition that is used to decide whether or not a field or top-level
- * class should be serialized or deserialized as part of the JSON output/input. For serialization,
```
Marcono1234, line 21: Have moved some of this documentation to `GsonBuilder` because there (or rather in the internal `Excluder`) class this logic is implemented.

gson/src/main/java/com/google/gson/JsonPrimitive.java
```diff
@@ -260,6 +264,11 @@ public int hashCode() {
     return value.hashCode();
   }
 
+  /**
+   * Returns whether the other object is equal to this. This method only considers
+   * the other object to be equal if it is an instance of {@code JsonPrimitive} and
+   * has an equal value.
```
Marcono1234, line 270: I did not want to describe in detail how number comparison is performed because there are multiple open issues which show that there are potential flaws with the current logic.

eamonnmcmanus:
This all looks great. Thanks!

PR #2196, opened 2022-09-19
Only create one UnsafeAllocator instance

kb-1000:
This reduces unnecessary allocations, and may potentially even help the JIT (at least for regular JVMs, unsure about Android) optimize the code a little bit more.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -61,6 +61,25 @@ public final class ConstructorConstructor {
     this.reflectionFilters = reflectionFilters;
   }
 
+  /**
+   * Check if the class can be instantiated by Unsafe allocator. If the instance has interface or abstract modifiers
+   * return an exception message.
+   * @param c instance of the class to be checked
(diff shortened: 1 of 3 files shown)


eamonnmcmanus:
Thanks, I agree this seems like a good idea. One small thing: [this line](https://github.com/google/gson/blob/2154e468b490c8a6576db13eca31c9d1afe6bf78/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L113)
```java
    final String exceptionMessage = UnsafeAllocator.checkInstantiable(rawType);
```
causes the static initializer for `UnsafeAllocator` to run and that will now attempt to create an `UnsafeAllocator`. Could you wrap the code-block containing the above line in `if (useJdkUnsafe)`?

Marcono1234:
> Could you wrap the code-block containing the above line in `if (useJdkUnsafe)`?

Maybe it would be better to move the method `checkInstantiable` to `ConstructorConstructor` instead? The point of this check is to also provide a correct exception message if `useJdkUnsafe == false`, see comment above the `checkInstantiable` call.

Marcono1234 (approved):

PR #2197, opened 2022-09-22
Support Java Records when present in JVM.

staale:
Fixes google/gson#1794

Added a TypeAdapterFactory that deals specifcally with Java 17 Records. It uses reflection to detect if the JVM supports records, and from there accesses the RecordComponent array to serialize and deserialize objects.

This new TypeAdapterFactory will only be added when Records are actually supported on the JVM.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -31,6 +31,7 @@ import com.google.gson.internal.bind.JsonTreeWriter;
 import com.google.gson.internal.bind.MapTypeAdapterFactory;
 import com.google.gson.internal.bind.NumberTypeAdapter;
 import com.google.gson.internal.bind.ObjectTypeAdapter;
+import com.google.gson.internal.bind.RecordTypeAdapterFactory;
 import com.google.gson.internal.bind.ReflectiveTypeAdapterFactory;
 import com.google.gson.internal.bind.TypeAdapters;
 import com.google.gson.internal.sql.SqlTypesSupport;
@@ -332,8 +333,16 @@ public final class Gson {
     this.jsonAdapterFactory = new JsonAdapterAnnotationTypeAdapterFactory(constructorConstructor);
     factories.add(jsonAdapterFactory);
     factories.add(TypeAdapters.ENUM_FACTORY);
+
+    // If we are on Java 17, we want to include the RecordTypeAdapterFactory before the ReflectiveTypeAdapterFactory,
+    // so that we intercept all Record types first.
+    if (RecordTypeAdapterFactory.SUPPORTS_RECORD_TYPES) {
+      factories.add(new RecordTypeAdapterFactory(this.excluder, constructorConstructor, jsonAdapterFactory));
+    }
+
     factories.add(new ReflectiveTypeAdapterFactory(
         constructorConstructor, fieldNamingStrategy, excluder, jsonAdapterFactory, reflectionFilters));
+
 
     this.factories = Collections.unmodifiableList(factories);
   }
--- a/gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
@@ -1,0 +1,480 @@
+package com.google.gson.internal.bind;
+
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.annotations.JsonAdapter;
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.internal.ConstructorConstructor;
+import com.google.gson.internal.Excluder;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * {@link RecordFieldFactory} to create adapters for Java 17 records.
+ *
+ * This class makes the following assumptions about records:
+ * - every record field is private final
+ * - for every field there exists a corresponding method with the same name
+ * - the order of the fields and the constructor arguments match exactly
+ * This means that this adapter will also be used for classes that are not records, but fulfill these
+ * requirements. There should be no issues here though, as we will just use constructors to create instances
+ * rather than the Gson Invoke magic.
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
+// unchecked, rawtypes: Unavoidable, as we are dealing with inferred generics.
+public class RecordTypeAdapterFactory implements TypeAdapterFactory {
+
+    public static boolean SUPPORTS_RECORD_TYPES;
+
+    // Find the isRecord method from the Class - using we do not need to handle NoSuchMethodException.
+    private static final RecordHelper RECORD_HELPER;
+
+    static {
+        Method isRecord;
+        try {
+            isRecord = Class.class.getDeclaredMethod("isRecord");
+        }
+        catch (NoSuchMethodException e) {
+            isRecord = null;
+        }
+        RECORD_HELPER = (isRecord == null) ? null : new RecordHelper(isRecord);
+        SUPPORTS_RECORD_TYPES = RECORD_HELPER != null;
+    }
+
+    private final Excluder excluder;
+    private final ConstructorConstructor constructorConstructor;
+    private final JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory;
+
+    public RecordTypeAdapterFactory(Excluder excluder,
+            ConstructorConstructor constructorConstructor,
+            JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory) {
+        this.excluder = excluder;
+        this.constructorConstructor = constructorConstructor;
+        this.jsonAdapterFactory = jsonAdapterFactory;
+    }
+
+    /**
+     * Internal helper class to manage access to Class.isRecord, and RecordComponent instances. Since this compiles
+     * on Java 8, we need to use reflection to handle RecordComponent instances.
+     */
+    private static final class RecordHelper {
+        private final Method isRecord;
+        private final Method getRecordComponents;
+        private final Class<?> recordComponentType;
+        // RecordComponent methods
+        private final Method getName;
+        private final Method getType;
+        private final Method getGenericType;
+        // getAccessor returns a method, that in turn can be invoked on the Record to read out its value
+        private final Method getAccessor;
+        private final Method getAnnotation;
+
+        /**
+         * Create a new RecordHelper to handle reflection for RecordComponent on Java 17.
+         */
+        private RecordHelper(Method isRecord) {
+            this.isRecord = Objects.requireNonNull(isRecord, "isRecord must not be null");
+            try {
+                getRecordComponents = Class.class.getDeclaredMethod("getRecordComponents");
+                recordComponentType = getRecordComponents.getReturnType().getComponentType();
+                getName = recordComponentType.getDeclaredMethod("getName");
+                getType = recordComponentType.getDeclaredMethod("getType");
+                getGenericType = recordComponentType.getDeclaredMethod("getGenericType");
+                getAccessor = recordComponentType.getDeclaredMethod("getAccessor");
+                getAnnotation = recordComponentType.getDeclaredMethod("getAnnotation", Class.class);
+
+            }
+            catch (NoSuchMethodException e) {
+                throw new TypeAdapterReflectionException(
+                        "Expected to find method getRecordComponents when the isRecord method is present in Class", e);
+            }
+        }
+
+        boolean isRecord(TypeToken<?> type) {
+            try {
+                return isRecord != null && Boolean.TRUE.equals(isRecord.invoke(type));
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException("Unable to create TypeAdapter for [" + type + "]", e);
+            }
+        }
+
+        Object[] getRecordComponents(TypeToken<?> type) {
+            try {
+                return (Object[]) getRecordComponents.invoke(type.getRawType());
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException("Unable to invoke getRecordComponents", e);
+            }
+        }
+
+        String getName(Object recordComponent) {
+            try {
+                return (String) getName.invoke(recordComponent);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException("Failed to invoke method [getName] on RecordComponent", e);
+            }
+        }
+        Class<?> getType(Object recordComponent) {
+            try {
+                return (Class<?>) getType.invoke(recordComponent);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException(
+                        "Failed to invoke method [getGenericType] on RecordComponent", e);
+            }
+        }
+        Type getGenericType(Object recordComponent) {
+            try {
+                return (Type) getGenericType.invoke(recordComponent);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException(
+                        "Failed to invoke method [getGenericType] on RecordComponent", e);
+            }
+        }
+        // getAccessor returns a method, that in turn can be invoked on the Record to read out its value
+        Method getAccessor(Object recordComponent) {
+            try {
+                return (Method) getAccessor.invoke(recordComponent);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException("Failed to invoke method [getAccessor] on RecordComponent", e);
+            }
+        }
+        <A extends Annotation> A getAnnotation(Object recordComponent, Class<A> annotation) {
+            try {
+                return (A) getAnnotation.invoke(recordComponent, annotation);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new TypeAdapterReflectionException(
+                        "Failed to invoke method [getAnnotation] on RecordComponent", e);
+            }
+        }
+
+
+        public <T> Field getField(Object recordComponent, TypeToken<T> type) {
+            // There is nothing in the RecordComponent class to access the underlying field, so we do this based
+            // on the component name.
+            String fieldName = getName(recordComponent);
+            try {
+                return type.getRawType().getDeclaredField(fieldName);
+            }
+            catch (NoSuchFieldException e) {
+                throw new TypeAdapterReflectionException(
+                        "Expected to find field [" + fieldName + "] on recordComponent [" + recordComponent + "]"
+                        + " on type [" + type + "]. Somehow there is a discrepancy between the record components"
+                        + " and the fields on the Class.", e);
+            }
+        }
+
+    }
+
+    @Override
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+        // ?: Is this not a record?
+        if (!RECORD_HELPER.isRecord(type)) {
+            // Yes -> Not supported, we return null as per the TypeAdapterFactory contract
+            return null;
+        }
+
+        try {
+            // To construct record fields, we actually need to do some reflection on the Gson instance. This
+            // logic is contained in the RecordFieldFactory.
+            RecordFieldFactory recordFieldFactory =
+                    new RecordFieldFactory(gson, excluder, jsonAdapterFactory, constructorConstructor);
+
+            Object[] recordComponents = RECORD_HELPER.getRecordComponents(type);
+            Class[] componentTypes = new Class[recordComponents.length];
+            for (int i = 0; i < recordComponents.length; i++) {
+                componentTypes[i] = RECORD_HELPER.getType(recordComponents[i]);
+            }
+
+            // Find the canonical constructor on the Record that corresponds to the record components.
+            // There is no method in the Java API to do the equivialent, instead we rely on the constructor and
+            // recordComponent order to be the same. This construct matches the StackOverflow answer here:
+            // https://stackoverflow.com/a/67127067
+            Constructor<T> recordConstructor = (Constructor<T>) type.getRawType().getConstructor(componentTypes);
+
+            RecordField[] recordFields = new RecordField[recordComponents.length];
+            for (int i = 0; i < recordComponents.length; i++) {
+                // We need the field for the Gson Excluder, so that we can correctly determine if a field should
+                // be included or not.
+                Field field = RECORD_HELPER.getField(recordComponents[i], type);
+                recordFields[i] = recordFieldFactory.createRecordField(field, recordComponents[i]);
+            }
+            return new RecordTypeAdapterImpl<>(recordFields, recordConstructor);
+        }
+        catch (NoSuchMethodException e) {
+            // We hit this either because we do not find the expected constructor, or we do not find a method
+            // name that matches the field name. In either of these classes, this is not a record, and we
+            // do not support serialization of it. The contract for TypeAdapterFactory is to return null for
+            // unsupported types. Gson will then attempt other means of creating a TypeAdapter. If all fails, an
+            // exception will be created where Gson was invoked to serialize/deserialize a type.
+            return null;
+        }
+    }
+
+    /**
+     * Factory to construct our own {@link RecordField} instances. This requires some reflection into the internal
+     * of Gson, which has logic to handle the {@link JsonAdapter} annotation. We also need to reflect into Gson
+     * to fetch the {@link Excluder}, that handles the {@link com.google.gson.annotations.Expose},
+     * {@link com.google.gson.annotations.Since} and {@link com.google.gson.annotations.Until} annotations. This
+     * allows us to closely follow the Gson mechanisms wrt. serialization/deserialization.
+     */
+    private static final class RecordFieldFactory {
+
+
+        private final Gson gson;
+        private final Excluder excluder;
+        private final JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory;
+        private final ConstructorConstructor constructorConstructor;
+
+        public RecordFieldFactory(Gson gson, Excluder excluder,
+                JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory,
+                ConstructorConstructor constructorConstructor) {
+            this.gson = gson;
+            this.excluder = excluder;
+            this.jsonAdapterFactory = jsonAdapterFactory;
+            this.constructorConstructor = constructorConstructor;
+        }
+
+        private RecordField createRecordField(Field field, Object recordComponent) {
+            TypeToken<?> fieldType = TypeToken.get(RECORD_HELPER.getGenericType(recordComponent));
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            this.serializeNulls = serializeNulls;
+            this.excludeOnSerialize = excludeOnSerialize;
+            this.excludeOnDeSerialize = excludeOnDeSerialize;
+            this.accessor = accessor;
+            this.serializedName = serializedName;
+            this.deSerializedNames = deSerializedNames;
+            this.typeAdapter = typeAdapter;
+        }
+
+        private void write(JsonWriter out, Object value) throws IOException {
```
eamonnmcmanus, line 335: The names `write` and `read` suggest a kind of symmetry which is a bit misleading, since `write` writes both the name and the value and `read` reads just the value. Perhaps rename them `writeNameAndValue` and `readValue`?

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            }
+            else if (in.peek() != JsonToken.BEGIN_OBJECT) {
+                throw new DeSerializeException("Expecting null or begin object at [" + in.getPath() + "]");
+            }
+            in.beginObject();
+            // This array will hold the value for each json field in the src object that we find, deserialized
+            // using a RecordField, that will also respect all Gson annotations. For fields that are not deserialized,
+            // the default null value will be used.
+            Object[] values = new Object[recordFields.length];
+            nextField:
```
eamonnmcmanus, line 421: I don't think this label is used?

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+                if (fieldIndex != null) {
+                    values[fieldIndex] = recordFields[fieldIndex].read(in);
+                }
+                else {
+                    // Be sure to skip values we do not consume, otherwise the state of the parser will be wrong.
+                    in.skipValue();
+                }
+            }
+            in.endObject();
+            // At this point, we have not verified that all fields are present, we treat absent values as null
```
eamonnmcmanus, line 438: Not sure about this. If any of the missing fields have primitive type then we will get a NullPointerException. It's probably better to throw an exception always rather than just sometimes? Or we need some more logic to supply a zero value for the missing fields.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+    public static class TypeAdapterReflectionException extends RuntimeException {
+
+        private static final long serialVersionUID = -3905048094470359103L;
+
+        public TypeAdapterReflectionException(String message, Throwable cause) {
+            super(message, cause);
+        }
+    }
+
+    public static class DeSerializeAccessException extends RuntimeException {
```
eamonnmcmanus, line 464: `DeserializeAccessException`, and `DeserializeException` below.

eamonnmcmanus (changes requested):
Thanks for doing this! It's great to have it.

Ideally we would use a multi-release jar file and some build magic, so you wouldn't have to do all that reflection. But that supposes quite a lot of mechanism that we don't have. Meanwhile the reflective approach is way better than what we have, which is nothing.

I have a lot of detailed comments, mostly minor. Some high-level comments:

* Could you run [google-java-format](https://github.com/google/google-java-format) over `RecordTypeAdapterFactory`? That's much easier than me quibbling over the various divergences from Google Java Style. You might want to make the change I requested to use `<ul>` in a javadoc comment first, because otherwise g-j-f will make a mess of that comment.
* Ideally we'd have some sort of test. Perhaps it could do `assumeTrue(RecordTypeAdapterFactory.SUPPORTS_RECORD_TYPES)` and then use reflection to make an instance of [`UnixDomainPrincipal`](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.net/jdk/net/UnixDomainPrincipal.html), which seems to be the only record type in the Java 17 API. If we can serialize and deserialize that, that's at least basic test coverage. Later we can think about ways to test more things, like `@SerializedName`, but let's at least do the `UnixDomainPrincipal` thing here.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -332,9 +333,17 @@ public Gson() {
     this.jsonAdapterFactory = new JsonAdapterAnnotationTypeAdapterFactory(constructorConstructor);
     factories.add(jsonAdapterFactory);
     factories.add(TypeAdapters.ENUM_FACTORY);
+
+    // If we are on Java 17, we want to include the RecordTypeAdapterFactory before the ReflectiveTypeAdapterFactory,
+    // so that we intercept all Record types first.
+    if (RecordTypeAdapterFactory.SUPPORTS_RECORD_TYPES) {
+      factories.add(new RecordTypeAdapterFactory(this.excluder, constructorConstructor, jsonAdapterFactory));
```
Marcono1234, line 340: For consistency:
```suggestion
      factories.add(new RecordTypeAdapterFactory(excluder, constructorConstructor, jsonAdapterFactory));
```

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            this.jsonAdapterFactory = jsonAdapterFactory;
+            this.constructorConstructor = constructorConstructor;
+        }
+
+        private RecordField createRecordField(Field field, Object recordComponent) {
+            TypeToken<?> fieldType = TypeToken.get(RECORD_HELPER.getGenericType(recordComponent));
+            JsonAdapter jsonAdapter = RECORD_HELPER.getAnnotation(recordComponent, JsonAdapter.class);
+            TypeAdapter<?> typeAdapter;
+            if (jsonAdapter == null) {
+                typeAdapter = gson.getAdapter(fieldType);
```
Marcono1234, line 262: Similar to `ReflectiveTypeAdapterFactory` this should probably wrap the adapter in a `TypeAdapterRuntimeTypeWrapper`.

Marcono1234 (commented):
Thanks also from my side!

My initial approach would have been to 'hack' this into `ReflectiveTypeAdapterFactory` to avoid duplicating much of its logic, but on the other hand that might have been very messy. However, I had not actually tried that yet and I think your approach is fine as well. We just need to pay attention in the future that `ReflectiveTypeAdapterFactory` and this newly added `RecordTypeAdapterFactory` don't diverge.

I am not a direct member of this project so feel free to consider my comments mostly suggestions (though I hope @eamonnmcmanus agrees with most of them).

Feel free to work on them one by one; you don't have to resolve all at once.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.internal.ConstructorConstructor;
+import com.google.gson.internal.Excluder;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * {@link RecordFieldFactory} to create adapters for Java 17 records.
```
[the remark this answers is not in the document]
eamonnmcmanus: If the code here doesn't work with Java 15 and `--enable-preview`, I think that is absolutely fine. People shouldn't be using that configuration in production.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+                JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory,
+                ConstructorConstructor constructorConstructor) {
+            this.gson = gson;
+            this.excluder = excluder;
+            this.jsonAdapterFactory = jsonAdapterFactory;
+            this.constructorConstructor = constructorConstructor;
+        }
+
+        private RecordField createRecordField(Field field, Object recordComponent) {
+            TypeToken<?> fieldType = TypeToken.get(RECORD_HELPER.getGenericType(recordComponent));
```
Marcono1234, line 258: Probably has to resolve the field type against the parameterized record type (in case the record type is generic), see `ReflectiveTypeAdapterFactory` code:
https://github.com/google/gson/blob/2154e468b490c8a6576db13eca31c9d1afe6bf78/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L205

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+                        fieldType,
+                        jsonAdapter
+                );
+            }
+
+            // Determine the serialized and deserialized names. If there is an SerializedName annotation
+            // present, we need to respect it. This might also mean that multiple keys acts as aliases
+            // when reading Json.
+            SerializedName serializedNameAnnotation =
+                    RECORD_HELPER.getAnnotation(recordComponent, SerializedName.class);
```
Marcono1234, line 277: Not sure if that actually works. `SerializedName` does not have `RECORD_COMPONENT` as target (and cannot have it as target if we want to still compile against Java 7).

However, you can obtain them from the private fields because they are propagated, see [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10.3):
> A component field is annotated with the annotations, if any, that appear on the corresponding record component and whose annotation interfaces are applicable in the field declaration context

eamonnmcmanus: Oh, that's unfortunate. It's still worth adding support for records, but I suppose we won't be able to serialize their members using a name other than their own. One possible workaround would be to look for an annotation on the property method:

```
record Foo(int bar) {
  @SerializedName("baz")
  int bar() {
    return bar;
  }
}
```

Another possibility would be to remove the `@Target` restriction on `@SerializedName`, but that would be a bit sad.

I'm not sure I understand the comment about obtaining the annotation from the private field. If we could get `@SerializedName` onto the record component then it would indeed be copied onto the field, but how would we do that?

Anyway, I'm fine with this initial version not supporting names other than the Java name of the record component.

This is all making me think that we should have tests that run on Java ≥16, presumably using Maven profiles. Again that doesn't have to be in this PR, though perhaps you could add a test and then just exclude it for now. The exclusion could look like [this](https://www.codejava.net/tools/maven/maven-how-to-exclude-tests-from-build) and go [here](https://github.com/google/gson/blob/2154e468b490c8a6576db13eca31c9d1afe6bf78/gson/pom.xml#L85).

Marcono1234: > If we could get `@SerializedName` onto the record component then it would indeed be copied onto the field, but how would we do that?

Java allows to place `@SerializedName` on the record component because it has `FIELD` as target. It won't be available on the component then, but on the backing field.
Just try it out (e.g. with `jshell`):
```java
import java.lang.annotation.*;

@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
@interface MyAnnotation {}

R.class.getRecordComponents()[0].getAnnotations() // empty array
R.class.getDeclaredFields()[0].getAnnotations() // Annotation[1] { @MyAnnotation() }
```

The annotation is also propagated to the accessor method if it has `METHOD` as target, however only if it is _implicit_. If the user explicitly defines the accessor method the annotation won't be propagated. So the more reliable approach would be to obtain the annotation from the field.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            String serializedName;
+            String[] deserializedNames;
+            if (serializedNameAnnotation == null) {
+                serializedName = RECORD_HELPER.getName(recordComponent);
```
Marcono1234, line 281: Maybe this should pass the private field to the `FieldNamingStrategy` to behave similarly to `ReflectiveTypeAdapterFactory`.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    @Override
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+        // ?: Is this not a record?
+        if (!RECORD_HELPER.isRecord(type)) {
+            // Yes -> Not supported, we return null as per the TypeAdapterFactory contract
+            return null;
+        }
+
```
Marcono1234, line 195: Probably this should also check the `ReflectionAccessFilter`:
https://github.com/google/gson/blob/2154e468b490c8a6576db13eca31c9d1afe6bf78/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L104-L108

For example a user might only permit reflection based serialization of classes in a `model` package but disallow (accidental) serialization of classes in an `implementation` package.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            Class[] componentTypes = new Class[recordComponents.length];
+            for (int i = 0; i < recordComponents.length; i++) {
+                componentTypes[i] = RECORD_HELPER.getType(recordComponents[i]);
+            }
+
+            // Find the canonical constructor on the Record that corresponds to the record components.
+            // There is no method in the Java API to do the equivialent, instead we rely on the constructor and
+            // recordComponent order to be the same. This construct matches the StackOverflow answer here:
+            // https://stackoverflow.com/a/67127067
+            Constructor<T> recordConstructor = (Constructor<T>) type.getRawType().getConstructor(componentTypes);
```
Marcono1234, line 212: The implicit canonical constructor has the same visibility as the record class, see https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10.4.
So maybe it would make sense to try make the constructor accessible to support non-public records as well without requiring the user to specify an explicit `public` constructor. Possibly in a similar way to how [`ConstructorConstructor.newDefaultConstructor`](https://github.com/google/gson/blob/2154e468b490c8a6576db13eca31c9d1afe6bf78/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L185) does it (with the `ReflectionAccessFilter` checks).

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+                ConstructorConstructor constructorConstructor) {
+            this.gson = gson;
+            this.excluder = excluder;
+            this.jsonAdapterFactory = jsonAdapterFactory;
+            this.constructorConstructor = constructorConstructor;
+        }
+
+        private RecordField createRecordField(Field field, Object recordComponent) {
+            TypeToken<?> fieldType = TypeToken.get(RECORD_HELPER.getGenericType(recordComponent));
+            JsonAdapter jsonAdapter = RECORD_HELPER.getAnnotation(recordComponent, JsonAdapter.class);
```
Marcono1234, line 259: Similar to https://github.com/google/gson/pull/2197/files#r978712348, I think the annotation has to be obtained from private field instead of the component.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
eamonnmcmanus: Oh, well that's OK. We should do that.

I've now configured the GitHub actions so that builds are done on both JDK 11 and JDK 17. So if we add a reflection-based test it should actually run on JDK 17. Ideally we would instead add a test that has actual records in it, and arrange for that to be excluded on Java 11. That would allow us to test the `@SerializedName` logic.

Marcono1234: Later if desired some of the [tests from my `gson-record-type-adapter-factory`](https://github.com/Marcono1234/gson-record-type-adapter-factory/blob/main/src/test/java/marcono1234/gson/recordadapter/SerializedNameTest.java) can be used. Though not all test cases are relevant for this implementation here.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * {@link RecordFieldFactory} to create adapters for Java 17 records.
+ *
+ * This class makes the following assumptions about records:
+ * - every record field is private final
+ * - for every field there exists a corresponding method with the same name
+ * - the order of the fields and the constructor arguments match exactly
+ * This means that this adapter will also be used for classes that are not records, but fulfill these
```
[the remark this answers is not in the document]
staale: The class has gone through some revisions internally, and these comments no longer hold. will fix.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+ * requirements. There should be no issues here though, as we will just use constructors to create instances
+ * rather than the Gson Invoke magic.
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
+// unchecked, rawtypes: Unavoidable, as we are dealing with inferred generics.
+public class RecordTypeAdapterFactory implements TypeAdapterFactory {
+
+    public static boolean SUPPORTS_RECORD_TYPES;
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+ *
+ * This class makes the following assumptions about records:
+ * - every record field is private final
+ * - for every field there exists a corresponding method with the same name
+ * - the order of the fields and the constructor arguments match exactly
+ * This means that this adapter will also be used for classes that are not records, but fulfill these
+ * requirements. There should be no issues here though, as we will just use constructors to create instances
+ * rather than the Gson Invoke magic.
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
```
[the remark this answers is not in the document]
staale: Fixed.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+                        "Expected to find field [" + fieldName + "] on recordComponent [" + recordComponent + "]"
+                        + " on type [" + type + "]. Somehow there is a discrepancy between the record components"
+                        + " and the fields on the Class.", e);
+            }
+        }
+
+    }
+
+    @Override
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
```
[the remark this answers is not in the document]
staale: Done.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new RuntimeException(
+                        "Failed to serialize field [" + accessor + "] on [" + value + "]", e);
+            }
+        }
+
+        private Object read(JsonReader in) throws IOException {
+            if (excludeOnDeSerialize) {
+                // Since we are always in an object context, this is a safe way to skip values
+                // we do not wish to de serialize.
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
staale: Done.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
staale: Fixed.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
staale: Fixed to initialize some default values, to ensure we have non-null values for primitives.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
staale: Fixed.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+            else if (in.peek() != JsonToken.BEGIN_OBJECT) {
+                throw new DeSerializeException("Expecting null or begin object at [" + in.getPath() + "]");
+            }
+            in.beginObject();
+            // This array will hold the value for each json field in the src object that we find, deserialized
+            // using a RecordField, that will also respect all Gson annotations. For fields that are not deserialized,
+            // the default null value will be used.
+            Object[] values = new Object[recordFields.length];
+            nextField:
+            while (in.peek() != JsonToken.END_OBJECT) {
```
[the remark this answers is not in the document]
staale: Fixed.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,472 @@
...
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * {@link RecordFieldFactory} to create adapters for Java 17 records.
+ *
+ * This class makes the following assumptions about records:
+ * - every record field is private final
```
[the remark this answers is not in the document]
staale: Fixed

staale:
> Ideally we would use a multi-release jar file and some build magic, so you wouldn't have to do all that reflection. But that supposes quite a lot of mechanism that we don't have. Meanwhile the reflective approach is way better than what we have, which is nothing.

I think this class is small enough to not impose any costs. Multi release-jars are a bit of a pain IMHO., harder to onboard new developers with multiple source trees, harder to reason about when files are spread out, and one more consideration for anyone consuming the library, including if they update their own build from for example Java 11 to 17.
> 
> I have a lot of detailed comments, mostly minor. Some high-level comments:
> 
>     * Could you run [google-java-format](https://github.com/google/google-java-format) over `RecordTypeAdapterFactory`? That's much easier than me quibbling over the various divergences from Google Java Style. You might want to make the change I requested to use `<ul>` in a javadoc comment first, because otherwise g-j-f will make a mess of that comment.

I have fixed this and your other comments, as well as run Google code format.

>     * Ideally we'd have some sort of test. Perhaps it could do `assumeTrue(RecordTypeAdapterFactory.SUPPORTS_RECORD_TYPES)` and then use reflection to make an instance of [`UnixDomainPrincipal`](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.net/jdk/net/UnixDomainPrincipal.html), which seems to be the only record type in the Java 17 API. If we can serialize and deserialize that, that's at least basic test coverage. Later we can think about ways to test more things, like `@SerializedName`, but let's at least do the `UnixDomainPrincipal` thing here.
I will look into this. But this would require to run the build on Java 17 unless I am mistaken. There is something called toolchains for Maven that supports different compilers, and could be set up to use Java 17 on tests. However, the plugin does not support download, and this would require additional configuration on each user by adding a toolchain.xml. to their .m2 folder. This is actually a bit easier on Gradle (where I have done something similar), as it can actually download the required jdks from public sources.

staale:
> Thanks also from my side!
> 
> My initial approach would have been to 'hack' this into `ReflectiveTypeAdapterFactory` to avoid duplicating much of its logic, but on the other hand that might have been very messy. However, I had not actually tried that yet and I think your approach is fine as well. We just need to pay attention in the future that `ReflectiveTypeAdapterFactory` and this newly added `RecordTypeAdapterFactory` don't diverge.
> 

I have created a new pull request with that solution instead: https://github.com/google/gson/pull/2201
That approach is different enough that it I feel a new pull request is more appropriate, and the better solution can be made the final solution. I did start that one from scratch, but it is actually less code. This solution was developed outside the Gson codebase, so I created it without reusing the internals in Gson.

I have not looked at your code comments yet, but will get to those later in this pull request.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,480 @@
...
+ * - for every field there exists a corresponding method with the same name
+ * - the order of the fields and the constructor arguments match exactly
```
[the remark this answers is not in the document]
staale: Fixed in last pull request to actually match implementation :). Old comments that have not been updated.

Marcono1234:
> I have created a new pull request with that solution instead: #2201 That approach is different enough that it I feel a new pull request is more appropriate, and the better solution can be made the final solution. I did start that one from scratch, but it is actually less code. This solution was developed outside the Gson codebase, so I created it without reusing the internals in Gson.
> 
> I have not looked at your code comments yet, but will get to those later in this pull request.

Thanks! That pull request looks a bit more promising in my opinion because it avoids duplicating `ReflectiveTypeAdapterFactory`'s logic. Do you want to continue working on both nonetheless (in case you have the time), or primarily on the new pull request now?

If you only want to continue working on the new pull request, it might be good at the end to have a look at the review comments here again nonetheless; maybe some of them also apply to the new pull request.

gson/src/main/java/com/google/gson/internal/bind/RecordTypeAdapterFactory.java
```diff
@@ -0,0 +1,514 @@
...
+
+      while (in.peek() != JsonToken.END_OBJECT) {
+        String fieldName = in.nextName();
+        // Check if we know which field this name should deserialize to. If we do not find a
+        // matching index, then this field is one that is not present in our record, and ignored. We
+        // do not treat unknown object fields as an error, because the version of the class we have
+        // might differ from the source.
+        Integer fieldIndex = recordFieldIndex.get(fieldName);
+        if (fieldIndex != null) {
+          values[fieldIndex] = recordFields[fieldIndex].readValueForKey(in);
```
gbjordal, line 461: Hi, I was also working on a record type adapter when I came across this PR. 
After playing around a bit with it, I think there is a missing null check here:
```suggestion
          Object value = recordFields[fieldIndex].readValueForKey(in);
          if(value != null){
              values[fieldIndex] = value
          }
```

Without this, the code overrides the default values for primitives with null.

Marcono1234: Possibly the `null` check should only be applied if the component type is a primitive to allow duplicate non-primitive members to overwrite each others values, to be consistent with `ReflectiveTypeAdapterFactory`'s behavior.

staale: > If you only want to continue working on the new pull request, it might be good at the end to have a look at the review comments here again nonetheless; maybe some of them also apply to the new pull request.

I think the new pull request is the way to go, as it is a bit more integrated with the normal reflection flow. I will still check these reviews to see if any apply. Thanks for the input.

staale:
Closing this in favor of https://github.com/google/gson/pull/2201 - which is a bit more integrated with the rest of the Gson codebase.

PR #2198, opened 2022-09-23
Build on JDK 17 as well as 11.

eamonnmcmanus:
Also fix some HTML javadoc warnings that JDK 17 chokes on. It really doesn't like you to write `<h3>` if the previous heading was `<h1>`.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -221,7 +221,7 @@ public final class GsonBuilder {
    * on the key; however, when this is called then one of the following cases
    * apply:
    *
-   * <h3>Maps as JSON objects</h3>
+   * <h4>Maps as JSON objects</h4>
    * For this case, assume that a type adapter is registered to serialize and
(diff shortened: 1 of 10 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -221,8 +221,9 @@ public GsonBuilder serializeNulls() {
    * on the key; however, when this is called then one of the following cases
    * apply:
    *
-   * <h3>Maps as JSON objects</h3>
-   * For this case, assume that a type adapter is registered to serialize and
+   * <p><b>Maps as JSON objects</b>
```
Marcono1234, line 224: The proper heading would be `<h4>`, but probably due to [JDK-8220379](https://bugs.openjdk.org/browse/JDK-8220379) `javadoc` of JDK 11 does not permit using it. I guess using `<b>` for now is fine though.

eamonnmcmanus: Yes, `<h3>` was rejected on JDK 11 and `<h4>` on JDK 17 (or maybe the other way around), so I did this annoying thing instead.

PR #2200, opened 2022-09-24
Fix incorrect HTML headings in Javadoc

Marcono1234:
Follow-up for #2198

It looks like in that pull request only the first incorrect heading was updated, but headings which should be at the same level were not updated, causing them to become sub-headings.


gson/src/main/java/com/google/gson/TypeAdapterFactory.java
```diff
@@ -22,7 +22,8 @@
  * Creates type adapters for set of related types. Type adapter factories are
  * most useful when several types share similar structure in their JSON form.
  *
- * <h2>Example: Converting enums to lowercase</h2>
+ * <h2>Examples</h2>
```
Marcono1234, line 25: Have added a new parent heading here to group the examples a bit, I hope that is fine.

eamonnmcmanus:
Thanks, that looks great!

PR #2201, opened 2022-09-25
Support Java Records when present in JVM.

staale:
Fixes google/gson#1794

Added support in the ReflectionHelper to detect if a class is a record on the JVM (via reflection), and if so, we will create a special RecordAdapter to deserialize records, using the canoncial constructor.

The ReflectionTypeAdapterFactory had to be refactored a bit to support this. The Adapter class inside the factory is now abstract, with concrete implementations for normal field reflection and for Records. The common code is in the Adapter, with each implementation deserializing values into an intermediary object.

For the FieldReflectionAdapter, the intermediary is actually the final result, and field access is used to write to fields as before. For the RecordAdapter the intermediary is the Object[] to pass to the Record constructor.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -227,8 +227,10 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         }
         // The accessor method is only used for records. If the type is a record, we will read out values
         // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
+        // If there is a static field on a record, there will not be an accessor. Instead we will use the default
+        // field logic for dealing with statics.
         Method accessor = null;
-        if (isRecord) {
+        if (isRecord && !Modifier.isStatic(field.getModifiers())) {
           accessor = ReflectionHelper.getAccessor(raw, field);
         }
 
@@ -264,18 +266,18 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
   static abstract class BoundField {
     final String name;
     /** Name of the underlying field */
-    final String componentName;
+    final String fieldName;
     final boolean serialized;
     final boolean deserialized;
 
-    protected BoundField(String name, String componentName, boolean serialized, boolean deserialized) {
+    protected BoundField(String name, String fieldName, boolean serialized, boolean deserialized) {
       this.name = name;
-      this.componentName = componentName;
+      this.fieldName = fieldName;
       this.serialized = serialized;
       this.deserialized = deserialized;
     }
 
-    /** Read this field value from the source, and append its json value to the writer */
+    /** Read this field value from the source, and append its JSON value to the writer */
     abstract void write(JsonWriter writer, Object source) throws IOException, ReflectiveOperationException;
 
     /** Read the value into the target array, used to provide constructor arguments for records */
@@ -291,7 +293,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
    * <p>The {@link RecordAdapter} is a special case to handle records for JVMs that support it, for
    * all other types we use the {@link FieldReflectionAdapter}. This class encapsulates the common
    * logic for serialization and deserialization. During deserialization, we construct an
-   * accumulator A, which we use to accumulate values from the source Json. After the object has been read in
+   * accumulator A, which we use to accumulate values from the source JSON. After the object has been read in
    * full, the {@link #finalize(Object)} method is used to convert the accumulator to an instance
    * of T.
    *
@@ -357,7 +359,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     /** Create the Object that will be used to collect each field value */
     abstract A createAccumulator();
     /**
-     * Read a single Bounded field into the accumulator. The JsonReader will be pointed at the
+     * Read a single BoundedField into the accumulator. The JsonReader will be pointed at the
      * start of the value for the BoundField to read from.
      */
     abstract void readField(A accumulator, JsonReader in, BoundField field)
@@ -396,15 +398,18 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     private final Constructor<? super T> constructor;
     // Array of arguments to the constructor, initialized with default values for primitives
     private final Object[] constructorArgsDefaults;
-    // Map from field names to index into the constructors arguments.
+    // Map from component names to index into the constructors arguments.
     private final Map<String, Integer> componentIndices = new HashMap<>();
 
     RecordAdapter(Class<? super T> raw, Map<String, BoundField> boundFields) {
       super(boundFields);
       this.constructor = ReflectionHelper.getCanonicalRecordConstructor(raw);
-      String[] recordFields = ReflectionHelper.getRecordComponentNames(raw);
-      for (int i = 0; i < recordFields.length; i++) {
-        componentIndices.put(recordFields[i], i);
+      // Ensure the constructor is accessible
+      ReflectionHelper.makeAccessible(this.constructor);
+
+      String[] componentNames = ReflectionHelper.getRecordComponentNames(raw);
+      for (int i = 0; i < componentNames.length; i++) {
+        componentIndices.put(componentNames[i], i);
       }
       Class<?>[] parameterTypes = constructor.getParameterTypes();
 
@@ -429,7 +434,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
 
     @Override
     void readField(Object[] accumulator, JsonReader in, BoundField field) throws IOException {
-      Integer fieldIndex = componentIndices.get(field.componentName);
+      Integer fieldIndex = componentIndices.get(field.fieldName);
       if (fieldIndex == null) {
         throw new IllegalStateException(
             "Could not find the index in the constructor "
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -40,6 +40,7 @@ import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.lang.reflect.Array;
@@ -106,12 +107,6 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       return null; // it's a primitive!
     }
 
-    if (ReflectionHelper.isRecord(raw)) {
-      // We can assume blockInaccessible here, because it will never kick inn. We are not doing
-      // direct field access, and are instead constructing records via their canonical constructor.
-      return new RecordAdapter<>(raw, getBoundFields(gson, type, raw, true));
-    }
-
     FilterResult filterResult =
         ReflectionAccessFilterHelper.getFilterResult(reflectionFilters, raw);
     if (filterResult == FilterResult.BLOCK_ALL) {
@@ -122,8 +117,14 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     }
     boolean blockInaccessible = filterResult == FilterResult.BLOCK_INACCESSIBLE;
 
+    // If the type is actually a Java Record, we need to use the RecordAdapter instead. This will always be false
+    // on JVMs that do not support records.
+    if (ReflectionHelper.isRecord(raw)) {
+      return new RecordAdapter<>(raw, getBoundFields(gson, type, raw, true, true));
+    }
+
     ObjectConstructor<T> constructor = constructorConstructor.get(type);
-    return new FieldReflectionAdapter<>(constructor, getBoundFields(gson, type, raw, blockInaccessible));
+    return new FieldReflectionAdapter<>(constructor, getBoundFields(gson, type, raw, blockInaccessible, false));
   }
 
   private static void checkAccessible(Object object, Field field) {
@@ -136,7 +137,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
   }
 
   private ReflectiveTypeAdapterFactory.BoundField createBoundField(
-      final Gson context, final Field field, final String name,
+      final Gson context, final Field field, final Method accessor, final String name,
       final TypeToken<?> fieldType, boolean serialize, boolean deserialize,
       final boolean blockInaccessible) {
     final boolean isPrimitive = Primitives.isPrimitive(fieldType.getRawType());
@@ -152,16 +153,18 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
 
     @SuppressWarnings("unchecked")
     final TypeAdapter<Object> typeAdapter = (TypeAdapter<Object>) mapped;
-    return new ReflectiveTypeAdapterFactory.BoundField(name, serialize, deserialize) {
-      @Override void write(JsonWriter writer, Object value)
-          throws IOException, IllegalAccessException {
+    return new ReflectiveTypeAdapterFactory.BoundField(name, field.getName(), serialize, deserialize) {
+      @Override void write(JsonWriter writer, Object source)
+          throws IOException, ReflectiveOperationException {
         if (!serialized) return;
-        if (blockInaccessible) {
-          checkAccessible(value, field);
+        if (blockInaccessible && accessor == null) {
+          checkAccessible(source, field);
         }
 
-        Object fieldValue = field.get(value);
-        if (fieldValue == value) {
+        Object fieldValue = (accessor != null)
+          ? accessor.invoke(source)
+          : field.get(source);
+        if (fieldValue == source) {
           // avoid direct recursion
           return;
         }
@@ -172,25 +175,29 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       }
 
       @Override
-      Object read(JsonReader reader) throws IOException {
-        return typeAdapter.read(reader);
+      void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException {
+        Object fieldValue = typeAdapter.read(reader);
+        if (fieldValue != null || !isPrimitive) {
+          target[index] = fieldValue;
+        }
       }
 
       @Override
-      void read(JsonReader reader, Object value)
+      void readIntoField(JsonReader reader, Object target)
           throws IOException, IllegalAccessException {
         Object fieldValue = typeAdapter.read(reader);
         if (fieldValue != null || !isPrimitive) {
           if (blockInaccessible) {
-            checkAccessible(value, field);
+            checkAccessible(target, field);
           }
-          field.set(value, fieldValue);
+          field.set(target, fieldValue);
         }
       }
     };
   }
 
-  private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type, Class<?> raw, boolean blockInaccessible) {
+  private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type, Class<?> raw,
+                                                 boolean blockInaccessible, boolean isRecord) {
     Map<String, BoundField> result = new LinkedHashMap<>();
     if (raw.isInterface()) {
       return result;
@@ -218,8 +225,17 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         if (!serialize && !deserialize) {
           continue;
         }
+        // The accessor method is only used for records. If the type is a record, we will read out values
+        // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
+        Method accessor = null;
+        if (isRecord) {
+          accessor = ReflectionHelper.getAccessor(raw, field);
+        }
 
-        // If blockInaccessible, skip and perform access check later
+        // If blockInaccessible, skip and perform access check later. When constructing a BoundedField for a Record
+        // field, blockInaccessible is always true, thus makeAccessible will never get called. This is not an issue
+        // though, as we will use the accessor method instead for reading record fields, and the constructor for
+        // writing fields.
         if (!blockInaccessible) {
           ReflectionHelper.makeAccessible(field);
         }
@@ -229,7 +245,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         for (int i = 0, size = fieldNames.size(); i < size; ++i) {
           String name = fieldNames.get(i);
           if (i != 0) serialize = false; // only serialize the default name
-          BoundField boundField = createBoundField(context, field, name,
+          BoundField boundField = createBoundField(context, field, accessor, name,
               TypeToken.get(fieldType), serialize, deserialize, blockInaccessible);
           BoundField replaced = result.put(name, boundField);
           if (previous == null) previous = replaced;
@@ -247,35 +263,42 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
 
   static abstract class BoundField {
     final String name;
+    /** Name of the underlying field */
+    final String componentName;
     final boolean serialized;
     final boolean deserialized;
 
-    protected BoundField(String name, boolean serialized, boolean deserialized) {
+    protected BoundField(String name, String componentName, boolean serialized, boolean deserialized) {
       this.name = name;
+      this.componentName = componentName;
       this.serialized = serialized;
       this.deserialized = deserialized;
     }
 
-    abstract void write(JsonWriter writer, Object value) throws IOException, IllegalAccessException;
+    /** Read this field value from the source, and append its json value to the writer */
+    abstract void write(JsonWriter writer, Object source) throws IOException, ReflectiveOperationException;
 
-    abstract Object read(JsonReader reader) throws IOException;
+    /** Read the value into the target array, used to provide constructor arguments for records */
+    abstract void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException;
 
-    abstract void read(JsonReader reader, Object value) throws IOException, IllegalAccessException;
+    /** Read the value from the reader, and set it on the corresponding field on target via reflection */
+    abstract void readIntoField(JsonReader reader, Object target) throws IOException, IllegalAccessException;
   }
 
   /**
    * Base class for Adapters produced by this factory.
    *
-   * <p>The {@link RecordAdapter} is a special case to handle Java 17 and later records, for all
-   * other types we use the {@link FieldReflectionAdapter}. This class encapuslates the common logic
-   * for serialization and deserialization. During deserialization, we construct an intermediary
-   * ResultHolder R, which we append values to. After the object has been read in full, the {@link
-   * #finalize(Object)} method is used to convert the intermediary to an instance of T.
+   * <p>The {@link RecordAdapter} is a special case to handle records for JVMs that support it, for
+   * all other types we use the {@link FieldReflectionAdapter}. This class encapsulates the common
+   * logic for serialization and deserialization. During deserialization, we construct an
+   * accumulator A, which we use to accumulate values from the source Json. After the object has been read in
+   * full, the {@link #finalize(Object)} method is used to convert the accumulator to an instance
+   * of T.
    *
    * @param <T> type of objects that this Adapter creates.
-   * @param <I> intermediary type used to build the deserialization result.
+   * @param <A> type of accumulator used to build the deserialization result.
    */
-  abstract static class Adapter<T, I> extends TypeAdapter<T> {
+  abstract static class Adapter<T, A> extends TypeAdapter<T> {
     protected final Map<String, BoundField> boundFields;
 
     protected Adapter(Map<String, BoundField> boundFields) {
@@ -296,6 +319,8 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         }
       } catch (IllegalAccessException e) {
         throw ReflectionHelper.createExceptionForUnexpectedIllegalAccess(e);
+      } catch (ReflectiveOperationException e) {
+        throw ReflectionHelper.createExceptionForRecordReflectionException(e);
       }
       out.endObject();
     }
@@ -307,7 +332,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         return null;
       }
 
-      I intermediary = createIntermediary();
+      A accumulator = createAccumulator();
 
       try {
         in.beginObject();
@@ -317,7 +342,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
           if (field == null || !field.deserialized) {
             in.skipValue();
           } else {
-            readField(intermediary, in, field);
+            readField(accumulator, in, field);
           }
         }
       } catch (IllegalStateException e) {
@@ -326,18 +351,19 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         throw ReflectionHelper.createExceptionForUnexpectedIllegalAccess(e);
       }
       in.endObject();
-      return finalize(intermediary);
+      return finalize(accumulator);
     }
 
-    // Create the Object that will be used to collect each field value
-    abstract I createIntermediary();
-    // Read a single Bounded field into the intermediary. The JsonReader will be pointed at the
-    // start of the value
-    // for the BoundField to read from.
-    abstract void readField(I intermediary, JsonReader in, BoundField field)
+    /** Create the Object that will be used to collect each field value */
+    abstract A createAccumulator();
+    /**
+     * Read a single Bounded field into the accumulator. The JsonReader will be pointed at the
+     * start of the value for the BoundField to read from.
+     */
+    abstract void readField(A accumulator, JsonReader in, BoundField field)
         throws IllegalAccessException, IOException;
-    // Convert the intermediary to a final instance of T.
-    abstract T finalize(I intermediary);
+    /** Convert the accumulator to a final instance of T. */
+    abstract T finalize(A accumulator);
   }
 
   private static final class FieldReflectionAdapter<T> extends Adapter<T, T> {
@@ -349,19 +375,19 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     }
 
     @Override
-    T createIntermediary() {
+    T createAccumulator() {
       return constructor.construct();
     }
 
     @Override
-    void readField(T intermediary, JsonReader in, BoundField field)
+    void readField(T accumulator, JsonReader in, BoundField field)
         throws IllegalAccessException, IOException {
-      field.read(in, intermediary);
+      field.readIntoField(in, accumulator);
     }
 
     @Override
-    T finalize(T intermediary) {
-      return intermediary;
+    T finalize(T accumulator) {
+      return accumulator;
     }
   }
 
@@ -371,34 +397,39 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     // Array of arguments to the constructor, initialized with default values for primitives
     private final Object[] constructorArgsDefaults;
     // Map from field names to index into the constructors arguments.
-    private final Map<String, Integer> fieldIndices = new HashMap<>();
+    private final Map<String, Integer> componentIndices = new HashMap<>();
 
     RecordAdapter(Class<? super T> raw, Map<String, BoundField> boundFields) {
       super(boundFields);
       this.constructor = ReflectionHelper.getCanonicalRecordConstructor(raw);
-      String[] recordFields = ReflectionHelper.recordFields(raw);
+      String[] recordFields = ReflectionHelper.getRecordComponentNames(raw);
       for (int i = 0; i < recordFields.length; i++) {
-        fieldIndices.put(recordFields[i], i);
+        componentIndices.put(recordFields[i], i);
       }
       Class<?>[] parameterTypes = constructor.getParameterTypes();
+
+      // We need to ensure that we are passing non-null values to primitive fields in the constructor. To do this,
+      // we create an Object[] where all primitives are initialized to non-null values.
       constructorArgsDefaults = new Object[parameterTypes.length];
       for (int i = 0; i < parameterTypes.length; i++) {
         if (parameterTypes[i].isPrimitive()) {
+          // Voodoo magic, we create a new instance of this primitive type using reflection via an
+          // array. The array has 1 element, that of course will be initialized to the primitives
+          // default value. We then retrieve this value back from the array to get the properly
+          // initialized default value for the primitve type.
           constructorArgsDefaults[i] = Array.get(Array.newInstance(parameterTypes[i], 1), 0);
         }
       }
     }
 
     @Override
-    Object[] createIntermediary() {
-      Object[] constructorArgs = new Object[constructorArgsDefaults.length];
-      System.arraycopy(constructorArgsDefaults, 0, constructorArgs, 0, constructorArgs.length);
-      return constructorArgs;
+    Object[] createAccumulator() {
+      return constructorArgsDefaults.clone();
     }
 
     @Override
-    void readField(Object[] intermediary, JsonReader in, BoundField field) throws IOException {
-      Integer fieldIndex = fieldIndices.get(field.name);
+    void readField(Object[] accumulator, JsonReader in, BoundField field) throws IOException {
+      Integer fieldIndex = componentIndices.get(field.componentName);
       if (fieldIndex == null) {
         throw new IllegalStateException(
             "Could not find the index in the constructor "
@@ -410,17 +441,17 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
                 + " same names as the fields in the Java class, and that the order of the"
                 + " RecordComponents is the same as the order of the canonical arguments.");
       }
-      intermediary[fieldIndex] = field.read(in);
+      field.readIntoArray(in, fieldIndex, accumulator);
     }
 
     @Override
     @SuppressWarnings("unchecked")
-    T finalize(Object[] intermediary) {
+    T finalize(Object[] accumulator) {
       try {
-        return (T) constructor.newInstance(intermediary);
+        return (T) constructor.newInstance(accumulator);
       } catch (ReflectiveOperationException e) {
         throw new RuntimeException(
-            "Failed to invoke " + constructor + " with args " + Arrays.toString(intermediary), e);
+            "Failed to invoke " + constructor + " with args " + Arrays.toString(accumulator), e);
       }
     }
   }
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -38,12 +38,16 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
+import java.lang.reflect.Array;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -94,22 +98,32 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     return fieldNames;
   }
 
-  @Override public <T> TypeAdapter<T> create(Gson gson, final TypeToken<T> type) {
+  @Override
+  public <T> TypeAdapter<T> create(Gson gson, final TypeToken<T> type) {
     Class<? super T> raw = type.getRawType();
 
     if (!Object.class.isAssignableFrom(raw)) {
       return null; // it's a primitive!
     }
 
-    FilterResult filterResult = ReflectionAccessFilterHelper.getFilterResult(reflectionFilters, raw);
+    if (ReflectionHelper.isRecord(raw)) {
+      // We can assume blockInaccessible here, because it will never kick inn. We are not doing
+      // direct field access, and are instead constructing records via their canonical constructor.
+      return new RecordAdapter<>(raw, getBoundFields(gson, type, raw, true));
+    }
+
+    FilterResult filterResult =
+        ReflectionAccessFilterHelper.getFilterResult(reflectionFilters, raw);
     if (filterResult == FilterResult.BLOCK_ALL) {
-      throw new JsonIOException("ReflectionAccessFilter does not permit using reflection for "
-          + raw + ". Register a TypeAdapter for this type or adjust the access filter.");
+      throw new JsonIOException(
+          "ReflectionAccessFilter does not permit using reflection for "
+              + raw
+              + ". Register a TypeAdapter for this type or adjust the access filter.");
     }
     boolean blockInaccessible = filterResult == FilterResult.BLOCK_INACCESSIBLE;
 
     ObjectConstructor<T> constructor = constructorConstructor.get(type);
-    return new Adapter<>(constructor, getBoundFields(gson, type, raw, blockInaccessible));
+    return new FieldReflectionAdapter<>(constructor, getBoundFields(gson, type, raw, blockInaccessible));
   }
 
   private static void checkAccessible(Object object, Field field) {
@@ -156,7 +170,14 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
             : new TypeAdapterRuntimeTypeWrapper<>(context, typeAdapter, fieldType.getType());
         t.write(writer, fieldValue);
       }
-      @Override void read(JsonReader reader, Object value)
+
+      @Override
+      Object read(JsonReader reader) throws IOException {
+        return typeAdapter.read(reader);
+      }
+
+      @Override
+      void read(JsonReader reader, Object value)
           throws IOException, IllegalAccessException {
         Object fieldValue = typeAdapter.read(reader);
         if (fieldValue != null || !isPrimitive) {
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+    // Map from field names to index into the constructors arguments.
+    private final Map<String, Integer> fieldIndices = new HashMap<>();
+
+    RecordAdapter(Class<? super T> raw, Map<String, BoundField> boundFields) {
+      super(boundFields);
+      this.constructor = ReflectionHelper.getCanonicalRecordConstructor(raw);
+      String[] recordFields = ReflectionHelper.recordFields(raw);
+      for (int i = 0; i < recordFields.length; i++) {
+        fieldIndices.put(recordFields[i], i);
+      }
```
Marcono1234, line 382: I don't think this properly handles renamed fields (`@SerializedName` or `FieldNamingStrategy`). You probably have to map `BoundField.name -> real field / component name -> constructor arg index`.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+            "Could not find the index in the constructor "
+                + constructor
+                + " for field with name "
+                + field.name
+                + ", unable to determine which argument in the constructor the field corresponds"
+                + " to. This is unexpected behaviour, as we expect the RecordComponents to have the"
+                + " same names as the fields in the Java class, and that the order of the"
+                + " RecordComponents is the same as the order of the canonical arguments.");
+      }
+      intermediary[fieldIndex] = field.read(in);
```
Marcono1234, line 413: Might have to handle the case where the JSON contains an explicit `null` for a primitive record component. Maybe similar to `ReflectiveTypeAdapterFactory`'s current behavior, the `null` should be ignored in that case:
https://github.com/google/gson/blob/2591ede59bf7e4cef496a01e82282c88e256cec8/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L162

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+    // The actual record constructor.
+    private final Constructor<? super T> constructor;
+    // Array of arguments to the constructor, initialized with default values for primitives
+    private final Object[] constructorArgsDefaults;
+    // Map from field names to index into the constructors arguments.
+    private final Map<String, Integer> fieldIndices = new HashMap<>();
+
+    RecordAdapter(Class<? super T> raw, Map<String, BoundField> boundFields) {
+      super(boundFields);
+      this.constructor = ReflectionHelper.getCanonicalRecordConstructor(raw);
```
Marcono1234, line 378: As mentioned in https://github.com/google/gson/pull/2197#discussion_r978871132, it might be necessary to make the constructor accessible.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -94,22 +98,32 @@ private List<String> getFieldNames(Field f) {
...
     Class<? super T> raw = type.getRawType();
 
     if (!Object.class.isAssignableFrom(raw)) {
       return null; // it's a primitive!
     }
 
-    FilterResult filterResult = ReflectionAccessFilterHelper.getFilterResult(reflectionFilters, raw);
+    if (ReflectionHelper.isRecord(raw)) {
+      // We can assume blockInaccessible here, because it will never kick inn. We are not doing
+      // direct field access, and are instead constructing records via their canonical constructor.
```
Marcono1234, line 111: > We are not doing direct field access

This seems to be wrong; for deserialization this implementation currently obtains the component values from the `private` fields.
Maybe it would be worth considering obtaining them from the accessor methods instead?

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
     }
   }
 
-  public static RuntimeException createExceptionForUnexpectedIllegalAccess(IllegalAccessException exception) {
+  public static boolean isRecord(Class<?> raw) {
+    return RECORD_HELPER.isRecord(raw);
+  }
+
+  public static String[] recordFields(Class<?> raw) {
+    return RECORD_HELPER.recordFields(raw);
```
Marcono1234, line 90: Maybe both methods should be renamed to `getComponentNames`.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+    private final Method getRecordComponents;
+    private final Method getName;
+    private final Method getType;
+
+    private RecordSupportedHelper() throws NoSuchMethodException {
+      isRecord = Class.class.getMethod("isRecord");
+      getRecordComponents = Class.class.getDeclaredMethod("getRecordComponents");
+      Class<?> recordComponentType = getRecordComponents.getReturnType().getComponentType();
+      getName = recordComponentType.getDeclaredMethod("getName");
+      getType = recordComponentType.getDeclaredMethod("getType");
```
Marcono1234, line 127:
```suggestion
      getRecordComponents = Class.class.getMethod("getRecordComponents");
      Class<?> recordComponentType = getRecordComponents.getReturnType().getComponentType();
      getName = recordComponentType.getMethod("getName");
      getType = recordComponentType.getMethod("getType");
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+    }
+
+    @Override
+    boolean isRecord(Class<?> raw) {
+      try {
+        return boolean.class.cast(isRecord.invoke(raw));
```
Marcono1234, line 133:
```suggestion
        return (boolean) isRecord.invoke(raw);
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+      } catch (IllegalAccessException e) {
+        throw createExceptionForUnexpectedIllegalAccess(e);
```
Marcono1234, line 135: `createExceptionForUnexpectedIllegalAccess` is only intended for exceptions which occur when trying to access members of user provided classes. Its exception message mentions `ReflectionAccessFilter` which does not apply here.

Would probably be best to merge this with the `catch (InvocationTargetException ...` below instead.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+                + " not throw when they are defined on Class",
+                e
+        );
+      }
+    }
+
+    @Override
+    String[] recordFields(Class<?> raw) {
+      try {
+        Object recordComponents = getRecordComponents.invoke(raw);
```
Marcono1234, line 148: Might simplify the code below a bit to cast this here to `Object[]`.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+        int componentCount = Array.getLength(recordComponents);
+        String[] recordFields = new String[componentCount];
+        for (int i = 0; i < componentCount; i++) {
+          recordFields[i] =
+                  String.class.cast(getName.invoke(Array.get(recordComponents, i)));
```
Marcono1234, line 153:
```suggestion
                  (String) getName.invoke(Array.get(recordComponents, i));
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+        }
+        return recordFields;
+      } catch (IllegalAccessException e) {
+        throw createExceptionForUnexpectedIllegalAccess(e);
```
Marcono1234, line 157: Similar to above, `createExceptionForUnexpectedIllegalAccess` should not be used here.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+      } catch (ReflectiveOperationException e) {
+        throw new RuntimeException("Failed to reflect into record components on"
+                + " class [" + raw + "], this is unexpected behaviour, as these methods should"
+                + " not throw when they are defined on Class",
```
Marcono1234, line 161: If you want you can include `(Gson " + GsonBuildConfig.VERSION + ")` in the message; maybe that will help troubleshooting in the future.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+    }
+
+    @Override
+    boolean isRecord(Class<?> raw) {
+      try {
+        return boolean.class.cast(isRecord.invoke(raw));
+      } catch (IllegalAccessException e) {
+        throw createExceptionForUnexpectedIllegalAccess(e);
+      } catch (InvocationTargetException e) {
+        throw new RuntimeException("Failed to reflect into record components on"
```
Marcono1234, line 137: > into record components

This looks incorrect

Also, if you want you can include `(Gson " + GsonBuildConfig.VERSION + ")` in the message; maybe that will help troubleshooting in the future.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+                + " not throw when they are defined on Class",
+                e
+        );
+      }
+    }
+
+    @Override
+    public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
+      try {
+        Object recordComponents = getRecordComponents.invoke(raw);
```
Marcono1234, line 170: Might simplify the code below a bit to cast this here to `Object[]`.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+        int componentCount = Array.getLength(recordComponents);
+        Class[] recordFields = new Class[componentCount];
+        for (int i = 0; i < componentCount; i++) {
+          recordFields[i] =
+                  Class.class.cast(getType.invoke(Array.get(recordComponents, i)));
```
Marcono1234, line 175:
```suggestion
                  (Class<?>) getType.invoke(Array.get(recordComponents, i));
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+        );
+      }
+    }
+
+    @Override
+    public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
+      try {
+        Object recordComponents = getRecordComponents.invoke(raw);
+        int componentCount = Array.getLength(recordComponents);
+        Class[] recordFields = new Class[componentCount];
```
Marcono1234, line 172:
```suggestion
        Class<?>[] recordFields = new Class<?>[componentCount];
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+      getType = recordComponentType.getDeclaredMethod("getType");
+    }
+
+    @Override
+    boolean isRecord(Class<?> raw) {
+      try {
+        return boolean.class.cast(isRecord.invoke(raw));
+      } catch (IllegalAccessException e) {
+        throw createExceptionForUnexpectedIllegalAccess(e);
+      } catch (InvocationTargetException e) {
```
Marcono1234, line 136: Why not catch `ReflectiveOperationException` here instead, similar to the calls below?

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
Marcono1234, line 172: Maybe this variable should rather be named `componentTypes`?

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+        Class[] recordFields = new Class[componentCount];
+        for (int i = 0; i < componentCount; i++) {
+          recordFields[i] =
+                  Class.class.cast(getType.invoke(Array.get(recordComponents, i)));
+        }
+        return raw.getConstructor(recordFields);
+      } catch (IllegalAccessException e) {
+        throw createExceptionForUnexpectedIllegalAccess(e);
+      } catch (ReflectiveOperationException e) {
+        throw new RuntimeException("Failed to reflect into record components on"
```
Marcono1234, line 181: > reflect into record components

Could be incorrect in case obtaining the constructor fails. Maybe a more generic message such as "Failed to obtain canonical constructor on class ..." would work.

If you want you can include `(Gson " + GsonBuildConfig.VERSION + ")` in the message; maybe that will help troubleshooting in the future.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +82,131 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+    public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
+      try {
+        Object recordComponents = getRecordComponents.invoke(raw);
+        int componentCount = Array.getLength(recordComponents);
+        Class[] recordFields = new Class[componentCount];
+        for (int i = 0; i < componentCount; i++) {
+          recordFields[i] =
+                  Class.class.cast(getType.invoke(Array.get(recordComponents, i)));
+        }
+        return raw.getConstructor(recordFields);
```
Marcono1234, line 177: Might have to use `getDeclaredConstructor` because the constructor might not be `public`, see also https://github.com/google/gson/pull/2197#discussion_r978871132.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -94,22 +98,32 @@ private List<String> getFieldNames(Field f) {
...
+  public <T> TypeAdapter<T> create(Gson gson, final TypeToken<T> type) {
     Class<? super T> raw = type.getRawType();
 
     if (!Object.class.isAssignableFrom(raw)) {
       return null; // it's a primitive!
     }
 
-    FilterResult filterResult = ReflectionAccessFilterHelper.getFilterResult(reflectionFilters, raw);
+    if (ReflectionHelper.isRecord(raw)) {
+      // We can assume blockInaccessible here, because it will never kick inn. We are not doing
```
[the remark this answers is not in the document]
staale: I first thought this would only apply to writing fields, but I see the accessible flag is also needed to read values via BoundedField. The RecordComponent actually has an accessor field, that returns the getter method, and I think the most correct approach here would actually be to use the accessor method. Then we do not need to use the set accessible at all. Still, I guess we need to evaluate the reflection filtering result. I will push a fix to this.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -234,26 +255,59 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
     abstract void read(JsonReader reader, Object value) throws IOException, IllegalAccessException;
   }
 
-  public static final class Adapter<T> extends TypeAdapter<T> {
-    private final ObjectConstructor<T> constructor;
-    private final Map<String, BoundField> boundFields;
+  /**
+   * Base class for Adapters produced by this factory.
+   *
+   * <p>The {@link RecordAdapter} is a special case to handle Java 17 and later records, for all
```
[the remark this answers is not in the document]
staale: Reworded to not set a specific version.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
 
-    @Override public void write(JsonWriter out, T value) throws IOException {
-      if (value == null) {
-        out.nullValue();
-        return;
+    // Create the Object that will be used to collect each field value
+    abstract I createIntermediary();
+    // Read a single Bounded field into the intermediary. The JsonReader will be pointed at the
+    // start of the value
+    // for the BoundField to read from.
```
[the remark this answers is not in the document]
staale: Result of google formatter, fixed.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+    abstract void readField(I intermediary, JsonReader in, BoundField field)
+        throws IllegalAccessException, IOException;
+    // Convert the intermediary to a final instance of T.
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+      this.constructor = ReflectionHelper.getCanonicalRecordConstructor(raw);
+      String[] recordFields = ReflectionHelper.recordFields(raw);
+      for (int i = 0; i < recordFields.length; i++) {
+        fieldIndices.put(recordFields[i], i);
+      }
+      Class<?>[] parameterTypes = constructor.getParameterTypes();
+      constructorArgsDefaults = new Object[parameterTypes.length];
+      for (int i = 0; i < parameterTypes.length; i++) {
+        if (parameterTypes[i].isPrimitive()) {
+          constructorArgsDefaults[i] = Array.get(Array.newInstance(parameterTypes[i], 1), 0);
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done, created a common handler to produce an exception for all Record-related ReflectionOperationExceptions.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Improved naming.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Good catch, fixed.

staale:
> I really like the intermediary approach you took here to store the canonical constructor arguments.

Thanks, it is in part inspired by how Collectors work (when applying them, not implementing). I have now also fixed so that I use the Record accessor method to read out values. However, I am a bit uncertain about the approach I have now, because it does not take into consideration that records can be non-public. In those cases the constructor and accessor must be made visible I think. Would be good to have an additional source tree for tests that only applied to Java 17.

Marcono1234:
Unfortunately GitHub has the bad habit of hiding unresolved comments, but there are a few more review comments:
Hidden conversations link

In case you are not aware of it, you can also show all unresolved conversations on the "Files" tab, see [documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#discovering-and-navigating-conversations).

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -111,34 +127,32 @@ private abstract static class RecordHelper {
...
-        return boolean.class.cast(isRecord.invoke(raw));
-      } catch (IllegalAccessException e) {
-        throw createExceptionForUnexpectedIllegalAccess(e);
-      } catch (InvocationTargetException e) {
-        throw new RuntimeException("Failed to reflect into record components on"
-                + " class [" + raw + "], this is unexpected behaviour, as these methods should"
-                + " not throw when they are defined on Class",
-                e
-        );
+        return Boolean.class.cast(isRecord.invoke(raw)).booleanValue();
```
Marcono1234, line 153: I think this should work as well:
```suggestion
        return (boolean) isRecord.invoke(raw);
```

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -169,22 +177,34 @@ public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
       try {
         Object recordComponents = getRecordComponents.invoke(raw);
         int componentCount = Array.getLength(recordComponents);
-        Class[] recordFields = new Class[componentCount];
+        Class<?>[] recordComponentType = new Class<?>[componentCount];
```
Marcono1234, line 180: The name should probably be plural:
```suggestion
        Class<?>[] recordComponentTypes = new Class<?>[componentCount];
```

staale:
> Unfortunately GitHub has the bad habit of hiding unresolved comments, but there are a few more review comments: Hidden conversations link

I thought those where hidden because they were resolved. I will work through that stack next then.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -169,22 +177,34 @@ public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
...
+        Object recordComponents = getRecordComponents.invoke(raw);
+        int componentCount = Array.getLength(recordComponents);
+        for (int i = 0; i < componentCount; i++) {
+          Object recordComponent = Array.get(recordComponents, i);
+          String componentName = String.class.cast(getName.invoke(recordComponent));
+          if (componentName.equals(field.getName())) {
+            return Method.class.cast(getAccessor.invoke(recordComponent));
+          }
+        }
+        return null;
```
Marcono1234, line 202: Would probably make sense to throw an exception here, otherwise this might fail in unexpected ways for `ReflectiveTypeAdapterFactory`.
Though if this fails here, it would most likely be unexpected because the language specification makes guarantees about the accessor method and field name.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -169,22 +177,34 @@ public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
...
-                e
-        );
+        throw createExceptionForRecordReflectionException(e);
       }
     }
+
+    @Override
+    public Method getAccessor(Class<?> raw, Field field) {
+      try {
+        Object recordComponents = getRecordComponents.invoke(raw);
```
Marcono1234, line 193: Would probably simplify the code below to cast this to `Object[]`.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -169,22 +177,34 @@ public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
...
+        int componentCount = Array.getLength(recordComponents);
+        for (int i = 0; i < componentCount; i++) {
+          Object recordComponent = Array.get(recordComponents, i);
+          String componentName = String.class.cast(getName.invoke(recordComponent));
+          if (componentName.equals(field.getName())) {
+            return Method.class.cast(getAccessor.invoke(recordComponent));
```
Marcono1234, line 199:
```suggestion
          String componentName = (String) getName.invoke(recordComponent);
          if (componentName.equals(field.getName())) {
            return (Method) getAccessor.invoke(recordComponent);
```

Is there are a reason why you are writing so often `...class.cast(...)` here? I don't think it provides any advantage over regular casts and is a bit more verbose.

Marcono1234 (commented):
> In those cases the constructor and accessor must be made visible I think.

Accessor methods will be fine I think because they are guaranteed to be `public` (see [Language Specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10.3)). But as mentioned earlier the canonical constructor will probably be an issue. Though you can address that later if you want.

Also no hurry, take your time.

eamonnmcmanus:
In #2203 I'm adding a skeletal test for records which I'm hoping you will be able to fill out to get some good test coverage here. I did the necessary work to ensure that the new test is only compiled in the JDK 17 build.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +326,102 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+        }
       }
+    }
 
-      out.beginObject();
+    @Override
+    Object[] createIntermediary() {
+      Object[] constructorArgs = new Object[constructorArgsDefaults.length];
+      System.arraycopy(constructorArgsDefaults, 0, constructorArgs, 0, constructorArgs.length);
+      return constructorArgs;
```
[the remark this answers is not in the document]
staale: I assume clone is then implemented for Arrays. Fixed, was not aware of this possibility.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -156,7 +170,14 @@ private ReflectiveTypeAdapterFactory.BoundField createBoundField(
...
+
+      @Override
+      Object read(JsonReader reader) throws IOException {
+        return typeAdapter.read(reader);
+      }
+
+      @Override
+      void read(JsonReader reader, Object value)
           throws IOException, IllegalAccessException {
         Object fieldValue = typeAdapter.read(reader);
```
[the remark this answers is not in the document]
staale: I rewrote this a bit, so that I could localise the logic wrt. primitve fields and null values.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -234,26 +255,59 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
       this.serialized = serialized;
       this.deserialized = deserialized;
     }
+
     abstract void write(JsonWriter writer, Object value) throws IOException, IllegalAccessException;
+
+    abstract Object read(JsonReader reader) throws IOException;
+
     abstract void read(JsonReader reader, Object value) throws IOException, IllegalAccessException;
```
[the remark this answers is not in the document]
staale: Fixed

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Fixed this by adding the componentName to the BoundedField

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Fixed

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Fixed

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: This has been changed.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Fixed

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Fixed

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Fixed

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -197,8 +221,17 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
         if (!serialize && !deserialize) {
           continue;
         }
+        // The accessor method is only used for records. If the type is a record, we will read out values
+        // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
+        Method accessor = null;
+        if (ReflectionHelper.isRecord(raw)) {
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -226,34 +263,75 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
...
-    abstract void write(JsonWriter writer, Object value) throws IOException, IllegalAccessException;
-    abstract void read(JsonReader reader, Object value) throws IOException, IllegalAccessException;
+
+    abstract void write(JsonWriter writer, Object value) throws IOException, ReflectiveOperationException;
+
+    /** Read the value into an array of values, used to construct records */
+    abstract void readIntoArray(JsonReader reader, int index, Object[] values) throws IOException
+
+    /** Read the value from the reader, and set it on the corresponding field on value via reflection */
+    abstract void readIntoField(JsonReader reader, Object value) throws IOException, IllegalAccessException;
```
Marcono1234, line 284: I think here and in the override the `value` parameter should be renamed to `instance` (or something else). Personally I feel `value` is extremely misleading in the context of a field.
```suggestion
    /** Read the value from the reader, and set it on the corresponding field on instance via reflection */
    abstract void readIntoField(JsonReader reader, Object instance) throws IOException, IllegalAccessException;
```

The same might also apply to `write(JsonWriter writer, Object value)`.

(Though to be fair this code already existed before your PR.)

What do you think?

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -226,34 +263,75 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
...
-  public static final class Adapter<T> extends TypeAdapter<T> {
-    private final ObjectConstructor<T> constructor;
-    private final Map<String, BoundField> boundFields;
+  /**
+   * Base class for Adapters produced by this factory.
+   *
+   * <p>The {@link RecordAdapter} is a special case to handle records for JVMs that support it, for
+   * all other types we use the {@link FieldReflectionAdapter}. This class encapuslates the common
+   * logic for serialization and deserialization. During deserialization, we construct an
+   * intermediary ResultHolder R, which we append values to. After the object has been read in
```
Marcono1234, line 293: Should this be `I` to match the parameter name?
```suggestion
   * intermediary result holder I, which we append values to. After the object has been read in
```

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +350,108 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+    }
+  }
+
+  private static final class RecordAdapter<T> extends Adapter<T, Object[]> {
+    // The actual record constructor.
+    private final Constructor<? super T> constructor;
+    // Array of arguments to the constructor, initialized with default values for primitives
+    private final Object[] constructorArgsDefaults;
+    // Map from field names to index into the constructors arguments.
+    private final Map<String, Integer> fieldIndices = new HashMap<>();
```
Marcono1234, line 399: Rather minor, but maybe here and also below it would be better to use the name "component index". Fields (even for records) don't have any guaranteed order or index, but record components do have a specified order.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
Marcono1234: @staale, this is not resolve yet, is it?

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Fixed this by just invoking getMethod instead.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Done

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
staale: Agree, I guess I was just in a mode to use Reflection for everything :)

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Renamed to source and target, as then these match with each other on read and write.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: I have renamed this to accumulator since I think that name fits better. We accumulate values into the accumulator, then finalize it to create an instance.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Agree.

staale:
I looked at https://www.baeldung.com/maven-multi-release-jars to create a multi-release jar, and have a separate Java16 source folder. Unfortunately this seems to conflict with bnd https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin
`[ERROR] Failed to execute goal biz.aQute.bnd:bnd-maven-plugin:6.3.1:bnd-process (default) on project gson: Classes found in the wrong directory: {META-INF/versions/16/com/google/gson/internal/reflect/RecordHelper.class=com.google.gson.internal.reflect.RecordHelper} -> [Help 1]`

Marcono1234 (commented):
A few more smaller suggestions, but I think this is looking really good! Thanks a lot for your work so far.

> I looked at https://www.baeldung.com/maven-multi-release-jars to create a multi-release jar

I think for now this reflection-based approach is fine. Even though in the past for other occasions I suggested using multi-release JARs as well, it also has it flaws. The main problem is when users create a JAR with dependencies (also called "fat JAR" or "uber JAR" sometimes) because most build tool plugins (e.g. Maven Shade Plugin or Gradle Shadow plugin) do not set `Multi-Release: true` in the `MANIFEST.MF`, and most users are not aware of this or forget it. So Gson would manually have to check whether it is running with JDK > X or whether certain features (such as records here) are supported and verify that the multi-release classes were properly loaded.

Also, I think there is no need to force-push to this GitHub pull request; @eamonnmcmanus will most likely squash the commits anyway on merge. Please let us know though if you don't want this and want the individual commits to be preserved.

staale:
> In #2203 I'm adding a skeletal test for records which I'm hoping you will be able to fill out to get some good test coverage here. I did the necessary work to ensure that the new test is only compiled in the JDK 17 build.

Thanks, I see that it is merged, so I can rebase on top of master to also have these tests actually run.

Marcono1234:
> so I can rebase on top of master

As mentioned before, unless you want the Git history of this PR to be preserved you can also just merge in `master` into this branch. Otherwise when you rebase and force-push GitHub will probably be unable to match the open review comments to the rebased commits, and it makes comparing the changes slightly more cumbersome.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -197,8 +225,17 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
         if (!serialize && !deserialize) {
           continue;
         }
+        // The accessor method is only used for records. If the type is a record, we will read out values
+        // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
+        Method accessor = null;
+        if (isRecord) {
```
Marcono1234, line 231: By using `GsonBuilder.excludeFieldsWithModifiers` a user might include `static` fields for serialization. That would break this logic in case a record class declares a `static` field.

Maybe it would be possible to handle this with special-casing, but personally I think it would also be fine and easier to just skip `static` fields here for now if this is a record class. If necessary `static` field support could be added in a later PR, though I assume such use cases are rather rare.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -226,34 +263,76 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
 
   static abstract class BoundField {
     final String name;
+    /** Name of the underlying field */
+    final String componentName;
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -226,34 +263,76 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
...
-  public static final class Adapter<T> extends TypeAdapter<T> {
-    private final ObjectConstructor<T> constructor;
-    private final Map<String, BoundField> boundFields;
+  /**
+   * Base class for Adapters produced by this factory.
+   *
+   * <p>The {@link RecordAdapter} is a special case to handle records for JVMs that support it, for
+   * all other types we use the {@link FieldReflectionAdapter}. This class encapsulates the common
+   * logic for serialization and deserialization. During deserialization, we construct an
+   * accumulator A, which we use to accumulate values from the source Json. After the object has been read in
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -272,24 +351,108 @@ protected BoundField(String name, boolean serialized, boolean deserialized) {
...
+      return accumulator;
+    }
+  }
+
+  private static final class RecordAdapter<T> extends Adapter<T, Object[]> {
+    // The actual record constructor.
+    private final Constructor<? super T> constructor;
+    // Array of arguments to the constructor, initialized with default values for primitives
+    private final Object[] constructorArgsDefaults;
+    // Map from field names to index into the constructors arguments.
```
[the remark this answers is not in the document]
staale: Done

gson/src/test/java/com/google/gson/internal/reflect/ReflectionHelperTest.java
```diff
@@ -0,0 +1,39 @@
...
+            throw new AssumptionViolatedException("java.lang.Record not supported");
+        }
+    }
+
+    @Test
+    public void testJava17Record() throws ClassNotFoundException {
+        Class<?> unixDomainPrincipalClass = Class.forName("jdk.net.UnixDomainPrincipal");
+        // UnixDomainPrincipal is a record
+        assertTrue(ReflectionHelper.isRecord(unixDomainPrincipalClass));
+        // with 2 fields
```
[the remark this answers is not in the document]
staale: Done

gson/src/test/java/com/google/gson/internal/reflect/ReflectionHelperTest.java
```diff
@@ -0,0 +1,39 @@
...
+        assertArrayEquals(new String[] {"user", "group"}, ReflectionHelper.getRecordComponentNames(unixDomainPrincipalClass));
+        // Check canonical constructor
+        Constructor<?> constructor = ReflectionHelper.getCanonicalRecordConstructor(unixDomainPrincipalClass);
+        assertNotNull(constructor);
+        assertArrayEquals(
+                new Class<?>[] { UserPrincipal.class, GroupPrincipal.class },
+                constructor.getParameterTypes());
+    }
```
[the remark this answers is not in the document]
staale: Done. though this was already at least covered in the serialization tests.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -65,10 +81,147 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
...
+
+    @Override
+    public <T> Constructor<T> getCanonicalRecordConstructor(Class<T> raw) {
+      try {
+        Object[] recordComponents = (Object[]) getRecordComponents.invoke(raw);
+        Class<?>[] recordComponentTypes = new Class<?>[recordComponents.length];
+        for (int i = 0; i < recordComponents.length; i++) {
+          recordComponentTypes[i] = (Class<?>) getType.invoke(recordComponents[i]);
+        }
+        return raw.getDeclaredConstructor(recordComponentTypes);
```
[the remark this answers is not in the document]
staale: Done

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Would it not be better to just not set the accessor if the field is static. I am assuming here, but I think if the field is static, then normal reflection to write to the field would still work. It is only reflective writes to record fields that are constrained.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Done

gson/src/test/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactoryTest.java
```diff
@@ -0,0 +1,132 @@
...
+                        final String name = in.nextString();
+                        return new GroupPrincipalImpl(name);
+                    }
+                })
+                .create();
+        UserPrincipal userPrincipal = gson.fromJson("\"user\"", UserPrincipal.class);
+        GroupPrincipal groupPrincipal = gson.fromJson("\"group\"", GroupPrincipal.class);
+        Object recordInstance = unixDomainPrincipalClass.getDeclaredConstructor(UserPrincipal.class, GroupPrincipal.class)
+                .newInstance(userPrincipal, groupPrincipal);
+        String serialized = gson.toJson(recordInstance);
```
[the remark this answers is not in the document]
staale: Done

gson/src/test/java/com/google/gson/internal/reflect/ReflectionHelperTest.java
```diff
@@ -0,0 +1,39 @@
+package com.google.gson.internal.reflect;
```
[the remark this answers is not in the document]
staale: Applied google format

staale (commented):
Fixed review comments.

Marcono1234:
Thanks for your changes! But *please avoid force-pushing this branch* (as mentioned before). The commits of this pull request will be squashed anyways, (maybe) unless you explicitly say you don't want this. But even in that case, please squash and force-push once all comments are resolved at the end.

With force-pushing (especially when modifying the previous commits) I cannot easily review the incremental changes you did but I would have to review all of your changes all over again... (or I have to compare the differences between the force-pushes)
It also renders GitHub's review comments ineffective because they link to commits which don't exist anymore. So I would have to search the comments here on the conversation tab to find out their context and then manually try to figure out if they still apply.

I think these changes are fine now and further review rounds would just cause this to drag on. For any smaller remaining things I can try to create a follow-up pull request.

Marcono1234:

eamonnmcmanus:
Great! I'm just going to run this through all of Google's tests. I don't really expect that to reveal any problems but I've been surprised before.

staale:
> Thanks for your changes! But please avoid force-pushing this branch (as mentioned before).

Sorry, will not do so again. I am used to Gerrit workflow at work, where we try to keep a clean history. I did one force push to fix the committer of the first commit, as that was using my work email, and not my private email. If this is squash merged, that's great.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -226,34 +265,76 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
...
+   * all other types we use the {@link FieldReflectionAdapter}. This class encapsulates the common
+   * logic for serialization and deserialization. During deserialization, we construct an
+   * accumulator A, which we use to accumulate values from the source JSON. After the object has been read in
+   * full, the {@link #finalize(Object)} method is used to convert the accumulator to an instance
+   * of T.
+   *
+   * @param <T> type of objects that this Adapter creates.
+   * @param <A> type of accumulator used to build the deserialization result.
+   */
+  abstract static class Adapter<T, A> extends TypeAdapter<T> {
```
eamonnmcmanus, line 303: I found a number of places in Google's source base that are doing this:
```
if (delegate instanceof ReflectiveTypeAdapterFactory.Adapter) {...}
```

Can you make this public so those don't break?

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
staale: Fixed

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks again! This will be very helpful for users. There are a few minor things but it's probably easier for me to address them in a followup.

Marcono1234:
Thanks again @staale for your work on this! Sorry that my last comment was a bit harsh and that my review comments were a bit excessive. Your changes integrated really well with the existing reflective adapter. Hopefully users will enjoy the full Record class support!

GeorgeCodes17:
Unless I am mistaken this doesn't work I have code like `data = """
                {"end": "09:00:00", "__pk": 1, "start": "08:00:00", "teacher": "00u8ociwo4drfKwS85d7", "_fk_lesson": 1, "created_at": "2023-03-26 09:32:22.000000", "updated_at": "2023-03-26 09:32:29.000000", "assigned_by": "john@email.com", "day_of_week": "wed"}""";
        TimetableObject timetableObject = new Gson().fromJson(data, TimetableObject.class);` and the object looks like `public record TimetableObject(
    String end,
    int __pk,
    String start,
    String teacher,
    int _fk_lesson,
    String created_at,
    String updated_at,
    String assigned_by,
    String day_of_week
) {
}` yet I am still getting this error **Caused by: java.lang.AssertionError: java.lang.IllegalAccessException: Can not set final java.lang.String field com.Schoolio.objects.TimetableObject.end to java.lang.String
** What am I doing wrong?

GeorgeCodes17:
Screenshot 2023-04-09 at 10 19 03
Screenshot 2023-04-09 at 10 20 21
I am a bit confused why using this code is giving "**java.lang.AssertionError: java.lang.IllegalAccessException: Can not set final java.lang.String field com.Schoolio.objects.TimetableObject.end to java.lang.String**".
Can anyone help, what is wrong with my implementation?

Marcono1234:
@GeorgeCodes17, which Gson version are you using? Support for Record classes was added in Gson 2.10; the exception message you showed looks like the one reported for older Gson versions which did not properly support Record classes.

GeorgeCodes17:
> @GeorgeCodes17, which Gson version are you using? Support for Record classes was added in Gson 2.10; the exception message you showed looks like the one reported for older Gson versions which did not properly support Record classes.

Gson 2.8.5 version

PR #2203, opened 2022-09-27
Add a currently-failing test for Java records and `@Ignore` it.

eamonnmcmanus:
Also do the Maven gymastics required to ensure that this test only compiles on Java versions ≥17. (It would also work on Java 16, but 17 is all we have in the CI.)

Fix some compilation problems I saw when running locally, which for some reason don't show up in the CI.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -247,6 +247,7 @@
       <properties>
         <javaVersion>17</javaVersion>
         <excludeTestCompilation></excludeTestCompilation>
+        <extraLintSuppressions>,-exports,-missing-explicit-ctor,-removal</extraLintSuppressions>
       </properties>
     </profile>
   </profiles>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -17,6 +17,10 @@
     </license>
   </licenses>
 
+  <properties>
+    <excludeTestCompilation>**/Java17*</excludeTestCompilation>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>junit</groupId>
@@ -58,6 +62,18 @@
                 <!-- module-info.java is compiled using ModiTect -->
                 <exclude>module-info.java</exclude>
               </excludes>
+            </configuration>
+          </execution>
+          <execution>
+            <id>default-testCompile</id>
+            <phase>test-compile</phase>
+            <goals>
+              <goal>testCompile</goal>
+            </goals>
+            <configuration>
+              <testExcludes>
+                <exclude>${excludeTestCompilation}</exclude>
+              </testExcludes>
             </configuration>
           </execution>
         </executions>
@@ -222,4 +238,16 @@
       </plugin>
     </plugins>
   </build>
+  <profiles>
+    <profile>
+      <id>JDK17</id>
+      <activation>
+        <jdk>[17,)</jdk>
+      </activation>
+      <properties>
+        <javaVersion>17</javaVersion>
+        <excludeTestCompilation></excludeTestCompilation>
+      </properties>
+    </profile>
+  </profiles>
 </project>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -245,7 +245,7 @@
         <jdk>[17,)</jdk>
       </activation>
       <properties>
-        <javaVersion>17</javaVersion>
+        <maven.compiler.testRelease>17</maven.compiler.testRelease>
         <excludeTestCompilation></excludeTestCompilation>
         <extraLintSuppressions>,-exports,-missing-explicit-ctor,-removal</extraLintSuppressions>
       </properties>
--- a/gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
+++ b/gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2022 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gson.functional;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+import com.google.gson.Gson;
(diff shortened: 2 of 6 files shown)


gson/pom.xml
```diff
@@ -222,4 +238,17 @@
...
+  <profiles>
+    <profile>
+      <id>JDK17</id>
+      <activation>
+        <jdk>[17,)</jdk>
+      </activation>
+      <properties>
+        <javaVersion>17</javaVersion>
+        <excludeTestCompilation></excludeTestCompilation>
+        <extraLintSuppressions>,-exports,-missing-explicit-ctor,-removal</extraLintSuppressions>
```
Marcono1234, line 250: I think the cleaner solution would be to have a separate test compilation.

This approach here has the following disadvantages:
- When a JDK 17 is on `PATH` and `JAVA_HOME`, but a `.m2/toolchains.xml` file exists, the current toolchain configuration seems to prefer a JDK 11 (see root `pom.xml`) which then causes build failures because at the same time this profile here is active
- When a JDK 17 is on `PATH` and `JAVA_HOME` this profile causes the build to produce a Gson JAR with target version 17 instead of 7. This also has these other side effects:
  - The JAR file contains an additional `module-info.class` file at the top level
  - `MANIFEST.MF` contains additional OSGi metadata for `Import-Package`, which now seems to define a non-optional dependency on the `java.sql` package
- It erroneously makes Eclipse IDE think the complete project is compiled against JDK 17 (not sure if that can be solved though)

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.google.gson.Gson;
+import com.google.gson.annotations.SerializedName;
+import java.util.Objects;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
```
Marcono1234, line 29: Looks like this can be omitted. Or is there a reason why you added this?

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
```diff
@@ -58,9 +58,20 @@ public void testBlockInaccessibleJava() {
...
 
-    // But serialization should succeed for classes with only public fields
-    String json = gson.toJson(new Point(1, 2));
-    assertEquals("{\"x\":1,\"y\":2}", json);
+
+    // But serialization should succeed for classes with only public fields.
+    // Not many JDK classes have mutable public fields, thank goodness, but java.awt.Point does.
+    Class<?> pointClass = null;
+    try {
+      pointClass = Class.forName("java.awt.Point");
```
Marcono1234, line 66: Why are you loading this class with `Class.forName`? Was the class not available for you when you tested this (maybe because the module was missing)?

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
```diff
@@ -58,9 +58,20 @@ public void testBlockInaccessibleJava() {
...
+    } catch (ClassNotFoundException e) {
+    }
+    if (pointClass != null) {
```
Marcono1234, line 69: Might be better to use `assumeNotNull`.

gson/pom.xml
Marcono1234, line 250: Would be good to fix these issues, I think the warnings are reasonable:
- `TypeAdapter` should have an explicit no-args constructor
- `TypeAdapters` should suppress warnings for usage of `AccessController.doPrivileged` (related to `SecurityManager` deprecation)
- `Gson.excluder()` should have `@SuppressWarnings("exports")` (see also #1903)

For the module related warnings for test compilation, it looks like Maven erroneously considers `module-info.java` also for the test classes? Might be good to investigate that. Or then at least only disable these warnings for test compilation.

eamonnmcmanus: I'm not sure exactly what you mean by a separate test compilation?

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
eamonnmcmanus: Yeah there was a module problem I was seeing locally that didn't seem to be resolved in the obvious way, doing `require static java.desktop`, so I was impatient and just did this instead. As the comment suggests, it's not too concerning if we don't have this test coverage on Java 17.

gson/pom.xml
eamonnmcmanus: I agree. I was going to do that in a separate PR. The problems don't seem to show up all the time, and I'm not sure I fully understand why these started cropping up with this PR. Maybe the version of JDK 17 in the CI has changed very recently?

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
eamonnmcmanus: OK.

gson/pom.xml
eamonnmcmanus: Yes, that's a good idea. Now that we are assuming Java 11+ for the build, we should just be able to set `maven.compiler.release` and `maven.compiler.testRelease`. Though I'm not sure if that's why we're suddenly seeing javadoc failures. :-(

pom.xml
```diff
@@ -21,7 +21,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <javaVersion>7</javaVersion>
+    <maven.compiler.release>7</maven.compiler.release>
```
Marcono1234, line 24: The code below still refers to `${javaVersion}` (though that does not seem to have any effect?).

gson/pom.xml
Marcono1234: > Though I'm not sure if that's why we're suddenly seeing javadoc failures. :-(

Probably previously it was implicit using Java 11 as target because `javadoc` was from JDK 11. Now it uses the `maven.compiler.release` value and fails due to references to `AccessibleObject.canAccess` (added in Java 9) and module issues.

I think the easiest and best solution would be to add `<release>11</release>` (and a short comment explaining the rationale) to the configuration for `maven-javadoc-plugin`. Then we will also get the module overview again in the generated Javadoc.

(You can then probably also revert the Javadoc change for `AccessibleObject#canAccess` again.)

pom.xml
eamonnmcmanus: Hmm, yes, I think we should be able to delete that.

gson/pom.xml
eamonnmcmanus: Good tips for the javadoc. I tried something else first but now I'm doing what you suggested here and it's looking good. Thanks!

eamonnmcmanus:
I think this is close enough now. Thanks for the feedback, @Marcono1234! We can tweak further as needed.

PR #2206, opened 2022-09-27
Disable Maven transfer progress for GitHub workflows

Marcono1234:
Disables the download transfer progress which is shown when Maven downloads (or uploads) artifacts which are not available in the local repository. This download progress can be quite verbose and is normally not that relevant.

This is mainly noticable when changing the `pom.xml` files which causes the GitHub workflow cache containing the local repository to be discarded.


eamonnmcmanus:
[Why do we even have that lever?](https://www.youtube.com/watch?v=v4IECphJ2RM)

PR #2207, opened 2022-09-27
Follow-up for record test build changes

Marcono1234:
Follow up for #2203

It looks like the extra lint suppressions are not actually needed anymore.


eamonnmcmanus:
Great, thanks!

PR #2208, opened 2022-09-29
Missing `getType()` in Maps Examples

jonathan-benn:
The code as-written does not compile! :(

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -263,7 +263,7 @@ For deserialization Gson uses the `read` method of the `TypeAdapter` registered 
 
 ```java
 Gson gson = new Gson();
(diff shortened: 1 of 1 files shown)


Marcono1234:
This was actually a deliberate change done by #1700 because in the next release there will be `fromJson` overloads with `TypeToken` parameter which will be the recommended approach because they are more type-safe (see #1700 for details).

Though I understand that this change to the user guide is misleading when those changes have not been released yet or for users of older Gson versions; I have tried to address this in #2209.

eamonnmcmanus:
I think this indeed addressed by #2209, though we should probably also make a new release fairly soon that includes the API changes.

PR #2209, opened 2022-09-29
Mention in User Guide alternative for Gson versions without fromJson(..., TypeToken)

Marcono1234:
Follow-up for #1700

Inspired by #2208; the changes done by #1700 to the User Guide will be confusing for users who are not using the latest Gson version (especially since those changes have not been released yet). This pull request tries to adjust the User Guide to make this clearer.


eamonnmcmanus:
Great idea! In general people might be reading the current user guide online but working with an older Gson version for whatever reason. While we don't want to make the guide unreadable with notes here and there about how things _used_ to work, occasional comments like this in key places are worthwhile.

PR #2212, opened 2022-10-01
Automatically replace version references on `release:prepare`

Marcono1234:
When running `mvn release:prepare` automatically updates the version references in `README.md` and `UserGuide.md`, and also replaces `$next-version$` with the new version. This is mainly useful for Javadoc (e.g. `@since $next-version$`) because it is not known in advance in which version a feature will be integrated.

The current solution seems to work (have tested it locally), but it is rather hacky and might rely on some implementation details / quirks of Maven and the Maven Release Plugin. So no worries if you don't want to integrate this. There might be better ways to solve this; any feedback is appreciated.

It appears Guava has a special `@since NEXT` and also automatically adds missing `@since` on release, but I was unable to find where exactly this is configured. Possibly it is using [`javadoc:fix`](https://maven.apache.org/plugins/maven-javadoc-plugin/fix-mojo.html), but I was not successful with applying this for Gson.


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1299,10 +1302,10 @@ public <T> T fromJson(JsonElement json, Type typeOfT) throws JsonSyntaxException
    * @return an object of type T from the JSON. Returns {@code null} if {@code json} is {@code null}
    * or if {@code json} is empty.
    * @throws JsonSyntaxException if json is not a valid representation for an object of type typeOfT
-   * @since 1.3
```
Marcono1234, line 1302: This was a copy & paste error I made in #1700.

ReleaseProcess.md
```diff
@@ -6,7 +6,6 @@ The following is a step-by-step procedure for releasing a new version of Google-
 1. Ensure all changelists are code-reviewed and have +1
 1. `cd gson` to the parent directory; ensure there are no open files and all changes are committed.
 1. Run `mvn release:clean`
-1. Do a dry run: `mvn release:prepare -DdryRun=true`
```
Marcono1234, line 9: Have removed this because it looks like this would prevent subsequent (non-dry-run) `mvn release:prepare` from working properly?

eamonnmcmanus:
OK, I think I should be ready to give this a spin shortly.

Marcono1234:
Thanks! The disadvantage of this manual approach seem to be that the rollback does not revert the additionally committed files (see 7bca5c4061726164e8b68f2496118b7b5243eefe). But luckily that did not matter here. Another thing to look out for is that it runs `git add .`, so it would always be good to verify that no unrelated files were committed accidentally (was not a problem for this release either).

Also, do you happen to know how Guava does this? (quoting from the description)

> It appears Guava has a special `@since NEXT` and also automatically adds missing `@since` on release, but I was unable to find where exactly this is configured. Possibly it is using [`javadoc:fix`](https://maven.apache.org/plugins/maven-javadoc-plugin/fix-mojo.html), but I was not successful with applying this for Gson.

Marcono1234:
One disadvantage might also be that `mvn release:prepare` directly pushes the changes, and there seems to be no easy way to prevent this / defer this until having reviewed the changes locally. When using [`<pushChanges>false</pushChanges>`](https://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html#pushChanges) it looks like you would have to manually push the commits and tag.

eamonnmcmanus:
> It appears Guava has a special @since NEXT and also automatically adds missing @since on release, but I was unable to find where exactly this is configured.

Both Gson and Guava are mirrored between GitHub and Piper, Google's internal monorepo. But the mirroring is different: for Gson we copy from GitHub to Piper while for Guava we copy from Piper to GitHub. Before making a Guava release, we use a shell script to update the `@since` tags in Piper, and those changes then get copied to GitHub. The process of adding `@since NEXT` tags in the first place is somewhat manual at the moment. All that to say that I don't think there's useful precedent there.

PR #2215, opened 2022-10-02
Clarify in documentation that Gson might cache strategy results

### LINKED ISSUE #2205: ExclusionStrategy cannot depend on threadlocal
In a Spring environment (or equivalent), it is common practice to implement security by checking on the current security context, which is a threadlocal object. Gson exclusion strategies are only evaluated once, it seems. I understand this is a performance concern, but would appreciate if this could be made configurable. At the minimum a warning could be placed on the ExclusionStrategy interface.

Marcono1234:
Resolves #2205

Maybe it could also be considered whether that behavior could be customized in the future, but for now it might suffice to document it.

Feedback, especially regarding the wording, is appreciated!

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -349,6 +349,10 @@ public final class GsonBuilder {
    * Configures Gson to apply a specific naming strategy to an object's fields during
    * serialization and deserialization.
    *
+   * <p>The Gson instance might only use the field naming strategy once for a field and
+   * cache the result. It is not guaranteed that the strategy will be used again every
+   * time the value of a field is serialized or deserialized.
+   *
    * @param fieldNamingStrategy the naming strategy to apply to the fields
    * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
    * @since 1.3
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -349,6 +349,10 @@ public GsonBuilder setFieldNamingPolicy(FieldNamingPolicy namingConvention) {
    * Configures Gson to apply a specific naming strategy to an object's fields during
    * serialization and deserialization.
    *
+   * <p>The Gson instance might only use the field naming strategy once for a field and
```
eamonnmcmanus, line 352: I think I would use `{@link Gson}` here to make it clear that we are talking about the class here (in the previous paragraph we are talking about Gson the project).

eamonnmcmanus (commented):
Seems good. Just one tiny comment (which applies several times).

gson/src/main/java/com/google/gson/GsonBuilder.java
Marcono1234: I had a similar concern, being afraid that users might think Gson statically (for all `Gson` instances) caches the results. That is why I chose "Gson instance", but I have now changed it to "The created Gson instance". In the context of this builder I hope this should be clear enough.

In this specific line where you added the comment I don't think a `{@link Gson}` would be needed. The previous paragraph is also about the `Gson` instance and not about specifying a naming strategy globally. Or did I misunderstand you?

eamonnmcmanus: This is fine now. I would probably use `{@link}` or `{@code}` every time we're talking about `Gson` the class, since Gson is also the name of the project. But that's probably overly fussy.

PR #2219, opened 2022-10-11
Small adjustments to the new record code.

eamonnmcmanus:
* Replace wildcard imports with single imports.
* Enable `Java17RecordTest` and fix its many previously-hidden problems.
* Use a `Map` to get primitive zero values rather than a potentially-expensive reflective trick.
* Apply some automated code fixes.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -43,7 +43,6 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
-import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -93,9 +92,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
 
     List<String> fieldNames = new ArrayList<>(alternates.length + 1);
     fieldNames.add(serializedName);
-    for (String alternate : alternates) {
-      fieldNames.add(alternate);
-    }
+    Collections.addAll(fieldNames, alternates);
     return fieldNames;
   }
 
@@ -394,6 +391,8 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
   }
 
   private static final class RecordAdapter<T> extends Adapter<T, Object[]> {
+    static Map<Class<?>, Object> ZEROES = primitiveZeroes();
+
     // The actual record constructor.
     private final Constructor<? super T> constructor;
     // Array of arguments to the constructor, initialized with default values for primitives
@@ -417,14 +416,22 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       // we create an Object[] where all primitives are initialized to non-null values.
       constructorArgsDefaults = new Object[parameterTypes.length];
       for (int i = 0; i < parameterTypes.length; i++) {
-        if (parameterTypes[i].isPrimitive()) {
-          // Voodoo magic, we create a new instance of this primitive type using reflection via an
-          // array. The array has 1 element, that of course will be initialized to the primitives
-          // default value. We then retrieve this value back from the array to get the properly
-          // initialized default value for the primitve type.
-          constructorArgsDefaults[i] = Array.get(Array.newInstance(parameterTypes[i], 1), 0);
-        }
+        constructorArgsDefaults[i] = ZEROES.get(parameterTypes[i]);
+        // This will correctly be null for non-primitive types.
       }
+    }
+
+    private static Map<Class<?>, Object> primitiveZeroes() {
+      Map<Class<?>, Object> zeroes = new HashMap<>();
+      zeroes.put(byte.class, (byte) 0);
+      zeroes.put(short.class, (short) 0);
+      zeroes.put(int.class, 0);
+      zeroes.put(long.class, 0L);
+      zeroes.put(float.class, 0F);
+      zeroes.put(double.class, 0D);
+      zeroes.put(char.class, '\0');
+      zeroes.put(boolean.class, false);
+      return zeroes;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


eamonnmcmanus:
@staale FYI, very minor things.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -394,6 +391,8 @@ T finalize(T accumulator) {
   }
 
   private static final class RecordAdapter<T> extends Adapter<T, Object[]> {
+    static Map<Class<?>, Object> ZEROES = primitiveZeroes();
```
Marcono1234, line 394: Maybe something like `PRIMITIVE_DEFAULTS` or similar would be a bit more descriptive but it does not matter much.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -417,16 +416,24 @@ T finalize(T accumulator) {
...
       for (int i = 0; i < parameterTypes.length; i++) {
-        if (parameterTypes[i].isPrimitive()) {
-          // Voodoo magic, we create a new instance of this primitive type using reflection via an
-          // array. The array has 1 element, that of course will be initialized to the primitives
-          // default value. We then retrieve this value back from the array to get the properly
-          // initialized default value for the primitve type.
-          constructorArgsDefaults[i] = Array.get(Array.newInstance(parameterTypes[i], 1), 0);
-        }
+        constructorArgsDefaults[i] = ZEROES.get(parameterTypes[i]);
+        // This will correctly be null for non-primitive types.
```
Marcono1234, line 420: Shouldn't this comment be _above_ the line it is describing?

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -43,28 +42,42 @@ public void testMultipleNamesDeserializedCorrectly() {
...
   @Test
   public void testConstructorRuns() {
-    assertThrows(NullPointerException.class,
-        () -> gson.fromJson("{'name1': null, 'name2': null", MyRecord.class));
+    Throwable e = assertThrows(RuntimeException.class,
+        () -> gson.fromJson("{'name1': null, 'name2': null}", MyRecord.class));
+    while (e != null && !(e instanceof NullPointerException)) {
+      e = e.getCause();
+    }
+    assertNotNull(e);
```
Marcono1234, line 64: Rather than using an exception to signal that the constructor was called it might be a bit cleaner to just have the constructor modify the provided value, e.g.:
```java
public record MyRecord(String a) {
  public MyRecord {
    a = "constructor-" + a;
  }
}
```

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -43,28 +42,42 @@ public void testMultipleNamesDeserializedCorrectly() {
...
+  public record MyRecord(
       @SerializedName("name") String a,
       @SerializedName(value = "name1", alternate = {"name2", "name3"}) String b) {
-    MyRecord {
+    public MyRecord {
       Objects.requireNonNull(a);
     }
   }
+
+  public record RecordWithPrimitives(
```
Marcono1234, line 81: Just to verify that the constructor was actually called, what about adding an additional `String s` component. Then the primitives test could for example deserialize `{"s": "missing"}` and create the expected record instance with `"missing"` as value for `s` as well.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -43,28 +42,42 @@ public void testMultipleNamesDeserializedCorrectly() {
...
+      e = e.getCause();
+    }
+    assertNotNull(e);
   }
 
-  private static record MyRecord(
+  @Test
+  public void testPrimitiveDefaultValues() {
+    RecordWithPrimitives zero = new RecordWithPrimitives((byte) 0, (short) 0, 0, 0, 0, 0, '\0', false);
+    assertEquals(zero, gson.fromJson("{}", RecordWithPrimitives.class));
```
Marcono1234, line 70: Maybe for completeness one test with explicit JSON `null` for each component would be useful as well? (combined with the comment above, it could then for example use `{"s": "nulls", "aByte": null, ...}`)

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
eamonnmcmanus: OK, done.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
eamonnmcmanus: I always hesitate between the two. :-) It feels a bit odd to talk about "this" when it is something we haven't seen yet. But with a trailing colon I think it is OK. Done.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
eamonnmcmanus: Yes, the exception stuff was clunky. I think I was actually using this test to check how to rewrite `@AutoValue` classes as records, which is hardly the right place for it. Done.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
eamonnmcmanus: Done.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
eamonnmcmanus: Done, though I'm kind of unhappy that an explicit `null` in the JSON produces these default values. I think that should probably produce a `JsonParseException`.

Marcono1234 (approved):

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
Marcono1234: That is a good point. It would probably be safe to make it strict initially and in case users complain we could still consider restoring the behavior with default values for JSON null.

Should I make that change in a separate PR?

eamonnmcmanus: I would be very happy to see that!

PR #2221, opened 2022-10-12
Simplify JsonReader

ThirdGoddess:
I think it can be modified here. What's your opinion?


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -737,7 +737,7 @@ private int peekNumber() throws IOException {
     }
 
     // We've read a complete number. Decide if it's a PEEKED_LONG or a PEEKED_NUMBER.
-    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative) && (value!=0 || false==negative)) {
+    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative) && (value!=0 || !negative)) {
```
Marcono1234, line 740: The negation could also be moved outside the parentheses and it might then be worth adding a comment explaining why this check exists:
```suggestion
    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative)
        // Don't lose information about -0, user might want to read it as double -0.0
        && !(value == 0 && negative)
    ) {
```
(not sure if this exactly follows the style guide)

Marcono1234 (commented):
Looks good, but please edit the title of your pull request to be more meaningful, for example: "Simplify JsonReader.peekNumber()"
(or similar)

gson/src/main/java/com/google/gson/stream/JsonReader.java
eamonnmcmanus: `false==negative` is certainly a bit eccentric. I agree that we should at least write `!negative`.

I think this complicated boolean expression can be simplified slightly:
```java
last == NUMBER_CHAR_DIGIT && fitsInLong && (negative ? value != 0 : value != Long.MIN_VALUE)
```
My reasoning is that if `negative` is true, `(value != Long.MIN_VALUE || negative)` is also true, and if `negative` is false, `(value != 0 || !negative)` is true. So we only need to check the other one of those two conditions in each case. We could also refactor further, to `value != (negative ? 0 : Long.MIN_VALUE)`, but I'm not sure that's an improvement.

Marcono1234: > I think this complicated boolean expression can be simplified slightly:
> 
> ```java
> last == NUMBER_CHAR_DIGIT && fitsInLong && (negative ? value != 0 : value != Long.MIN_VALUE)
> ```

But a comment to explain this condition would be good nonetheless, maybe something like this?
```
// Don't store -0 as long; user might want to read it as double -0.0
// Don't try to convert Long.MIN_VALUE to positive long; it would overflow MAX_VALUE
```

eamonnmcmanus: Agreed. Those comments look good.

Marcono1234:
@ThirdGoddess, do you want to include the changes we mentioned in the review thread above?

Marcono1234:
This change was implemented now by #2282, therefore I am going to close your pull request. Thank you nontheless for your work on this!

(Though I guess the condition in `JsonReader` could still be simplified as mentioned in https://github.com/google/gson/pull/2221#discussion_r993964588, in case someone wants to do that.)

PR #2224, opened 2022-10-14
Adjust Record adapter and extend test coverage

Marcono1234:
Changes:
- Make record constructor and accessor accessible (unless reflection access filter prevents it)
- Disallow `null` for primitive component, see https://github.com/google/gson/pull/2219#discussion_r992866260
- Skip `static` fields when deserializing records
Would not work without special casing and I am not sure if serializing or deserializing `static` fields was ever officially documented as supported (just happens to work by using `GsonBuilder.excludeFieldsWithModifiers`). For serialization there might be legit use cases (e.g. to encode a version number), but deserialization it rather questionable.
- Disallow `@SerializedName` placed only on accessor method
Implemented this because we have the accessor method available for this check anyway, and because for records the accessor method is actually called on serialization so users might expect `@SerializedName` to work on them.
- Adjust `ReflectionHelper.makeAccessible` to create description of object only when exception is thrown instead of eagerly
- Prefix names of test classes `ReflectiveTypeAdapterFactoryTest` and `ReflectionHelperTest` with `Java17`
This makes sure that we notice when the checked JDK class `UnixDomainPrincipal` is missing for some reason even though the test runs on Java >= 17. Otherwise these tests might just be silently skipped without us noticing.
Arguably these tests could also be rewritten to use custom record classes now that the tests are only compiled on Java 17 or newer, but on the other hand it might also be valuable that they test against a record class which is not part of Gson's tests.
- ... for further changes see messages of subsequent commits

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -19,6 +19,7 @@ package com.google.gson.internal.bind;
 import com.google.gson.FieldNamingStrategy;
 import com.google.gson.Gson;
 import com.google.gson.JsonIOException;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.ReflectionAccessFilter;
 import com.google.gson.ReflectionAccessFilter.FilterResult;
@@ -38,16 +39,18 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
+import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
-import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
-import java.util.LinkedHashMap;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -117,19 +120,22 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     // If the type is actually a Java Record, we need to use the RecordAdapter instead. This will always be false
     // on JVMs that do not support records.
     if (ReflectionHelper.isRecord(raw)) {
-      return new RecordAdapter<>(raw, getBoundFields(gson, type, raw, true, true));
+      @SuppressWarnings("unchecked")
+      TypeAdapter<T> adapter = (TypeAdapter<T>) new RecordAdapter<>(raw,
+          getBoundFields(gson, type, raw, blockInaccessible, true), blockInaccessible);
+      return adapter;
     }
 
     ObjectConstructor<T> constructor = constructorConstructor.get(type);
     return new FieldReflectionAdapter<>(constructor, getBoundFields(gson, type, raw, blockInaccessible, false));
   }
 
-  private static void checkAccessible(Object object, Field field) {
-    if (!ReflectionAccessFilterHelper.canAccess(field, Modifier.isStatic(field.getModifiers()) ? null : object)) {
-      throw new JsonIOException("Field '" + field.getDeclaringClass().getName() + "#"
-          + field.getName() + "' is not accessible and ReflectionAccessFilter does not "
-          + "permit making it accessible. Register a TypeAdapter for the declaring type "
-          + "or adjust the access filter.");
+  private static <M extends AccessibleObject & Member> void checkAccessible(Object object, M member) {
+    if (!ReflectionAccessFilterHelper.canAccess(member, Modifier.isStatic(member.getModifiers()) ? null : object)) {
+      String memberDescription = ReflectionHelper.getAccessibleObjectDescription(member, true);
+      throw new JsonIOException(memberDescription + " is not accessible and ReflectionAccessFilter does not "
+          + "permit making it accessible. Register a TypeAdapter for the declaring type, adjust the "
+          + "access filter or increase the visibility of the element and its declaring type.");
     }
   }
 
@@ -154,8 +160,14 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       @Override void write(JsonWriter writer, Object source)
           throws IOException, ReflectiveOperationException {
         if (!serialized) return;
-        if (blockInaccessible && accessor == null) {
-          checkAccessible(source, field);
+        if (blockInaccessible) {
+          if (accessor == null) {
+            checkAccessible(source, field);
+          } else {
+            // Note: This check might actually be redundant because access check for canonical
+            // constructor should have failed already
+            checkAccessible(source, accessor);
+          }
         }
 
         Object fieldValue = (accessor != null)
@@ -172,11 +184,13 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       }
 
       @Override
-      void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException {
+      void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException, JsonParseException {
         Object fieldValue = typeAdapter.read(reader);
-        if (fieldValue != null || !isPrimitive) {
-          target[index] = fieldValue;
+        if (fieldValue == null && isPrimitive) {
+          throw new JsonParseException("null is not allowed as value for record component '" + fieldName + "' "
+              + "of primitive type; at path " + reader.getPath());
         }
+        target[index] = fieldValue;
       }
 
       @Override
@@ -224,18 +238,33 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
         }
         // The accessor method is only used for records. If the type is a record, we will read out values
         // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
-        // If there is a static field on a record, there will not be an accessor. Instead we will use the default
-        // field logic for dealing with statics.
         Method accessor = null;
-        if (isRecord && !Modifier.isStatic(field.getModifiers())) {
-          accessor = ReflectionHelper.getAccessor(raw, field);
+        if (isRecord) {
+          // If there is a static field on a record, there will not be an accessor. Instead we will use the default
+          // field logic for dealing with statics. For deserialization the field is ignored.
+          if (Modifier.isStatic(field.getModifiers())) {
+            deserialize = false;
+          } else {
+            accessor = ReflectionHelper.getAccessor(raw, field);
+            // If blockInaccessible, skip and perform access check later
+            if (!blockInaccessible) {
+              ReflectionHelper.makeAccessible(accessor);
+            }
+
+            // @SerializedName can be placed on accessor method, but it is not supported there
+            // If field and method have annotation it is not easily possible to determine if accessor method
+            // is implicit and has inherited annotation, or if it is explicitly declared with custom annotation
+            if (accessor.getAnnotation(SerializedName.class) != null
+                && field.getAnnotation(SerializedName.class) == null) {
+              String methodDescription = ReflectionHelper.getAccessibleObjectDescription(accessor, false);
+              throw new JsonIOException("@SerializedName on " + methodDescription + " is not supported");
+            }
+          }
         }
 
-        // If blockInaccessible, skip and perform access check later. When constructing a BoundedField for a Record
-        // field, blockInaccessible is always true, thus makeAccessible will never get called. This is not an issue
-        // though, as we will use the accessor method instead for reading record fields, and the constructor for
-        // writing fields.
-        if (!blockInaccessible) {
+        // If blockInaccessible, skip and perform access check later
+        // For Records if the accessor method is used the field does not have to be made accessible
+        if (!blockInaccessible && accessor == null) {
           ReflectionHelper.makeAccessible(field);
         }
         Type fieldType = $Gson$Types.resolve(type.getType(), raw, field.getGenericType());
@@ -278,7 +307,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     abstract void write(JsonWriter writer, Object source) throws IOException, ReflectiveOperationException;
 
     /** Read the value into the target array, used to provide constructor arguments for records */
-    abstract void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException;
+    abstract void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException, JsonParseException;
 
     /** Read the value from the reader, and set it on the corresponding field on target via reflection */
     abstract void readIntoField(JsonReader reader, Object target) throws IOException, IllegalAccessException;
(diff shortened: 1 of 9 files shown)


gson/src/test/java/com/google/gson/internal/bind/Java17ReflectiveTypeAdapterFactoryTest.java
```diff
@@ -6,42 +6,39 @@
...
+    @SuppressWarnings("unused")
+    public String s;
   }
 
   @Test
   public void testCustomAdapterForRecords() {
     Gson gson = new Gson();
     TypeAdapter<?> recordAdapter = gson.getAdapter(unixDomainPrincipalClass);
-    TypeAdapter<?> defaultReflectionAdapter = gson.getAdapter(UserPrincipal.class);
+    TypeAdapter<?> defaultReflectionAdapter = gson.getAdapter(DummyClass.class);
```
Marcono1234, line 41: Have changed this because `UserPrincipal` is an interface without any fields, so assuming that the normal reflection based adapter is used for it might be a bit brittle.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -117,19 +120,22 @@ public <T> TypeAdapter<T> create(Gson gson, final TypeToken<T> type) {
...
-    if (!ReflectionAccessFilterHelper.canAccess(field, Modifier.isStatic(field.getModifiers()) ? null : object)) {
-      throw new JsonIOException("Field '" + field.getDeclaringClass().getName() + "#"
-          + field.getName() + "' is not accessible and ReflectionAccessFilter does not "
-          + "permit making it accessible. Register a TypeAdapter for the declaring type "
-          + "or adjust the access filter.");
+  private static <M extends AccessibleObject & Member> void checkAccessible(Object object, M member) {
+    if (!ReflectionAccessFilterHelper.canAccess(member, Modifier.isStatic(member.getModifiers()) ? null : object)) {
+      String memberDescription = ReflectionHelper.getAccessibleObjectDescription(member, true);
+      throw new JsonIOException(memberDescription + " is not accessible and ReflectionAccessFilter does not "
+          + "permit making it accessible. Register a TypeAdapter for the declaring type, adjust the "
```
eamonnmcmanus, line 137: While you're changing this, could you move the spaces to the beginning of each new string rather than the end? That makes it more obvious that we haven't forgotten one.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -172,11 +184,13 @@ private ReflectiveTypeAdapterFactory.BoundField createBoundField(
...
 
       @Override
-      void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException {
+      void readIntoArray(JsonReader reader, int index, Object[] target) throws IOException, JsonParseException {
         Object fieldValue = typeAdapter.read(reader);
-        if (fieldValue != null || !isPrimitive) {
-          target[index] = fieldValue;
+        if (fieldValue == null && isPrimitive) {
+          throw new JsonParseException("null is not allowed as value for record component '" + fieldName + "' "
+              + "of primitive type; at path " + reader.getPath());
```
eamonnmcmanus, line 191: Same remark about space here.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -297,10 +326,11 @@ protected BoundField(String name, String fieldName, boolean serialized, boolean
    * @param <T> type of objects that this Adapter creates.
    * @param <A> type of accumulator used to build the deserialization result.
    */
+  // This class is public because external projects check for this class (even though it is internal)
```
eamonnmcmanus, line 329: ...check for this class _with `instanceof`_...

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -24,65 +24,81 @@ public class ReflectionHelper {
...
-  private static void makeAccessible(String description, AccessibleObject object) throws JsonIOException {
+  public static void makeAccessible(AccessibleObject object) throws JsonIOException {
     try {
       object.setAccessible(true);
     } catch (Exception exception) {
-      throw new JsonIOException("Failed making " + description + "' accessible; either change its visibility "
-              + "or write a custom TypeAdapter for its declaring type", exception);
+      String description = getAccessibleObjectDescription(object, false);
+      throw new JsonIOException("Failed making " + description + " accessible; either increase its visibility "
+              + "or write a custom TypeAdapter for its declaring type.", exception);
```
eamonnmcmanus, line 39: Space at start (again, while you are here).

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -24,65 +24,81 @@ public class ReflectionHelper {
...
+    appendExecutableParameters(constructor, stringBuilder);
+
+    return stringBuilder.toString();
+  }
+
+  // Note: Ideally parameter type would be java.lang.reflect.Executable, but that was added in Java 8
+  private static void appendExecutableParameters(AccessibleObject executable, StringBuilder stringBuilder) {
+    stringBuilder.append('(');
+
+    Class<?>[] parameters = (executable instanceof Method) ? ((Method) executable).getParameterTypes()
```
eamonnmcmanus, line 92: I think this would be easier to read if `?` were on a new line.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -24,65 +24,81 @@ public class ReflectionHelper {
...
+  public static String getAccessibleObjectDescription(AccessibleObject object, boolean uppercaseFirstLetter) {
+    String description;
+
+    if (object instanceof Field) {
+      Field field = (Field) object;
+      description = "field '" + field.getDeclaringClass().getName() + "#" + field.getName() + "'";
+    } else if (object instanceof Method) {
+      Method method = (Method) object;
+
+      StringBuilder methodSignatureBuilder = new StringBuilder(method.getName());
```
eamonnmcmanus, line 58: This all seems like a lot of trouble to go to. It seems to me that the default `toString()` of `Method` and `Constructor` is good enough, isn't it?

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
```diff
@@ -99,7 +115,7 @@ public static String tryMakeAccessible(Constructor<?> constructor) {
       return null;
     } catch (Exception exception) {
       return "Failed making constructor '" + constructorToString(constructor) + "' accessible; "
-          + "either change its visibility or write a custom InstanceCreator or TypeAdapter for its declaring type: "
+          + "either increase its visibility or write a custom InstanceCreator or TypeAdapter for its declaring type: "
```
eamonnmcmanus, line 118: Spaces again. :-)

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -224,18 +238,33 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
...
         // The accessor method is only used for records. If the type is a record, we will read out values
         // via its accessor method instead of via reflection. This way we will bypass the accessible restrictions
-        // If there is a static field on a record, there will not be an accessor. Instead we will use the default
-        // field logic for dealing with statics.
         Method accessor = null;
-        if (isRecord && !Modifier.isStatic(field.getModifiers())) {
-          accessor = ReflectionHelper.getAccessor(raw, field);
+        if (isRecord) {
+          // If there is a static field on a record, there will not be an accessor. Instead we will use the default
+          // field logic for dealing with statics. For deserialization the field is ignored.
```
eamonnmcmanus, line 244: Just to clarify: by default, static fields in records will be neither serialized nor deserialized, right? Only if that weird `excludeFieldsWithModifiers` method is called? Because I definitely don't think static fields should be serialized.

eamonnmcmanus (changes requested):
Thanks for this!

This is my first pass. I think I will need to spend more time looking at this, though.

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
Marcono1234: The reason why I added this is that `toString()` can be a bit verbose because it:
- includes all modifiers (`public`, `final`, `transient`, ...)
- uses qualified names for parameter types
- for fields also includes the field type

This implementation here omits modifiers and uses simple names. Though on the other hand maybe there are situations where at least some modifiers are relevant?

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
Marcono1234: Yes that is correct, see default excluded modifiers here:
https://github.com/google/gson/blob/c2458bf876ef52a4489af937f7dd7c2709b74fb9/gson/src/main/java/com/google/gson/internal/Excluder.java#L55

I have mainly added this to handle it in a reasonable way instead of failing with some obscure exception.

For static fields maybe there are legit use cases, e.g. to automatically encode a version number:
```java
record Response(
   ...
) {
    public static final int VERSION = 1;
}
```
Though I am not sure if someone is actually using this, and it is certainly not obvious (and probably not even officially supported) to use `excludeFieldsWithModifiers` for this.

If you want I can adjust it to also always exclude static fields on serialization, and we can then wait for someone to request this feature.

eamonnmcmanus: Yes, I think by default static fields should not be serialized. I suppose they should be if someone calls `excludeFieldsWithModifiers(0)`, though even then I don't really understand why people would want that. I'm not really convinced by the `VERSION` use case. We would serialize that field, and then what would we do on deserialization? Do we check that the version matches? Do we try to overwrite the `VERSION` field? If we don't, isn't that inconsistent? I think if you really want a version then it should either be an instance field or you should use a custom `TypeAdapter`.

(I looked at the history, and apparently `excludeFieldsWithModifiers` was already present in the first version on GitHub, dated 2008. So who knows why it was added. For what it's worth, Google's code has only a handful of calls to this method, and exactly one that is including `STATIC`, perhaps accidentally.)

Marcono1234: > I'm not really convinced by the `VERSION` use case. We would serialize that field, and then what would we do on deserialization?

It would probably be most useful for services which only serialize data, e.g. a REST API which generates the response. At least that is the most reasonable situation I can come up with at the moment.

It appears the `excludeFieldsWithModifiers` trick is known, see [this StackOverflow answer](https://stackoverflow.com/a/33666187).

gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java
eamonnmcmanus: OK, I still think `toString()` would be good enough, but I agree with you about the noise it can introduce.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -67,21 +175,201 @@ public void testPrimitiveDefaultValues() {
...
+   * Tests serialization of a record with {@code static} field.
+   *
+   * <p>Important: It is not documented that this is officially supported; this
+   * test just checks the current behavior.
+   */
+  @Test
+  public void testStaticFieldSerialization() {
+    Gson gson = new GsonBuilder()
+        // Include static fields
+        .excludeFieldsWithModifiers(0)
```
eamonnmcmanus, line 232: Could you also test that the field is _not_ serialized without this?

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -67,21 +175,201 @@ public void testPrimitiveDefaultValues() {
...
+        .create();
+
+    String json = gson.toJson(new RecordWithStaticField());
+    assertEquals("{\"s\":\"initial\"}", json);
+  }
+
+  /**
+   * Tests deserialization of a record with {@code static} field.
```
eamonnmcmanus, line 240: Is this consistent with existing reflective deserialization? Maybe I'm misreading the code, but it doesn't look to me as if it treats deserialization differently from static. If someone really wants to serialize static fields (which as I've said I don't think they should) then maybe they want to deserialize them too?

Marcono1234: No, it is not consistent. For simplicity I changed it to always ignore static fields for record deserialization: [relevant lines](https://github.com/google/gson/pull/2224/files/e1f696c641d793081d16c95b22969f136837f37a#diff-925cdeedc560bb3d51ddc67046b656b95247a2a819b4b4b9dd6b4d139526b465R242-R246)

It would be possible to also support deserialization but that would require some adjustments due to record deserialization normally only invoking the canonical constructor. The current `master` state would just throw an exception because the static field has no corresponding canonical constructor parameter.

Though personally I don't think we should allow static field deserialization. It can cause thread safety issues and cause integrity issues when a `static final` field is modified whose constant value was inlined by the compiler at other locations. For records we can still prohibit (respectively ignore) static field deserialization without breaking backward compatibility because record deserialization was mostly not possible before (#1794).

Though maybe this test here does not make the current behavior clear enough?

eamonnmcmanus: OK, the point about the constructor is an important one. I still hate the idea of serializing _or_ deserializing static fields [:-)] but I can live with serializing them only following explicit configuration, and never deserializing them. Writing static fields is way worse than just reading them.

eamonnmcmanus (commented):
(It was unclear to me whether there is in fact test coverage for static fields not being serialized by default.)

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
Marcono1234: It looks like there was already a test for this: `com.google.gson.functional.UncategorizedTest.testStaticFieldsAreNotSerialized()`
And additionally multiple tests implicitly relied on `static` fields being ignored by default. However, I have now extended these new test methods to also test the default Gson behavior.

Also, I have noticed that `Field.set` per documentation does not support `static final` fields. So to avoid Gson claiming "Unexpected IllegalAccessException occurred...", I have added custom handling for that case to provide a better exception message.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
Marcono1234: > and never deserializing them

But only for records, right? This pull request keeps the behavior for the regular reflective adapter which supports deserialization of static fields.

Marcono1234:
Sorry for these additional commits. I did not consider these cases initially and only noticed now that they are not covered by any tests. I hope the additional commits I pushed are fine.

gson/src/test/java/com/google/gson/functional/Java17RecordTest.java
```diff
@@ -32,32 +53,167 @@ public final class Java17RecordTest {
...
+
+    LocalRecord deserialized = gson.fromJson("{\"s\": null}", LocalRecord.class);
+    assertEquals(new LocalRecord(null), deserialized);
+    assertEquals("custom-null", deserialized.s());
+  }
+
+  /** Tests behavior when the canonical constructor throws an exception */
+  @Test
+  public void testThrowingConstructor() {
+    record LocalRecord(String s) {
```
Marcono1234, line 142: It appears the Eclipse compiler produces malformed byte code for this record class (have not investigated that further yet). Just as forewarning in case you are experiencing a `java.lang.VerifyError` when running this test in Eclipse.

eamonnmcmanus (approved):
Looks good to me. Let me know if you think it is ready to be merged.

Marcono1234:
It should be ready to merge.

In case you plan to create a release soon (per Semantic Versioning that would be 2.10.0 I assume), it would be good if you could take a look at #2212. If those changes are too experimental / hacky in your opinion, then at least during the release process the `@since` Javadoc tags for the new methods would have to be added manually.

Issue #1755, opened 2020-08-18
Missing methods


priand:
Would be great if you can add a few missing methods:
JsonObject/JsonArray: clear() -> empty the container, similar to Collection.clear()
JsonArray: addAt() -> insert an element at a defined index, similar to List addAt()

Would also help to have a JsonPrimitive get() that doesn't do any check, but just returns the raw value (Boolean, Number or String)

These are easy, non breaking add-ons that can simplify some code :-)

Marcono1234:
#2225 has added the methods `JsonArray.asList` and `JsonObject.asMap` so it is at least now possible to use `clear()` and `addAt` (respectively `add(int, JsonElement)`) with the `List` and `Map` views from these methods.

PR #2225, opened 2022-10-15
Add `JsonArray.asList` and `JsonObject.asMap` view methods

Marcono1234:
Supersedes #1209, #1619 (and maybe also #1667, see comments)
Resolves potentially #1755; can be achieved with `asList()`

See discussion in https://github.com/google/gson/pull/2169#issuecomment-1221370207

I hope the tests cover everything relevant. I have also merged `JsonArrayTest` from the `functional` package with the other `JsonArrayTest` class.


gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -28,11 +29,14 @@
  * elements are added is preserved. This class does not support {@code null} values. If {@code null}
  * is provided as element argument to any of the methods, it is converted to a {@link JsonNull}.
  *
+ * <p>{@code JsonArray} only implements the {@link Iterable} interface but not the {@link List}
+ * interface, but a {@code List} view of it can be obtained with {@link #asList()}.
```
eamonnmcmanus, line 33: I would start a new sentence here.
> ...interface. A {@code List} view...

gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override public boolean equals(Object o) {
+    return delegate.equals(o);
+  }
+
+  @Override public int hashCode() {
+    return delegate.hashCode();
+  }
+
+  // TODO: Once Gson targets Java 8 also override List.sort
```
eamonnmcmanus, line 97: I think you could override it now, couldn't you? Just don't use the `@Override` annotation. Maybe that will trigger annoying warnings, though.

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -393,6 +397,20 @@ public boolean getAsBoolean() {
     return getAsSingleElement().getAsBoolean();
   }
 
+  /**
+   * Returns a mutable {@link List} view of this {@code JsonArray}. Changes to the {@code List}
+   * are visible for this {@code JsonArray} and the other way around.
```
eamonnmcmanus, line 402: I think I would write "visible in this" rather than "visible for this".

eamonnmcmanus (changes requested):
Thanks, this looks great! Only minor comments.

gson/src/main/java/com/google/gson/JsonArray.java
Marcono1234: Thanks! I wasn't really happy with the previous wording either.

gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java
Marcono1234: Good point, I did not think of this. Unfortunately `ArrayList.sort` does not exist in Java 7 so it would not provide a great advantage to override this now.

The reason why I added this comment is that `ArrayList.sort` directly [sorts the internal array](https://github.com/openjdk/jdk/blob/d3781ac8a38943d8a20304e770b01d5418ee33d0/src/java.base/share/classes/java/util/ArrayList.java#L1719-L1721), whereas the default `List.sort` [creates a new array and re-orders the elements through the iterator](https://github.com/openjdk/jdk/blob/d3781ac8a38943d8a20304e770b01d5418ee33d0/src/java.base/share/classes/java/util/List.java#L508-L516), so it would be nice to delegate to the more efficient `ArrayList` implementation.

Either way, this is only a nice-to-have feature but it is not required for correctness.

eamonnmcmanus: Yes, I was thinking you could call `Collections.sort` but that actually just delegates back to `List.sort`. So there isn't really anything you can usefully do without requiring Java 8.

eamonnmcmanus (approved):

PR #2226, opened 2022-10-19
fix(sec): upgrade com.fasterxml.jackson.core:jackson-databind to 2.14.0-rc1

Super-Sky:
### What happened？
There are 1 security vulnerabilities found in com.fasterxml.jackson.core:jackson-databind 2.13.4.2
- [CVE-2022-42003](https://www.oscs1024.com/hd/CVE-2022-42003)

### What did I do？
Upgrade com.fasterxml.jackson.core:jackson-databind from 2.13.4.2 to 2.14.0-rc1 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/metrics/pom.xml
+++ b/metrics/pom.xml
@@ -32,7 +32,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <version>2.13.4.2</version>
+      <version>2.14.0-rc1</version>
     </dependency>
     <dependency>
       <groupId>com.google.caliper</groupId>


Marcono1234:
Thanks, but it looks like version 2.13.4.2 is actually _not_ vulnerable, see [CVE description](https://nvd.nist.gov/vuln/detail/CVE-2022-42003):

> Additional fix version in 2.13.4.1 and 2.12.17.1
> [...]
> Up to (excluding) 2.13.4.1

Additionally the `metrics` module where you upgraded the dependency is not published and is only for local testing. Additionally and it does not use untrusted input as far as I know.

@eamonnmcmanus, do you agree that this dependency upgrade can be skipped?

eamonnmcmanus:
We apply the version updates from Dependabot punctually and I don't generally see a need to do more than that. I'd assume that if there is a vulnerability in Gson's dependencies, Dependabot will update past it.

PR #2227, opened 2022-10-24
Small tweaks to fix Error Prone warnings.

eamonnmcmanus:


gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -277,7 +277,7 @@ JsonElement nextJsonElement() throws IOException {
     JsonToken peeked = peek();
     switch (peeked) {
       case NAME:
-        nextName(true);
+        String unused = nextName(true);
```
Marcono1234, line 280: Should we add `@SuppressWarnings("unused")` to avoid IDE warnings?

eamonnmcmanus: If you send a PR for that, I will merge it. :-)
Error Prone doesn't warn about an unused variable if its name begins with `unused`.
Eventually I might want to add Error Prone's `@CanIgnoreReturnValue` to methods like this `nextName`, and then we wouldn't need to have this assignment. But that would mean introducing a dependency on Error Prone's annotations artifact.

PR #2228, opened 2022-10-25
Revise the version regex in `GsonVersionDiagnosticsTest`.

eamonnmcmanus:
Before we can release 2.10 we must support two-digit components. Additionally, there's no reason to require a patch number (2.10.0 rather than 2.10).

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/GsonVersionDiagnosticsTest.java
+++ b/gson/src/test/java/com/google/gson/functional/GsonVersionDiagnosticsTest.java
@@ -35,7 +35,7 @@ import junit.framework.TestCase;
  * @author Inderjeet Singh
  */
 public class GsonVersionDiagnosticsTest extends TestCase {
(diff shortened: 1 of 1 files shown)


Marcono1234:
If we were strictly following Semantic Versioning then the patch version [is required](https://semver.org/#spec-item-2):
> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes.

Though I am not sure if that matters much. Also thanks a lot for creating the release!

eamonnmcmanus:
Hmm, yes, starting with 2.8.0 in 2016 we've always included the trailing `.0`, so it was a mistake to have this release be 2.10 rather than 2.10.0. I think I'll undo that part of this PR and add a comment.

eamonnmcmanus:
Sent #2237 to fix the test and prepare for following the SemVer spec strictly again.

PR #2229, opened 2022-10-25
Add 2.10 changes to CHANGELOG; minor release follow-ups

Marcono1234:
Not sure though if it is worth keeping both the `CHANGELOG.md` and the GitHub releases up to date.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,7 @@
 
   <developers>
     <developer>
+      <id>google</id>
       <organization>Google</organization>
       <organizationUrl>https://www.google.com</organizationUrl>
     </developer>
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
   <developers>
     <developer>
       <organization>Google</organization>
-      <organizationUrl>http://www.google.com</organizationUrl>
+      <organizationUrl>https://www.google.com</organizationUrl>
     </developer>
   </developers>
 
@@ -211,6 +211,8 @@
                   <exec executable="git" dir="${project.basedir}" failonerror="true">
                     <arg value="add" />
                     <arg value="." />
+                    <!-- Don't add (unrelated) not yet tracked files -->
+                    <arg value="--update" />
                   </exec>
                 </target>
               </configuration>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -211,6 +212,8 @@
                   <exec executable="git" dir="${project.basedir}" failonerror="true">
                     <arg value="add" />
                     <arg value="." />
+                    <!-- Don't add (unrelated) not yet tracked files -->
+                    <arg value="--update" />
```
Marcono1234, line 216: Uses the [`--update` option](https://git-scm.com/docs/git-add#Documentation/git-add.txt---update) to not add untracked files. The Maven release plugin does not seem to detect those for its clean worktree check and this could otherwise lead to unrelated files being added accidentally.

Using this option should be fine because these steps for replacing version references are only expected to affect already tracked files.

pom.xml
```diff
@@ -33,8 +33,9 @@
 
   <developers>
     <developer>
+      <id>google</id>
```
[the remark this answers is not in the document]
eamonnmcmanus: I don't know what "the unique ID of the developer in the SCM" means exactly, but it could in fact be `google` given that this repo is `https://github.com/google/gson`?

pom.xml
eamonnmcmanus: OK

eamonnmcmanus (approved):
Thanks!

eamonnmcmanus:
Indeed, in some other Google projects we have given up on maintaining a change log since it is redundant with the GitHub release notes.

Marcono1234:
> Indeed, in some other Google projects we have given up on maintaining a change log since it is redundant with the GitHub release notes.

Would you prefer to do this here as well? We could keep `CHANGELOG.md` around for the previous releases, but at the beginning of it mention that for newer versions the changelog is documented on the GitHub releases.

eamonnmcmanus:
To keep things simple, why don't we merge this PR and then make a separate PR that adds some text to the top of `CHANGELOG.md` explaining that releases past 2.10 won't be documented there.

Marcono1234:
Have created #2230 for that. I hope that is what you had in mind, if not feel free to suggest adjustments.

PR #2231, opened 2022-11-01
Fixes #776: Add settings for the kind of newline to use

### LINKED ISSUE #776: Allow configuring the newLine used on Gson.toJson
Currently (version 2.5) there's no way to use Windows new lines on Gson, being able to configure this would be nice.

Also some old software doesn't like files that don't have a dangling new line at the end of the JSON, being able to configure this would be nice as well.

mihnita:
Addresses https://github.com/google/gson/issues/776

I don't think that `setFormattingStyles(String style)` (as proposed) should affect the newline, that would be orthogonal.

That idea has no explanation, but I'm thinking that it would be something similar to the formatting styles in programming languages (where to break and how to indent, for example K&R vs Linux kernel vs BSD KNF, etc.)
If that is the idea then the kind of newline to use is not related to the style.

The "dangling new line at the end of the JSON" would indeed be part of the style.


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -886,6 +892,21 @@ public JsonWriter newJsonWriter(Writer writer) throws IOException {
...
+        jsonWriter.setNewline(System.lineSeparator());
+        break;
+      case WINDOWS:
+        jsonWriter.setNewline("\r\n");
+        break;
+      case OLD_MACOS:
+        jsonWriter.setNewline("\r");
+        break;
+      case MACOS_AND_LINUX: // intentional fallthrough
+      default:
```
mihnita, line 906: If this would be a new library then the best `default` would probably be the same as `CURRENT_OS`.
But changing this after 14 years would probably break too many things.

For what ECMAScript says about line terminators see
https://262.ecma-international.org/5.1/#sec-7.3
There is nothing there giving special status to `LF` (`U+000A`)

Marcono1234: Just my personal opinion, but I am rather skeptical towards OS defaults. They have caused a lot trouble for the Java ecosystem in the past, and are still causing trouble (e.g. `String.toUpperCase()` or default encoding). If users want to have OS defaults, then they should explicitly specify that.

So I think `LF` is a good default.

Marcono1234 (commented):
Thanks a lot for this well written pull request!

Personally your reasoning regarding #776 makes sense to me. However, I am not a direct member of this project, so I won't be making the final decision.

My review comments are mostly intended as suggestions; I hope they are helpful. Also as warning (in case you have not encountered that yet): GitHub has the bad habbit to collapse unresolved review comments as part of its "... hidden conversations".

I am not sure if the current `NewlineStyle` constants are a good idea. The advantage is that it makes it easy to select a style for an OS, but on the other hand this binds Gson to something external: Which OS uses which newline. This could be a bit problematic if an OS changes its newline (quite unlikely though) or if a new OS becomes popular or one of these becomes obsolete (also rather unlikely). Maybe just using `LF`, `CRLF`, `CR` and `CURRENT_OS` would be better, assuming that users which want to use this feature know which newline Linux or Windows uses. What do you think?

(Also could you please edit the description of this pull request to say "Fixes #..." / "Resolves #..." instead of "Addresses #..." so that [GitHub understands](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) that this PR closes the issue.)

gson/src/main/java/com/google/gson/Gson.java
Marcono1234, line 906: Because this `switch` statement covers all constants, it might be good to omit the `default` case to get IDE warnings if a constant is not covered (not sure though if `javac` or CodeQL also warn about this), or use `throw new AssertionError()`.

mihnita: Ack.

We might disagree on the OS defaults.
But since we both end up with the same conclusion (using different arguments), it's all good :-)

Thanks,
M.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,52 @@
...
+ * @see <a href="https://en.wikipedia.org/wiki/Newline">Wikipedia Newline article</a>
+ *
+ * @since 2.10.1
+ */
+public enum NewlineStyle {
+  /**
+   * Using this style will result in the same kind of new lines that the current environment uses.
+   *
+   * <p>So it will produce {@code "\r\n"} when running on Windows, and {@code "\n"} when running
+   * on macOS &amp; Linux.</p>
```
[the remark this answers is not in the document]
mihnita: Done.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,52 @@
...
+   */
+  CURRENT_OS,
+
+  /**
+   * Using this style will result in the same new line convention that Windows uses
```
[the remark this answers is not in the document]
mihnita: Done

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -225,6 +230,22 @@ public final void setIndent(String indent) {
     }
   }
 
+  /**
+   * Sets the newline string to be used in the encoded document.
+   *
+   * @param newline the string that will be used for newline.
+   */
+  public final void setNewline(String newline) {
+    this.newline = newline == null ? "\n": newline;
```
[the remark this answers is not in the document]
mihnita: Done.

gson/src/test/java/com/google/gson/functional/NewlineTest.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.gson.common.TestTypes.BagOfPrimitives;
+import com.google.gson.reflect.TypeToken;
+
+/**
+ * Functional tests for pretty printing option.
+ *
+ * @author Inderjeet Singh
+ * @author Joel Leitch
+ */
+public class NewlineTest extends TestCase {
```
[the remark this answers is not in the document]
mihnita: Done

gson/src/main/java/com/google/gson/Gson.java
mihnita: I've checked to see who complains about an enum value not covered by a switch, and:
* just `javac` (no options) does not complain
* the regular `maven install` does not complain
* Visual Studio Code does not complain
* IntelliJ only complains if you run a code analyze, but not for a regular build
* Eclipse gives a warning, out of the box, there is no need to configure anything

So I would rather keep this "as is", if you don't mind.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Using this style will result in the same new line convention that macOS, Linux, and UNIX-like
+   * systems use. This is {@code LF} ({@code 0A}, {@code "\n"})
+   */
+  MACOS_AND_LINUX,
+
+  /**
+   * Using this style will result in the same new line convention that classic Mac OS used.
+   * Rarely needed. This is {@code CR} ({@code 0D}, {@code "\r"})
```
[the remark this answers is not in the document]
mihnita: Done

Sorry, I missed to accept the suggestion, because GitHub decided to hide this comment.
I know Github does that... and I hate it, but I still fall for it :-)

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -225,6 +230,22 @@ public final void setIndent(String indent) {
...
+  }
+
+  /**
+   * Returns the newline string used by this writer.
+   */
+  public final String getNewline() {
```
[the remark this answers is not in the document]
mihnita: I am not sure how to test it in a way that is not 100% trivial (set, then get and compare).

So I checked how the other getters are tested (`isLenient()`, `isHtmlSafe()`, `getSerializeNulls()`).
And they are not.

But if you think that something like this adds any value I can add this:
```
String nl = "foo";
writer.setNewline(nl);
assertEquals(nl, writer.getNewline());
```
I can even call `setNewline(null)`.

gson/src/main/java/com/google/gson/Gson.java
Marcono1234: Thanks for testing this so extensively!

What do you think about `default: throw new AssertionError()`? My concern is that the `default` case is really not supposed to be reached.

mihnita:
Implemented everything, but I still need to check that GitHub didn't hide anything from me :-)

And I did not resolve any conversation because I don't know what the convention is for this project.

I've contributed to various projects with different conventions.
I some I'm suppose to resolve to indicate I'm done.
In others I'm supposed to say "Done", and the reviewer resolves to show that they agree and it is indeed done.

Thanks for the review, and please don't hesitate to push back on the one or two items that I didn't do,
M.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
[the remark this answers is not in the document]
Marcono1234: Hmm, I haven't fully thought this through, sorry. Maybe there is indeed not much value in a test for this method then. But if you want feel free to add a trivial test for it nonetheless, I guess it does not hurt.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -225,6 +230,22 @@ public final void setIndent(String indent) {
     }
   }
 
+  /**
+   * Sets the newline string to be used in the encoded document.
```
Marcono1234, line 234: Would be good to clarify that this has no effect if no indent (with link to `#setIndent`) is set.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -488,6 +492,17 @@ public GsonBuilder setPrettyPrinting() {
     return this;
   }
 
+  /**
+   * Configures Gson to output JSON that uses a certain kind of newline.
+   * This option only affects JSON serialization.
```
Marcono1234, line 497: Would be good to clarify that this has no effect unless pretty printing (with link to `#setPrettyPrinting`) is used. Or should this implicitly activate pretty printing (if so that would have to be documented here)?

mihnita:
> I am not sure if the current NewlineStyle constants are a good idea.
> The advantage is that it makes it easy to select a style for an OS,
> but on the other hand this binds Gson to something external:
> Which OS uses which newline. This could be a bit problematic if
> an OS changes its newline (quite unlikely though) or if a new OS
> becomes popular or one of these becomes obsolete (also rather unlikely).
> Maybe just using LF, CRLF, CR and CURRENT_OS would be better, assuming
> that users which want to use this feature know which newline Linux or Windows uses.
> What do you think?

I think there are some good points here, but unfortunately I've seen a lot of developers that
(surprisingly) don't know what kind of end-of-line conventions are used for various systems.
Or even that there are differences :-(

A few alternative ideas:
* Use CR, LF, CRLF, CURRENT_OS, and add comments with the OS
* Keep it enum, and define CR, CRLF, etc, and also LINUX, MAC, etc.
  This means that `NewlineStyle.CRLF != NewlineStyle.WINDOWS`, complicating switches and what not.
* Define CR, LF, CRLF, CURRENT_OS, and add "aliases" (LINUX = LF)
  It might mean that the enum should be replace with a class with constants
* Make this a class with strings constants, and change `setNewlineStyle` to `setNewlineStyle(String)`.
  This means a user of this library can do `setNewlineStyle("\u2028")` and it would work, without a need to change the library.
  They can also define their own friendly constants outside the library.

Of course, each one has pros and cons :-(

I kind of like the last one though...
But the price is that the use in a `switch` would become harder to validate.

---

I think that indeed an OS changing the convention is a very-very unlikely event.
It would be a very disrupting event. And using CR and CRLF would probably not help.
Imagine macOS changing the to CRLF, and Linux staying the same.
I would have to grep the my code for LF and "sprinkle" if(s) everywhere, based on the OS.
Worse, figure out what version of macOS I'm dealing with.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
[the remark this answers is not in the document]
mihnita: Done.

gson/src/main/java/com/google/gson/Gson.java
mihnita: Done.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
mihnita: Done.

I used the more vague description _"Has no effect if the serialized format is a single line."_
because I am not sure on all the conditions.
For example `gson.toJson("test")` does not need a newline, no matter if we are using indent or not.

And if a condition changes, then the doc will be out of sync.
In the future the example above this might change to "use trailing newline if pretty-print", who knows :-)

mihnita:
Done with he implementation of the second round of review.
I've also checked that there are no "hidden comments" :-)

Unfortunately there are still a couple of items that I couldn't get myself to implement, as I still have some doubts.
I've explained what those are, but if you say "it's ok, just do it," I will.
There is nothing that I think is bad and I can't live with.

For example CRLF vs WINDOWS.
I think it might confuse some people? Yes.
Do I have data to show it? No.
Is it bad if they are a bit confused? Not really.
It is better than no setting at all, like now. They can just read the doc :-)

Thanks,
Mihai

gson/src/test/java/com/google/gson/functional/NewlineTest.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.gson.NewlineStyle;
+import com.google.gson.common.TestTypes.ArrayOfObjects;
+import com.google.gson.common.TestTypes.BagOfPrimitives;
+import com.google.gson.reflect.TypeToken;
+
+/**
+ * Functional tests for pretty printing option.
+ *
+ * @author Inderjeet Singh
+ * @author Joel Leitch
```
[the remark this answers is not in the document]
mihnita: Done

gson/src/main/java/com/google/gson/GsonBuilder.java
mihnita: Added a more generic description: _"Has no effect if the serialized format is a single line"_

So that it does not depend on the exact settings, which might change.
For example right now `Gson().tojson("foo")` is one line, pretty-print or not.
But if someone adds a `setTrailingNewline()` setting 2 years from now this comment would still be true.

mihnita:
I've slept a bit on this over night ;-)

What about this: I add the standard "glue" for an enum with a string value and I define all the values in the ECMAScript spec:
```
    CR("\r"), LF("\n"), CRLF("\r\n"), LS("\u2028"), PS("\u2029");
```
With comments specifying the OS using each (if any).
This way there are fewer chances that the enum will need changes anytime soon, because that would mean ECMAScript spec changes.

And the string values eliminates the need of a switch in `Gson`.
I can just do `jsonWriter.setNewline(newlineStyle.getValue())`

mihnita:
UGH :-(
Bad idea to define `PS` and `LS`.

From https://en.wikipedia.org/wiki/Newline#Unicode
_"JSON allows LS and PS characters within strings, while ECMAScript prior to ES2019 treated them as newlines, and therefore illegal syntax."_

And the fact that C transparently maps "\n" to whatever the OS uses as newline when doing IO. So "\n" on Windows actually means CR+LF, unless you open the file in binary mode. Which is something I know, as I've spent a lot of time on Windows. But a lot of people don't.

From readying the wiki article it looks like thinks are messy, with JSON using different rules than ECMAScript, "\n" not always == CR in some programming languages, and what not.

I also did an (informal) survey with some colleagues.
One out of three said they don't know what OS uses what line ending.
So we have to trust that people read the documentation.

mihnita:
I think it is all done form my side.

I've changed the enum to use `CURRENT_OS`, `CRLF`, `LF`, `CR`
With associated string part of the enum, so there is less of a need for a switch.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -488,6 +492,19 @@ public GsonBuilder setPrettyPrinting() {
...
+   *
+   * <p>Has no effect if the serialized format is a single line.</p>
+   *
+   * @return a reference to this {@code GsonBuilder} object to fulfill the "Builder" pattern
+   */
```
Marcono1234, line 502: Would probably be good to add `@since $next-version$` here as well.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,69 @@
...
+  private final String value;
+
+  private NewlineStyle(String value) {
+    this.value = value;
+  }
+
+  /**
+   * The string value that will be used as newline, not the name of the constant.
+   *
+   * @return the string value of the enum constant.
```
Marcono1234, line 64: What do you think about the following wording? "string value" (without more information) might be a bit too unspecific.
```suggestion
   * @return the newline value.
```

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -225,6 +230,24 @@ public final void setIndent(String indent) {
...
+   *
+   * <p>Has no effect if the serialized format is a single line.</p>
+   *
+   * @param newline the string that will be used for newline.
+   */
+  public final void setNewline(String newline) {
```
Marcono1234, line 240: Would probably be good to add `@since $next-version$` here as well.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -225,6 +230,24 @@ public final void setIndent(String indent) {
...
+    this.newline = Objects.requireNonNull(newline, "newline == null");
+  }
+
+  /**
+   * Returns the newline string used by this writer.
+   */
+  public final String getNewline() {
```
Marcono1234, line 247: Would probably be good to add `@since $next-version$` here as well.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,69 @@
...
+public enum NewlineStyle {
+  /**
+   * Using this style will result in the same kind of newline that the current environment uses.
+   *
+   * <p>So it will produce {@code "\r\n"} when running on Windows, and {@code "\n"} when running
+   * on macOS &amp; Linux.</p>
+   *
+   * @see System#lineSeparator()
+   */
+  CURRENT_OS(""),
```
Marcono1234, line 35: What about initializing this with `System.lineSeparator()` instead of `""`? That would avoid the special casing in `Gson` and would also lead to a more correct result for `getValue()`.

Or is there a reason why you did not do that?

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,69 @@
...
+  CR("\r");
+
+  private final String value;
+
+  private NewlineStyle(String value) {
+    this.value = value;
+  }
+
+  /**
+   * The string value that will be used as newline, not the name of the constant.
```
Marcono1234, line 62: What do you think about `getNewline()` (or `getNewlineValue()`) as name for this method? That would probably also render the "not the name of the constant" part obsolete.

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,69 @@
...
+ * @see <a href="https://en.wikipedia.org/wiki/Newline">Wikipedia Newline article</a>
+ *
+ * @since $next-version$
+ */
+public enum NewlineStyle {
+  /**
+   * Using this style will result in the same kind of newline that the current environment uses.
+   *
+   * <p>So it will produce {@code "\r\n"} when running on Windows, and {@code "\n"} when running
+   * on macOS &amp; Linux.</p>
```
Marcono1234, line 31: Maybe it would be good to recommend here to not use `CURRENT_OS` when exchanging the JSON data between systems.

Marcono1234 (commented):
Thanks a lot for the great amount of work and thoughts you have put into this! I have only a few minor comments but I think this looks very good otherwise.

Regarding the `NewlineStyle` constants, I think your current solution with `LF`, `CR`, ... is good, but I am not sure either what the "best" solution would be.

Though as mentioned before, I am not a direct member of this project. So @eamonnmcmanus, what you think about this pull request?

gson/src/test/java/com/google/gson/functional/NewlineTest.java
```diff
@@ -0,0 +1,95 @@
...
+  public void testNewlineCrLf() {
+    Gson gson = new GsonBuilder()
+        .setPrettyPrinting()
+        .setNewlineStyle(NewlineStyle.WINDOWS)
+        .create();
+    String json = gson.toJson(map);
+    assertEquals(EXPECTED.replace("<EOL>", "\r\n"), json);
+  }
+
+  public void testNewlineLF() {
```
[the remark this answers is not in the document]
mihnita: Done

gson/src/main/java/com/google/gson/stream/JsonWriter.java
mihnita: Done.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
mihnita: Done.

gson/src/main/java/com/google/gson/NewlineStyle.java
mihnita: Done.

> Or is there a reason why you did not do that?

Because it didn't cross my mind :-)

gson/src/main/java/com/google/gson/NewlineStyle.java
mihnita: Done.

Used `getNewlineValue()`

gson/src/main/java/com/google/gson/GsonBuilder.java
mihnita: Done

gson/src/main/java/com/google/gson/NewlineStyle.java
mihnita: Done

gson/src/main/java/com/google/gson/NewlineStyle.java
mihnita: Done.

Also added a test (`NewlineTest.testNewlineInterop`) making sure that all
the combinations of parsers and inputs work, regardless the newline style.

mihnita:
Done, all feedback implemented (2022/11/03)

gson/src/main/java/com/google/gson/NewlineStyle.java
Marcono1234: Thanks, I think this can be simplified now:
```suggestion
   * The string value that will be used as newline.
```

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * <p>It might be good discipline to not use {@code CURRENT_OS} when exchanging JSON data
+   * between systems, or storing them long term.<br>
+   * Gson will parse any valid JSON regardless the newline style.<br>
+   * But mixing newline styles might afect other kind of operations
```
Marcono1234, line 36:
```suggestion
   * But mixing newline styles might affect other kind of operations
```

gson/src/main/java/com/google/gson/NewlineStyle.java
```diff
@@ -0,0 +1,75 @@
...
+public enum NewlineStyle {
+  /**
+   * Using this style will result in the same kind of newline that the current environment uses.
+   *
+   * <p>So it will produce {@code "\r\n"} when running on Windows, and {@code "\n"} when running
+   * on macOS &amp; Linux.</p>
+   *
+   * <p>It might be good discipline to not use {@code CURRENT_OS} when exchanging JSON data
+   * between systems, or storing them long term.<br>
+   * Gson will parse any valid JSON regardless the newline style.<br>
```
Marcono1234, line 35: Really minor, but maybe the `<br>` here could be omitted since the next sentence is still closely related to the previous one.
```suggestion
   * Gson will parse any valid JSON regardless the newline style.
```

Marcono1234 (commented):
Thanks a lot! Only a few minor things which came up now, but feel free to wait until Éamonn has given feedback on this pull request as a whole as well.

mihnita:
Thanks!
Submitted all to get them out of the way :-)
Waiting for Éamonn. No rush.
Mihai

mihnita:
Gentle ping?

Do you think that adding Éamonn as reviewer might help?
Maybe just mentioning (with  @eamonnmcmanus) is easier to miss...

Thank you,
Mihai

eamonnmcmanus:
Hi, just back from vacation. :-)

My initial reactions:

1. Rather than `GsonBuilder.setNewlineStyle` which is only relevant if `GsonBuilder.setPrettyPrinting()` has been called, I think it might be better to overload `GsonBuilder.setPrettyPrinting` with a version that takes a parameter describing the pretty-printing style. We could imagine that this might later include things like indentation (currently [hardcoded](https://github.com/google/gson/blob/3f1e62d1e1b43f1d76af5106f9484e940bf74606/gson/src/main/java/com/google/gson/Gson.java#L887) to two spaces).
2. I'm not really convinced that the `NewlineStyle` class is justified. If we had a `FormattingStyle` class then we might imagine it looking like this initially:

```java
public class FormattingStyle {
  private final String newline;

  private FormattingStyle(String newline) {
    this.newline = newline;
  }

  public static final FormattingStyle DEFAULT = new FormattingStyle("\n");

  public FormattingStyle withNewline(String newline) {
    return new FormattingStyle(newline);
  }
}
```

which would evolve in the expected way if we add other parameters like indent. I don't think it is really necessary to have constants for the kinds of newline style. We can certainly document the two prevailing styles, and perhaps require that the `newline` string match `[\r\n]+`. If people need a `CRLF` constant they can trivially define one themselves.

I'll also note, playing Devil's Advocate a bit, that it's pretty easy to do `gson.toJson(...).replace("\n", "\r\n")`.

mihnita:
> I'll also note, playing Devil's Advocate a bit, that it's pretty easy to do `gson.toJson(...).replace("\n", "\r\n")`.

That's true, and it is something I did for years (and it annoyed me for years :-).
It feels "unclean" and it renders all the `toJson` methods that take a `JsonWriter` useless, and probably doubles the memory consumption (probably not an issue).
One can also bring the same "devil's advocate" argument about the `toJson` + `JsonWriter`: you can easily toJson to a String and write it yourself in a writer.
To apply the workaround you have to either do it every time you invoke `toJson`, or wrap the whole gson in a custom class that does that.

So, there are workarounds, but at the core this makes the people who need Windows EOL feel like second class citizens.

When the JSON spec (and ECMAScript spec) does not specify a certain EOL as preferred, they accept all the valid combinations.

In fact there is an effort to update the spec to also allow for `U+2028` (LINE SEPARATOR) and `U+2029` (PARAGRAPH SEPARATOR), to bring JSON closer to ECMAScript:
https://github.com/tc39/proposal-json-superset/blob/master/README.md
Apparently _"It is already implemented in Chrome 66 and Safari TP49"_ (https://bugzilla.mozilla.org/show_bug.cgi?id=1435828). And in progress (or done?) in Firefox.
So being able to specify that as EOL would be nice. If it's valid / standard, why not?

---

Anyway...
About the implementation, if you are open to accept a contribution,
I would be happy to make this change follow the direction you suggest.

Maybe move this discussion into an issue, so that you can review the direction before I implement something debatable?

I would probably go with a more "fluent" `FormattingStyle`?
So that in time one can do `style.newline(CRLF).indent("\t").bracketStyle(KandR).someOtherOption(opt)`?
(TBD if newline would be an enum or just a string)

Thank you,
Mihai

mihnita:
Copied the last set of comments (about `FormattingStyle`) to issue #776

Hoping we can outline there what the preferred way would be, and I can follow up with a new PR or update this one. Your call.

Thanks,
Mihai

eamonnmcmanus:
Yes, for a nontrivial API change like this it's probably better to agree on the API in an issue first before committing to an implementation with tests and everything.

mihnita:
Updated to `FormattingStyle`

Marcono1234:
Just to make sure there is no misunderstanding; I hope no one was waiting for my review. For me the initial implementation was fine, however I did not consider back then how the implementation can be made more generic as done now with the `FormattingStyle` (and as discussed in the linked issue, and the above comments).

@eamonnmcmanus, are these changes similar to what you had in mind? (though hopefully this also has time until next year)

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+  }
+
+  @Test
+  public void testStyleValidations() {
+    try {
+      // TBD if we want to accept \u2028 and \u2029. For now we don't.
+      FormattingStyle.DEFAULT.withNewline("\u2028");
+      fail("Gson should not accept anything but \r and \r for newline");
```
eamonnmcmanus, line 132: I think these messages should say
  ...anything but \\r and \\n...
So double backslash, and the second one should be n rather than r.

eamonnmcmanus (changes requested):
Sorry to have lost track of this. It looks great! Just some small comments.

gson/src/main/java/com/google/gson/FormattingStyle.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * The string value that will be used as newline.
+   *
+   * @return the newline value.
+  */
+  public String getNewline() {
+    return this.newline;
+  }
+
+  public String getIndent() {
```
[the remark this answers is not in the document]
mihnita: Done

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
mihnita: Done

gson/src/main/java/com/google/gson/FormattingStyle.java
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson;
+
+import java.util.Objects;
+
+/**
+ * An enumeration that defines the kind of newline to use for serialization.
```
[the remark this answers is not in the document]
mihnita: Done

mihnita:
> Sorry to have lost track of this. It looks great! Just some small comments.

Don't worry, I know what is like :-)

I've added comments to all public methods, generated the javadoc and checked what they look like, checked the links, etc.

Thank you,
Mihai

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks for doing this!

mihnita:
Thank you very much for taking it! 
:-)

PR #2233, opened 2022-11-04
Added JsonObject method isEmpty()

dhoard:
**Code change**

Added JsonObject method `isEmpty()`.

**Rationale**

There are scenarios where there is a need to check whether a `JsonObject` contains any key/value pairs.

This mirrors the existing functionality of `JsonArray.isEmpty()`

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -148,11 +148,21 @@ public final class JsonObject extends JsonElement {
   /**
    * Returns the number of key/value pairs in the object.
    *
-   * @return the number of key/value pairs in the object.
+   * @return the number of key/value pairs in the objezct.
    * @since 2.7
    */
   public int size() {
     return members.size();
+  }
+
+  /**
+   * Returns true if the number of key/value pairs in the object is zero.
+   *
+   * @return true if the number of key/value pairs in the object is zero.
+   * @since $next-version$
+   */
+  public boolean isEmpty() {
+    return members.size() == 0;
   }
 
   /**
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -148,7 +148,7 @@ public final class JsonObject extends JsonElement {
   /**
    * Returns the number of key/value pairs in the object.
    *
-   * @return the number of key/value pairs in the objezct.
+   * @return the number of key/value pairs in the object.
    * @since 2.7
    */
   public int size() {
--- a/gson/src/test/java/com/google/gson/JsonObjectTest.java
+++ b/gson/src/test/java/com/google/gson/JsonObjectTest.java
@@ -225,13 +225,13 @@ public class JsonObjectTest {
   @Test
   public void testIsEmpty() {
     JsonObject o = new JsonObject();
-    assertEquals(true, o.isEmpty());
+    assertTrue(o.isEmpty());
 
     o.add("Hello", new JsonPrimitive(1));
-    assertEquals(false, o.isEmpty());
+    assertFalse(o.isEmpty());
 
     o.remove("Hello");
-    assertEquals(true, o.isEmpty());
+    assertTrue(o.isEmpty());
   }
 
   @Test
--- a/gson/src/test/java/com/google/gson/JsonObjectTest.java
+++ b/gson/src/test/java/com/google/gson/JsonObjectTest.java
@@ -223,6 +223,18 @@ public class JsonObjectTest {
   }
 
   @Test
+  public void testIsEmpty() {
+    JsonObject o = new JsonObject();
+    assertEquals(true, o.isEmpty());
+
+    o.add("Hello", new JsonPrimitive(1));
+    assertEquals(false, o.isEmpty());
+
+    o.remove("Hello");
+    assertEquals(true, o.isEmpty());
+  }
+
+  @Test
   public void testDeepCopy() {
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -155,6 +155,15 @@ public int size() {
     return members.size();
   }
 
+  /**
+   * Returns true if the number of key/value pairs in the object is zero.
+   *
+   * @return true if the number of key/value pairs in the object isgit sd zero.
```
Marcono1234, line 161:
```suggestion
   * @return true if the number of key/value pairs in the object is zero.
```
Or alternatively maybe you could also omit the `@return` since it does not add much value, see also [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment), which implicitly shows that `@return` can be omitted (if I understand it correctly).

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -155,6 +155,15 @@ public int size() {
...
+   */
```
Marcono1234, line 162: Could you please also add `@since` (see #2212):
```suggestion
   * @since $next-version$
   */
```

Marcono1234 (commented):
Thanks for this pull request!

Could you please for completeness sake also add a small test to `JsonObjectTest`?

Your argument about consistency with `JsonArray.isEmpty()` is reasonable, but on the other hand the question might be how many `Map` convenience methods we want to add, see also related https://github.com/google/gson/pull/1667#issuecomment-1207502023.

gson/src/main/java/com/google/gson/JsonObject.java
dhoard: Done.

gson/src/main/java/com/google/gson/JsonObject.java
dhoard: Done.

dhoard:
> Your argument about consistency with `JsonArray.isEmpty()` is reasonable, but on the other hand the question might be how many `Map` convenience methods we want to add, see also related [#1667 (comment)](https://github.com/google/gson/pull/1667#issuecomment-1207502023).

I feel that `isEmpty()` and `clear()` ( mentioned in https://github.com/google/gson/pull/1667#issuecomment-1207502023 ) make sense because they are around the object's internal state.

Other `Map` methods are around conversion, which I see as a different aspect,  and agree they can become unwieldy.

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -148,13 +148,23 @@ public Set<String> keySet() {
   /**
    * Returns the number of key/value pairs in the object.
    *
-   * @return the number of key/value pairs in the object.
+   * @return the number of key/value pairs in the objezct.
```
Marcono1234, line 151: Looks like you introduced a different typo by accident:
```suggestion
   * @return the number of key/value pairs in the object.
```

Also, there is no need to force-push. In case these changes are integrated, the commits will be squashed anyways.

dhoard: Appologies. resolved

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -222,6 +222,18 @@ public void testSize() {
     assertEquals(1, o.size());
   }
 
+  @Test
+  public void testIsEmpty() {
+    JsonObject o = new JsonObject();
+    assertEquals(true, o.isEmpty());
```
mariocandela, line 228:
```suggestion
    assertTrue(o.isEmpty());
```

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -222,6 +222,18 @@ public void testSize() {
...
+
+    o.add("Hello", new JsonPrimitive(1));
+    assertEquals(false, o.isEmpty());
```
mariocandela, line 231:
```suggestion
    assertFalse(o.isEmpty());
```

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -222,6 +222,18 @@ public void testSize() {
...
+
+    o.remove("Hello");
+    assertEquals(true, o.isEmpty());
```
mariocandela, line 234:
```suggestion
    assertTrue(o.isEmpty());
```

mariocandela (commented):
What do you think about it ?

gson/src/test/java/com/google/gson/JsonObjectTest.java
dhoard: Changed.

gson/src/test/java/com/google/gson/JsonObjectTest.java
dhoard: Changed.

gson/src/test/java/com/google/gson/JsonObjectTest.java
dhoard: Changed.

mariocandela (approved):

eamonnmcmanus:
Given that we already have both `JsonArray.isEmpty()` and `JsonObject.size()`, I agree that this method makes sense for consistency. Additionally, it appears that most of Google's internal uses of `JsonObject.size()` in non-test code are comparing the size against 0.

PR #2243, opened 2022-11-24
Fix the javadoc of JsonDeserializer.deserialize()

MaicolAntali:
In the javadoc of the JsonDeserializer.deserialize() there was a double "the"


eamonnmcmanus:
Thanks!

PR #2251, opened 2022-12-03
Improve exception message for duplicate field names

Marcono1234:
This hopefully makes it easier to troubleshoot this issue, especially when there is a conflict between `@SerializedName` and an inherited field, or when users depend on Java or Android platform types. In the latter case it will be easier as external person to spot the underlying cause of the issue (i.e. serializing third-party types).

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -139,7 +139,7 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
     }
   }
 
-  private ReflectiveTypeAdapterFactory.BoundField createBoundField(
+  private BoundField createBoundField(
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -310,14 +310,16 @@ private Map<String, BoundField> getBoundFields(Gson context, TypeToken<?> type,
 
   static abstract class BoundField {
     final String name;
+    final Field field;
```
Marcono1234, line 313: As side note: Most likely `BoundField` (or rather its anonymous subclass) already had an implicit `Field` field since it captured the `Field` from the enclosing scope:
https://github.com/google/gson/blob/bc04dd20b969cb8a6999282b6749266b9f477bcf/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L164-L170

eamonnmcmanus:
This looks great, thanks!

PR #2271, opened 2022-12-06
Remove covered condition in JsonNull.equals()

MaicolAntali:
The Condition `this == other` is covered by subsequent condition `other instanceof JsonNull`

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonNull.java
+++ b/gson/src/main/java/com/google/gson/JsonNull.java
@@ -64,6 +64,6 @@ public final class JsonNull extends JsonElement {
    */
   @Override
   public boolean equals(Object other) {
-    return this == other || other instanceof JsonNull;
+    return other instanceof JsonNull;
   }
 }


eamonnmcmanus:
Fair point! Thanks for the contribution.

Marcono1234:
One question would be though whether `this == other` is faster than `other instanceof JsonNull`. That is relevant because code would mostly work with the `JsonNull.INSTANCE` singleton. But caring about that might be premature optimization.

eamonnmcmanus:
Well, `this == other` would only be faster when comparing against `JsonNull.INSTANCE`, right? Whereas I think you might just as often be comparing against instances of other `JsonElement` classes. Anyway I found only 3 production classes in all of Google's source that did `.equals(JsonNull.INSTANCE)`. Hundreds are doing `.isJsonNull()`, which sidesteps the question. (To be fair, I also didn't find many `Set<JsonElement>` or `Map<JsonElement,` cases either.)

On the whole the performance difference is likely to be absolutely negligible so we might as well simplify.

PR #2276, opened 2022-12-09
Remove the `final` keyword from `private` method

MaicolAntali:
Remove the `final` keyword from `private` method. `private` methods are implicitly `final`


eamonnmcmanus:
Ah yes, that other method in #2275 is `@SafeVarargs` and in the older Java versions we build with such a method must be either `static` or `final`. :-( Looks like Java 9 also allows `private` but we build with Java 8.

PR #2277, opened 2022-12-09
Unnecessary unboxing at JsonPrimitive.getAsBoolean()

MaicolAntali:
The `value` is cast to a `Boolean`. The `.booleanValue()` is unnecessary.


eamonnmcmanus:
It seems you're doing a number of different small cleanups to the code base, and thanks for that! I wonder if you could consolidate some of them into bigger PRs, though?

MaicolAntali:
@eamonnmcmanus yes make sense! 😄 

If for you it's ok merge this one PR. The next PRs I’ll make them bigger and more consolidate

PR #2279, opened 2022-12-10
No-args constructor may be shrinked by obfuscator. Need try newUnsafeAllocator

beavi5:
No-args constructor may be shrinked by obfuscator. Need try newUnsafeAllocator or some else.


Marcono1234:
I am going to close this pull request because it seems to be identical to your other pull request #2280.

PR #2280, opened 2022-12-10
Because no-args constructor may be shrinked by obfuscator. Need try newUnsafeAllocator or some else.

beavi5:
Because no-args constructor may be shrinked by obfuscator. Need try newUnsafeAllocator or some else.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -265,13 +265,13 @@ public final class ConstructorConstructor {
         // Note: InstantiationException should be impossible because check at start of method made sure
         //   that class is not abstract
         catch (InstantiationException e) {
-          throw new RuntimeException("Failed to invoke constructor '" + ReflectionHelper.constructorToString(constructor) + "'"
-              + " with no args", e);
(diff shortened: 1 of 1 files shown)


Marcono1234:
Thanks for the pull request! How does this one differ from your other open one #2279?

Also could you please share the full stack trace which would normally occur without this fix? I can imagine that ProGuard / R8 might remove the no-args constructor but in that case the code you changed would not run because Gson would not find the no-args constructor in the first place. I wouldn't expect that ProGuard / R8 preserves the no-args constructor but corrupts it in some way, causing it to throw exceptions.

I am also not quite sure if the change you are proposing here has the desired effect of falling back to using `Unsafe`; for that `newDefaultConstructor` would have to return `null`, and not a `ObjectConstructor` whose `construct` method returns `null`. I assume that would only lead to a `NullPointerException` later on.

Marcono1234:
@beavi5, as mentioned above, please clarify which problem this pull request tries to solve. Otherwise we will probably have to close this pull request.

Marcono1234:
I am going to close this pull request because it is still not clear to me what the intention behind these changes is, and if they would really work as desired, see comments above.

@beavi5, feel free to comment here if you want to provide more information or want to continue working on this.

PR #2281, opened 2022-12-10
Remove EOFException special casing of JsonStreamParser.next()

### LINKED ISSUE #2110: OSS-Fuzz issue 40999
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5741446309740544 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40999 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

Marcono1234:
The previous behavior violated the Iterator contract where for `JsonStreamParser("[")` a call to `hasNext()` would return true, but `next()` would throw a NoSuchElementException.
I currently cannot think of a use case where such behavior would be desired for incomplete JSON. Maybe that code is just a historic remnant from when `next()` did not delegate to `hasNext()` which itself would run into the `EOFException`.

This should hopefully fix issue #2110 and #2176, though we can also keep them open and (hopefully) let oss-fuzz detect on its own that they are resolved.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonStreamParser.java
+++ b/gson/src/main/java/com/google/gson/JsonStreamParser.java
@@ -75,7 +75,7 @@ public final class JsonStreamParser implements Iterator<JsonElement> {
    * {@link NoSuchElementException} if no element is available.
    *
    * @return the next available {@code JsonElement} on the reader.
-   * @throws JsonSyntaxException if the incoming stream is malformed JSON.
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Thanks! I ran this past all of Google's internal tests and didn't find any problems.

PR #2282, opened 2022-12-11
Code cleanup

MaicolAntali:
I have cleanup the codebase

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -47,14 +47,14 @@ public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
   public Date read(JsonReader in) throws IOException {
     try {
         switch (in.peek()) {
-            case NULL:
-                in.nextNull();
-                return null;
-            default:
-                String date = in.nextString();
-                // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
-                // This is because Android doesn't support XXX because it is JDK 1.6
-                return parse(date, new ParsePosition(0));
+        case NULL:
+            in.nextNull();
+            return null;
+        default:
+            String date = in.nextString();
+            // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
+            // This is because Android doesn't support XXX because it is JDK 1.6
+            return parse(date, new ParsePosition(0));
         }
     } catch (ParseException e) {
       throw new JsonParseException(e);
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -19,8 +19,11 @@ package com.google.gson.typeadapters;
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.ParsePosition;
-import java.util.*;
-
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.Locale;
+import java.util.TimeZone;
 import com.google.gson.JsonParseException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.stream.JsonReader;
@@ -43,14 +46,16 @@ public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
   @Override
   public Date read(JsonReader in) throws IOException {
     try {
-        if (Objects.requireNonNull(in.peek()) == JsonToken.NULL) {
-            in.nextNull();
-            return null;
+        switch (in.peek()) {
+            case NULL:
+                in.nextNull();
+                return null;
+            default:
+                String date = in.nextString();
+                // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
+                // This is because Android doesn't support XXX because it is JDK 1.6
+                return parse(date, new ParsePosition(0));
         }
-        String date = in.nextString();
-        // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
-        // This is because Android doesn't support XXX because it is JDK 1.6
-        return parse(date, new ParsePosition(0));
     } catch (ParseException e) {
       throw new JsonParseException(e);
     }
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -46,16 +46,16 @@ public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
   @Override
   public Date read(JsonReader in) throws IOException {
     try {
-        switch (in.peek()) {
+      switch (in.peek()) {
         case NULL:
-            in.nextNull();
-            return null;
+          in.nextNull();
+          return null;
         default:
-            String date = in.nextString();
-            // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
-            // This is because Android doesn't support XXX because it is JDK 1.6
-            return parse(date, new ParsePosition(0));
-        }
+          String date = in.nextString();
+          // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
+          // This is because Android doesn't support XXX because it is JDK 1.6
+          return parse(date, new ParsePosition(0));
+      }
     } catch (ParseException e) {
       throw new JsonParseException(e);
     }
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -19,15 +19,12 @@ package com.google.gson.typeadapters;
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.ParsePosition;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.GregorianCalendar;
-import java.util.Locale;
-import java.util.TimeZone;
+import java.util.*;
 
 import com.google.gson.JsonParseException;
 import com.google.gson.TypeAdapter;
 import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 
 public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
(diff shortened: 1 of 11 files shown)


extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
```diff
@@ -19,15 +19,12 @@
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.ParsePosition;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.GregorianCalendar;
-import java.util.Locale;
-import java.util.TimeZone;
+import java.util.*;
```
Marcono1234, line 22: Wildcard imports [should be avoided](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
```diff
@@ -46,16 +43,14 @@ public void write(JsonWriter out, Date date) throws IOException {
   @Override
   public Date read(JsonReader in) throws IOException {
     try {
-      switch (in.peek()) {
-      case NULL:
-        in.nextNull();
-        return null;
-      default:
+        if (Objects.requireNonNull(in.peek()) == JsonToken.NULL) {
```
Marcono1234, line 46: `Objects.requireNonNull` seems to be redundant; if `in.peek()` returns `null`, then there is an implementation bug in `JsonReader`

gson/src/main/java/com/google/gson/internal/Excluder.java
```diff
@@ -198,11 +198,7 @@ private boolean excludeClassChecks(Class<?> clazz) {
           return true;
       }
 
-      if (isAnonymousOrNonStaticLocal(clazz)) {
-          return true;
-      }
-
-      return false;
+    return isAnonymousOrNonStaticLocal(clazz);
```
Marcono1234, line 201: The indentation here is inconsistent with the indentation of the other statements of this method. Though maybe it would be best to just fix the indentation of `excludeClassChecks`, `excludeClass` and `excludeClassInStrategy` to use the expected 2 spaces (instead of 4).

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -737,7 +737,7 @@ private int peekNumber() throws IOException {
     }
 
     // We've read a complete number. Decide if it's a PEEKED_LONG or a PEEKED_NUMBER.
-    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative) && (value!=0 || false==negative)) {
+    if (last == NUMBER_CHAR_DIGIT && fitsInLong && (value != Long.MIN_VALUE || negative) && (value!=0 || !negative)) {
```
Marcono1234, line 740: In a different pull request there was a discussion about further simplifying this and also adding comments explaining these checks: https://github.com/google/gson/pull/2221#discussion_r993964588

Would you mind implementing that? It is not clear if the author of that pull request intends to update the pull request.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1381,11 +1374,9 @@ private TypeAdapter<T> delegate() {
...
   @Override
   public String toString() {
-    return new StringBuilder("{serializeNulls:")
-        .append(serializeNulls)
-        .append(",factories:").append(factories)
-        .append(",instanceCreators:").append(constructorConstructor)
-        .append("}")
-        .toString();
+    return "{serializeNulls:" + serializeNulls +
+            ",factories:" + factories +
```
Marcono1234, line 1378: Not sure if the indentation here is correct, I think this should have 4 spaces relative from the start of `return` (instead of 8).

metrics/src/main/java/com/google/gson/metrics/BagOfPrimitives.java
```diff
@@ -43,14 +43,12 @@ public int getIntValue() {
...
-    StringBuilder sb = new StringBuilder();
-    sb.append("{");
-    sb.append("\"longValue\":").append(longValue).append(",");
-    sb.append("\"intValue\":").append(intValue).append(",");
-    sb.append("\"booleanValue\":").append(booleanValue).append(",");
-    sb.append("\"stringValue\":\"").append(stringValue).append("\"");
-    sb.append("}");
-    return sb.toString();
+    return "{" +
+            "\"longValue\":" + longValue + "," +
```
Marcono1234, line 47: Not sure if the indentation here is correct, I think this should have 4 spaces relative from the start of `return` (instead of 8).

Marcono1234 (commented):
Thanks a lot for these changes!

Feel free to consider these review comments mostly as suggestions; Éamonn will probably provide the main review.

metrics/src/main/java/com/google/gson/metrics/BagOfPrimitives.java
eamonnmcmanus: Yes. Google Java Style (which admittedly is not always followed throughout the Gson code) suggests 4-space continuation lines, and also moving `+` from the end of each line to the start of the following line.

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
eamonnmcmanus: +1

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
eamonnmcmanus: I think this particular change isn't a clear improvement so I'd be inclined to keep the original `switch`.

eamonnmcmanus (changes requested):
Thanks, some of these changes are clear improvements! I agree with @Marcono1234 about the formatting concerns, though, so can you fix those?

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
```diff
@@ -46,16 +46,16 @@ public void write(JsonWriter out, Date date) throws IOException {
...
-      case NULL:
-        in.nextNull();
-        return null;
-      default:
-        String date = in.nextString();
-        // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
-        // This is because Android doesn't support XXX because it is JDK 1.6
-        return parse(date, new ParsePosition(0));
-      }
+        switch (in.peek()) {
```
eamonnmcmanus, line 49: Could you revert these lines? I'm just seeing indentation changes, and they aren't consistent with the two-space indentation used elsewhere.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1381,11 +1374,9 @@ private TypeAdapter<T> delegate() {
...
   @Override
   public String toString() {
-    return new StringBuilder("{serializeNulls:")
-        .append(serializeNulls)
-        .append(",factories:").append(factories)
-        .append(",instanceCreators:").append(constructorConstructor)
-        .append("}")
-        .toString();
+    return "{serializeNulls:" + serializeNulls
+      + ",factories:" + factories
```
eamonnmcmanus, line 1378: These continuation lines should be indented +4 rather than +2. You might find [google-java-format](https://github.com/google/google-java-format) useful, though you probably only want to run it on changed lines rather than the whole file.

gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java
```diff
@@ -22,14 +22,7 @@
...
 import java.io.Serializable;
-import java.util.AbstractMap;
-import java.util.AbstractSet;
-import java.util.Comparator;
-import java.util.ConcurrentModificationException;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.NoSuchElementException;
-import java.util.Set;
+import java.util.*;
```
eamonnmcmanus, line 25: Still need to put back individual imports here.

metrics/src/main/java/com/google/gson/metrics/BagOfPrimitives.java
```diff
@@ -43,14 +43,12 @@ public int getIntValue() {
...
-    StringBuilder sb = new StringBuilder();
-    sb.append("{");
-    sb.append("\"longValue\":").append(longValue).append(",");
-    sb.append("\"intValue\":").append(intValue).append(",");
-    sb.append("\"booleanValue\":").append(booleanValue).append(",");
-    sb.append("\"stringValue\":\"").append(stringValue).append("\"");
-    sb.append("}");
-    return sb.toString();
+    return "{"
+      + "\"longValue\":" + longValue + ","
```
eamonnmcmanus, line 47: Continuation +4 again.

eamonnmcmanus (changes requested):
There are quite a few good improvements here! Still a few nits, though.

MaicolAntali:
If you want I can run in the google-java-format in the project so everything could be formatted. Let me know

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
eamonnmcmanus: I'm still seeing changed indentation.

eamonnmcmanus:
> If you want I can run in the google-java-format in the project so everything could be formatted.

No, please don't do that. We should probably do it at some point, in a PR with no other changes, but we've hesitated to because it would mean any outstanding PRs would probably get a bunch of annoying merge work to do.

eamonnmcmanus (commented):
This is looking good! Just one thing.

MaicolAntali:
> No, please don't do that.

Ohh yes, make sense. Thank you for explaining

----

> I'm still seeing changed indentation.

I don't know why Github don't show the indentation changes. If you look at this [commit](https://github.com/google/gson/pull/2282/commits/fa2dd8c60a42338cbe73c9a2974f8e9bdbc47867), for me it's fine

eamonnmcmanus:
The indentation problem is still there, including in the [commit](https://github.com/google/gson/pull/2282/commits/fa2dd8c60a42338cbe73c9a2974f8e9bdbc47867) you referenced. It should look like this:
```java
..public Date read(JsonReader in) throws IOException {
....try {
......switch (in.peek()) {
........case NULL:
..........in.nextNull();
```
but obviously with spaces instead of dots.

It doesn't look as if there are any real changes to that file anyway, so maybe just restore it to what it was?

(Also make sure that you don't have whitespace changes hidden: Files changed > ⚙️ > Hide whitespace.)

eamonnmcmanus (approved):
Thanks! I hadn't realized that the indentation of the switch in `UtcDateTypeAdapter` did not follow Google style, since the case labels were at the same indentation as the `switch`. So thanks for fixing that.

MaicolAntali:
Most welcome! i'm happy to help 😃

PR #2283, opened 2022-12-12
Update UserGuide.md

lengyijun:
use `false` as a variable name in java8, is an error


eamonnmcmanus:
For what it's worth, I think this was a reasonable change.

PR #2285, opened 2022-12-14
Add troubleshooting guide

Marcono1234:
There are some issues which users encounter (somewhat) frequently so I thought a troubleshooting guide might be useful to list them and to describe how to solve these issues. Any feedback (e.g. regarding wording, formatting, ...) is appreciated.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Report a Gson bug.
+about: Report a Gson bug. Please have a look at the troubleshooting guide (Troubleshooting.md) first.
 title: ''
 labels: bug
 assignees: ''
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Report a Gson bug.
+about: Report a Gson bug. Please have a look at the troubleshooting guide (Troubleshooting.md) first.
```
Marcono1234, line 3: Have used plain text here instead of a proper link to the guide because apparently this 'about' text does not support Markdown.

eamonnmcmanus:
This is spectacularly helpful! Thank you so much for putting it together.

I didn't see anything that needed to be changed. We can always make minor tweaks later. For now I think this is good to go.

Issue #2284, opened 2022-12-13
Clean up Gson source code


eamonnmcmanus:
We've been reluctant to make large-scale changes because they are likely to cause pain for people with unmerged PRs. But at this point I think there are relatively few open PRs that we are likely to merge, thanks largely to the heroic work of @Marcono1234. So this would be as good a time as any to consider at least:

* Port tests from JUnit 3 to JUnit 4. (Google's internal systems don't support JUnit 5 so we can't use that.)
* Use [Truth](https://truth.dev/) for clearer assertions in tests.
* Apply various automated fixes suggested by Error Prone and the like.
* Reformat all source files with [google-java-format](https://github.com/google/google-java-format). Possibly also install a check that ensures that future PRs are also correctly formatted.

MaicolAntali:
For me, sound very good. If you need a help with the port of the tests or with the code reformat I'm available.

Marcono1234:
> Reformat all source files with [google-java-format](https://github.com/google/google-java-format)

We should probably create a [`.git-blame-ignore-revs` file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) then to avoid that commit from appearing in `git blame` (and on GitHub's "Blame" feature). That would require that the formatting changes would have to be in its own pull request and not be combined with any other clean up (or the commits of the pull request should not be squashed).

> Port tests from JUnit 3 to JUnit 4

Maybe some of the changes from #2048 could be useful for that since JUnit 4 and JUnit 5 API is pretty similar. Though on the other hand it might be easier to port the tests from scratch since that pull request likely has multiple conflicts by now.

eamonnmcmanus:
> We should probably create a [`.git-blame-ignore-revs` file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) then to avoid that commit from appearing in `git blame`

Excellent idea! I think we should probably do this reformatting last.

> Maybe then also configure Error Prone to run as part of the build?

That also seems like a great idea.

One other thing I forgot to mention: change the test assertions to use [Truth](https://github.com/google/truth).

eamonnmcmanus:
Some ideas on how to migrate JUnit 3 tests to JUnit 4 are [here](https://stackoverflow.com/questions/264680/best-way-to-automagically-migrate-tests-from-junit-3-to-junit-4).

MaicolAntali:
[...]

Issue #617, opened 2015-03-19
Serialization in order


GoogleCodeExporter:
```
What steps will reproduce the problem?
1.I am using Gson library for serialization, In gson after serialization fields 
are not in order(I mean order however it was in that model class).But it gives 
some random order.


What is the expected output? What do you see instead?
I want Serialization to be done in that same order(fields order inside that 
model class)

What version of the product are you using? On what operating system?


Please provide any additional information below.
```

Original issue reported on code.google.com by `harshith...@gmail.com` on 30 Dec 2014 at 4:58

Marcono1234:
The issue is that Gson cannot determine the order itself, see [`Class.getDeclaredFields()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getDeclaredFields()):
> [...] The elements in the returned array are not sorted and are not in any particular order.

Even when trying to inspect the bytecode of a class, there appears to be no requirement that the compiler has to emit fields in declaration order.

For reference, Jackson implemented this using (at least) the two annotations [`@JsonProperty#index`](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-annotations/2.12.2/com/fasterxml/jackson/annotation/JsonProperty.html#index--) and [`@JsonPropertyOrder`](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-annotations/2.12.2/com/fasterxml/jackson/annotation/JsonPropertyOrder.html).

MurtuzaSrashtaSoft:
shode I use both libraries at the same time in android kotlin.

Marcono1234:
> shode I use both libraries at the same time in android kotlin.

Using two libraries for the same purpose is probably not a good idea; it is more difficult to maintain and makes your application larger. Since you are using Kotlin, maybe consider switching to [Moshi](https://github.com/square/moshi), which seems to support preserving field order, see https://github.com/square/moshi/issues/753. Though I cannot really give you any further advice regarding Moshi. If you have questions it might be best to ask them on StackOverflow.

PR #2288, opened 2022-12-16
Format the code following Google Java Style Guide

MaicolAntali:
Reformatted all source files with [google-java-format](https://github.com/google/google-java-format) as mentioned in #2284

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,4 @@
+# .git-blame-ignore-revs
+
+#Format the code following Google Java Style Guide
+3bb4dcfd31382437cc206f24fbeedc239b56b4cc
--- a/examples/android-proguard-example/src/com/google/gson/examples/android/model/Cart.java
+++ b/examples/android-proguard-example/src/com/google/gson/examples/android/model/Cart.java
@@ -41,6 +41,34 @@ public class Cart {
     this.creditCard = creditCard;
   }
 
+  @SuppressWarnings("unchecked")
+  public static String getSimpleTypeName(Type type) {
+    if (type == null) {
+      return "null";
+    }
+    if (type instanceof Class) {
+      return ((Class)type).getSimpleName();
+    } else if (type instanceof ParameterizedType) {
+      ParameterizedType pType = (ParameterizedType) type;
+      StringBuilder sb = new StringBuilder(getSimpleTypeName(pType.getRawType()));
+      sb.append('<');
+      boolean first = true;
+      for (Type argumentType : pType.getActualTypeArguments()) {
+        if (first) {
+          first = false;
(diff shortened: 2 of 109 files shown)


MaicolAntali:
Some tests fails, but the issue it's related with the order of the object.

For exemple the test: `ExposeFieldsTest.testArrayWithOneNullExposeFieldObjectSerialization` fails due: `expected:<[{"[a":1,"d":2.0},{"d":2.0},{"a":2,"d":2.0]}]> but was:<[{"[d":2.0,"a":1},{"d":2.0},{"d":2.0,"a":2]}]>`

The issue here come from the object field order. The class before formatting following the google style giude:
```java

  private static class ClassWithExposedFields {
    @Expose private final Integer a;
    private final Integer b;
    @Expose(serialize = false) final long c;
    @Expose(deserialize = false) final double d;
    @Expose(serialize = false, deserialize = false) final char e;

...
```

and after:

```java 
private static class ClassWithExposedFields {
    @Expose(serialize = false) final long c;
    @Expose(deserialize = false) final double d;
    @Expose(serialize = false, deserialize = false) final char e;
    @Expose private final Integer a;
    private final Integer b;

...
```

In this case is it better to keep the class in the "old" way or fix the test?

eamonnmcmanus:
As [mentioned](https://github.com/google/gson/issues/2284#issuecomment-1350196671) (somewhat hidden) in the issue, I think we should do this reformatting last, after we've addressed the other things such as migrating tests to JUnit 4. But this is still a useful experiment because we certainly wouldn't expect test failures. Did google-java-format reorder those fields? I don't think the Google style guide requires package fields to come before private ones, so that would be a bug. But also, the order in which fields are returned by reflection is [explicitly not ordered](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getDeclaredFields()) so it may be that the failing tests are making invalid assumptions.

Marcono1234:
> Did google-java-format reorder those fields? I don't think the Google style guide requires package fields to come before private ones, so that would be a bug.

It looks like multiple methods and nested classes were reordered as well. The [style guide says](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents):
> However, there's no single correct recipe for how to do it; different classes may order their contents in different ways.

Personally I also think that no specific ordering should be enforced; that would be too strict and would be more than just formatting.

----

> But also, the order in which fields are returned by reflection is [explicitly not ordered](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getDeclaredFields()) so it may be that the failing tests are making invalid assumptions.

Probably related to #617. It looks like the majority of Gson's reflection based tests depend on a consistent order.

PR #2289, opened 2022-12-17
Replace custom user guide header anchors

### LINKED ISSUE #2286: User guide "TOC-" section anchors don't work anymore
# Description
It looks like the custom anchors such as `<a name="TOC-Overview">` in `UserGuide.md` don't work anymore. For example when clicking on any of the links in the "Overview" section nothing happens.

Would be good if someone else could check this as well in case there is an issue with my browsers (tested Firefox and Edge).

Maybe we should then simply edit the user guide (and any other documents referring to it) to use the automatically generated anchors instead.

Marcono1234:
Fixes #2286

Looks like using the HTML `id` attribute instead of `name` does not work either so I just removed the HTML anchors and used the implicit ones.


eamonnmcmanus:
Thanks!

PR #2294, opened 2022-12-21
Port tests from JUnit 3 to JUnit 4

MaicolAntali:
I have port tests from JUnit 3 to JUnit 4

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
+++ b/gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
@@ -17,6 +17,10 @@
 package com.google.gson.functional;
 
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -31,7 +35,8 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
-import junit.framework.TestCase;
+import org.junit.Before;
+import org.junit.Test;
 
 /**
  * Functional tests for Json primitive values: integers, and floating point numbers.
@@ -39,19 +44,20 @@ import junit.framework.TestCase;
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
-public class PrimitiveTest extends TestCase {
+public class PrimitiveTest {
   private Gson gson;
 
-  @Override
-  protected void setUp() throws Exception {
-    super.setUp();
+  @Before
+  public void setUp() throws Exception {
     gson = new Gson();
   }
 
+  @Test
   public void testPrimitiveIntegerAutoboxedSerialization() {
     assertEquals("1", gson.toJson(1));
   }
 
+  @Test
   public void testPrimitiveIntegerAutoboxedDeserialization() {
     int expected = 1;
     int actual = gson.fromJson("1", int.class);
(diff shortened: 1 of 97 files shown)


MaicolAntali:
I would suggest this next steps:

-  Cleanup the test (Some example:  methods with redundant `throws` clause, there are some redundant local variable,  ... )
-  Simplify the asserts

gson/src/test/java/com/google/gson/metrics/PerformanceTest.java
```diff
@@ -37,23 +40,24 @@
...
+  public void setUp() throws Exception {
     gson = new Gson();
   }
   
+  @Test
   public void testDummy() {
     // This is here to prevent Junit for complaining when we disable all tests.
   }
 
+  @Ignore
```
eamonnmcmanus, line 60: Maybe `@Test @Ignore`? `@Ignore` on its own doesn't mean much.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -672,6 +721,7 @@ public void testNegativeZero() throws Exception {
    * This test fails because there's no double for 9223372036854775808, and our
    * long parsing uses Double.parseDouble() for fractional values.
    */
+  @Ignore
```
eamonnmcmanus, line 724: Same here.

eamonnmcmanus (commented):
This is great, thanks! And thanks too for your patience.

MaicolAntali:
I have added `@Test` above every `@Ignore`

I don't simplify the asserts cause, as you said, we switch to [Truth](https://truth.dev/). 


> This is great, thanks! And thanks too for your patience.

No problems, I’m here to help and learn 😄

gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
```diff
@@ -280,86 +307,97 @@ public void testNumberDeserialization() {
...
               + "0.0011904761904761906, 0.0]";
       double[] values = gson.fromJson(json, double[].class);
-      assertEquals(6, values.length);
-      assertEquals(0.0, values[0]);
-      assertEquals(0.004761904761904762, values[1]);
-      assertEquals(3.4013606962703525E-4, values[2]);
-      assertEquals(7.936508173034305E-4, values[3]);
-      assertEquals(0.0011904761904761906, values[4]);
-      assertEquals(0.0, values[5]);
+      assertEquals(6, values.length, 0);
```
Marcono1234, line 366: This looks incorrect / unnecessary since the values here are `int` so there is no need to use `assertEquals(double, double, double)`:
```suggestion
      assertEquals(6, values.length);
```

In case you do want to change this, maybe the following should also be changed to use `assertEquals(float, float, float)` instead of the `double` variant by appending`F` to `0.00001`:
https://github.com/google/gson/blob/1a2170b99c9d293e825bf8f511b191196326ea03/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java#L100
(though this is not an issue you introduced)

Marcono1234 (commented):
Thanks a lot for these changes! There might be one small oversight (see review comment), but besides that this pull request has really good quality; based on the Maven test output it appears you did not overlook a single test. It still runs the same amount of tests (or actually a few more since those are now properly reported as 'skipped').

Out of curiousity, did you adjust these test classes by hand? I had a look at OpenRewrite, but it appears they only have a recipe for [migrating from JUnit 4 to JUnit 5](https://docs.openrewrite.org/reference/recipes/java/testing/junit5/junit4to5migration).

gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
MaicolAntali: Ohh yes, good point. 😄 
If we don’t change all the assertions to Truth, the next PR will be "simplify assertions". I’ll fix it in that PR.

MaicolAntali:
> Out of curiousity, did you adjust these test classes by hand?

No, I used some regex to adjust classes, methods and imports. Then I ran the tests and corrected the remaining errors. 

Here you can find some regex that i have used: [link](https://stackoverflow.com/a/4535170)(If you want to use it, the regexs need some small adjustments. It's not just a copy-paste)

PR #2303, opened 2023-01-26
add copyright to pom.xml

tobi5775:
For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar file. Therefore it would be great to have the copyright and license information included in the Jar file as COPYRIGHT and LICENSE files.


Marcono1234:
> to extract the copyright and licence information from our dependencies

Are you using a build tool such as Maven or Gradle? Then it is usually easier to use a plugin to extract the license information from the dependency metadata than having to rely on special files inside the JAR files. I assume Gson is not the only project which does not include these files in the JAR.

I am not a lawyer, but the COPYRIGHT file might be a bit pointless (at least for Gson) because the important information is the type of license being used; the COPYRIGHT file just duplicates some of that.

Though the Gson project members might see this differently.

eamonnmcmanus:
I think we would be fine with adding `LICENSE`. Several other Google projects already do this. I searched about 2500 jars that I happen to have in `$HOME/.m2/repository` and 1288 of them had a file with `LICENSE` in the name. On the other hand, exactly _one_ had a file with `COPYRIGHT` in the name, so I would suggest you reformulate what you are trying to do so it doesn't require that file. That avoids the question of what year should go into the copyright notice, among other things.

I'm fine with this PR if it contains just `LICENSE`.

tobi5775:
Besides the license, our lawyers also want to see the copyright. To extract the license and copyright we use the [ScanCode](https://github.com/nexB/scancode-toolkit) tool. 

I have changed the PR so that the copyright information can now be found in the pom.xml. The ScanCode tool gets the license information from the pom.xml and can also extract the copyright from there. This would also put the copyright in a central location.

eamonnmcmanus:
Thanks, putting the copyright notices in `pom.xml` is much more in line with standard practice. Many other Google projects already have such notices. I also realize that there are a few source files that are missing copyright notices, so I'm proposing to fix those and the `pom.xml` files in #2307.

PR #2305, opened 2023-01-28
Add Maven Wrapper

sconvent:
This adds a Maven Wrapper.
Right now, the used Maven version to build this project is unclear and to make reproducible builds, a wrapper should be added.

More info: [Maven Wrapper Documentation](https://maven.apache.org/wrapper/)

Right now, the `check-api-compatibility` Github Action cannot be changed to use the wrapper as it operates on the base commit which does not yet have a Maven Wrapper.


Marcono1234:
Disclaimer: I am not a direct project member; they might think differently about this pull request.

----

Thanks, this looks interesting! One issue might be that there is no official "wrapper validation GitHub action" yet (see [MNG-6887](https://issues.apache.org/jira/browse/MNG-6887)), so in the future the Gson maintainers would have to be careful when someone creates a pull request claiming to "upgrade the wrapper JAR".

It might be good to cache the Maven distribution downloaded by the wrapper because the setup-java action does not do that yet (https://github.com/actions/setup-java/issues/448). I assume the build logs for this pull request did not indicate that the wrapper is downloaded because info log messages are [discarded by the wrapper by default](https://github.com/apache/maven-wrapper/blob/56fc8968933e6e16efcff01a5126e136b85185f3/maven-wrapper/src/main/java/org/apache/maven/wrapper/Logger.java#L27-L31).

In case this is merged, it might also be useful to use the wrapper for the [OSS-Fuzz configuration](https://github.com/google/oss-fuzz/tree/master/projects/gson) (probably requires changes to `Dockerfile` and `build.sh`).

Marcono1234:
There are also variants for the wrapper setup which don't involve any binary files, see https://maven.apache.org/wrapper/index.html#usage-without-binary-jar
Especially `only-script` sounds tempting I think because it also does not involve downloading the wrapper JAR. But apparently that variant can fail in some cases, see https://github.com/google/guava/pull/6783 (to be investigated?).

The question would also be whether the checksums of the downloaded files should be checked, see https://maven.apache.org/wrapper/index.html#checksum-verification-of-downloaded-binaries and also https://github.com/google/guava/issues/6805.

cpovirk:
> see https://github.com/google/guava/pull/6783 (to be investigated?), and

I was planning to be lazy and not do anything about it... :) Ideally, I would report it upstream (or I guess even propose a fix).

I got the impression that the only problem was that we had Windows CI that was using `mvnw` (rather than `mvnw.cmd`). So if Gson has no Windows CI or it at least would configure that CI to use the right script, `only-script` would likely be fine.

Marcono1234:
> I got the impression that the only problem was that we had Windows CI that was using `mvnw` (rather than `mvnw.cmd`).

It looks like I reproduced it now locally in a test Maven project by running the Linux `mvnw` script on Windows using Git Bash (and probably due to having `git core.autocrlf true` (or similar) / the properties file using CRLF). Though that explicitly required me to use Git Bash.

PR #2308, opened 2023-02-01
Adds Error Prone to the `maven-compiler-plugin`

MaicolAntali:
First at all, I have set up the Error Prone in the `pom.xml`.  I had to remove the `<failOnWarning>true</failOnWarning>` otherwise the compiler would stop at every warning that Error Prone gives us.

After, I have add some ` @SuppressWarnings(...)` to avoid errors from Error Prone (that stop the compiler)

### DIFF
--- a/extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
+++ b/extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
@@ -57,8 +57,8 @@ public class PostConstructAdapterFactoryTest {
         assertEquals(sandwiches, sandwichesFromJson);
     }
 
-    @SuppressWarnings("overrides") // for missing hashCode() override
-    static class Sandwich {
+  @SuppressWarnings({"overrides", "EqualsHashCode"}) // for missing hashCode() override
+  static class Sandwich {
         public String bread;
         public String cheese;
 
@@ -92,7 +92,7 @@ public class PostConstructAdapterFactoryTest {
         }
     }
 
-    @SuppressWarnings("overrides") // for missing hashCode() override
+    @SuppressWarnings({"overrides", "EqualsHashCode"}) // for missing hashCode() override
     static class MultipleSandwiches {
         public List<Sandwich> sandwiches;
 
--- a/gson/src/test/java/com/google/gson/JsonArrayTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayTest.java
@@ -35,6 +35,7 @@ public final class JsonArrayTest {
   }
 
   @Test
+  @SuppressWarnings("TruthSelfEquals")
   public void testEqualsNonEmptyArray() {
     JsonArray a = new JsonArray();
     JsonArray b = new JsonArray();
--- a/gson/src/test/java/com/google/gson/JsonObjectTest.java
+++ b/gson/src/test/java/com/google/gson/JsonObjectTest.java
@@ -163,6 +163,7 @@ public class JsonObjectTest {
   }
 
   @Test
+  @SuppressWarnings("TruthSelfEquals")
   public void testEqualsNonEmptyObject() {
     JsonObject a = new JsonObject();
     JsonObject b = new JsonObject();
--- a/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
+++ b/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
@@ -101,10 +101,9 @@ public class DefaultTypeAdaptersTest {
   }
 
   @Test
-  @SuppressWarnings("GetClassOnClass")
   public void testClassDeserialization() {
     try {
-      gson.fromJson("String.class", String.class.getClass());
+      gson.fromJson("String.class", Class.class);
       fail();
     } catch (UnsupportedOperationException expected) {
     }
--- a/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
+++ b/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
@@ -101,6 +101,7 @@ public class DefaultTypeAdaptersTest {
   }
 
   @Test
+  @SuppressWarnings("GetClassOnClass")
   public void testClassDeserialization() {
     try {
       gson.fromJson("String.class", String.class.getClass());
--- a/pom.xml
+++ b/pom.xml
@@ -311,7 +311,6 @@
         </plugins>
       </build>
     </profile>
-
     <!-- Profile defining additional plugins to be executed for release -->
     <profile>
       <id>release</id>
--- a/pom.xml
+++ b/pom.xml
@@ -291,6 +291,27 @@
   </build>
 
   <profiles>
+    <!-- Disable Error Prone in Java 15 -->
+    <profile>
+      <id>jdk15</id>
+      <activation>
+        <jdk>15</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+             <compilerArgs combine.self="override">
+               <compilerArg>-Xlint:all,-options</compilerArg>
+             </compilerArgs>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
     <!-- Profile defining additional plugins to be executed for release -->
     <profile>
       <id>release</id>
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
             <showDeprecation>true</showDeprecation>
             <fork>true</fork>
             <compilerArgs>
-              <!-- Args related to Error Prone, see: https://github.com/google/error-prone/blob/f8e33bc460be82ab22256a7ef8b979d7a2cacaba/docs/installation.md -->
+              <!-- Args related to Error Prone, see: https://errorprone.info/docs/installation#maven -->
               <arg>-XDcompilePolicy=simple</arg>
               <arg>-Xplugin:ErrorProne</arg>
               <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
--- a/pom.xml
+++ b/pom.xml
@@ -290,6 +290,27 @@
   </build>
 
   <profiles>
+    <!-- using github.com/google/error-prone-javac is required when running on JDK 8 -->
+    <profile>
+      <id>jdk8</id>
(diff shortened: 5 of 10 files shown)


eamonnmcmanus:
Thanks, this is great!

> Any idea why CIFuzz is failing?

I think the key thing is this:

> An exception has occurred in the compiler (15.0.2)

It's running with JDK 15, and I don't think Error Prone supports that. Interestingly I found an old [PR](https://github.com/google/oss-fuzz/pull/8980) that I had lost track of, where I was trying to update Guava fuzzing to use JDK 17 instead of 15. Maybe we can do something like that here too. (It would be more convincing if I had actually made that PR work.)

As a short-term alternative, perhaps you could fiddle with Maven profiles so that Error Prone is not enabled if we're on JDK 15.

MaicolAntali:
I have forked [oss-fuzz ](https://github.com/google/oss-fuzz). I will look into the `Dockerfile` of the Gson project to see if i can run it under `jdk-17`

MaicolAntali:
Hello @eamonnmcmanus, the PR google/oss-fuzz#9554 have passed all tests. Can you take a look and tell me if it’s everything good? If yes, I will mark the PR as `ready to review` so that it will no longer a `draft`

eamonnmcmanus:
That's fantastic! Thanks for taking the time to do it.

MaicolAntali:
Most welcome, now we have to wait until the PR will be merged

pom.xml
```diff
@@ -94,12 +94,31 @@
           <configuration>
             <showWarnings>true</showWarnings>
             <showDeprecation>true</showDeprecation>
-            <failOnWarning>true</failOnWarning>
+            <fork>true</fork>
             <compilerArgs>
+              <arg>-XDcompilePolicy=simple</arg>
+              <arg>-Xplugin:ErrorProne</arg>
```
Marcono1234, line 100: Should probably suppress warnings for classes generated by `proto` module:
```suggestion
              <arg>-Xplugin:ErrorProne -XepExcludedPaths:.*/generated-test-sources/protobuf/.*</arg>
```

Error Prone's `-XepDisableWarningsInGeneratedCode` does not seem to work because it relies on `@Generated` and the protobuf plugin does not add that yet, see https://github.com/xolstice/protobuf-maven-plugin/issues/29.

Marcono1234 (commented):
Is it risky to use Error Prone with the regular compilation run, and should it be executed in a separate compilation run instead? I am a bit concerned that it hooks into compilation of classes which then end up in the final JAR. Though I am not very familiar with Error Prone, so maybe this is not an issue.

> otherwise the compiler would stop at every warning that Error Prone gives us

The question might be though whether the other warnings should be fixed, suppressed or whether the severity should be adjusted for them (possibly in a follow-up PR?). Otherwise it might be difficult to notice newly introduced warnings when there are already a bunch of previously existing warnings.

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -35,6 +35,7 @@ public void testEqualsOnEmptyArray() {
   }
 
   @Test
+  @SuppressWarnings("TruthSelfEquals")
```
Marcono1234, line 38: This should not be suppressed, see [its description](https://errorprone.info/bugpattern/TruthSelfEquals).

`assertThat(a).isEqualTo(a);` below should be changed to `assertThat(a.equals(a)).isTrue()` (or `MoreAsserts.assertEqualsAndHashCode(a, a)`?) since this is intended to explicitly test the `equals` implementation.

gson/src/test/java/com/google/gson/JsonObjectTest.java
```diff
@@ -163,6 +163,7 @@ public void testEqualsOnEmptyObject() {
   }
 
   @Test
+  @SuppressWarnings("TruthSelfEquals")
```
Marcono1234, line 166: This should not be suppressed, see [its description](https://errorprone.info/bugpattern/TruthSelfEquals).

MaicolAntali:
> The question might be though whether the other warnings should be fixed, suppressed or whether the severity should be adjusted for them (possibly in a follow-up PR?). Otherwise it might be difficult to notice newly introduced warnings when there are already a bunch of previously existing warnings.

Every warn could be suppressed or fixed. If you look the build log, Error prone returns to you something like that for each warn:

```
Warning:  /home/runner/work/gson/gson/gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java:[568,12] [ReferenceEquality] Comparison using reference equality instead of value equality
    (see https://errorprone.info/bugpattern/ReferenceEquality)
  Did you mean 'if (Objects.equals(e, header)) {' or 'if (e.equals(header)) {'?
```

You have 2 possibility: Fix it  or Suppress it. If you follow the link `https://errorprone.info/bugpattern/ReferenceEquality` ([Link](https://errorprone.info/bugpattern/ReferenceEquality)) there is an explanation of the problem, a way to fix it and a way to suppress it.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -101,6 +101,7 @@ public void testClassSerialization() {
   }
 
   @Test
+  @SuppressWarnings("GetClassOnClass")
   public void testClassDeserialization() {
     try {
       gson.fromJson("String.class", String.class.getClass());
```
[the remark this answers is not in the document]
MaicolAntali: I thought the same thing, but when I removed the `.getClass()` the test failed.

gson/src/test/java/com/google/gson/JsonArrayTest.java
MaicolAntali: In #2299, Éamonn said we could use `com.google.guava:guava-testlib` to simplify assert like this. 

[Link to comment](https://github.com/google/gson/pull/2299#discussion_r1070699132)

I never use `guava-testlib` so I'm not sure how to do that. I have looked around for some documentations/examples, but I don't find nothing.

Marcono1234: Hmm, right. Personally I think for this PR it would be best to omit these suppressions and change the code to `a.equals(a)` to fix this issue. If desired we could then add `guava-testlib` in a separate PR.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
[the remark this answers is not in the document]
Marcono1234: Just to make sure there is no misunderstanding, did you only remove the `.getClass()` (i.e. `fromJson(..., String.class)`) or replace the argument with `Class.class` (i.e. `fromJson(..., Class.class)`).

If I replace it with `Class.class` the test still seems to pass. Could you please share the assertion error message / exception stack trace (in case there is one)?

eamonnmcmanus:
I see this in the CIFuzz logs:

> Step 2/8 : RUN apt-get update && apt-get install -y make autoconf automake libtool wget

which I think means that your oss-fuzz [change](https://github.com/google/oss-fuzz/commit/6e214481347cca5a0fd4344f3236cd9f3d22130c) isn't yet active. Otherwise we would [see](https://github.com/google/oss-fuzz/commit/6e214481347cca5a0fd4344f3236cd9f3d22130c#diff-90c16b96000b6e3a2367c0a5b851bd3feb9c16a163256491e561cf1e141961b6L18) `openjdk-17-jdk` at the end of that line.

We don't really care about running Error Prone in the CIFuzz build. Would it be very hassly to omit it when running on JDK 15? As a workaround until the CIFuzz thing is working properly.

extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
```diff
@@ -57,8 +57,8 @@ public void testList() {
         assertEquals(sandwiches, sandwichesFromJson);
     }
 
-    @SuppressWarnings("overrides") // for missing hashCode() override
-    static class Sandwich {
+  @SuppressWarnings({"overrides", "EqualsHashCode"}) // for missing hashCode() override
+  static class Sandwich {
```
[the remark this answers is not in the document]
eamonnmcmanus: I'm fine either way. If @MaicolAntali prefers to add suppressions in this PR and remove them later by fixing the code in one or more later PRs, that seems like a good approach.

gson/src/test/java/com/google/gson/JsonArrayTest.java
eamonnmcmanus: Yes, sorry, I should have been more explicit. That is exactly what you can write, and alternatively you can have just one test method with all the different kinds of objects in different equality groups. Something like this:

```java
new EqualsTester()
    .addEqualityGroup(array1, array2) // with equal contents
    .addEqualityGroup(array3, array4) // equal to each other but not array1 or array2
    .addEqualityGroup(object1, object2)
    ...
   .testEquals();
```

`EqualsTester` tests that each object in an equality group is equal to itself and to each other object in the group, and not equal to any object in any other group. It also checks that equal objects have equal hashcodes. So it removes a lot of drudgery.

I think that would be a fine project for a separate PR.

pom.xml
```diff
@@ -94,12 +94,32 @@
           <configuration>
             <showWarnings>true</showWarnings>
             <showDeprecation>true</showDeprecation>
-            <failOnWarning>true</failOnWarning>
+            <fork>true</fork>
             <compilerArgs>
+              <!-- Args related to Error Prone, see: https://github.com/google/error-prone/blob/f8e33bc460be82ab22256a7ef8b979d7a2cacaba/docs/installation.md -->
```
eamonnmcmanus, line 99: I think you could probably have a shorter link than this? The one @Marcono1234 gave seems fine to me, https://errorprone.info/docs/installation#maven.

eamonnmcmanus:
> Is it risky to use Error Prone with the regular compilation run, and should it be executed in a separate compilation run instead? I am a bit concerned that it hooks into compilation of classes which then end up in the final JAR. Though I am not very familiar with Error Prone, so maybe this is not an issue.

I think it's fine. Error Prone doesn't affect the generated code, except that there won't be any if it detects an error. We use it in all Java compilations at Google.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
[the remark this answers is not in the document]
MaicolAntali: I had removed just the `.getClass()`. Now, I have put the argument to `Class.class` and works fine.

eamonnmcmanus:
This is looking good. Thanks for the CIFuzz workaround! Let me know if this is ready to be merged.

MaicolAntali:
✅  Ready to merge

PR #2316, opened 2023-02-14
Fix error prone warnings

MaicolAntali:
In this PR I have fixed some ErrorProne warns. 

I have also add the flag `-XepExcludedPaths:` to the ErrorProne plugin to exclude tests and proto path

### DIFF
--- a/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
+++ b/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
@@ -79,7 +79,7 @@ public final class GraphAdapterBuilder {
 
   static class Factory implements TypeAdapterFactory, InstanceCreator<Object> {
     private final Map<Type, InstanceCreator<?>> instanceCreators;
-    private final ThreadLocal<Graph> graphThreadLocal = new ThreadLocal<>();
+    private static final ThreadLocal<Graph> graphThreadLocal = new ThreadLocal<>();
 
     Factory(Map<Type, InstanceCreator<?>> instanceCreators) {
       this.instanceCreators = instanceCreators;
--- a/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
+++ b/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
@@ -79,7 +79,8 @@ public final class GraphAdapterBuilder {
 
   static class Factory implements TypeAdapterFactory, InstanceCreator<Object> {
     private final Map<Type, InstanceCreator<?>> instanceCreators;
-    private static final ThreadLocal<Graph> graphThreadLocal = new ThreadLocal<>();
+    @SuppressWarnings("ThreadLocalUsage")
+    private final ThreadLocal<Graph> graphThreadLocal = new ThreadLocal<>();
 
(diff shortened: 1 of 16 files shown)


gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -637,7 +637,7 @@ public JsonWriter value(Number value) throws IOException {
     out.close();
 
     int size = stackSize;
-    if (size > 1 || size == 1 && stack[size - 1] != NONEMPTY_DOCUMENT) {
+    if ((size > 1 || size == 1) && stack[size - 1] != NONEMPTY_DOCUMENT) {
```
eamonnmcmanus, line 640: This doesn't look right. The existing code is equivalent to
```
if (size > 1 || (size == 1 && stack[size - 1] != NONEMPTY_DOCUMENT)) {
```
Parenthesizing it like that would be helpful, but this version changes the meaning.

pom.xml
```diff
@@ -98,7 +98,7 @@
             <compilerArgs>
               <!-- Args related to Error Prone, see: https://errorprone.info/docs/installation#maven -->
               <arg>-XDcompilePolicy=simple</arg>
-              <arg>-Xplugin:ErrorProne</arg>
+              <arg>-Xplugin:ErrorProne -XepExcludedPaths:.*gson/src/test.*|.*extras/src/test.*|.*proto.*</arg>
```
eamonnmcmanus, line 101: Eventually we might want to have the tests pass Error Prone too, but I agree that is a much lower priority.

eamonnmcmanus (changes requested):
Thanks for doing this!

pom.xml
Marcono1234, line 101: Is it intended that you disable warnings for the complete `proto` module? Maybe it would be better to only disable warnings for the generated source code, as suggested in https://github.com/google/gson/pull/2308#discussion_r1096470904.

Marcono1234 (commented):
Sorry to bring this up again, but would you mind fixing (respectively removing) the incorrect `@SuppressWarnings("TruthSelfEquals")` suppressions as mentioned in https://github.com/google/gson/pull/2308/files#r1096471723 (simply change it to `a.equals(a)` for now)? The warnings are true-positives; regardless of whether `guava-testlib` will be added in the future, the current test code on `master` is incorrect and the `@SuppressWarnings` hide that, suggesting the code is correct.

My concern is that we forget about this, or for whatever reason `guava-testlib` is not integrated, and in that case the incorrect code remains unnoticed.

extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
```diff
@@ -31,10 +31,10 @@
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.util.ArrayDeque;
```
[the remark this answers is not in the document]
MaicolAntali: Exactly, I think we can use it safely

eamonnmcmanus (approved):
I'm merging this snapshot under the assumption that if there are further changes in the pipeline they can just be applied in a followup PR.

PR #2319, opened 2023-02-17
Fix error prone warns

MaicolAntali:
I have fixed some other warns.

### DIFF
--- a/gson/src/main/java/com/google/gson/ExclusionStrategy.java
+++ b/gson/src/main/java/com/google/gson/ExclusionStrategy.java
@@ -93,14 +93,16 @@ package com.google.gson;
 public interface ExclusionStrategy {
 
   /**
-   * Method use to decide if a field should be serialized or deserialized
+   * Decides if a field should be skipped during serialization or deserialization.
+   *
    * @param f the field object that is under test
    * @return true if the field should be ignored; otherwise false
    */
   public boolean shouldSkipField(FieldAttributes f);
 
   /**
-   * Method use to decide if a class should be serialized or deserialized
+   * Decides if a class should be serialized or deserialized
+   *
    * @param clazz the class object that is under test
    * @return true if the class should be ignored; otherwise false
    */
--- a/gson/src/main/java/com/google/gson/ExclusionStrategy.java
+++ b/gson/src/main/java/com/google/gson/ExclusionStrategy.java
@@ -93,12 +93,14 @@ package com.google.gson;
 public interface ExclusionStrategy {
 
   /**
+   * Method use to decide if a field should be serialized or deserialized
    * @param f the field object that is under test
    * @return true if the field should be ignored; otherwise false
    */
   public boolean shouldSkipField(FieldAttributes f);
 
   /**
+   * Method use to decide it a class should be serialized or deserialized
    * @param clazz the class object that is under test
    * @return true if the class should be ignored; otherwise false
    */
--- a/gson/src/main/java/com/google/gson/ExclusionStrategy.java
+++ b/gson/src/main/java/com/google/gson/ExclusionStrategy.java
@@ -100,7 +100,7 @@ public interface ExclusionStrategy {
   public boolean shouldSkipField(FieldAttributes f);
 
   /**
-   * Method use to decide it a class should be serialized or deserialized
+   * Method use to decide if a class should be serialized or deserialized
    * @param clazz the class object that is under test
    * @return true if the class should be ignored; otherwise false
    */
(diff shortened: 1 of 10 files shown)


gson/src/main/java/com/google/gson/ExclusionStrategy.java
```diff
@@ -93,12 +93,14 @@
 public interface ExclusionStrategy {
 
   /**
+   * Method use to decide if a field should be serialized or deserialized
```
eamonnmcmanus, line 96: I would write:

```java
 * Decides if a field should be skipped during serialization or deserialization.
 *
```

The main changes:
* blank line before the `@` tags;
* no need to say that this is a method, just what it does;
* explicitly talk about "skipping" since it is sort of the opposite of what you might expect from a method that is "deciding".

gson/src/main/java/com/google/gson/annotations/SerializedName.java
```diff
@@ -83,10 +83,12 @@
 public @interface SerializedName {
 
   /**
+   * A String that indicates the desired name of the field when it is serialized or deserialized.
```
eamonnmcmanus, line 86: I'd go for brevity here too:

`The desired name of the field...`.

No need to say that it is a `String` since that is in the declared type.

gson/src/main/java/com/google/gson/annotations/SerializedName.java
```diff
@@ -83,10 +83,12 @@
...
    * @return the desired name of the field when it is serialized or deserialized
    */
   String value();
   /**
+   * An array of strings that indicates the alternative names of the field when it is deserialized
```
eamonnmcmanus, line 91: Likewise:

`The alternative names...`

gson/src/main/java/com/google/gson/internal/ReflectionAccessFilterHelper.java
```diff
@@ -98,6 +98,7 @@ private static abstract class AccessChecker {
             }
           };
         } catch (NoSuchMethodException ignored) {
+          // Ignored exception
```
eamonnmcmanus, line 101: I usually use something like `// OK: <reason why it is OK>`. Here, maybe:

`// OK: will assume everything is accessible`

Though I'm not sure it really _is_ OK. I think it might be better to wrap the exception in an `IllegalStateException` or something. But if we do that it should be in a separate PR.

gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java
```diff
@@ -64,6 +64,7 @@ public <T> T newInstance(Class<T> c) throws Exception {
         }
       };
     } catch (Exception ignored) {
+      // Ignored exception
```
eamonnmcmanus, line 67: For these ones, I would write:

`// OK: try the next way`

gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java
```diff
@@ -81,7 +81,9 @@ private Date deserializeToDate(JsonReader in) throws IOException {
       for (DateFormat dateFormat : dateFormats) {
         try {
           return dateFormat.parse(s);
-        } catch (ParseException ignored) {}
+        } catch (ParseException ignored) {
+          // Ignored exception
```
eamonnmcmanus, line 85: `// OK: try the next format`

gson/src/main/java/com/google/gson/internal/bind/DefaultDateTypeAdapter.java
```diff
@@ -154,7 +154,9 @@ private Date deserializeToDate(JsonReader in) throws IOException {
       for (DateFormat dateFormat : dateFormats) {
         try {
           return dateFormat.parse(s);
-        } catch (ParseException ignored) {}
+        } catch (ParseException ignored) {
+          // Ignored exception
```
eamonnmcmanus, line 158: `// OK: try the next format`

eamonnmcmanus (changes requested):
Thanks for doing this! Just some wording suggestions.

MaicolAntali:
Thanks!

gson/src/main/java/com/google/gson/internal/JavaVersion.java
```diff
@@ -75,13 +75,15 @@ private static int extractBeginningInt(String javaVersion) {
   }
 
   /**
+   * Gets the major Java version
```
eamonnmcmanus, line 78: Blank line before `@` tag. (I'm surprised Error Prone doesn't mention this.)

eamonnmcmanus (approved):
Thanks again!

haaami01 (commented):
Thank You!!
Fixes the `EmptyCatch` warn given by ErrorProne

PR #2322, opened 2023-02-24
3 jsonreader should not accept the escape sequence representing a in strict mode

carlp02:


carlp02:
Accidently opened on non-fork repo

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -50,6 +38,8 @@
 import java.util.Map;
 import java.util.Objects;
 
+import static com.google.gson.Gson.*;
```
MaicolAntali, line 41: Wildcard imports [should be avoided](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)

PR #2323, opened 2023-02-24
Strict mode for JSON parsing

### LINKED ISSUE #2295: Strict mode for JSON parsing
As noted in the documentation for [`JsonReader.setLenient`](https://www.javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/stream/JsonReader.html#setLenient(boolean)), there are a number of areas where Gson accepts JSON inputs that are not strictly conformant, even when lenient mode is off:

> * JsonReader allows the literals `true`, `false` and `null` to have any capitalization, for example `fAlSe`
> * JsonReader supports the escape sequence `\'`, representing a `'`
> * JsonReader supports the escape sequence `\LF` (with `LF` being the Unicode character U+000A), resulting in a LF within the read JSON string
> * JsonReader allows unescaped control characters (U+0000 through U+001F)

To which we could add that it allows a trailing comma in JSON arrays, for example `["foo", "bar",]` as noted in #494.

We could imagine that `GsonBuilder` would acquire a method `setStrict()`, mutually exclusive with `setLenient()` and that `JsonReader` would likewise acquire `setStrict(boolean)` and `isStrict()`. We can't make strict mode the default, for fear of breaking existing users, but we could recommend it.

marten-voorberg:
Note that this PR is still a work in progress.

Aims to resolve #2295.

In a non-lenient JSON parser, certain non-spec compliant values are accepted. Thie code in this PR allows you to set a JSONReader to strict mode in which non-spec compliant values are not allowed. 

The following non-spec compliant values are accepted by the default JSONReader but not a JSONReader in strict mode:
- Capitalization in keywords. In normal mode, any combination of lower and uppercase letters in the keywords `false`, `true`, and `null` is allowed (e.g. `nUlL` or `FALSE`). In strict mode, only the fully lower-case values are allowed.
- JsonReader allows unescaped control characters (U+0000 through U+001F) in normal mode but not in strict mode.
- JsonReader supports the escape sequence `\'`, representing a `'` in normal mode. In strict mode, this raises an exception.
- JsonReader supports the escape sequence \LF (with LF being the Unicode character U+000A), resulting in a LF within the read JSON string. In strict mode, this raises an exception.


marten-voorberg:
The API for `setStrict` is not final and will definitely be changed to one of the options discussed in #2295.

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -50,6 +38,8 @@
 import java.util.Map;
 import java.util.Objects;
 
+import static com.google.gson.Gson.*;
```
MaicolAntali, line 41: [Wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports) should be not used.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -38,6 +38,103 @@
...
+  @Test
+  public void testEscapedNewlineNotAllowedInStrictMode() throws IOException {
+    String json = "\"\\\n\"";
+    JsonReader reader = new JsonReader(reader(json));
+    reader.setStrict(true);
+    try {
+      reader.nextString();
+      fail();
+    } catch (IOException expected) {
+
```
MaicolAntali, line 50: If you want to leave an empty catch block you shoud put a comment inside. Sometime like:

```
// OK: reason why the block is empty
```

Otherwise you can check if the exception has the correct message:
```java
assertThat(e).hasMessageThat().isEqualTo("...");
```

---

Here some exemple:

https://github.com/google/gson/blob/742fd50ff74350462d8dcf80941efce219e187dd/gson/src/test/java/com/google/gson/JsonPrimitiveTest.java#L113-L115

https://github.com/google/gson/blob/742fd50ff74350462d8dcf80941efce219e187dd/gson/src/main/java/com/google/gson/internal/ReflectionAccessFilterHelper.java#L100-L102

MaicolAntali (commented):
Thanks! 😄

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -609,15 +620,17 @@ private int peekKeyword() throws IOException {
     String keyword;
     String keywordUpper;
     int peeking;
-    if (c == 't' || c == 'T') {
+
+    // Uppercase letters are not recognized if strict mode is used.
+    if (c == 't' || (!strict && c == 'T')) {
```
eamonnmcmanus, line 625: Rather than checking `strict` in all these ifs, I think it would be simpler to start the `for` loop below at 0 instead of 1. Also, instead of checking `strict` in the loop, before the loop you could set `keywordUpper` to be the same as `keyword` if `strict` is true. (Maybe renaming it to something like `alternativeKeyword`.)

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1007,7 +1021,9 @@ private String nextQuotedValue(char quote) throws IOException {
       while (p < l) {
         int c = buffer[p++];
 
-        if (c == quote) {
+        if (strict && c <= '\u001F') {
```
eamonnmcmanus, line 1024: Slight preference for `c < 0x20` here.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1636,11 +1652,15 @@ private char readEscapeCharacter() throws IOException {
       return '\f';
 
     case '\n':
+      if (strict) {
+        throw syntaxError("Cannot espace a newline character in strict mode!");
```
eamonnmcmanus, line 1656: Typo: escape

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1636,11 +1652,15 @@ private char readEscapeCharacter() throws IOException {
...
+      }
       lineNumber++;
       lineStart = pos;
       // fall-through
 
     case '\'':
+      if (this.strict) throw syntaxError("Invalid escaped character \"'\" in strict mode");
```
eamonnmcmanus, line 1663: Google Java style requires braces here.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
eamonnmcmanus: I think we can use `assertThrows` here, and yes, it would be good to check the exception message.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -38,6 +38,103 @@
...
+  @Test
+  public void testEscapedNewlineAllowedInDefaultMode() throws IOException {
+    String json = "\"\\\n\"";
+    JsonReader reader = new JsonReader(reader(json));
+    assertThat(reader.nextString()).isEqualTo("\n");
+  }
+
+  @Test
+  public void testStrictModeFailsToParseUnespacedControlCharacter() {
```
eamonnmcmanus, line 62: Typo: Unescaped (also in the next method)

gson/src/main/java/com/google/gson/GsonBuilder.java
eamonnmcmanus: Yes, the imports were fine as they were. Probably you need to configure your IDE so it won't do this.

eamonnmcmanus (changes requested):
Thanks, this is a good start!

We do require new code to follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html), even if not all existing code does. You may be able to configure your IDE to use [google-java-format](https://github.com/google/google-java-format).

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
eamonnmcmanus: Sorry, I forgot we're compiling tests in Java 7 mode, so lambdas aren't allowed and `assertThrows` is inconvenient. I think we could actually compile the tests in Java 11 mode, with Gson itself still being Java 7. But that's a different project.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1198,7 +1198,7 @@ public <T> T fromJson(JsonReader reader, Type typeOfT) throws JsonIOException, J
    * <p>Unlike the other {@code fromJson} methods, no exception is thrown if the JSON data has
    * multiple top-level JSON elements, or if there is trailing data.
    *
-   * <p>The JSON data is parsed in {@linkplain JsonReader#setLenient(boolean) lenient mode},
+   * <p>//todo: update javadoc The JSON data is parsed in {@ linkplain JsonReader#setLenient(boolean) lenient mode},
```
MaicolAntali, line 1201: The build fails because there is a whitespace  between `@` and `linkplain`

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
MaicolAntali: Now that this PR #2324 has been merged the tests are compiled with `JDK-11`. So, you can use lambda (in tests). 😄

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -539,7 +571,7 @@ int doPeek() throws IOException {
         throw syntaxError("Expected ':'");
       }
     } else if (peekStack == JsonScope.EMPTY_DOCUMENT) {
-      if (lenient) {
+      if (strictness == Strictness.LENIENT) {
```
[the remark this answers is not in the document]
marten-voorberg: I'm not sure I agree. If you do `isLenient()` in all such places, then you would make sense to call `isStrict()` and `isDefualt()` in other places. However, I think that creating methods for each enum value is a bit overkill as `strict == val` or `getStrictness() == val` seems fine to me.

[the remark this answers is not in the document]
MaicolAntali: The `isStrict()` method can be useful to users. Maybe someone wants to change the level of strictness with some logic and could be relevant know the level of strictness set.

It's true that users can call the `getStrictness()` method and do a `if (getStrictness () ==  ...)`. But this bring me a question: why I can do `if (isLenient()) {...}` and for others level of strictness has to do a little bit more work?

In addition, users, could assume that there is a method for each level of strictness if they see `isLenient()`. 

What do you think @eamonnmcmanus?

gson/src/main/java/com/google/gson/stream/JsonReader.java
marten-voorberg: I've rewritten this in the most recent version. Personally, I'm not convinced its cleaner than what was there before so let me know what you think.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
marten-voorberg: I think we now assert the commit message, without using a lambda.

marten-voorberg:
I think that the current version is very close to ready. Any additional feedback would of course be welcome.

gson/src/main/java/com/google/gson/stream/JsonReader.java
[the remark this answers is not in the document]
marten-voorberg: I think that the reason for being able to call `isLenient` is mostly for backwards compatibility. An extreme option might be to deprecate `isLenient` and `setLenient` in favor of `getStrictness` and `setStrictness`. Not sure what makes the most sense in a situation like this.

gson/src/main/java/com/google/gson/Strictness.java
```diff
@@ -0,0 +1,20 @@
+package com.google.gson;
+
+/**
+ * Modes that indicate how strictly a JSON {@linkplain com.google.gson.stream.JsonReader reader} or
+ * {@linkplain com.google.gson.stream.JsonWriter writer} follows the syntax laid out in the
+ * <a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627 JSON specification</a>.
```
eamonnmcmanus, line 6: I think this should be RFC 8259. (Throughout the PR.)

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
-   * <ul>
-   *   <li>JsonReader allows the literals {@code true}, {@code false} and {@code null}
-   *       to have any capitalization, for example {@code fAlSe}
-   *   <li>JsonReader supports the escape sequence {@code \'}, representing a {@code '}
-   *   <li>JsonReader supports the escape sequence <code>\<i>LF</i></code> (with {@code LF}
-   *       being the Unicode character U+000A), resulting in a {@code LF} within the
-   *       read JSON string
-   *   <li>JsonReader allows unescaped control characters (U+0000 through U+001F)
-   * </ul>
+   * @param lenient provoding <code>true</code> will set the strictness of this reader to {@link Strictness#LENIENT} and
```
eamonnmcmanus, line 299: Typo: providing. Also I think it would make sense to use the same verb for the true and false cases, probably "passing".

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   * Configure how liberal this parser is in what it accepts.
+   *
+   * <p>In {@linkplain Strictness#STRICT strict} mode, the
+   * parser only accepts JSON in accordance with <a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627</a>.
+   * {@linkplain Strictness#LENIENT lenient} mode, all sort of non-spec compliant JSON is accepted (see below).
+   * In {@linkplain Strictness#DEFAULT default} mode, only JSON in accordance with <a
+   * href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627</a> is accepted, with a few exceptions denoted below
+   * for backwards compatibility reasons.</p>
+   *
+   * <ul>
```
eamonnmcmanus, line 326: Perhaps `<dl>` would work better here.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   *     <li>
+   *         {@link Strictness#LENIENT}: In lenient mode, the following syntax erros are accepted:
```
eamonnmcmanus, line 328: Typo: errors. But anyway, I think I would phrase this a bit differently, for example "the following departures from the specification are accepted".

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   *         <ul>
+   *             <li>Streams that start with the <a href="#nonexecuteprefix">non-execute prefix</a>, <code>")]}'\n"
+   *                 </code>.
+   *             <li>Streams that include multiple top-level values. With default or strict parsing,
+   *                 each stream must contain exactly one top-level value.
+   *             <li>Numbers may be {@link Double#isNaN() NaNs} or {@link Double#isInfinite() infinities}.
```
eamonnmcmanus, line 334: I think we should say how these are spelled in our pseudo-JSON.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   *             <li>Array elements separated by {@code ;} instead of {@code ,}.
+   *             <li>Unnecessary array separators. These are interpreted as if null
+   *                 was the omitted value.
+   *             <li>Names and values separated by {@code =} or {@code =>} instead of
+   *                 {@code :}.
+   *             <li>Name/value pairs separated by {@code ;} instead of {@code ,}.
+   *         </ul>
+   *     </li>
+   *     <li>
+   *         {@link Strictness#STRICT}: In strict mode, the only <a href="http://www.ietf.org/rfc/rfc4627.txt">RFC 4627
```
eamonnmcmanus, line 349: I would write, "In strict mode, only input compliant with RFC 8259 is accepted."

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   *         </a> is parsed. This means that the following syntax errors that are allowed in default and lenient mode
```
eamonnmcmanus, line 350: default and lenient modes

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -292,55 +293,94 @@ public JsonReader(Reader in) {
...
+   *                 to have any capitalization, for example {@code fAlSe}
+   *             <li>JsonReader supports the escape sequence {@code \'}, representing a {@code '}
+   *             <li>JsonReader supports the escape sequence <code>\<i>LF</i></code> (with {@code LF}
+   *                 being the Unicode character U+000A), resulting in a {@code LF} within the
+   *                 read JSON string
+   *             <li>JsonReader allows unescaped control characters (U+0000 through U+001F)
+   *         </ul>
+   *     </li>
+   *     <li>
+   *         {@link Strictness#DEFAULT}: In default mode, the syntax errors allowed in leninet mode are not allowed.
```
eamonnmcmanus, line 363: I think it might be easier to follow if the presentation were ordered on the lines of "STRICT mode accepts only the spec; DEFAULT mode is like STRICT mode but also accepts [these four things]; LENIENT mode is like DEFAULT mode but also accepts [this horrifying list of random crap]."

gson/src/main/java/com/google/gson/stream/JsonReader.java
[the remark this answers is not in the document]
eamonnmcmanus: I agree with Marten's assessment. We only keep `isLenient` for compatibility. If you want to know whether your reader is default or strict then it's no great hardship to call `getStrictness`.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -257,24 +258,51 @@ public final FormattingStyle getFormattingStyle() {
...
+  }
+
+  /**
+   * Configure how strict this writer is with regard to the syntax rules specified in <a
+   * href="http://www.ietf.org/rfc/rfc7159.txt">RFC 7159</a>. By default, {@link Strictness#DEFAULT} is used.
+   *
+   * <ul>
+   *     <li>{@link Strictness#DEFAULT} and {@link Strictness#DEFAULT}: The behavior of these
+   *     is currently identical. In these strictness modes, the writer only writes JSON in accordance to
+   *     <a href="http://www.ietf.org/rfc/rfc7159.txt">RFC 7159</a>.</li>
```
eamonnmcmanus, line 287: 8259

eamonnmcmanus (changes requested):
Thanks, this looks very good! For now I mostly have comments on the documentation.

gson/src/main/java/com/google/gson/stream/JsonReader.java
marten-voorberg: I also wasn't super happy with the way I had originally written it. I took this suggestion and I think it's much clearer now.

gson/src/main/java/com/google/gson/stream/JsonReader.java
marten-voorberg: I added that they are represented as `NaN` and `(-)Infinity` respectively

gson/src/main/java/com/google/gson/stream/JsonReader.java
marten-voorberg: I did not even know that HTML tag existed. Changed it now.

gson/src/main/java/com/google/gson/Strictness.java
marten-voorberg: I think I updated this in all places in the PR.

gson/src/main/java/com/google/gson/stream/JsonReader.java
marten-voorberg: @eamonnmcmanus What do you think of this bit of logic has been rewritten?

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -26,18 +26,163 @@
...
+    reader.setStrictness(Strictness.STRICT);
+    assertThat(reader.getStrictness()).isEqualTo(Strictness.STRICT);
+  }
+
+  @Test
+  public void testSetStrictnessNull() {
+    JsonReader reader = new JsonReader(reader("{}"));
+    try {
+      reader.setStrictness(null);
+      fail();
```
eamonnmcmanus, line 71: We should be able to use `assertThrows` and lambdas now that tests run with `--release 11`. That should simplify tests like this drastically.

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -107,8 +107,11 @@
  *
  * <h2>Lenient JSON handling</h2>
  * For legacy reasons most of the {@code Gson} methods allow JSON data which does not
- * comply with the JSON specification, regardless of whether {@link GsonBuilder#setLenient()}
- * is used or not. If this behavior is not desired, the following workarounds can be used:
+ * comply with the JSON specification, even when the strictness {@link Strictness#DEFAULT}
+ * is used. To restrict a {@code Gson} instance from parsing JSON that
+ * does not comply with the JSON specification, {@linkplain Strictness#STRICT strict} mode should be used.
+ * Alternatively, if you do not with to use {@linkplain Strictness#STRICT strict} mode,
```
eamonnmcmanus, line 113: Typo: s/with/wish/ (or want)

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -822,24 +826,37 @@ public void toJson(Object src, Type typeOfSrc, Appendable writer) throws JsonIOE
...
    * {@code writer}.
    *
-   * <p>The JSON data is written in {@linkplain JsonWriter#setLenient(boolean) lenient mode},
-   * regardless of the lenient mode setting of the provided writer. The lenient mode setting
-   * of the writer is restored once this method returns.
+   * <p> The JSON is written in the {@linkplain Strictness strictness} mode of the
+   * the provided {@link JsonWriter JsonWriter} except when the writer is in
+   * {@linkplain Strictness#DEFAULT default} mode. In that case, for legacy reasons,
+   * the JSON is written in {@link Strictness#LENIENT lenient} mode.
+   * The original strictness mode of the writer is returned when this method returns.
```
eamonnmcmanus, line 833: s/returned/restored/

eamonnmcmanus (commented):
A couple of small notes, plus you need to rebase to get the latest changes.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -26,18 +26,135 @@
...
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.junit.Assert.assertThrows;
 
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
 import java.util.Arrays;
+
```
Marcono1234, line 38: No blank line between regular imports (only between `static` and non-`static`)
```suggestion
```

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -26,18 +26,135 @@
 import static com.google.gson.stream.JsonToken.NUMBER;
 import static com.google.gson.stream.JsonToken.STRING;
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertTrue;
```
Marcono1234, line 29: Unused import?

gson/src/test/java/com/google/gson/stream/JsonWriterTest.java
```diff
@@ -30,6 +32,52 @@
...
+    jsonWriter.setStrictness(Strictness.STRICT);
+    assertThat(jsonWriter.getStrictness()).isEqualTo(Strictness.STRICT);
+    jsonWriter.value(false);
+    jsonWriter.close();
+  }
+
+  @Test
+  public void testSetStrictnessNull() throws IOException {
+    JsonWriter jsonWriter = new JsonWriter(new StringWriter());
+    try {
```
Marcono1234, line 73: Probably no need for `try`-`finally` here

Marcono1234 (commented):
I hope it is fine that I also added some review comments; most are regarding some smaller documentation issues. Hopefully the comments are useful.

Also, be wary of GitHub's "hidden conversations":
GitHub hidden conversations screenshot
GitHub has the bad habit of hiding unresolved conversations. On the "Files changed" tab of this PR you can see all unresolved conversations, see the [GitHub documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#discovering-and-navigating-conversations).

Also, it looks like your commit https://github.com/marten-voorberg/DD2480-gson/commit/4f6c4e0f54f0aed850667f8494a9aa864e7bc94c is missing on this pull request, is that intentional?

Marcono1234:
@marten-voorberg, thanks to you and your team for your work on this so far!

It looks like a few comments are still unresolved. Do you want and have the time to address them in the future? Since this pull request targets branch other than `master` (branch `strictness`), an option might also be to merge the changes as is (if you want you can add TODO comments first), and then either you or we can address the remaining comments in follow-up pull requests.

marten-voorberg:
> @marten-voorberg, thanks to you and your team for your work on this so far!
> 
> It looks like a few comments are still unresolved. Do you want and have the time to address them in the future? Since this pull request targets branch other than `master` (branch `strictness`), an option might also be to merge the changes as is (if you want you can add TODO comments first), and then either you or we can address the remaining comments in follow-up pull requests.

I think the PR is almost done. I have time to finalize the last few things. 

For this I was mostly waiting on a response from @eamonnmcmanus about a matter of API design. Namely the fact that `Gson.toJson` and `Gson.fromJson` seem to ignore the strictness setting of the `Gson` instance. See the discussion above for details. @Marcono1234 Do you have any opinion on what to do with such situations in the API?

Besides that, I think all other comments should have been resolved but the PR discussion has become so large that I've sort of lost the overview so it could be that there are a few comments I've missed so feel free to point those out so I can get to them :)

Marcono1234:
For simplicity I have created a pull request against your fork (https://github.com/marten-voorberg/DD2480-gson/pull/49) which should address most of the unresolved comments and with some additional changes. Once you merge that it should also appear here in this pull request.

One open question was whether you missed https://github.com/marten-voorberg/DD2480-gson/commit/4f6c4e0f54f0aed850667f8494a9aa864e7bc94c by accident in your pull request. (Might be good to check if any other commit is missing as well.)

> For this I was mostly waiting on a response from @eamonnmcmanus about a matter of API design. Namely the fact that `Gson.toJson` and `Gson.fromJson` seem to ignore the strictness setting of the `Gson` instance. See the discussion above for details. @Marcono1234 Do you have any opinion on what to do with such situations in the API?

That is probably related to Éamonn's comment https://github.com/google/gson/pull/2323#discussion_r1122175196 (and related: https://github.com/google/gson/pull/2323#discussion_r1125454341, https://github.com/google/gson/pull/2323#discussion_r1125453890).

Personally I think that calling `GsonBuilder.setStrictness` should apply that strictness (regardless which mode) to `toJson` and `fromJson`. The only thing that should remain unchanged for backward compatibility is the default behavior of `Gson` when `setStrictness` was not called. So internally it might make sense to allow `Gson.strictness` to be `null` which represents the default?

I am not completely happy with the name of `Strictness.DEFAULT` though, especially in the context of `Gson`'s default, and with the changes mentioned in the paragraph above, calling `GsonBuilder.setStrictness(Strictness.DEFAULT)` would then be confusing because that would not achieve the default `Gson` strictness mode (which is `LENIENT`). But I don't have a good idea for an alternative name either at the moment, maybe `LEGACY_STRICT`? What do you all think?

Either way, maybe it would make sense to cover any renaming of `DEFAULT` and changes to `toJson` and `fromJson` in a separate pull request.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -609,29 +653,37 @@ private int peekKeyword() throws IOException {
     String keyword;
     String keywordUpper;
     int peeking;
-    if (c == 't' || c == 'T') {
```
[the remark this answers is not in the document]
eamonnmcmanus: FWIW, `Character.toLowerCase(char)` does not depend on the locale and the only characters that map to `t` are `t` and `T`. Same story for `f` and `n`. We might feel a bit icky about it, but it's not wrong.

eamonnmcmanus:
I was going to say the same thing as @Marcono1234: if someone calls `GsonBuilder.setStrictness`, a new method, then we should do the right thing and apply that to `toJson` and `fromJson`. (I might have implied otherwise earlier but I have come around to this way of thinking.)

marten-voorberg:
Thanks for the response. I won't be able to work on this in the coming week
but I'll be sure to take look when I'm back.

gson/src/main/java/com/google/gson/stream/JsonReader.java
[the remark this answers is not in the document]
Marcono1234: Yes you are right, I wasn't very precise with my message. For `keyword.toUpperCase()` it would definitely be necessary to specify a `Locale` though.

> and the only characters that map to `t` are `t` and `T`. Same story for `f` and `n`

I would add here "... at the moment" though. The JDK is regularly updated with new Unicode data, and there is for example already the Kelvin sign character `U+212A` for which `Character.toLowerCase('\u212A') == 'k'` is true. So it is not completely unrealistic that a future Unicode version includes a character whose lowercase variant is `t`, `f` or `n`.

Either way, I have adjusted the code in https://github.com/marten-voorberg/DD2480-gson/pull/49 (mentioned below) to be close to the original code again and to not perform any Unicode case conversions anymore (I assume that would also be more performant, but I have not verified it).

marten-voorberg:
I comitted some changes where the strictness now works as follows:
- By default, a GSON object has a `strictness == null`. When this is the case, the GSON instance _should_ behave exactly as a GSON instance does in the current release to maintain backwards compatibility. 
- If the GSON strictness is set, then this strictness value will be imposed on any writers and readers that are used through the instance, but there old values will be restored when the method returns.

I think if we chose to go this way, we should indeed rename `Strictness.DEFAULT` to something along the lines of `Strictness.LEGACY_STRICT`. 

I have not updated any of the JavaDoc but will do so once we agree on this approach (or decide to go some other way).

eamonnmcmanus:
That approach sounds good to me.

marten-voorberg:
> A couple things which we might want to consider:
> 
>     * Should it be possible to set the strictness to `null` manually by calling `GsonBuilder.setStrictness(null)`? I am inclined to disallow this as calling this method means you are actively thinking about stricntess and you should just chose which one you prefer.
> 
>     * What should `setLenient(boolean)` do for the `JsonWriter` and `JsonReader`? The functionality of passing true is obvious but what should the value be if false is passed? I guess it should be `LEGACY_STRICT` for backwards compatability reasons although I also see the case for making this method parameterless for consistency with the `GsonBuilder` or deprecating it as it now seems to be preferable to use `setStrictness(LENIENT)` instead.

@Marcono1234 @eamonnmcmanus Could you comment on the above? Once this is sorted, we can finalize that JavaDoc and then merge it in :).

Marcono1234:
Sorry for the late response; I don't know what Éamonn thinks about this, but personally I agree with you:
- `GsonBuilder.setStrictness(null)` should not be possible but instead throw a `NullPointerException`
- For `JsonReader` / `JsonWriter` `setLenient(false)` use `LEGACY_STRICT` to remain backward compatible, but deprecate these `setLenient` methods and in the `@deprecated` Javadoc tag refer to `setStrictness`


Also, as side note, this question is still unanswered I think:
> One open question was whether you missed https://github.com/marten-voorberg/DD2480-gson/commit/4f6c4e0f54f0aed850667f8494a9aa864e7bc94c by accident in your pull request. (Might be good to check if any other commit is missing as well.)

Marcono1234:
@eamonnmcmanus, do you agree with that?

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -625,14 +670,17 @@ private int peekKeyword() throws IOException {
...
+    // Confirm that chars [0..length) match the keyword.
     int length = keyword.length();
-    for (int i = 1; i < length; i++) {
+    for (int i = 0; i < length; i++) {
       if (pos + i >= limit && !fillBuffer(i + 1)) {
         return PEEKED_NONE;
       }
       c = buffer[pos + i];
-      if (c != keyword.charAt(i) && c != keywordUpper.charAt(i)) {
+      if (c != keyword.charAt(i) && !(allowsUpperCased && c == keywordUpper.charAt(i))) {
```
eamonnmcmanus, line 683: The negation makes this condition a little hard to follow. What would you think of something like this?
```java
boolean matched = c == keyword.charAt(i) || (allowsUpperCased && c == keywordUpper.charAt(i));
if (!matched) {
  return PEEKED_NONE;
}
```

eamonnmcmanus:
Yes, I agree your [earlier comment](https://github.com/google/gson/pull/2323#issuecomment-1535396944).

Sorry I haven't been following this more closely. Is it ready to be merged into the `strictness` branch?

marten-voorberg:
> Sorry I haven't been following this more closely. Is it ready to be merged into the `strictness` branch?

No problem. I think this can be merged into the branch! @eamonnmcmanus 

> One open question was whether you missed [marten-voorberg@4f6c4e0](https://github.com/marten-voorberg/DD2480-gson/commit/4f6c4e0f54f0aed850667f8494a9aa864e7bc94c) by accident in your pull request. (Might be good to check if any other commit is missing as well.)

@Marcono1234 Good point. I've not directly added the commit back in but I have added an equivalent test case that uses the more modern `assertThrows()`. As far as I can tell, there are no other missing commits. 

I've now deprecated `setLenient` in `JsonReader`, `JsonWriter`, and `GsonBuilder`. I think the next step is to merge it into the strictness branch. After that, we should probably make sure that the JavaDoc is up-to-date everywhere as we changed the functionality quite a bit througout this PR so there might be some lingering, outdated documentation.

Marcono1234 (approved):
Thanks for the changes! Sorry, I hope you weren't waiting for my response. Just to make sure there is no misunderstanding, to me the changes look good.

There might be a few small documentation changes I want to make, and maybe extending the U+2028 and U+2029 tests, but this is definitely not blocking this PR.

As you mentioned, it would be good to merge this to the `strictness` branch already and then perform any additional changes in separate PRs. The first thing would be merging `master` back into the `strictness` branch.

eamonnmcmanus (approved):

eamonnmcmanus:
OK, my attempts to merge the `strictness` branch seem to have left it in a not-great state. I will see if I can fix it tomorrow.

eamonnmcmanus:
I believe the new `strictness2` branch now reflects what I _should_ have done with `strictness`. Please confirm. For now I have removed the deprecation tags from `setLenient` since they cause compilation failures now that we compile with `-Werror`. It would make sense to restore the deprecation and fix those failures, either by calling the replacement methods or by using `@SuppressWarnings("deprecation")`.

Marcono1234:
Thanks a lot for resolving the merge conflicts! Comparing the original state of `strictness` after the merge with `strictness2` using https://github.com/google/gson/compare/9ce668e10bb2aed284cb8b5ad0345f0dbea8670b..strictness2 (I hope that URL is correct), it looks to me like nothing was lost (except the `setLenient` changes you mentioned).

In the end it would be good to compare `strictness2` and `master` again before merging, just to make sure it does not accidentally revert something from `master`.

PR #2324, opened 2023-02-24
Adds the `<maven.compiler.testRelease>11</maven.compiler.testRelease>`

MaicolAntali:
Adds the `<maven.compiler.testRelease>11</maven.compiler.testRelease>` property to run test with JDK 11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.release>7</maven.compiler.release>
+    <maven.compiler.testRelease>11</maven.compiler.testRelease>
   </properties>
 
   <scm>
@@ -291,15 +292,6 @@
   </build>
 
   <profiles>
-    <profile>
-      <id>JDK11</id>
-      <activation>
-        <jdk>(,11]</jdk>
-      </activation>
-      <properties>
-        <maven.compiler.testRelease>11</maven.compiler.testRelease>
-      </properties>
-    </profile>
     <!-- Disable Error Prone in Java 15 -->
     <profile>
       <id>jdk15</id>
--- a/pom.xml
+++ b/pom.xml
@@ -291,6 +291,15 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>JDK11</id>
+      <activation>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -291,6 +291,15 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>JDK11</id>
+      <activation>
+        <jdk>(,11]</jdk>
```
Marcono1234, line 297: The compiler plugin is already executed with JDK 11 or newer, see
https://github.com/google/gson/blob/742fd50ff74350462d8dcf80941efce219e187dd/pom.xml#L123-L125

You should therefore be able to specify `maven.compiler.testRelease` in the main properties, without having to define a separate profile.

MaicolAntali: So, I can put `<maven.compiler.testRelease>11</maven.compiler.release> ` here:
https://github.com/google/gson/blob/742fd50ff74350462d8dcf80941efce219e187dd/pom.xml#L36-L39

and everything should work?

Marcono1234: Yes I think so. The project already requires JDK 11 to build and fails to compile if you use JDK 8.
However, unfortunately the current error message is a bit cryptic because the Maven plugins do not seem to notice that the used JDK is < JDK 11.

Maybe it would be useful to add the Maven Enforcer Plugin to fail fast when users try to run with < JDK 11, e.g.:
```xml
<plugin>
  <artifactId>maven-enforcer-plugin</artifactId>
  <version>3.2.1</version>
  <executions>
    <execution>
      <id>enforce-jdk-version</id>
      <goals>
        <goal>enforce</goal>
      </goals>
      <configuration>
        <rules>
          <requireJavaVersion>
            <version>[11,)</version>
          </requireJavaVersion>
        </rules>
      </configuration>
    </execution>
  </executions>
</plugin>
```

Marcono1234:
One issue might be how well different target versions for main and test code are supported by IDEs, hopefully in the worst case IDEs choose the greater version for both main and test code (here Java 11). That might make users write code which is not actually valid for Java 7, but the Maven build will then at least catch that.

eamonnmcmanus:
I think this is good to merge. If there are unforeseen problems we can address them later.

PR #2327, opened 2023-02-26
FormattingStyle follow-up

Marcono1234:
Follow-up for #2231

- Extends and adds missing documentation
- Fixes some small typos
- Adds tests for `null` being used as formatting style

### DIFF
--- a/gson/src/main/java/com/google/gson/FormattingStyle.java
+++ b/gson/src/main/java/com/google/gson/FormattingStyle.java
@@ -16,6 +16,7 @@
 
 package com.google.gson;
 
+import com.google.gson.stream.JsonWriter;
 import java.util.Objects;
 
 /**
@@ -24,6 +25,8 @@ import java.util.Objects;
  * <p>It currently defines the kind of newline to use, and the indent, but
  * might add more in the future.</p>
  *
+ * @see GsonBuilder#setPrettyPrinting(FormattingStyle)
+ * @see JsonWriter#setFormattingStyle(FormattingStyle)
  * @see <a href="https://en.wikipedia.org/wiki/Newline">Wikipedia Newline article</a>
  *
  * @since $next-version$
--- a/gson/src/main/java/com/google/gson/FormattingStyle.java
+++ b/gson/src/main/java/com/google/gson/FormattingStyle.java
@@ -19,7 +19,7 @@ package com.google.gson;
 import java.util.Objects;
 
 /**
- * A class used to control what the serialization looks like.
+ * A class used to control what the serialization output looks like.
  *
  * <p>It currently defines the kind of newline to use, and the indent, but
  * might add more in the future.</p>
@@ -32,7 +32,11 @@ public class FormattingStyle {
   private final String newline;
   private final String indent;
 
-  static public final FormattingStyle DEFAULT =
+  /**
+   * The default pretty printing formatting style using {@code "\n"} as
+   * newline and two spaces as indent.
+   */
+  public static final FormattingStyle DEFAULT =
       new FormattingStyle("\n", "  ");
 
   private FormattingStyle(String newline, String indent) {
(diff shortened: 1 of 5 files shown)


gson/src/main/java/com/google/gson/FormattingStyle.java
```diff
@@ -44,35 +51,35 @@ private FormattingStyle(String newline, String indent) {
...
-          "Only combinations of spaces and tabs allowed in indent.");
+          "Only combinations of spaces and tabs are allowed in indent.");
     }
     this.newline = newline;
     this.indent = indent;
   }
 
   /**
-   * Creates a {@link FormattingStyle} with the specified newline setting.
+   * Creates a {@code FormattingStyle} with the specified newline setting.
```
Marcono1234, line 61: Have changed this to `@code` because linking to the class declaring this method is probably not necessary.

The (somewhat dated) ["How to Write Doc Comments for the Javadoc Tool" guide](https://www.oracle.com/de/technical-resources/articles/java/javadoc-tool.html) also says: "Use in-line links economically"

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
```diff
@@ -89,6 +90,13 @@ public void testNewlineOs() {
     assertEquals(EXPECTED_OS, json);
   }
 
+  @Test
+  public void testCompact() {
+    Gson gson = new GsonBuilder().setPrettyPrinting(null).create();
+    String json = gson.toJson(INPUT);
+    assertEquals(EXPECTED_COMPACT, json);
```
MaicolAntali, line 97: You can use Truth

```java
assertThat(json).isEqualTo(EXPECTED_COMPACT)
```

_Note: Some other above tests use `assertEquals(...)`, should be modified them too._

MaicolAntali (commented):
Looks nice, just a small review

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
Marcono1234: Good point, have changed it

eamonnmcmanus (commented):
Looking at these changes, it strikes me that it's kind of ugly to use `null` to mean no formatting. Could we have `FormattingStyle.NONE` instead? I think having empty strings for `newline` and `indent` should mean no formatting? Also, we should probably reject having an empty `newline` with a non-empty `indent`.

These changes are outside the scope of this PR but I think we should probably make them before freezing the API at the next release.

Marcono1234:
> Could we have `FormattingStyle.NONE` instead? I think having empty strings for `newline` and `indent` should mean no formatting?

Though currently `JsonWriter` sets `this.separator = ": "` for any non-`null` formatting style. Maybe there are used cases where that is wanted for empty newline and indentation, but then a special hardcoded check for `FormattingStyle.NONE` is needed, regardless of whether it has empty newline and indentation. What do you think?

eamonnmcmanus:
> Though currently `JsonWriter` sets `this.separator = ": "` for any non-`null` formatting style. Maybe there are use cases where that is wanted for empty newline and indentation, but then a special hardcoded check for `FormattingStyle.NONE` is needed, regardless of whether it has empty newline and indentation. What do you think?

I think that suggests that `FormattingStyle` should also allow users to specify what space comes after `:` and `,`? `FormattingStyle.NONE` would have an empty string there, and `FormattingStyle.DEFAULT` would have a single space.

Marcono1234:
> what space comes after `:` and `,`

Currently after a `,` there is either nothing (compact) or a newline:
https://github.com/google/gson/blob/ef35a34a6033914644410845c4b90f52247c763f/gson/src/main/java/com/google/gson/stream/JsonWriter.java#L727-L729

For space after `:` I am not sure if we really need to expose that, and if the use cases for customizing this are that common. Maybe for now we could store that internally and use no space for `NONE`, but any customization of `NONE` implicitly adds a space. If users really request this to be customized we can consider in the future adding API for that.

Also, what about naming `NONE` → `COMPACT`, and `DEFAULT` → `PRETTY`; that might make their meaning more clear? At least from the perspective that formatting is always applied, even if that means for `COMPACT` to add no whitespace.

eamonnmcmanus:
OK, well what do you think of this:

* For this PR, just address the minor issues with javadoc.
* In a later PR:
  * Rename `DEFAULT` to `PRETTY` and introduce `COMPACT` (née `NONE`). I agree that these are better names.
  * Introduce `FormattingStyle.equals` with the obvious behaviour.
  * Forbid a null `FormattingStyle`.

I agree that customizing what comes after `,` or `:` is probably not necessary, at least until people ask for it. If you're using a `FormattingStyle` that `equals(FormattingStyle.COMPACT)` then there's no whitespace; otherwise we use newline+indent after `,` and a single space after `:`. We could add a public method `FormattingStyle.isCompact()` that caches the value of `equals(COMPACT)`.

gson/src/main/java/com/google/gson/FormattingStyle.java
Marcono1234: The point about validating the referenced name makes sense, but to me that appears to be rather a workaround for missing Javadoc functionality (e.g. `@code` tag which represents a reference).

Personally I still think in the HTML output the links can be distracting / confusing, giving the impression that they refer to a different, similarly named class.

I have reverted these `@link` changes now though.

Marcono1234:
Have reduced this now to Javadoc and minimal code changes.

eamonnmcmanus:
LGTM

PR #2339, opened 2023-03-06
Specify Error Prone JVM args with `.mvn/jvm.config`

Marcono1234:
### Purpose
Specify JVM arguments needed for Error Prone using the [`.mvn/jvm.config` file](https://maven.apache.org/configure.html#mvn-jvm-config-file).

### Description
This allows avoiding fork mode for compilation, which:
- might slightly reduce compilation time
- guarantees that all compiler diagnostics are properly reported

See also https://github.com/google/gson/pull/2320#issuecomment-1456934862 and comments above.

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,7 +105,7 @@
         <version>3.2.1</version>
         <executions>
           <execution>
-            <id>enforce-maven-version</id>
+            <id>enforce-versions</id>
             <goals>
               <goal>enforce</goal>
             </goals>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -104,6 +104,20 @@
...
+          <execution>
+            <id>enforce-maven-version</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireMavenVersion>
+                  <!-- Usage of `.mvn/jvm.config` for Error Prone requires at least Maven 3.3.1 -->
+                  <version>[3.3.1,)</version>
```
Marcono1234, line 116: Possibly some plugins used during build already require a newer version, but version 3.3.1 (released in 2015) added support for `.mvn/jvm.config`, so it is definitely the lower bound.

pom.xml
```diff
@@ -104,6 +104,20 @@
         <artifactId>maven-enforcer-plugin</artifactId>
         <version>3.2.1</version>
         <executions>
+          <execution>
+            <id>enforce-maven-version</id>
```
eamonnmcmanus, line 108: I wonder if this can be merged with the existing `<execution>`? The example on the [usage page](https://maven.apache.org/enforcer/maven-enforcer-plugin/usage.html) makes it look as if that should be possible.

eamonnmcmanus (commented):
Thanks, this is great! One question, though.

eamonnmcmanus:
Thanks again!

PR #2346, opened 2023-03-18
Make dependency on Error Prone Annotations non-optional

Marcono1234:
### Purpose

Avoid problems for users which could occur due to dependency being optional

### Description

See discussion starting at https://github.com/google/gson/pull/2320#issuecomment-1455233938
Making the dependency non-optional seems to be safer and avoids confusing warnings for users.

### Checklist

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -38,6 +38,17 @@
   </properties>
 
   <dependencies>
+    <!-- This dependency can be considered optional; omitting it during runtime will most likely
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Thanks! As discussed previously, this does seem to be the right thing to do. As mentioned in that discussion, we _might_ later decide to have Gson-private versions of the annotations we use (`@Keep` and `@InlineMe`), since their analysers only look at the simple name of the annotation. But the overhead of an annotation definition is pretty small so it's probably not worth doing that.

PR #2351, opened 2023-03-22
Fix incorrect inherited URLs in `pom.xml`

Marcono1234:
### Purpose
Fixes incorrect inherited URLs in the `pom.xml`, mainly in the one of the `gson` Maven module

### Description
Previously Maven appended the artifact ID of the modules which lead for example for the `gson` module to the incorrect URL https://github.com/google/gson/gson (note the extra `/gson` at the end). This could be seen for example with `mvn help:effective-pom`:
```xml
...
 <url>https://github.com/google/gson/gson</url>
...
  <scm>
    <connection>scm:git:https://github.com/google/gson.git/gson</connection>
    <developerConnection>scm:git:git@github.com:google/gson.git/gson</developerConnection>
    <url>https://github.com/google/gson/gson/</url>
  </scm>
...
```
Other situations where this might have been visible are when using plugins to generate a summary of dependencies of a project and probably also the [Sonatype Maven Central browser](https://central.sonatype.com/artifact/com.google.code.gson/gson/2.10.1) (see "Project URL" on the right).

This pull request uses the attributes described [here](https://maven.apache.org/ref/3.9.1/maven-model-builder/index.html#inheritance-assembly) to disable appending the artifact ID of the modules. With these changes the output of `mvn help:effective-pom` looks correct:
```xml
...
  <url>https://github.com/google/gson</url>
...
  <scm child.scm.connection.inherit.append.path="false" child.scm.developerConnection.inherit.append.path="false" child.scm.url.inherit.append.path="false">
    <connection>scm:git:https://github.com/google/gson.git</connection>
    <developerConnection>scm:git:git@github.com:google/gson.git</developerConnection>
    <url>https://github.com/google/gson/</url>
  </scm>
...
```

See also the related Guava issue: https://github.com/google/guava/issues/5618

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,8 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd"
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Thanks!

PR #2357, opened 2023-03-25
Link to troubleshooting guide from exception messages

Marcono1234:
:warning: Maybe wait with merging this until #2323 is merged.

### Purpose
Adds troubleshooting guide links to some exception messages; improves unit tests

### Description
As shortly discussed in https://github.com/google/gson/issues/1219#issuecomment-1455164503 (`TypeToken` validation is *not* implemented here though) adding troubleshooting guide links to exception messages hopefully makes it easier for users to understand what the exception means and how to solve the issue.

This approach again uses custom HTML anchors in a Markdown document even though we had bad experience with this before (#2286), but it appears using only lowercase IDs is supported by GitHub (though possibly not officially, see open question https://github.com/orgs/community/discussions/50962). The custom HTML anchors allow us to produce short stable links.

Open questions:
- Possibly we could also consider including the troubleshooting section ID (e.g. "T1") in the troubleshooting guide headers, but that might become confusing if we decide to reorder, remove or merge sections in the future and the section IDs are then not in order anymore. What is your opinion?
- Currently for all `MalformedJsonException`s it refers to the troubleshooting guide; maybe we should reduce this only to specific variants?

### Checklist


Marcono1234:
Or maybe instead of the anchors `t1`, `t2`, ... it would be better to use more expressive ones, such as `unexpected-structure`, `class-unsupported`, .... What do you think?

eamonnmcmanus:
This looks great! I'm sure it will be very helpful for users.
I'm holding off on merging as requested.

eamonnmcmanus (approved):

Marcono1234:
> I'm holding off on merging as requested.

Or could you please merge it now already? My intention was to avoid conflicts for #2323, however it looks like that has already conflicts with `master` now, and also it targets the `strictness` branch, so it might take some time until it is finally merged to `master`.

PR #2358, opened 2023-03-26
Add strategies for unknown and missing fields

### LINKED ISSUE #188: Support handlers for unknown properties, useful for error handling

### LINKED ISSUE #1005: Default Value for Missing fields in Json

Marcono1234:
### Purpose
Fixes #188
Fixes #1005

Supersedes  #1006

### Description
Adds the interfaces `MissingFieldValueStrategy` and `UnknownFieldStrategy` to allow handling missing and unknown fields. The default behavior remains the same.

The API for `MissingFieldValueStrategy` was inspired by #1006. These interfaces are rather low level, but should hopefully allow implementing most strategies, such as throwing an exception or ignoring missing or unknown fields.

**Important:** `MissingFieldValueStrategy` makes the assumption that there is always a `Field` for a JSON object member. If we decide to take this path we might have a hard time supporting setter methods in the future (#232), in case we want that at some point. Should we instead of directly exposing `Field` use some wrapper type which in the future would support representing setter methods as well (maybe we could use the existing `FieldAttributes` for that)?

I have marked this as draft for now to gather some feedback, especially about API and implementation. Please feel free to try out the changes locally and let me know what functionality your a missing or if you encounter and issues.

I am a bit worried that the changes `ReflectiveTypeAdapterFactory` might negatively affect performance. Though maybe that is unavoidable to some extent.

### Checklist


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -388,6 +394,37 @@ public GsonBuilder setObjectToNumberStrategy(ToNumberStrategy objectToNumberStra
...
+  public GsonBuilder setMissingFieldValueStrategy(MissingFieldValueStrategy missingFieldValueStrategy) {
+    this.missingFieldValueStrategy = Objects.requireNonNull(missingFieldValueStrategy);
+    return this;
+  }
+
+  /**
+   * Configures Gson to apply a specific unknown field strategy during deserialization.
+   * The strategy is used during reflection-based deserialization when an unknown field
+   * is encountered in the JSON data. If a field which is excluded from deserialization
+   * appears in the JSON data it is considered unknown as well.
```
Marcono1234, line 416: There might be use cases where a field excluded from deserialization should not be considered unknown, but should just be silently ignored.

With the current `UnknownFieldStrategy` API users might be able to determine themselves if a field is excluded, but not in a very reliable way.

We could possibly add a `boolean excluded` parameter to `UnknownFieldStrategy.handleUnknownField` whose value is `true` if `ReflectiveTypeAdapterFactory` finds a `BoundField` but it has `!field.deserialized`. Though I am not sure if that is really worth it. What are your opinions?

gson/src/main/java/com/google/gson/MissingFieldValueStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * @param declaringType type declaring the field
+   * @param instance instance of the declaring type, {@code null} for Record classes
+   * @param field field whose value is missing
+   * @param resolvedFieldType resolved type of the field
+   * @return the field value, or {@code null}
+   */
+  // TODO: Should this really expose `instance`? Only use case would be to derive value from other fields
+  //   but besides that user should not directly manipulate `instance` but return new value instead
+  Object handleMissingField(TypeToken<?> declaringType, Object instance, Field field, TypeToken<?> resolvedFieldType);
```
Marcono1234, line 76: Not completely sure whether it is really necessary to expose `instance`.

gson/src/main/java/com/google/gson/MissingFieldValueStrategy.java
```diff
@@ -0,0 +1,77 @@
+package com.google.gson;
+
+import com.google.gson.internal.reflect.ReflectionHelper;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Field;
+
+/**
+ * A strategy defining how to handle missing field values during reflection-based deserialization.
+ *
```
Marcono1234, line 9: Maybe this should have a warning that users should be careful when implementing security critical validation with this, and add extensive tests. Because if they forget to call `setMissingFieldValueStrategy` or use a default `Gson` instance somewhere, then their custom strategy will not be used.

gson/src/main/java/com/google/gson/UnknownFieldStrategy.java
```diff
@@ -0,0 +1,79 @@
+package com.google.gson;
+
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import java.io.IOException;
+
+/**
+ * A strategy defining how to handle unknown fields during reflection-based deserialization.
+ *
```
Marcono1234, line 9: Maybe this should have a warning that users should be careful when implementing security critical validation with this, and should add extensive tests. Because if they forget to call `setUnknownFieldStrategy` or use a default `Gson` instance somewhere, then their custom strategy will not be used.

PR #2362, opened 2023-03-29
Fix incorrect sample code in User Guide

False-Mask:
1. gson is configured serializeNulls, so the classField will be serialized.
2. you forgot to initial the final class member—— clazzField.

### Checklist

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -651,7 +651,6 @@ public class SampleObjectForTest {
   @Foo private final int annotatedField;
   private final String stringField;
   private final long longField;
-  private final Class<?> clazzField;
 
   public SampleObjectForTest() {
     annotatedField = 5;


eamonnmcmanus:
Thanks! Definitely that `final` field needs to be initialized in the constructor.

We do need the CLA before we can merge this.

Marcono1234 (commented):
Good catch!

Though Gson does not support support serialization and deserialization of `Class` by default ([relevant code](https://github.com/google/gson/blob/1d9d9774be408d1f24cb7c7cb4f2050cdb265e2f/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java#L74-L82)). So even though this sample happens to work because the value of the field is `null`, maybe it would be better to remove the field to avoid confusion and because it is not related to exclusions at all.

Also, could you please update the title of this pull request to be more descriptive, e.g. "Fix incorrect sample code in User Guide"?

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks for this!

PR #740, opened 2015-11-23
ISO8601 Date deserialization

IgorMiller:
Added support for all ISO8601 formats, including milliseconds. You already have `UtcDateTypeAdapter` in `gson-extras` project, which was using the code from https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/util/ISO8601Utils.java to parse dates. Therefore I did the same, I created an utils class in gson project with that code and made it used by `DateTypeAdapter` and `DefaultDateTypeAdapter`

PR #2363, opened 2023-04-01
Fixes outdated URL (`ISO8601Utils.java`)

### LINKED ISSUE #1475: URL outdated (ISO8601Utils.java)
In [ISO8601Utils.java](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java) a file is linked in line 16, which has been removed and can't be accessed. I'm not completely sure what the file once contained, but I think [this file](https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/util/StdDateFormat.java) might be the new location. Should the link just be removed (maybe with line 15 referencing it) or should it be changed to the file I proposed or where is this file now located?

MaicolAntali:
Fixes the javadoc and the link to Jackson databind ISO8601Utils.java

### Purpose
Fixes #1475 

### Description
Fixes the outdated link to the Jackson databind `ISO8601Utils.java`

### Checklist

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -62,7 +62,7 @@ public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
   }
 
   // Date parsing code from Jackson databind ISO8601Utils.java
-  // https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/util/ISO8601Utils.java
+  // https://github.com/FasterXML/jackson-databind/blob/2.8/src/main/java/com/fasterxml/jackson/databind/util/ISO8601Utils.java
   private static final String GMT_ID = "GMT";
 
   /**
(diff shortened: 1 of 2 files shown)


Marcono1234:
Thanks!

I am not completely sure if Jackson 2.8 is really the correct version since that was released apparently in 2016 ([release notes](https://github.com/FasterXML/jackson-databind/blob/d03e6b5f9f8a152e71dbfc5970aaa7248fc26ec7/release-notes/VERSION-2.x#L1405)) whereas this code might be older. But it is probably good enough and trying to find the exact version is probably not worth it, especially since modifications here in the Gson code were made and as mentioned in #740 the code might (or might not?) come from Gson `extra`'s `UtcDateTypeAdapter` and therefore might be even older.

Speaking of `UtcDateTypeAdapter` though; could you please update the outdated URL there as well?
https://github.com/google/gson/blob/3adead6ab4f240ed53322eff34210ed4a69dd490/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java#L65

MaicolAntali:
> I am not completely sure if Jackson 2.8 is really the correct version

Yeah, I didn't find the exact same Class in the `Jackson` codebase so I put the most similar

> Speaking of UtcDateTypeAdapter though; could you please update the outdated URL there as well?

Of course 😄

Marcono1234 (approved):

eamonnmcmanus (approved):

PR #2370, opened 2023-04-10
Change in the "UserGuide.md" in the "Collections Example" section

avimathur009:
### Checklist

- [1] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)
- [1] If necessary, new public API validates arguments, for example rejects `null`
- [1] New public API has Javadoc
    - [1] Javadoc uses `@since $next-version$`  
- [0] If necessary, new unit tests have been added  
  - [0] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [0] No JUnit 3 features are used (such as extending class `TestCase`)
  - [0] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [1] `mvn clean verify javadoc:jar` passes without errors


Marcono1234:
Thanks for the suggestion!

> According to the code, and the use of the Gson class, the actual output as a json would be simply that collection in " "

Do you mean the `json is ` prefix is confusing? Is is this about the space ` ` between `is` and `[` (and therefore you removed the complete `json is `)?
Note that there are also other examples which use `json is ...` as well.

Marcono1234:
Ok; for consistency it might be good to also remove this in the "Maps Examples" section right below it as well.

There are a few more `json is` usages in that file but the formatting there is different so maybe they can remain as they are.

I assume you will have to sign the Contributor License Agreement (CLA) for this though, see the message by the bot above (or the failing pull request check).

avimathur009:
Oh okay, yeah I will remove it from "Maps Examples" as well. 
Though I am not trying to resolve the CLA/Google check, even though I signed it with correct email and username. 

May I make a new pull request for the same change? I am trying to sign the CLA but am not able to.

avimathur009:
Hi, yeah understand, just working on it right now.

avimathur009:
I am closing the PR for now, and setting up my email of 'no-reply' for my commits. Once I am done with that, I'll make a new PR with same PR name, to make all the commits associated with that email itself, and resolving the CLA error.

PR #2371, opened 2023-04-10
Gson ABCI Fella

EruSaki:
### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -222,6 +222,9 @@ public class JsonReader implements Closeable {
   private static final int NUMBER_CHAR_EXP_E = 5;
   private static final int NUMBER_CHAR_EXP_SIGN = 6;
   private static final int NUMBER_CHAR_EXP_DIGIT = 7;
+  private static final int MAX_STACK_SIZE = 32;
+  private static final int SECURITY_TOKEN_LENGTH = 5;
+
 
   /** The input JSON. */
   private final Reader in;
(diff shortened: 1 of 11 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -222,6 +222,9 @@ public class JsonReader implements Closeable {
   private static final int NUMBER_CHAR_EXP_E = 5;
   private static final int NUMBER_CHAR_EXP_SIGN = 6;
   private static final int NUMBER_CHAR_EXP_DIGIT = 7;
+  private static final int MAX_STACK_SIZE = 32;
```
Marcono1234, line 225: This is not the max stack size; it is the initial stack size, see `JsonReader.push`. I am also not sure if this value needs a dedicated `static final` field.

Marcono1234 (commented):
Hello,
could you please explain a bit the purpose of this pull request (and edit title and description)? I don't really understand what your title "ABCI Fella" means. Your pull request contains some good changes, but maybe it would be better to split it; for example the `README` changes could probably be in a separate pull request.
Also note that some of your changes modify the public API of Gson (removal of `Gson.excluder()`; renaming `GsonBuilder.create()`), we cannot do this since it would break backward compatibility.

It is also necessary to sign the Contributor License Agreement (CLA), as mentioned by the bot comment above.

jecisc:
Hi,

Sorry for the inconvenience, I am teaching a software engineering class where the students have to analyze an existing project and propose improvements to them in terme of software engineering.
I did not think that my students would do a PR to the projects. 

This can probably be closed, or some improvements could be cherry pick if you wish.

Marcono1234:
Thanks for the clarification! And no problem.

> or some improvements could be cherry pick if you wish

This would probably still require the CLA to be signed. But @EruSaki, feel free to create a new pull request where you exclude any breaking changes, if you want and if you can sign the CLA 🙂

@eamonnmcmanus, should we close this pull request then? Or of course @EruSaki you can also close it as well if you want.

Marcono1234:
Going to close this pull request then, as suggested in https://github.com/google/gson/pull/2371#issuecomment-1513435184.

PR #2372, opened 2023-04-10
Address review comments about the recent `@CanIgnoreReturnValue` changes.

eamonnmcmanus:


Marcono1234 (approved):
Thanks! (for context: https://github.com/google/gson/pull/2369#pullrequestreview-1378039009)

A bit nitpicky, but maybe it would be good if the PR title was a bit more specific, e.g.

> Address review comments for `@CanIgnoreReturnValue` usage

or similar. What do you think?

eamonnmcmanus:
Crap, I changed the PR title but not the commit message. Oh well.

Marcono1234:
At least it were only 4 files though 😄
With the next pull requests that commit message will be buried in the Git history.

PR #2374, opened 2023-04-14
Remove the comment strings " json is " from the UserGuide.md

avimathur009:
### Purpose
The purpose of this PR is to update the "UserGuide.md" to avoid confusion for the reader looking at the examples.

### Description
According to the code, and the use of the Gson class, the actual output as a json would be simply that collection

### Checklist

- [1] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)
- [1] If necessary, new public API validates arguments, for example rejects `null`
- [1] New public API has Javadoc
    - [1] Javadoc uses `@since $next-version$`  
- [0] If necessary, new unit tests have been added  
  - [0] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [0] No JUnit 3 features are used (such as extending class `TestCase`)
  - [0] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [1] `mvn clean verify javadoc:jar` passes without errors

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -135,7 +135,7 @@ BagOfPrimitives obj = new BagOfPrimitives();
 Gson gson = new Gson();
 String json = gson.toJson(obj);
 
-// ==> json is {"value1":1,"value2":"abc"}
+// ==>{"value1":1,"value2":"abc"}
 ```
 
 Note that you can not serialize objects with circular references since that will result in infinite recursion.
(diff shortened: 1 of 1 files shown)


Marcono1234:
Thanks! It looks like you overlooked two occurrences of `json is` though; for `// json is {"key":"value"}` could you please also add `==>` for consistency there?

And could you please also for consistency and for better readability add a space after `==>` for the cases where you removed `json is`? E.g. `==> {` instead of `==>{`

avimathur009:
Yeah I have made the additional changes as well,  two occurrences of `json is` 
And also made the change for `==> { instead of ==>{` at respective places

Marcono1234 (approved):

eamonnmcmanus (approved):

avimathur009:
Thank you for the merge! :)

PR #2375, opened 2023-04-15
Improve creation of `ParameterizedType`

Marcono1234:
### Purpose
Improve creation and validation of `ParameterizedType`

### Description
- Reject non-generic raw types for `TypeToken.getParameterized`
- Fix `ParameterizedTypeImpl` erroneously requiring owner type for types without owner (mainly / only local classes?)
- Extend tests and use `assertThrows`

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/reflect/TypeToken.java
+++ b/gson/src/main/java/com/google/gson/reflect/TypeToken.java
@@ -97,8 +97,8 @@ public class TypeToken<T> {
     }
     // Check for raw TypeToken as superclass
     else if (superclass == TypeToken.class) {
-      throw new IllegalStateException("TypeToken must be created with a type argument: new TypeToken<...>() {}; "
-          + "When using code shrinkers (ProGuard, R8, ...) make sure that generic signatures are preserved.");
+      throw new IllegalStateException("TypeToken must be created with a type argument: new TypeToken<...>() {};"
+          + " When using code shrinkers (ProGuard, R8, ...) make sure that generic signatures are preserved.");
     }
 
     // User created subclass of subclass of TypeToken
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -351,6 +351,18 @@ public static TypeToken<?> getParameterized(Type rawType, Type... typeArguments)
     Class<?> rawClass = (Class<?>) rawType;
     TypeVariable<?>[] typeVariables = rawClass.getTypeParameters();
 
+    // Note: Does not check if owner type of rawType is generic because this factory method
+    // does not support specifying owner type
+    if (typeVariables.length == 0) {
```
eamonnmcmanus, line 356: I ran #2376 against all of Google's internal tests and found one failure where someone was calling this method with a non-generic type and an empty `typeArguments` array. Of course then it is just the same as calling `get(NonGeneric.class)`, but I think this case should continue to work. It's something of a historical quirk that the reflection API represents classes with no type parameters so differently from classes with some.

There were no other failures, so I think the other changes are probably safe.

eamonnmcmanus: The code I was talking about is in a test, and looks something like this:

```java
    Type type = TypeToken.getParameterized(FooResponse.class).getType();
    return new GsonBuilder()
        .registerTypeAdapterFactory(AutoValueAdapterFactory.create())
        .registerTypeAdapter(ImmutableList.class, new ImmutableListDeserializer())
        .create()
        .fromJson(json, type);
```

I had not realized that `TypeToken.getParameterized` returns a `ParameterizedType` with no type arguments in this case. That is indeed a bizarre sort of beast, and I can't help feeling that this code works mostly by accident. Since it's test code, rather than production, and apparently nothing else in all of Google's code is doing this, I think it's probably OK to introduce this exception. We should make sure to mention it in the next release notes, though.

eamonnmcmanus (approved):

gson/src/main/java/com/google/gson/reflect/TypeToken.java
eamonnmcmanus: For the record, I did find another instance, in non-test code this time. I think the code in question happened to be broken for unrelated reasons when I did my earlier test, which is why I didn't see the failure then. I think it might be worth catching this case and returning `TypeToken.get(rawType)`, assuming the type-argument list is also empty.

Marcono1234: Ok, have created #2447 for that. I hope that is what you had in mind.

PR #2387, opened 2023-05-09
Update Error Prone, bypass a bug in it, fix some problems it flags

eamonnmcmanus:

PR #2390, opened 2023-05-11
Enable Error Prone HidingField bug pattern again

Marcono1234:
Bug which required disabling this bug pattern was fixed in 2.19.1 and Error Prone dependency was updated to that version in commit 3c4b0721005dbc5e0d7515b95a321715d48b3291.

See also pull request #2387 which disabled this bug pattern initially.


MaicolAntali (approved):

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks!

PR #2395, opened 2023-05-24
Remove accidental invisible characters from Troubleshooting Guide

Marcono1234:
Accidentally contained U+200C (Zero Width Non-Joiner) followed by U+200B (Zero Width Space) here: `cla#ss`


eamonnmcmanus:
I wonder how that happened!

Marcono1234:
> I wonder how that happened!

If I recall correctly when I added that entry to the Troubleshooting Guide I copied it from somewhere where I had just written something similar before, possibly [this Stack Overflow comment](
https://stackoverflow.com/questions/8119138/gson-not-parsing-class-variable/8120175#comment133755647_8120175). Maybe that introduced these characters, but I am not sure (and cannot reproduce it anymore).

As side note: A similar issue existed with the HTML pages generated by Javadoc which used the Zero Width Space for certain line break positions, which you then might have copied by accident (see [JDK-8266779](https://bugs.openjdk.org/browse/JDK-8266779)).

Issue #2310, opened 2023-02-05
java.lang.VerifyError


ylk2534246654:
# Gson version

2.10.1

# Java / Android version

Android 4.0.3

# Description

Error after updating from 2.9.1 to 2.10.1

# Exception stack trace

```
W/System.err: Caused by: java.lang.VerifyError: com/google/gson/internal/reflect/ReflectionHelper
W/System.err:     at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:122)
W/System.err:     at com.google.gson.Gson.getAdapter(Gson.java:556)
W/System.err:     at com.google.gson.internal.bind.CollectionTypeAdapterFactory.create(CollectionTypeAdapterFactory.java:53)
W/System.err:     at com.google.gson.Gson.getAdapter(Gson.java:556)
W/System.err:     at com.google.gson.Gson.fromJson(Gson.java:1226)
W/System.err:     at com.google.gson.Gson.fromJson(Gson.java:1137)
W/System.err:     at com.google.gson.Gson.fromJson(Gson.java:1047)
W/System.err:     at com.google.gson.Gson.fromJson(Gson.java:1014)
```

Marcono1234:
Does this occur on an actual Android device? The comments on #1353 suggest that this might (in the past) have been caused by Google Play Pre-Launch and does not occur on actual devices.

Also Android 4.0.3 is somewhat old; I think there are a few issue reports here where users encountered problems because some vendors included themselves Gson in their Android version, which caused conflicts.

For which device does this occur? Does this occur only for one device or also for other ones (with newer Android versions)? Is this the complete exception stack trace or is there more information about _why_ the verifier rejected the class?

ylk2534246654:
The Android simulator is used.

This problem only occurs in Gson version 2.10.

The problem cannot be found in Android 5.0 and above.

Marcono1234:
Thanks for the answer!

> The Android simulator is used.

Do you mean an Android Emulator started from the IDE (e.g. Android Studio)? Has this issue also appeared on a real device? Is there more information in the exception stack trace why the `VerifyError` occurred?

ylk2534246654:
```
        TypeToken<List<StringBean>> typeToken = new TypeToken<>() {};
        return gson.fromJson(databaseValue,typeToken.getType());
```

```
E/AndroidRuntime: FATAL EXCEPTION: main
[...]

PR #2398, opened 2023-05-27
Updated dependency

vishalsingh2972:
Solves issue #2310 as the issue occurs because the jsr250-api dependency is not compatible with the version of Java that we are using. The 1.0 version of the dependency is only compatible with Java 5 and earlier. The 1.0-20050927.133100 version of the dependency is compatible with Java 6 and later.

### DIFF
--- a/extras/pom.xml
+++ b/extras/pom.xml
@@ -48,7 +48,7 @@
     <dependency>
       <groupId>javax.annotation</groupId>
       <artifactId>jsr250-api</artifactId>
-      <version>1.0</version>
+      <version>1.0-20050927.133100</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>


Marcono1234:
Could you please provide a bit more information about this?

- > is not compatible with the version of Java that we are using

  Which Java version are you using? Are you using Android? If so, does this occur on a real physical Android device or only on an emulator?

- > The 1.0 version of the dependency is only compatible with Java 5 and earlier. The 1.0-20050927.133100 version of the dependency is compatible with Java 6 and later.

  That would surprise me a bit because 1.0-20050927.133100 is two years _older_.

- Why do you think this is related to jsr250-api in the first place? Could you please link to any resources (blog posts, GitHub issues, ...) which makes you think that. Or otherwise could you please create a minimal self-contained example which shows that jsr250-api is causing this.

- The `pom.xml` you are editing belongs to `gson-extras` which is not published to Maven Central. I haven't seen any indication in #2310 that the reporter is using that.

There must for sure be something special in your project setup and the one of #2310, because lots of users are using Gson successfully without any issues. Could you for completeness please also provide the full stack trace of the `VerifyError` you encounter?

PR #2431, opened 2023-07-06
Document minimum Android API level and add workflow to check compatibility

### LINKED ISSUE #2429: Specify minimum Android API level (+ add Animal Sniffer to verify this)
The README only specifies that Java 7 is the minimum version for the latest Gson versions. However, for Android developers this information is not that helpful because Java 7 API was added in multiple different Android API levels. For example `java.lang.ReflectiveOperationException` added in Java 7 was only added for Android API level 19 (Android 4.4), see also https://github.com/google/gson/issues/2310#issuecomment-1616848536.

# Feature description
- The README should explicitly state which Android API level is the minimum
- The build setup should verify that no newer Android APIs are used (respectively that desugaring is supported).
This can be verified using Animal Sniffer, see:
  - https://github.com/google/guava/issues/4005
  - Android Scents for Maven
    (to be verified that these artifacts are safe; but appears they are also used by other projects, e.g. OkHttp ([code](https://github.com/square/okhttp/blob/7cd7f9720dc7ff45663c29b7a15a42d27cc845d4/gradle/libs.versions.toml#L82-L83)))
    - https://sourceforge.net/projects/androidscents/
    - https://search.maven.org/search?q=g:net.sf.androidscents.signature
  - Animal Sniffer signatures, considering desugaring: https://github.com/open-toast/gummy-bears
    In that case would have to explicitly mention desugaring (using the latest Android build tools?) in the README

# Alternatives / workarounds
_none_

Marcono1234:
### Purpose
Resolves #2429

### Description
This is based on the discussion in #2429

- Documents minimum Android API level of 21 for the next Gson version (2.11.0 probably)
For older versions it claims the minimum API level is 19 because Animal Sniffer did not find any violations there for the latest code from `main`. If it turns out that previous versions were actually incompatible with API level 19 we will have to adjust the README.
- Executes Animal Sniffer (as separate GitHub workflow) to verify compatibility

### Checklist

### DIFF
--- a/.github/workflows/check-android-compatibility.yml
+++ b/.github/workflows/check-android-compatibility.yml
@@ -1,0 +1,29 @@
+# For security reasons this is a separate GitHub workflow, see https://github.com/google/gson/issues/2429#issuecomment-1622522842
+# Once https://github.com/mojohaus/animal-sniffer/issues/252 or https://github.com/mojohaus/animal-sniffer/pull/253
+# are resolved, can consider adjusting pom.xml to include this as part of normal Maven build
+
+name: Check Android compatibility
+
+on: [push, pull_request]
+
+permissions:
+  contents: read #  to fetch code (actions/checkout)
+
+jobs:
+  check-android-compatibility:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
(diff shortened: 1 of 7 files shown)


eamonnmcmanus (approved):
Thanks, this looks great!

PR #2434, opened 2023-07-15
Revised "has" method to "containsKey" and added "containsValue" metho…

elevne:
There were circumstances when I had to check if the specific JsonElement object is in JsonObject. 

I found out that there is already method with name "has" and thought it would not be consistent if just adding new method with name "containsValue". I thought that it would be better to change pre-existing method's name to "containsKey", too.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -173,9 +173,17 @@ public final class JsonObject extends JsonElement {
    * @param memberName name of the member that is being checked for presence.
    * @return true if there is a member with the specified name, false otherwise.
    */
-  public boolean has(String memberName) {
+  public boolean containsKey(String memberName) {
     return members.containsKey(memberName);
   }
+
+  /**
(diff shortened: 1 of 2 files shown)


Marcono1234:
Thanks for the pull request!

Unfortunately though we cannot rename the `has` method as it is part of the public API and would break backward compatibility for all users of that method.
The addition of a new `containsValue` method sounds interesting, but it has to be considered if the added functionality really justifies adding a new method to the public API, see also related discussions such as https://github.com/google/gson/pull/1667#issuecomment-1312821320. For your use case the existing method `asMap().containsValue(value)` could be used.

I am not a direct member of this project though, so they might think differently about this.

eamonnmcmanus:
Right, we're certainly not going to rename any existing public methods in the API. Furthermore, really the whole reason for adding `JsonObject.asMap()` and `JsonArray.asList()` was to provide an alternative to adding all the various methods of the `Map` and `List` APIs to `JsonObject` and `JsonArray` respectively.

elevne:
Got it!Thank you for replying! 


-----Original Message-----

PR #2440, opened 2023-07-20
Only create one `BoundField` instance per field in `ReflectiveTypeAdapterFactory`

### LINKED ISSUE #2438: JsonSerializers are constructed multiple times when associated with SerializedName with alternate names but only one is used
# Gson version
2.10.1

# Java / Android version
Java 17

# Used tools
- [ ] Maven; version: 3.8.1

# Description
When a class has a field with : 
 - @SerializedName annotation with alternate names
 - and with @JsonAdapter indicating a JsonSerializer class to use
 
Then a new JsonSerializer will instanciate for each alternate names, even if they will never be used ([serialize](https://github.com/google/gson/blob/6d9c3566b71900c54644a9f71ce028696ee5d4bd/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L167) is always false because of [this](https://github.com/google/gson/blob/6d9c3566b71900c54644a9f71ce028696ee5d4bd/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L301))

## Expected behavior
Instanciante a new JsonSerializer only for the main name, because created one can be an heavy operation.

## Actual behavior
A new JsonSerializer will instanciate for each alternate names

# Reproduction steps
Run the following code
```java
public static class Pojo {

        @SerializedName(value = "field1", alternate = {"field2", "field3"})
        @JsonAdapter(StringSerializer.class)
        public final String str;

        public Pojo(String str) {
            this.str = str;
        }

}

public static class StringSerializer implements JsonSerializer<String> {

        private static int count = 0;

        private final int c;

        public StringSerializer() {
            this.c = count;
            System.out.println("new StringSerializer " + this.c);
            count++;
        }

        @Override
        public JsonElement serialize(String value, Type type, JsonSerializationContext context) {
            System.out.println("serialize " + this.c);
            return new JsonPrimitive(value);
        }

}

        TypeAdapter<Pojo> adapter = new Gson().getAdapter(Pojo.class);
        Pojo pojo1 = new Pojo("my value 1");
        Pojo pojo2 = new Pojo("my value 2");
        adapter.toJson(pojo1);
        adapter.toJson(pojo2);
}
```

# Dev hint
Maybe we can give `serialize` value to [this](https://github.com/google/gson/blob/6d9c3566b71900c54644a9f71ce028696ee5d4bd/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L158) method call and used it in some way ? 

OR

Create [this](https://github.com/google/gson/blob/6d9c3566b71900c54644a9f71ce028696ee5d4bd/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L158) type adapter before [here](https://github.com/google/gson/blob/6d9c3566b71900c54644a9f71ce028696ee5d4bd/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L299)

Marcono1234:
### Purpose
Fixes #2438

### Description
Instead of creating a `BoundField` for every possible name of a field (for `@SerializedName` usage) and then storing for that `BoundField` whether it is serialized or deserialized, instead only create one `BoundField` and then have a separate `Map<String, BoundField>` for deserialized fields, and a separate `List<BoundField>` for serialized fields.

This shouldn't cause any changes visible to the user.

Not sure if this implementation is clean enough; feedback is appreciated!

### Checklist


gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -233,13 +231,35 @@ void readIntoField(JsonReader reader, Object target)
...
+
+    public FieldsData(Map<String, BoundField> deserializedFields, List<BoundField> serializedFields) {
+      this.deserializedFields = deserializedFields;
+      this.serializedFields = serializedFields;
+    }
+  }
+
+  private static IllegalArgumentException createDuplicateFieldException(Class<?> declaringType, String duplicateName, Field field1, Field field2) {
+    throw new IllegalArgumentException("Class " + declaringType.getName()
+    + " declares multiple JSON fields named '" + duplicateName + "'; conflict is caused"
```
eamonnmcmanus, line 249: Would you mind indenting these +4?

eamonnmcmanus (changes requested):
This looks great. I ran it past Google's tests and had no problems. (Except one place where people were accessing the internals of `ReflectiveTypeAdapterFactory` via reflection. I'll update that at the same time as I import these changes.)

Just one trivial request.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
Marcono1234: Thanks for noticing!

eamonnmcmanus (approved):

PR #2442, opened 2023-07-21
Improve `JsonAdapter` documentation and tests

Marcono1234:
### Purpose
Improve `@JsonAdapter` documentation and tests, especially regarding:
- how `@JsonAdapter` creates instances
- effect of `JsonAdapter.nullSafe`

### Checklist


eamonnmcmanus (approved):
Looks good. Thanks!

PR #2447, opened 2023-07-25
Support non-generic type for `TypeToken.getParameterized` for legacy reasons

Marcono1234:
### Purpose
Follow-up for https://github.com/google/gson/pull/2375#discussion_r1273736816

### Description
This partially restores the behavior before a589ef20087b4b0f1ec3048d3ceaef1eedccd09d, except that back then for a non-generic type a bogus `TypeToken(ParameterizedType)` was created, whereas now a `TypeToken(Class)` is created instead.

### Checklist


eamonnmcmanus (approved):
Thanks, that is indeed exactly what I had in mind!

PR #2448, opened 2023-07-26
Update ProGuard default shrinking rules

sgjesse:
This change deals correctly with classes without a no-args constructor. If a no-args constructor was not present the conditional keep rule would not trigger.

See https://issuetracker.google.com/150189783#comment11.

### DIFF
--- a/gson/src/main/resources/META-INF/proguard/gson.pro
+++ b/gson/src/main/resources/META-INF/proguard/gson.pro
@@ -65,6 +65,6 @@
 # See also https://github.com/google/gson/pull/2420#discussion_r1241813541 for a more detailed explanation
 -if class *
 -keepclasseswithmembers,allowobfuscation,allowoptimization class <1> {
-  <init>();
+  <init>(...);
   @com.google.gson.annotations.SerializedName <fields>;
 }


eamonnmcmanus (approved):

eamonnmcmanus (changes requested):
Looks like `ShrinkingIT` needs to be updated?

sgjesse:
Yes, and that makes sense. However, I cannot reproduce locally. If I run `mvn clean verify` I get 7 failures, but not the one on `ShrinkingIT`. What am I missing?

eamonnmcmanus:
> Yes, and that makes sense. However, I cannot reproduce locally. If I run `mvn clean verify` I get 7 failures, but not the one on `ShrinkingIT`. What am I missing?

I was seeing the same thing with JDK 11 on Google's internal Linux. #2450 should fix it. With that PR, I am able to reproduce the test failure in that environment.

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -65,6 +65,6 @@
 # See also https://github.com/google/gson/pull/2420#discussion_r1241813541 for a more detailed explanation
 -if class *
 -keepclasseswithmembers,allowobfuscation,allowoptimization class <1> {
-  <init>();
+  <init>(...);
```
Marcono1234, line 68: The comment above mentioning "and a no-args constructor" has to be adjusted.

Marcono1234:
This case of using `<init>()` vs. `<init>(...)` was what we were discussing in https://github.com/google/gson/pull/2420#discussion_r1238689431 I think.

Out of curiosity, could you please explain a bit why you think it is needed?

In the end it is still Éamonn's decission whether this is included, but I am a bit afraid that with this change we contribute to people unwittingly relying on `Unsafe`, while we would have the chance here to indicate to them that their class is misconfigured (missing a no-args constructor).

Marcono1234:
> If I run `mvn clean verify` I get 7 failures, but not the one on `ShrinkingIT`. What am I missing?

Just to be sure, you run this in the parent directory, right? `ShrinkingIT` is part of a separate Maven module, so it won't be executed if you run inside `/gson/gson`.

sgjesse:
After testing with the new configuration I realized that with just `<init>()` the conditional `-if` rule would not actually match classes _without_ a no-args constructor - which was also what we discussed and decided on. However, for common code patterns data classes will not have a no-args constructor, e.g.:

__Java__
```
public class Data {
  @SerializedName("s")
  private final String s;
  public Data(String s) { this.s = s; }
}
```

__Kotlin__
```
data class Data(
  @field:SerializedName("s") val s: String
)
```
To get no-args constructors requires effectively assigning default values to all fields for these patterns (either directly for the field or assigning in an explicit no-args constructor), which I don't think is feasible to require, and then the provided rules will not work for many developers.

I is still an option for developers to add no-args constructors to not use `sun.misc.Unsafe`, and if present they will still be kept.

sgjesse:
Tests updated, please review.

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -65,6 +65,6 @@
 # See also https://github.com/google/gson/pull/2420#discussion_r1241813541 for a more detailed explanation
 -if class *
```
sfreilich, line 66: I think this `if class *`, `-classeswithmembers... class <1>` doesn't do anything relative to `-classeswithmembers... class *` , after reading some of the follow-up threads on that change.

sfreilich (changes requested):
Three other issues stand out to me here:
* There are two rules about classes with `@SerializedName` fields, which I think could be consolidated into one `-keepclasseswithmembers` rule?
* More of these `-keep`/`-keepclassmembers` declarations could use `allowobfuscation`, which should work fine with reflection that refers to things by symbol instead of string name (including these annotations).
* Possibly this should have `-keep,allowobfuscation class` rules for the annotations used here. I think that can matter when there are multiple optimization passes? Not 100% sure, basing this on other examples that follow that pattern. Not sure if that would be best to do with a rule that matches `com.google.gson.annotations.*` or just `JsonAdapter` and `SerializedName` specifically.

sfreilich:
> Possibly this should have `-keep,allowobfuscation` class rules for the annotations used here.

Realized this bit probably doesn't matter, I was confused by examples for annotations that are just referenced by proguard config rules, and not actually otherwise referenced in the code.

Marcono1234 (changes requested):
@sgjesse, hmmm ok, let's hope then that those users don't encounter the pitfalls of JDK Unsafe (see documentation of [`GsonBuilder.disableJdkUnsafe()`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/GsonBuilder.html#disableJdkUnsafe())) (and also let's hope that neither the JDK nor Android remove the needed JDK Unsafe functionality any time soon).

<details>

<summary>Personal opinion on this (click to show)</summary>

Personally I am a bit indecisive on these changes; on one hand I can definitely see that it would be convenient for users if it just worked automatically. On the other hand, the problem is that JDK Unsafe usage is opt-out and I assume most users are not aware of it being used at all, nor what the implications of it are (if it was opt-in I wouldn't have a problem with that). If as user you want to use your classes with Gson, then you should (in my opinion) adjust them to follow Gson's requirements (or if necessary differentiate between application logic classes and DTO classes, or however you want to call them), and from that perspective I would appreciate any changes which make users less dependent on JDK Unsafe.

Though I am not a direct member of this project, so it is not up to me to decide this.

----

</details>

Thanks for the test changes, could you please adjust the comments which are now partially obsolete?

Also, ideally there would be a new test in `shrinker-test`'s `com.example.Main` which covers the successful case of a class without no-args constructor but with fields annotated with `@SerializedName`, which is now supported with your changes.
If you don't want to do that I can do it in a follow-up PR.

Technically `Troubleshooting.md` is now also a bit outdated (e.g. it says "has a no-args constructor"), but I would leave it that way for now to encourage usage of a no-args constructor to avoid relying on JDK Unsafe.

gson/src/main/resources/META-INF/proguard/gson.pro
Marcono1234: At least for R8 it does seem to make a difference: `-keepclasseswithmembers ... class *` always keeps the class, even if it isn't used. Whereas `-if class *  ... -keepclasseswithmembers ... class <1>` only keeps the class if it is actually used.

Unfortunately this is not covered by the tests yet; I will try to add a new test for that.

Marcono1234: Thanks also for the other points you raised, but unless sgjesse wants to address them here as well could you please create a separate GitHub issue or pull request? Otherwise the discussion on this pull request here might drift too far away from the original topic.

sfreilich:
I'll file a PR for whatever @sgjesse doesn't want to handle here. We were just discussing that since I was trying to solve some obstacles to adopting these rules downstream, and we're colleagues, we share a downstream.

Marcono1234:
@sgjesse, I have tried to address my review comments in https://github.com/sgjesse/gson/pull/1 now; I hope that is easier for you. That pull request targets your `fix-proguard-rules` branch, so in case you merge it, the changes should show up here as well.

shrinker-test/proguard.pro
```diff
@@ -15,3 +15,6 @@
 -keepclassmembernames class com.example.NoSerializedNameMain$TestClassWithoutDefaultConstructor {
   <fields>;
 }
+#-keep class com.example.ClassWithSerializedName {
+#  <init>(...);
+#}
```
Marcono1234, line 20: Is this needed?

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -59,13 +54,16 @@
...
 }
 
-# If a class is used in some way by the application and has fields annotated with @SerializedName,
-# keep those fields and the constructors of the class
-# For convenience this also matches classes without no-args constructor, in which case Gson uses JDK Unsafe
-# Based on https://issuetracker.google.com/issues/150189783#comment11
-# See also https://github.com/google/gson/pull/2420#discussion_r1241813541 for a more detailed explanation
+# Keep fields annotated with @SerializedName for classes which are present.
+# If classes with fields annotated with @SerializedName have a no-args
+# constructor keep that as well.
```
Marcono1234, line 59: Why did you remove the link to the pull request discussion?

At least in my opinion this `-if class * ...` is pretty cryptic and its intended behavior is not directly obvious, so it would be good to have a reference explaining what this actually does.

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -59,13 +54,16 @@
...
 -if class *
--keepclasseswithmembers,allowobfuscation,allowoptimization class <1> {
```
Marcono1234, line 67: Why did you remove `allowoptimization`?

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -59,13 +54,16 @@
...
-  <init>(...);
```
Marcono1234: Is it intentional that you removed `<init>(...)` again and used `<init>()` below? My personal opinion in https://github.com/google/gson/pull/2448#pullrequestreview-1553268244 was not blocking the merge of this pull request (sorry if I did not make that clear enough).

This change to use `<init>()` again might render some or all of the shrinker test changes obsolete.

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -15,25 +15,20 @@
...
 # Keep any (anonymous) classes extending TypeToken
 -keep,allowobfuscation class * extends com.google.gson.reflect.TypeToken
 
 # Keep classes with @JsonAdapter annotation
 -keep,allowobfuscation,allowoptimization @com.google.gson.annotations.JsonAdapter class *
 
-# Keep fields with @SerializedName annotation, but allow obfuscation of their names
--keepclassmembers,allowobfuscation class * {
-  @com.google.gson.annotations.SerializedName <fields>;
-}
```
Marcono1234, line 35: Out of curiosity, why did you remove this here and add it below guarded by a `-if class *`? Does that make a difference for the end result?

sgjesse: Yes, in R8 full mode we have a distinction between referenced and instantiated classes. For classes which are not seen by R8 as instantiated (e.g. if only referenced statically using `.class` as one will do in `fromJson` then `-keepclassmembers` will only keep static members and not instance members. For ProGuard and R8 in compatability mode the two will do the same.

gson/src/main/resources/META-INF/proguard/gson.pro
sgjesse: Good point, added those references back.

gson/src/main/resources/META-INF/proguard/gson.pro
sgjesse: Having `allowoptimization` can allow R8 to propagate field values, so if you have an explicit construction of a serialized class always setting a field to a specific value, then that single value can become the default value for the created instances instead of `0`/`null`.

shrinker-test/proguard.pro
sgjesse: Good catch, no removed the rules in comments. Also removed the rule above, as that was also not needed. I probably added it while testing different rules.

sgjesse (commented):
Addressed all comments.

Marcono1234 (commented):
Thanks for your update and also for your work on this!

Though it looks like some of the changes in your last commit might not have been intentional (see comments)?

Also, I think https://github.com/google/gson/pull/2448#discussion_r1292560101 is still open; just to clarify that nothing was overlooked.

gson/src/main/resources/META-INF/proguard/gson.pro
Marcono1234: Or at least I am a bit surprised that no exception occurs for `ClassWithoutDefaultConstructor`; I assume it might be related to your newly added `-keepclasseswithmembers`.

So, do I understand it correctly that your new approach of using `-keepclasseswithmembers` achieves the same thing as your initial `<init>(...)` approach (i.e. Gson can instantiate classes without no-args constructor using Unsafe; instead of failing because R8 made the class abstract). But this new approach is better than the `<init>(...)` rule because that would have unnecessarily kept the constructors with args around, even though they are not actually used by Gson?

gson/src/main/resources/META-INF/proguard/gson.pro
Marcono1234: Hmm thanks, ok that sounds like the previous rules would not have worked properly then, though I think the shrinker-test showed that it did work as expected?
I am bit confused, or am I misunderstanding your comment or are the shrinker-tests flawed and not close enough to real code?

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -54,16 +55,19 @@
...
+# for a more detailed explanation.
 -if class *
 -keepclasseswithmembers,allowobfuscation class <1> {
   @com.google.gson.annotations.SerializedName <fields>;
 }
 -if class * {
   @com.google.gson.annotations.SerializedName <fields>;
 }
--keepclassmembers,allowobfuscation class <1> {
+-keepclassmembers,allowobfuscation,allowoptimization class <1> {
```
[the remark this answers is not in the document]
sgjesse: This rule is only covering the no-args constructor (`<init>()`), so here there is no issue with `allowoptimization`. For the rule above for the annotated fields it should not be there as otherwise R8 can do value propagation.

shrinker-test/common.pro
```diff
@@ -38,3 +38,5 @@
 -keepclassmembernames class com.example.ClassWithVersionAnnotations {
   <fields>;
 }
+
+-dontobfuscate
```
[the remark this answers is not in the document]
sgjesse: No, this was all for some testing. These changes should not have been there.

sgjesse:
@Marcono1234 Thank you for being such an observant reviewer! Addressed my mistakes in my previous upload. Please take another look.

gson/src/main/resources/META-INF/proguard/gson.pro
```diff
@@ -22,6 +22,7 @@
 # Keep class TypeToken (respectively its generic signature) if present
 -if class com.google.gson.reflect.TypeToken
 -keep,allowobfuscation class com.google.gson.reflect.TypeToken
+#-keep class com.google.gson.reflect.TypeToken { *; }
```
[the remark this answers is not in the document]
sgjesse: No removed.

gson/src/main/resources/META-INF/proguard/gson.pro
sgjesse: One way to see the difference is to have an instance field annotated with `@SerializedName` which is not used in the code at all, then the field will be removed with the previous rule in R8 full mode (which is default from AGP 8.0.0). Then in a pass through from `fromJson` to `toJson` the field value will be lost. Of course it is a matter of taste if unused fields should be kept this way.

eamonnmcmanus (approved):
I think this is close enough now. If there are further changes needed, we can make them in a follow-up PR.

Marcono1234:
> @Marcono1234 Thank you for being such an observant reviewer!

Thank you for all the work you have put into this, and all the detailed explanations! That was helpful for me to understand at least a bit better how R8 works. And hopefully it will also be useful for future reference.

The changes look good, so I don't think there is any direct follow-up PR needed.

guyslumclubxx (approved):

PR #2450, opened 2023-07-26
Make date-formatting tests less fragile with regular expressions.

eamonnmcmanus:
This is not great. We should really ensure that formatted dates are the same regardless of JDK version. There is code that attempts to do that but it is not really effective. So for we use regular expressions to paper over the differences.

### Checklist


Marcono1234 (approved):
Thanks, looks good!

For future reference, for which JDK version did you encounter the "Narrow No Break Space"?

eamonnmcmanus:
> For future reference, for which JDK version did you encounter the "Narrow No Break Space" in front of `AM` / `PM`?

It was Google's JDK 11, which might have some localization differences from regular JDK 11. (I don't know, but that's my theory.)

PR #2452, opened 2023-07-26
Throw exception when using classes from other JSON libraries with Gson

### LINKED ISSUE #2445: Throw exception when using classes from other JSON libraries with Gson
Quite frequently users seem to accidentally use JSON classes from other libraries with Gson, such as JSON-java's `JSONObject`. This leads to incorrect results because Gson will then fall back to using reflection on these classes.

Recent examples:
- https://stackoverflow.com/q/76706456
- https://stackoverflow.com/q/76755061
- https://stackoverflow.com/q/76756548

# Feature description
Register a default `TypeAdapter` (or `TypeAdapterFactory`) which checks if the class is from a different popular JSON library and in the case throw an exception.
- The exception should only be thrown in the `read` and `write` methods of `TypeAdapter` to minimize the risk of backward compatibility when users don't actually serialize or deserialize the instances
- The adapter should act as fallback (internally registered right before the reflection-based factory); a user defined adapter should have higher precedence
- The exception message mention that the class is unsupported and the Gson equivalent should be used; optionally it should also point to a new troubleshooting guide entry

Classes which should be detected:
- JSON-java (these are also the ones available for Android)
- Jackson
- Vert.x ?
- json-simple ?
- ...?

Note that would break backward compatibility for applications which (accidentally) rely on this.

# Alternatives / workarounds
- Do nothing
- Add default adapters which support third-party JSON classes; probably not something we want to do?

Marcono1234:
### Purpose
Resolves #2445

### Description
See #2445 for the rationale.

With these changes an exception is thrown when users try to serialize or deserialize unsupported JSON classes from other libraries.

Currently detected libraries:
- JSON-java / Android (`org.json.JSONArray`, `org.json.JSONObject`)
- Vert.x

Also a new Troubleshooting Guide entry has been added explaining the issue and for JSON-java / Android suggesting backward compatible `TypeAdapterFactory` implementations.

### Checklist

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -49,6 +49,14 @@
       <version>2.20.0</version>
     </dependency>
 
+    <!-- For testing interoperability with other JSON libraries -->
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20230618</version>
+      <scope>test</scope>
+    </dependency>
+
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -53,7 +53,9 @@
     <dependency>
       <groupId>org.json</groupId>
       <artifactId>json</artifactId>
-      <version>20230618</version>
+      <!-- Uses an ancient version because it matches roughly the API of the same classes from Android
+        See also https://stackoverflow.com/q/39564936 -->
+      <version>20090211</version>
       <scope>test</scope>
     </dependency>
 
--- a/gson/src/test/java/com/google/gson/functional/JsonOrgInteropTest.java
+++ b/gson/src/test/java/com/google/gson/functional/JsonOrgInteropTest.java
@@ -1,0 +1,568 @@
+package com.google.gson.functional;
+
+import static com.google.common.truth.Truth.assertAbout;
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
+
+import com.google.common.truth.FailureMetadata;
+import com.google.common.truth.StandardSubjectBuilder;
+import com.google.common.truth.Subject;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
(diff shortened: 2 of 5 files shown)


gson/pom.xml
```diff
@@ -49,6 +49,16 @@
       <version>2.20.0</version>
     </dependency>
 
+    <!-- For testing interoperability with other JSON libraries -->
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <!-- Uses an ancient version because it matches roughly the API of the same classes from Android
+        See also https://stackoverflow.com/q/39564936 -->
+      <version>20090211</version>
```
Marcono1234, line 58: I am not very happy about using such an outdated test dependency, especially because that version does not even have source code or Javadoc available, but it matches the Android API closely and therefore makes it easier to verify that the code in the `JsonOrgInteropTest` test class (which is also shown in the Troubleshooting Guide) compiles for Android as well.

gson/src/test/java/com/google/gson/functional/JsonOrgInteropTest.java
```diff
@@ -0,0 +1,589 @@
...
+        builder.about(JSONArraySubject.jsonArrays()).that((JSONArray) actual).isEqualTo(expected);
+      } else if (actual instanceof JSONObject) {
+        builder.about(JSONObjectSubject.jsonObjects()).that((JSONObject) actual).isEqualTo(expected);
+      } else {
+        builder.that(actual).isEqualTo(expected);
+      }
+    }
+  }
+
+  private static class JSONArraySubject extends JsonOrgBaseSubject {
```
Marcono1234, line 112: Maybe having these custom Truth `Subject` classes is a bit exaggerated, and they might not follow Truth best practices (feedback is welcome). Though on the other hand they make comparison of `JSONArray` and `JSONObject` easier.

Marcono1234:
The code suggested in the troubleshooting guide to keep backward compatibility with the previous reflection-based output for `JSONArray` and `JSONObject` might be a bit too verbose. Maybe it should be removed (also from the unit test).

If necessary we could possibly consider adding a system property for now which still allows users to use reflection on those classes, but mention that the system property will be removed in future Gson versions.

PR #2455, opened 2023-07-29
Add test for unused class to shrinker-test

Marcono1234:
### Purpose
Extends tests of `shrinker-test` for unused classes

### Description
Based on #2448

Adds a test case for a class with a field annotated with `@SerializedName`, but which is not actually used anywhere. The default Gson ProGuard rules should allow this class to be removed.

Note: This actually works as intended for R8, but not for ProGuard which keeps the class. Not sure if it is worth investigating that further at the moment.

### Checklist

### DIFF
--- a/shrinker-test/common.pro
+++ b/shrinker-test/common.pro
@@ -38,3 +38,7 @@
 -keepclassmembernames class com.example.ClassWithVersionAnnotations {
   <fields>;
 }
+
+# Keep the name of the class to allow using reflection to check if this class still exists
+# after shrinking
+-keepnames class com.example.UnusedClass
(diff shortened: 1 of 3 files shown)


eamonnmcmanus (approved):

Marcono1234:
@eamonnmcmanus, just to avoid any confusion, this does not address the open points I mentioned in  #2448 regarding updating comments and adding a separate test for a class without no-args constructor.

This here adds a different test case, and I simply decided to base it on #2448 to reduce merge conflicts later.

eamonnmcmanus:
Thanks, still looks good!

PR #2465, opened 2023-08-10
Access `RecordComponent` via `Class.forName`.

eamonnmcmanus:
This should mean that GraalVM will understand the reflective lookup of its methods. See [documentation](https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/Reflection/#automatic-detection).

Fixes #2464.

PR #2466, opened 2023-08-12
Add integration test for GraalVM Native Image

Marcono1234:
### Purpose
Adds an integration test module for GaalVM Native Image, see also https://github.com/google/gson/pull/2465#issuecomment-1675164426

### Description
The integration test is a separate Maven module and uses the [GraalVM Native Maven Plugin](https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#agent-support-running-tests) for test execution. Because Native Image tests require GraalVM as JDK, the Native Image test execution is opt-in (but run by the GitHub build workflow by default). By default the tests are only executed as regular unit tests.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,3 +24,22 @@ jobs:
       - name: Build with Maven
         # This also runs javadoc:jar to detect any issues with the Javadoc generated during release
         run: mvn --batch-mode --update-snapshots --no-transfer-progress verify javadoc:jar
+
+  native-image-test:
+    name: "GraalVM Native Image test"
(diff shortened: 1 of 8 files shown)


Marcono1234:
@eamonnmcmanus, I hope the unit tests in the `graal-native-image-test` suffice for now, but feel free to suggest changes or additions to them.
And of course any other feedback regarding this pull request is welcome as well!

eamonnmcmanus (approved):

eamonnmcmanus:
This is amazing! Thanks so much for putting it together. We might end up refining it as we discover corner cases but I think it's already an excellent starting point.

PR #2478, opened 2023-08-25
Add shrinker-test examples with never-referenced constructors

sfreilich:
R8 can do some optimizations that assume values of types whose constructors are never referenced are `null`. (A bit confused why the rules here are sufficient to avoid that optimization in the case of a class whose sole constructor has arguments and is unreferenced, since the keep rule only refers to `<init>()` instead of `<init>(...)`. Does R8 have special behavior for `<init>()` where attempting to keep that makes it assume a class is constructable even if it doesn't have a no-args constructor?)

Also rename some test classes/names to refer to `NoArgs` instead of `Default` constructors. The examples here care about whether a no-argument constructor is present. While a no-argument constructor (explicitly defined or not) is used by default in some circumstances, the Java documentation consistently uses "default constructor" to refer only to constructors whose implementation is implicitly provided by the complier (all default constructors are no-argument constructors, but not vice versa).

### DIFF
--- a/shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
+++ b/shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
@@ -49,7 +49,7 @@ public class NoSerializedNameMain {
   }
 
   /**
-   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_ConstructorHasArgs()}.
+   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_HasArgsConstructor()}.
    */
   public static String runTestHasArgsConstructor() {
     TestClassHasArgsConstructor deserialized = new Gson().fromJson(
--- a/shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
+++ b/shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
@@ -10,7 +10,8 @@ import com.google.gson.GsonBuilder;
  * therefore not matched by the default {@code gson.pro} rules.
  */
 public class NoSerializedNameMain {
-  static class TestClass {
+  static class TestClassNoArgsConstructor {
+    // Has a no-args default constructor.
     public String s;
   }
 
@@ -19,11 +20,11 @@ public class NoSerializedNameMain {
     public String s;
   }
 
-  static class TestClassConstructorHasArgs {
+  static class TestClassHasArgsConstructor {
     public String s;
 
     // Specify explicit constructor with args to remove implicit no-args default constructor
-    public TestClassConstructorHasArgs(String s) {
+    public TestClassHasArgsConstructor(String s) {
       this.s = s;
     }
   }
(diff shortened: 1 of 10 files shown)


Marcono1234 (commented):
Thanks a lot! The changes look good, just a few small things (see comments below).

> R8 can do some optimizations that assume values of types whose constructors are never referenced are `null`

I think by accident I triggered this for ProGuard while initially creating the shrinker-tests, respectively I triggered the [cryptic `NullPointerException` this is causing](https://github.com/Guardsquare/proguard/blob/57d425046493723f2556e27602cdd579cfb963ad/base/src/main/java/proguard/optimize/peephole/NoConstructorReferenceReplacer.java#L99-L104) (with JVM's helpful `NullPointerException` messages it says something like 'Cannot throw exception because "null" is null'). Though I am not sure if I still have the code to cause this, and if it would still cause this with the latest `gson.pro` file.

> A bit confused why the rules here are sufficient to avoid that optimization...

I am not sure either, but I am also not very familiar with the details of R8.

> Also rename some test classes/names to refer to `NoArgs` instead of `Default` constructors

Good point! You are right, the [JLS seems to use the term "default constructor" only for the implicit one](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.8.9). I was mixing the terms a bit inconsistently.

shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
```diff
@@ -19,25 +19,25 @@ static class TestClassNotAbstract {
...
     // Specify explicit constructor with args to remove implicit no-args default constructor
-    public TestClassWithoutDefaultConstructor(String s) {
+    public TestClassConstructorHasArgs(String s) {
       this.s = s;
     }
   }
 
   /**
-   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_DefaultConstructor()}.
+   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_ConstructorHasArgs()}.
```
Marcono1234, line 32: I think this should be
```suggestion
   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_NoArgsConstructor()}.
```

Marcono1234, line 32: Or actually, maybe this specific instance should not have been renamed because `TestClass` does actually have an implicit default constructor.

sfreilich: I think I like NoArgsConstructor better because the code doesn't really care if it's default or not.

sfreilich: Adjusted the test method/class names for consistency.

shrinker-test/src/main/java/com/example/NoSerializedNameMain.java
```diff
@@ -19,37 +20,40 @@ static class TestClassNotAbstract {
...
     Gson gson = new GsonBuilder().disableJdkUnsafe().create();
-    TestClassNotAbstract deserialized = gson.fromJson("{\"s\": \"value\"}", same(TestClassNotAbstract.class));
+    TestClassNotAbstract deserialized = gson.fromJson(
+        "{\"s\": \"value\"}", same(TestClassNotAbstract.class));
     return deserialized.s;
   }
 
   /**
-   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_NoDefaultConstructor()}.
+   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_ConstructorHasArgs()}.
```
Marcono1234, line 52: The method name seems to be incorrect
```suggestion
   * Main entrypoint, called by {@code ShrinkingIT.testNoSerializedName_HasArgsConstructor()}.
```

sfreilich: Fixed, thanks.

Marcono1234 (approved):
Thanks a lot for the changes, they look good to me!

eamonnmcmanus (approved):

Issue #1511, opened 2019-04-11
ISO8601Utils assumes local timezone while parsing "yyyy-MM-dd" dates


roman-tikhonov:
Parsing a "yyyy-MM-dd" date without a timezone using

```
ISO8601Utils.parse(dateString, ParsePosition(0)).toInstant()
```

applies a timezone from a default locale which leads to an incorrect result.

Input: `2019-03-20`
Actual Output: `2019-03-19T23:00:00Z` when a system default locale is UTC+01
Expected Output: `2019-03-20T00:00:00Z` regardless of the system locale

I think the problem is that `GregorianCalendar` must be preconfigured with the UTC timezone, and it's not done in `ISO8601Utils`:
```
// if the value has no time component (and no time zone), we are done
            boolean hasT = checkOffset(date, offset, 'T');
            
            if (!hasT && (date.length() <= offset)) {
                Calendar calendar = new GregorianCalendar(year, month - 1, day);

                pos.setIndex(offset);
                return calendar.getTime();
            }
```

paulzin:
Any updates on this?

PR #2480, opened 2023-08-26
Update ISO8601Utils parser to default to UTC timezone when the value is not provided

HyperspaceAlmanac:
### Purpose

This is to create a fix for https://github.com/google/gson/issues/1511 which is marked as a bug. Currently the parser defaults to the local timezone when the value is not provided, but the issue marked as a bug specifies that it should instead parse the time using UTC timezone.

### Description

This is a fix for https://github.com/google/gson/issues/1511

There is existing unit test that checks the parsed time against local timezone. I updated that unit test to instead compare against time generated using UTC time zone.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
@@ -170,6 +170,7 @@ public class ISO8601Utils
 
             if (!hasT && (date.length() <= offset)) {
                 Calendar calendar = new GregorianCalendar(year, month - 1, day);
+                calendar.setTimeZone(TIMEZONE_UTC);
                 calendar.setLenient(false);
 
                 pos.setIndex(offset);
--- a/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
@@ -78,10 +78,12 @@ public class ISO8601UtilsTest {
 
   @Test
   @SuppressWarnings("UndefinedEquals")
-  public void testDateParseWithDefaultTimezone() throws ParseException {
+  public void testDateParseDefaultsToUTC() throws ParseException {
         String dateStr = "2018-06-25";
         Date date = ISO8601Utils.parse(dateStr, new ParsePosition(0));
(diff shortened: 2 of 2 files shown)


Marcono1234:
Thanks a lot for you pull request!

I am not quite sure what the correct behavior here would be. On one hand relying implicitly on system default values (such as the default timezone here) is most often not ideal and can lead to unexpected behavior and bugs. But on the other hand maybe in this case using the default timezone can be desired in some cases:

If we try to compare this with the `java.time` API, then
- the pattern `yyyy-MM-dd` is equivalent to `java.time.LocalDate`
- `java.util.Date` is equivalent to `java.time.Instant`

For `java.time` conversion from `LocalDate` to `Instant` explicitly requires specifying a timezone.

Since here for Gson the conversion happens implicitly, I guess the desired behavior might be application-specific, and there might be cases where the default timezone is desired.
This is only my personal opinion on this; and I don't have any specific examples at the moment for either of those cases.

Sorry if the "bug" label on the GitHub issue you mentioned gave a false impression. I added the label back then while trying to clean up stale and resolved GitHub issues. Similar to new GitHub issues which automatically get the label based on which issue type the reporter selects, it is mainly intended to indicate that the reporter considers this a bug, but not necessarily that this is "acknowledged" and will definitely be fixed.

Marcono1234:
In https://github.com/FasterXML/jackson-databind/issues/816 where this date format was requested for Jackson (and where this Gson class is coming from), the reporter explicitly said:
> ISO8601Utils.parse() should be modified to accept a date-only string (YYYY-MM-DD) with no time zone and return the value as a date corresponding to midnight on that day in the local (default) time zone.

So it seems there are indeed cases where the default time zone is desired.

Marcono1234:
@eamonnmcmanus what do you think?

eamonnmcmanus:
The state of date/time parsing in Gson is indeed not great. One of the reasons it's not great is that we can't easily change things without breaking deployments that were depending on the previous behaviour. That's the case here: people almost certainly are depending on the way this works currently. Changing it could cause things to fail in very subtle ways. We could conceivably have some sort of non-default configuration option to switch to UTC, but we can't just switch unconditionally.

I think it would be worth undertaking a significant effort to improve the date/time situation. Just fixing this one thing in isolation isn't going to help much.

PR #2505, opened 2023-10-03
Improve formatting

Marcono1234:
### Purpose
Fixes a few cases of wrong indentation and missing whitespace.

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -695,8 +695,8 @@ public final class GsonBuilder {
   }
 
   private boolean isTypeObjectOrJsonElement(Type type) {
-    return type instanceof Class &&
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -696,8 +696,8 @@ public GsonBuilder registerTypeAdapter(Type type, Object typeAdapter) {
 
   private boolean isTypeObjectOrJsonElement(Type type) {
     return type instanceof Class &&
-            (type == Object.class
-                    || JsonElement.class.isAssignableFrom((Class<?>) type));
+        (type == Object.class
```
eamonnmcmanus, line 699: Not very important, but if you wanted to follow Google Java Style here, the `&&` should be at the start of this line.

eamonnmcmanus (approved):

PR #2524, opened 2023-11-04
`Spotless` formatting

MaicolAntali:
### Purpose
Adds the `Spotless` plugin to check that formatting follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)

### Description
Related with #2284

⚠️ This is a daft PR to discuss the configuration and how run it (in a new workflow or use the `build` workflow). 
_(I have applied some default configurations)_

### Checklist


pom.xml
```diff
@@ -353,6 +358,38 @@
...
+            <formats>
+              <format>
+                <includes>
+                  <include>*.md</include>
+                  <include>.gitignore</include>
+                </includes>
+                <trimTrailingWhitespace/>
+                <endWithNewline/>
+                <indent>
+                  <tabs>true</tabs>
```
eamonnmcmanus, line 376: I don't think we want tabs.

pom.xml
```diff
@@ -353,6 +358,38 @@
...
+                </indent>
+              </format>
+            </formats>
+            <java>
+              <googleJavaFormat>
+                <style>GOOGLE</style>
+                <reflowLongStrings>true</reflowLongStrings>
+                <formatJavadoc>true</formatJavadoc>
+              </googleJavaFormat>
+              <importOrder />             <!-- Checks imports order -->
```
eamonnmcmanus, line 387: I would be inclined to omit the comments on these lines.

pom.xml
```diff
@@ -353,6 +358,38 @@
...
+              <removeUnusedImports />     <!-- Removes unused imports -->
+              <formatAnnotations />       <!-- Formats the annotations -->
```
eamonnmcmanus, line 389: ...except here, where if I'm reading the documentation correctly this is specifically about putting type annotations immediately before the type, rather than on the preceding line.

eamonnmcmanus (commented):
Thanks, this is very helpful! Just a few comments. If `mvn verify` passes then I think this should be good.

MaicolAntali:
> If `mvn verify` passes then I think this should be good.

I thought the same thing, also because the [documentation](https://github.com/diffplug/spotless/tree/main/plugin-maven#binding-to-maven-phase) says: 
> By default, `spotless: check` is bound to verify maven phase.

Until, I executed:`mvn spotless:check` and some error came out. I don’t know why they don’t come out with `mvn check` 🤔

pom.xml
MaicolAntali: Replaced with `<spaces>true</spaces>`

pom.xml
MaicolAntali: I’m not sure if I understood correctly, I removed the comments on `<importOrder />` and `<removeUnusedImports />`

eamonnmcmanus: I'm suggesting that the comment could read something like `<!-- Puts type annotations immediately before types. -->`. I'm not even sure that's necessary, because google-java-format may do it anyway, but it's what the documentation [says](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#formatAnnotations).

eamonnmcmanus:
> Until, I executed:`mvn spotless:check` and some error came out. I don’t know why they don’t come out with `mvn check` 🤔

Well, can you temporarily introduce a deliberate formatting error into the PR, and verify that the CI fails?

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -62,6 +62,9 @@
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicLongArray;
 
+// Temporary unused import
+import java.math.RoundingMode;
+
```
MaicolAntali, line 67: Temporary unused import ❗

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -147,7 +150,13 @@
...
  * @author Jesse Wilson
  */
-public final class Gson {
+
+// Temporary formatting issue
+public
+final
+class Gson
+
+{
```
MaicolAntali, line 159: Temporary unused import ❗

MaicolAntali:
For the first time, I'm happy that CI fails 🤣 

I will run `mvn spotless:apply`, check that every changes is correct and then I'll commit the changes.

pom.xml
```diff
@@ -92,6 +92,11 @@
         <artifactId>truth</artifactId>
         <version>1.1.5</version>
       </dependency>
+      <dependency>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <version>2.40.0</version>
+      </dependency>
```
Marcono1234, line 99: Is this used? I don't think it has an effect on the plugin configured below.

pom.xml
```diff
@@ -124,6 +129,45 @@
...
+                <include>.gitignore</include>
+              </includes>
+              <trimTrailingWhitespace/>
+              <endWithNewline/>
+              <indent>
+                <spaces>true</spaces>
+                <spacesPerTab>4</spacesPerTab>
+              </indent>
+            </format>
+          </formats>
```
Marcono1234, line 158: Is this `<formats>` section needed? If I understand it correctly that would only apply to `*.md` and `.gitignore` the way it is currently configured; and I am not sure if we should impose any formatting rules on them.

MaicolAntali: Formatting rules are very weak on`. md` and `.gitignore`. They replace tab characters with 4 spaces and trim trailing whitespace. 

I think they are good general rules, but if you prefer to remove them, let me know.

pom.xml
MaicolAntali: The official [docs](https://github.com/diffplug/spotless/tree/main/plugin-maven#quickstart) say:
> To use it in your pom, just [add the Spotless dependency](https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.diffplug.spotless%22%20AND%20a%3A%22spotless-maven-plugin%22), and configure it like so:

but you’re right, I removed it and everything still works. Good catch!

MaicolAntali:
I'm not sure why the `Build` fails. 
If I run `mvn clean verify javadoc:jar` on my machine, everything works fine

eamonnmcmanus:
It's great to see this progress!

A couple of things. First, I think reformatting the existing files should be a separate PR from the one that adds the Spotless plugin. Second, I'm pretty sure the CI failure is because `Java17RecordTest` uses Java `record`, and in the JDK 11 configuration that doesn't work. You could reasonably exclude that one file from formatting.

MaicolAntali:
Make sense, I reverted the formatting commits and I will move it in a new PR (when this PR will be merged, so I can edit the `pom.xml` to exclude `Java17RecordTest`)

eamonnmcmanus:
> Make sense, I reverted the formatting commits and I will move it in a new PR (when this PR will be merged, so I can edit the `pom.xml` to exclude `Java17RecordTest`)

I think you need to merge that PR first, so that you won't get errors from this one?

MaicolAntali:
All CI passed, but I think is better has a more tidy commits history.
Tomorrow:
-  I will move [edde156](https://github.com/google/gson/pull/2524/commits/edde1561529313934f58fc26a0b0d82d1eac3f7b) commit in a single PR
- I'll create a new PR to apply the `spotless` plugin.

And I'll close this PR.
I hope that will help to obtain a more tidy and clear commits history

pom.xml
Marcono1234: Ok, I just asked because this section looked like the section from the Spotless examples and it isn't needed for Java formatting.

The combination of `*.md` with `trimTrailingWhitespace` should be removed though because they can break Markdown formatting. Arguably the Spotless examples are also flawed (https://github.com/diffplug/spotless/issues/1877).

Marcono1234:
@eamonnmcmanus if possible please wait a bit with merging this (or any related formatting PR). #2531 'broke' some formatting of Markdown files which intentionally used two trailing spaces for [hard line breaks](https://spec.commonmark.org/0.30/#hard-line-breaks), and in theory someone would have to review all the 229 files changed by the last PR (in case you haven't done so already).

I can try to revert the Markdown formatting changes (and possibly any other unintentional changes), but only in the next days.

pom.xml
Marcono1234: I assume the wording there was just a bit misleading; have created https://github.com/diffplug/spotless/pull/1878 to change that. Let's see what the maintainers of Spotless think.

eamonnmcmanus:
> I can try to revert the Markdown formatting changes (and possibly any other unintentional changes), but only in the next days.

I can do that. I will refrain from sharing my thoughts on that particular Markdown syntax, except to say that the alternative syntax with `\` seems preferable.

Marcono1234:
> I can do that. I will refrain from sharing my thoughts on that particular Markdown syntax, except to say that the alternative syntax with `\` seems preferable.

Thanks! I completely overlooked in the specification that a trailing `\` achieves that as well; that is certainly better than the invisible trailing spaces.

Sorry that my previous comment was not that kind. I think it is great though that you both are working on making the formatting of this project consistent!

MaicolAntali:
This PR is now replaced by #2536 and #2537

Issue #2472, opened 2023-08-17
Change default `Date` serialization to ISO 8601 format


Marcono1234:
# Problem solved by the feature
The current default `java.util.Date` serialization uses `DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)` ([source line](https://github.com/google/gson/blob/gson-parent-2.10.1/gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java#L61)), which produces output like the following:
```json
"Jan 1, 1970, 1:00:00 AM"
```

The problem is that this is a locale dependent format intended to be human-readable, and not necessarily to be parsed by computers. The formatting is also dependent on the CLDR data, which changes between Java versions. This has happened for Java 9 already (#1210), and possibly again for Java 20 respectively also older versions using updated CLDR data (https://github.com/google/gson/pull/2450#issuecomment-1652362725, and [JDK-8304925](https://bugs.openjdk.org/browse/JDK-8304925)).

This leads to backward compatibility issues (#1682), and makes this format unsuitable for transferring data to other applications or services. However, users might not even be aware of these pitfalls until they encounter the issues when switching Java versions.

Related issues:
- #223
- #2365 (see also comments there)

# Feature description
- Change the default serialization format for `java.util.Date`, `java.sql.Date`, `java.sql.Timestamp` and `java.sql.Time` to use ISO 8601 format
For this can probably either construct a `SimpleDateFormat` with `Locale.ROOT` which matches the ISO 8601 format (in case that is possible), or use the existing internal Gson class `ISO8601Utils`.
- Deprecate `GsonBuilder` methods `setDateFormat(int)` and `setDateFormat(int, int)`
Because they encourage serializing using a non-stable format. The deprecation message for `setDateFormat(int, int)` could show how a user can write their own `TypeAdapter` if they really need this date formatting. `setDateFormat(String)` would not have to be marked as deprecated because the provided format might be stable and independent of locale settings and CLDR data (I hope).

The existing `Date` deserialization logic already has multiple fallbacks in case parsing fails, including `ISO8601Utils`, so that might work without issues.

Optional changes:
- Change the deserialization logic to try ISO 8601 deserialization first, and only if that fails fall back to the other `SimpleDateFormat` patterns. That might yield better performance.

Risks:
- In some cases applications might be directly displaying the date value to the user, and therefore a human readable format is appreciated
[...]

PR #2529, opened 2023-11-06
validating the date format, add test case, since NumberFormatException extends IllegalArgumentException, it is only necessary to write IllegalArgumentException.

Carpe-Wang:
validating the date format, add test case and since NumberFormatException extends IllegalArgumentException, it is only necessary to write IllegalArgumentException.

#### Purpose

* Throw exception if  `public GsonBuilder setDateFormat(String pattern) ` input pattern is invalid. [[Relate issue](https://github.com/google/gson/issues/2472)](https://github.com/google/gson/issues/2472)
* NumberFormatException extends IllegalArgumentException, it is only necessary to write IllegalArgumentException in `public class ISO8601Utils`

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -44,6 +44,7 @@ import com.google.gson.stream.JsonWriter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -595,7 +596,12 @@ public final class GsonBuilder {
    */
   @CanIgnoreReturnValue
   public GsonBuilder setDateFormat(String pattern) {
-    // TODO(Joel): Make this fail fast if it is an invalid date format
+    try {
+      new SimpleDateFormat(pattern);
+    } catch (IllegalArgumentException e) {
+      // throw exception if it is an invalid date format
+      throw new IllegalArgumentException("The date pattern '" + pattern + "' is not valid", e);
+    }
     this.datePattern = pattern;
     return this;
   }
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -596,11 +596,13 @@ public final class GsonBuilder {
    */
   @CanIgnoreReturnValue
   public GsonBuilder setDateFormat(String pattern) {
-    try {
-      new SimpleDateFormat(pattern);
-    } catch (IllegalArgumentException e) {
-      // throw exception if it is an invalid date format
-      throw new IllegalArgumentException("The date pattern '" + pattern + "' is not valid", e);
+    if (pattern != null) {
+      try {
+        new SimpleDateFormat(pattern);
+      } catch (IllegalArgumentException e) {
+        // Throw exception if it is an invalid date format
+        throw new IllegalArgumentException("The date pattern '" + pattern + "' is not valid", e);
+      }
     }
     this.datePattern = pattern;
     return this;
--- a/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
@@ -280,11 +280,7 @@ public class ISO8601Utils
             return calendar.getTime();
             // If we get a ParseException it'll already have the right message/offset.
             // Other exception types can convert here.
-        } catch (IndexOutOfBoundsException e) {
-            fail = e;
-        } catch (NumberFormatException e) {
-            fail = e;
-        } catch (IllegalArgumentException e) {
+        } catch (IndexOutOfBoundsException | IllegalArgumentException e) {
             fail = e;
         }
         String input = (date == null) ? null : ('"' + date + '"');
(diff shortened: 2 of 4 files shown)


eamonnmcmanus (changes requested):
This looks great, thanks! Could you add a test, for example in `DefaultTypeAdaptersTest`, that fails with the current code and passes with this change? You could use `assertThrows`: with the current code the test will fail because no exception will be thrown.

Carpe-Wang:
> This looks great, thanks! Could you add a test, for example in `DefaultTypeAdaptersTest`, that fails with the current code and passes with this change? You could use `assertThrows`: with the current code the test will fail because no exception will be thrown.

Yes, I add them.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -710,6 +711,20 @@ public void testStringBufferDeserialization() {
...
   }
+  @Test
+  public void testSetDateFormatWithInvalidPattern() {
+    GsonBuilder builder = new GsonBuilder();
+    String invalidPattern = "This is a invalid Pattern";
+    assertThrows(IllegalArgumentException.class, () -> {
+      builder.setDateFormat(invalidPattern);
+    });
+  }
+  @Test
```
eamonnmcmanus, line 722: Just one tiny thing: could you add a blank line before each of these new test methods?

Carpe-Wang: OK, I am sorry about that.

Carpe-Wang:
> This looks great, thanks! Could you add a test, for example in `DefaultTypeAdaptersTest`, that fails with the current code and passes with this change? You could use `assertThrows`: with the current code the test will fail because no exception will be thrown.

Please check again to ensure it meets the requirements.

eamonnmcmanus (approved):
Thanks! I am just checking this against Google's internal tests to see if they show up anything unexpected. If not, I'll go ahead and merge this.

Marcono1234:
Should the new tests maybe be moved to `GsonBuilderTest` instead since they seem to only test `GsonBuilder` functionality?

Also, could you please change the title of this PR, there is an "Edit" button at the top right of the GitHub UI[^1]. The main change here seems to be about validating the date format, and issue 2472 might not be that closely related that it is worth mentioning it in the final commit message after merge. What do you think about something like this as title (based on what you wrote in the description)?
> Throw exception if `GsonBuilder.setDateFormat` input pattern is invalid

[^1]: Or alternatively, @eamonnmcmanus could you please adjust the message when merging?

gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -595,7 +596,12 @@ public GsonBuilder disableHtmlEscaping() {
    */
   @CanIgnoreReturnValue
   public GsonBuilder setDateFormat(String pattern) {
-    // TODO(Joel): Make this fail fast if it is an invalid date format
+    try {
```
Marcono1234, line 599: It appears `null` might actually be a valid `pattern` value to clear any custom pattern previously set for the builder. The changes here would not support that anymore.

@eamonnmcmanus, should `null` be supported?

Apparently no unit test is covering that yet, might be good to add one (or I can add one in a separate PR if desired).

Carpe-Wang: if pattern is null, here will  throw null point exception.

Carpe-Wang:
> Should the new tests maybe be moved to `GsonBuilderTest` instead since they seem to only test `GsonBuilder` functionality?
> 
> Also, could you please change the title of this PR, there is an "Edit" button at the top right of the GitHub UI[1](#user-content-fn-1-eb8cfc75c4fc527652fd473304ede99c). The main change here seems to be about validating the date format, and issue 2472 might not be that closely related that it is worth mentioning it in the final commit message after merge. What do you think about something like this as title (based on what you wrote in the description)?
> 
> > Throw exception if `GsonBuilder.setDateFormat` input pattern is invalid
> 
> ## Footnotes
> 1. Or alternatively, @eamonnmcmanus could you please adjust the message when merging? [↩](#user-content-fnref-1-eb8cfc75c4fc527652fd473304ede99c)

I had change the title. Please check again to ensure it meets the requirements. In addition I'm not sure if I need to move the test cases to `GsonBuilderTest`, and I see a conflict with a small note that says "Only those with write access to this repository can merge pull requests." Do I need to resolve this conflict by myself?

eamonnmcmanus:
>  I see a conflict with a small note that says "Only those with write access to this repository can merge pull requests." Do I need to resolve this conflict by myself?

Yes, you should resolve the conflict. I will merge the PR when it is ready.

I don't have a strong opinion on where the new tests should be. We may move them later. On the other hand, I agree with @Marcono1234  that we probably shouldn't change the behaviour for `setDateFormat(null)`. So the call to `new SimpleDateFormat` should be guarded by a `!= null` check.

gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
```diff
@@ -141,169 +141,158 @@
...
+            int milliseconds = 0; // always use 0 otherwise returned date will include millis of current time
 
-        pos.setIndex(offset);
-        return calendar.getTime();
-      }
+            // if the value has no time component (and no time zone), we are done
+            boolean hasT = checkOffset(date, offset, 'T');
 
-      if (hasT) {
+            if (!hasT && (date.length() <= offset)) {
```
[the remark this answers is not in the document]
Carpe-Wang: I am trying to solve that

PR #2531, opened 2023-11-06
Formats codebase

MaicolAntali:
### Purpose
Reformat the Gson codebase to follow the Google Java Style Guide

### Checklist

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -16,6 +16,10 @@
 
 package com.google.gson.typeadapters;
 
+import com.google.gson.JsonParseException;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
 import java.text.ParseException;
 import java.text.ParsePosition;
@@ -24,11 +28,6 @@ import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.Locale;
 import java.util.TimeZone;
-import com.google.gson.JsonParseException;
-import com.google.gson.TypeAdapter;
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonToken;
-import com.google.gson.stream.JsonWriter;
 
 public final class UtcDateTypeAdapter extends TypeAdapter<Date> {
   private final TimeZone UTC_TIME_ZONE = TimeZone.getTimeZone("UTC");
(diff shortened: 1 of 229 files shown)


eamonnmcmanus (approved):
Thanks!

Marcono1234:
For anyone who wants to verify that these changes are legit, do the following:
1. Check out commit cb6643f148d553404754a785fda22d36531f178f (the commit before the formatting change)
2. Add the following to the root `pom.xml`:
    <details>
    
    <summary>Spotless plugin</summary>
    
    ```xml
    <!-- Spotless plugin: keeps the code formatted following the google-java-styleguide -->
    <plugin>
      <groupId>com.diffplug.spotless</groupId>
      <artifactId>spotless-maven-plugin</artifactId>
      <version>2.40.0</version>
      <executions>
        <execution>
          <goals>
            <goal>check</goal>
          </goals>
        </execution>
      </executions>
      <configuration>
        <formats>
          <format>
            <includes>
              <include>*.md</include>
              <include>.gitignore</include>
            </includes>
            <trimTrailingWhitespace/>
            <endWithNewline/>
            <indent>
              <spaces>true</spaces>
              <spacesPerTab>4</spacesPerTab>
            </indent>
          </format>
        </formats>
        <java>
          <googleJavaFormat>
            <style>GOOGLE</style>
            <reflowLongStrings>true</reflowLongStrings>
            <formatJavadoc>true</formatJavadoc>
          </googleJavaFormat>
          <importOrder />
          <removeUnusedImports />
          <formatAnnotations />     <!-- Puts type annotations immediately before types. -->.
        </java>
      </configuration>
    </plugin>
    ```
    
    </details>
3. Run `mvn spotless:apply`

You should obtain a state nearly identical to what this PR includes. The only changes are in 26 files were comments were modified (I assume because the Spotless plugin / Google Formatter did not properly wrap them).

@MaicolAntali, just for clarification, you edited those comments, right?

eamonnmcmanus:
FWIW the check I used was to import this change into Google's build system, which does not include timestamps in jar files. Then compile before and after with `-g:none` (since some line numbers changed) and verify that the resultant Gson jar file is identical.

MaicolAntali:
> The only changes are in 26 files were comments were modified (I assume because the Spotless plugin / Google Formatter did not properly wrap them).

Exactly, I edit some comments to avoid some wrong new lines.

extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
```diff
@@ -160,7 +160,8 @@ private static Date parse(String date, ParsePosition pos) throws ParseException
       int hour = 0;
       int minutes = 0;
       int seconds = 0;
-      int milliseconds = 0; // always use 0 otherwise returned date will include millis of current time
+      // always use 0 otherwise returned date will include millis of current time
```
MaicolAntali, line 163: Comment in a new line

Marcono1234:
Thanks for the confirmation from you both and thanks @MaicolAntali for your dedication with this by manually improving the comment formatting!

> FWIW the check I used was to import this change into Google's build system, which does not include timestamps in jar files. Then compile before and after with `-g:none` (since some line numbers changed) and verify that the resultant Gson jar file is identical.

Thanks for the hint. It looks like locally you can achieve this using the following Maven command:
```
mvn clean verify jar:test-jar javadoc:jar "-Dproguard.skip" -DskipTests "-Dmaven.compiler.debug=true" "-Dmaven.compiler.debuglevel=none" "-Dproject.build.outputTimestamp=2023-01-01T00:00:00Z"
```

I only tried this on the chances I obtained locally by running `mvn spotless:apply` (which as mentioned above slightly differ from the changes of this PR); but unless I messed up my test setup, to me it also looks like the main and test JARs were identical before and after the formatting.

PR #2538, opened 2023-11-07
validating the date format, add test case, since NumberFormatExceptio…

Carpe-Wang:
validating the date format, add test case and since NumberFormatException extends IllegalArgumentException, it is only necessary to write IllegalArgumentException.

#### Purpose

* Throw exception if  `public GsonBuilder setDateFormat(String pattern) ` input pattern is invalid.
* NumberFormatException extends llegalArgumentException, it is only necessary to write llegalArgumentException in `public class ISO8601Utils`

### Checklist


Carpe-Wang:
Sorry everyone, I make a mistake when I solve conflict. So I close first pr, and make a new pr. I am so sorry for my mistake.

Marcono1234:
@Carpe-Wang why did you close this PR? I think you can just run `mvn spotless:apply`, commit and push the changes and it should be fine then.

Sorry though for the trouble; in the meantime while your PR was open checks had been added which verify that the code is properly formatted.

Marcono1234 (approved):
Thanks!

Technically the current `GsonBuilder` implementation also allows blank strings as 'no-pattern':
https://github.com/google/gson/blob/bf572581b9156002ee98866620279d92ba8dc509/gson/src/main/java/com/google/gson/GsonBuilder.java#L889

But I am not sure if that is something we have to keep supporting and if anyone really relies on this (especially since it was not publicly documented).

@eamonnmcmanus, would it make sense to run the changes of this PR against the Google tests to be safe anyway?

eamonnmcmanus:
> @eamonnmcmanus, would it make sense to run the changes of this PR against the Google tests to be safe anyway?

I've already done that, and it didn't show up any problems.

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks for this contribution!

eamonnmcmanus:
> By the way out of curiosity, what do I need to do to become a contributor to this project? What more effort is required?

Well, you already _are_ a contributor, since you've just contributed this PR. Perhaps I'm not understanding the question?

eamonnmcmanus:
I think the people in that box are chosen automatically by GitHub based on who has contributed the most. If you open up where it says "+ 132 contributors" you should see everyone else. I don't see you there yet, but probably it only gets updated every so often.

PR #2537, opened 2023-11-07
Add `spotless` plugin

MaicolAntali:
### Purpose
Add the `spotless` plugins to enforce the Google Java Style Guide.

### Description
Related with #2284

PR #2540, opened 2023-11-14
Formatting follow-up

Marcono1234:
### Purpose
Follow-up for #2531, #2536 and #2537

### Description
- Adds formatting commits to `.git-blame-ignore-revs` so that they don't distract during Git blame; see also the [GitHub documentation](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)
- Restores hard line breaks in Troubleshooting.md using `\` instead of trailing spaces
- Changes formatting of some string literals and comments
- Fixes accidental Javadoc and comment issues introduced by manual changes of formatting commit
- Fixes license header in `$Gson$Types.java` erroneously being a Javadoc comment and being reformatted
- Slightly changes `JsonReader` `getPath` and `getPreviousPath` documentation to help Javadoc detect first sentence as summary
- Adjusts pull request template to mention Maven commands for checking and fixing formatting issues

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
@@ -160,10 +160,12 @@ public class ISO8601Utils {
 
       // extract day
       int day = parseInt(date, offset, offset += 2);
+
       // default time value
       int hour = 0;
       int minutes = 0;
       int seconds = 0;
+
       // always use 0 otherwise returned date will include millis of current time
       int milliseconds = 0;
 
--- a/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
@@ -164,8 +164,8 @@ public class ISO8601Utils {
       int hour = 0;
       int minutes = 0;
       int seconds = 0;
-      int milliseconds =
-          0; // always use 0 otherwise returned date will include millis of current time
+      // always use 0 otherwise returned date will include millis of current time
+      int milliseconds = 0;
 
       // if the value has no time component (and no time zone), we are done
       boolean hasT = checkOffset(date, offset, 'T');
--- a/gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
+++ b/gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
@@ -115,14 +115,12 @@ public class FormattingStyleTest {
     // Sanity check to verify that `buildExpected` works correctly
     assertThat(json)
         .isEqualTo(
-            // spotless:off
-            "{\n"
-            + "  \"a\": [\n"
-            + "    1,\n"
-            + "    2\n"
-            + "  ]\n"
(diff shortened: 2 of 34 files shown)


gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
```diff
@@ -164,8 +164,8 @@ public static Date parse(String date, ParsePosition pos) throws ParseException {
       int hour = 0;
       int minutes = 0;
       int seconds = 0;
-      int milliseconds =
-          0; // always use 0 otherwise returned date will include millis of current time
+      // always use 0 otherwise returned date will include millis of current time
```
eamonnmcmanus, line 167: I'm never sure whether a standalone `//` comment is supposed to apply to the previous line or the next one. Perhaps put a blank line before it to clarify?

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
```diff
@@ -113,7 +113,16 @@ public void testFormatPretty() {
     String expectedJson = buildExpected("\n", "  ", true);
     assertThat(json).isEqualTo(expectedJson);
     // Sanity check to verify that `buildExpected` works correctly
-    assertThat(json).isEqualTo("{\n" + "  \"a\": [\n" + "    1,\n" + "    2\n" + "  ]\n" + "}");
+    assertThat(json)
+        .isEqualTo(
+            // spotless:off
```
eamonnmcmanus, line 118: I don't think this should be necessary. If you put an empty `//` comment at the end of any line then google-java-format will not try to join them:
```java
assertThat(json).isEqualTo(
    "{\n" //
    + ...);
```
Though I don't really know whether the Spotless plugin applies exactly the same logic.

gson/src/test/java/com/google/gson/functional/LeniencyTest.java
```diff
@@ -40,7 +40,11 @@ public void setUp() throws Exception {
   public void testLenientFromJson() {
     List<String> json =
         gson.fromJson(
-            "" + "[ # One!\n" + "  'Hi' #Element!\n" + "] # Array!",
+            // spotless:off
```
eamonnmcmanus, line 43: Same comment here, and in several places below.

eamonnmcmanus (changes requested):
Thanks for putting in all this effort! Just some small remarks.

gson/src/test/java/com/google/gson/functional/FormattingStyleTest.java
Marcono1234: > If you put an empty `//` comment at the end of any line then google-java-format will not try to join them

Yes that works, and I guess here it is ok because there are always multiple lines with `//`, but a single trailing `//` on its own looks to me rather like someone forgot to finish a comment.

> Though I don't really know whether the Spotless plugin applies exactly the same logic.

I think it should. If I understand it correctly it just delegates the formatting to google-java-format (and optionally applies additional formatting steps).

eamonnmcmanus (approved):
Thanks again!

PR #2543, opened 2023-11-18
Make Maven build reproducible

Marcono1234:
### Purpose
Makes the Maven build reproducible

### Description
Makes the Maven build reproducible, see https://maven.apache.org/guides/mini/guide-reproducible-builds.html, and also https://github.com/google/guava/pull/6322. However, it might still be necessary to use the same OS and JDK version to actually be able to reproduce a build.

The changes also format the `pom.xml` files in the way the Maven Release Plugin would.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
           cache: 'maven'
       - name: Build with Maven
         # This also runs javadoc:jar to detect any issues with the Javadoc generated during release
-        run: mvn --batch-mode --update-snapshots --no-transfer-progress verify javadoc:jar
+        run: mvn --batch-mode --no-transfer-progress verify javadoc:jar
 
   native-image-test:
     name: "GraalVM Native Image test"
@@ -42,4 +42,29 @@ jobs:
       - name: Build and run tests
         # Only run tests in `graal-native-image-test` (and implicitly build and run tests in `gson`),
         # everything else is covered already by regular build job above
-        run: mvn test --batch-mode --update-snapshots --no-transfer-progress --activate-profiles native-image-test --projects graal-native-image-test --also-make
+        run: mvn test --batch-mode --no-transfer-progress --activate-profiles native-image-test --projects graal-native-image-test --also-make
+
+  verify-reproducible-build:
+    name: "Verify reproducible build"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: "Set up JDK 17"
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: 17
+          cache: 'maven'
+
+      - name: "Verify no plugin issues"
+        run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress
+
+      - name: "Verify reproducible build"
+        # See https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
+        run: |
+          mvn clean install --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests
+          # Run with `-Dbuildinfo.attach=false`; otherwise `artifact:compare` fails because it creates a `.buildinfo` file which
+          # erroneously references the existing `.buildinfo` file (respectively because it is overwriting it, a file with size 0)
+          # See https://issues.apache.org/jira/browse/MARTIFACT-57
+          mvn clean verify artifact:compare --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests -Dbuildinfo.attach=false
--- a/ReleaseProcess.md
+++ b/ReleaseProcess.md
@@ -37,6 +37,72 @@ This section was borrowed heavily from [Doclava release process](https://code.go
 
 See [OSSRH Publish Guide](https://central.sonatype.org/publish/publish-guide/).
 
+## Testing Maven release workflow locally
+
+The following describes how to perform the steps of the release locally to verify that they work as desired.
+
+**Warning:** Be careful with this, these steps might be outdated or incomplete. Doublecheck that you are working on a copy of your local Gson Git repository and make sure you have followed all steps. To be safe you can also temporarily turn off your internet connection to avoid accidentally pushing changes to the real remote Git or Maven repository.\
+As an alternative to the steps described below you can instead [perform a dry run](https://maven.apache.org/maven-release/maven-release-plugin/usage.html#do-a-dry-run), though this might not behave identical to a real release.
+
+1. Make a copy of your local Gson Git repository and only work with that copy
+2. Make sure you are on the `main` branch
+3. Create a temp directory outside the Gson directory\
+   In the following steps this will be called `#gson-remote-temp#`; replace this with the actual absolute file path of the directory, using only forward slashes. For example under Windows `C:\my-dir` becomes `C:/my-dir`.
+4. Create the directory `#gson-remote-temp#/git-repo`
+5. In that directory run
+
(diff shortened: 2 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -42,4 +42,29 @@ jobs:
       - name: Build and run tests
         # Only run tests in `graal-native-image-test` (and implicitly build and run tests in `gson`),
         # everything else is covered already by regular build job above
-        run: mvn test --batch-mode --update-snapshots --no-transfer-progress --activate-profiles native-image-test --projects graal-native-image-test --also-make
+        run: mvn test --batch-mode --no-transfer-progress --activate-profiles native-image-test --projects graal-native-image-test --also-make
+
+  verify-reproducible-build:
+    name: "Verify reproducible build"
```
Marcono1234, line 48: Have included this as separate job (instead of including it in the steps of the build above) because it requires running `mvn ... install` which is normally not needed for a regular build, and also to keep it independent from the 'real' build setup above.

However, this makes the GitHub workflow slightly slower though.

ReleaseProcess.md
```diff
@@ -37,6 +37,72 @@ This section was borrowed heavily from [Doclava release process](https://code.go
 
 See [OSSRH Publish Guide](https://central.sonatype.org/publish/publish-guide/).
 
+## Testing Maven release workflow locally
+
+The following describes how to perform the steps of the release locally to verify that they work as desired.
+
+**Warning:** Be careful with this, these steps might be outdated or incomplete. Doublecheck that you are working on a copy of your local Gson Git repository and make sure you have followed all steps. To be safe you can also temporarily turn off your internet connection to avoid accidentally pushing changes to the real remote Git or Maven repository.\
```
Marcono1234, line 44: This setup is a bit hacky and I am not sure if there is a proper better alternative to this, therefore I added this "Warning". But it does allow you to test the behavior of the release workflow quite accurately, which is why I added it.

pom.xml
```diff
@@ -175,10 +178,29 @@
...
+
+      <!-- Attaches a `.buildinfo` file which contains information for reproducing the build,
+        such as OS, JDK version, ...
+        Since this is a multi-module Maven project, only one aggregated file will be created for
+        the last module, see the note on https://maven.apache.org/plugins/maven-artifact-plugin/usage.html#recording-buildinfo-file -->
+      <!-- The other goals of this plugin are run by the GitHub workflow to verify that
+        the build is reproducible (see `artifact:...` usage in the workflow) -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-artifact-plugin</artifactId>
```
Marcono1234, line 194: This includes information about the build environment to make it easier to reproduce the build. I am not sure if you would consider some of this information sensitive (e.g. OS name and version or Java vendor and version).

For example for me the generated file looks like this:

<details>

<summary><code>gson-2.10.2.buildinfo</code> (click to expand)</summary>

```
# https://reproducible-builds.org/docs/jvm/
buildinfo.version=1.0-SNAPSHOT

name=Gson Parent
group-id=com.google.code.gson
artifact-id=gson-parent
version=2.10.2

# source information
source.scm.uri=scm:git:https://github.com/google/gson.git
source.scm.tag=gson-parent-2.10.2

# build instructions
build-tool=mvn

# effective build environment information
java.version=17.0.6
java.vendor=Eclipse Adoptium
os.name=Windows 10

# Maven rebuild instructions and effective environment
mvn.version=3.9.1
mvn.aggregate.artifact-id=gson

# aggregated output

outputs.0.coordinates=com.google.code.gson:gson-parent

outputs.0.0.groupId=com.google.code.gson
outputs.0.0.filename=gson-parent-2.10.2.pom
outputs.0.0.length=20999
outputs.0.0.checksums.sha512=36a916aea7d2e261e62a9287d0d9a2fcf9cdd94f1658bfba616dc3e9a08121b3ff080e5b574f32102311b8c6d286db977dd5b7038c5b4841c0fd63a164fecd93

outputs.1.coordinates=com.google.code.gson:gson

outputs.1.0.groupId=com.google.code.gson
outputs.1.0.filename=gson-2.10.2.pom
outputs.1.0.length=11602
outputs.1.0.checksums.sha512=e16df3d3a571fee3d6145007eb7eb20dcef0a7c9707d7ea347341cf3708ea73f197550076336dc3bb29c842ee4d00ad2ce5a9ae10302d98ac55d4437d3d7949c

outputs.1.1.groupId=com.google.code.gson
outputs.1.1.filename=gson-2.10.2.jar
outputs.1.1.length=307320
outputs.1.1.checksums.sha512=300dc667e042cb8dcc65fcf97b169b2b15bec96d4728164649f2e9dff2cc1a8b4b1f1053f12875fe59395fb42bac8abed49a5d410ecfcae8ca2287f9f8e9ada1

outputs.1.2.groupId=com.google.code.gson
outputs.1.2.filename=gson-2.10.2-sources.jar
outputs.1.2.length=202441
outputs.1.2.checksums.sha512=aa94a042ee8e3a985329eb421853668a2660ef5f1b70c5fdbcfa0b6f2cbb203ae07275ccaba02fb418773fc89ba36984e1671ee832ebe8eb31c97436e84761c9
```

</details>

pom.xml
```diff
@@ -175,10 +178,29 @@
...
+        <version>3.5.0</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>buildinfo</goal>
```
Marcono1234, line 199: This causes a warning during the build:
> [WARNING] SCM source tag in buildinfo source.scm.tag=HEAD does not permit rebuilders reproducible source checkout

But I guess that can be safely ignored because during release `source.scm.tag` will be temporarily changed by the Maven Release Plugin.

Marcono1234 (commented):
I have tested the release workflow locally and it seems to work as expected, but I **excluded the GPG plugin** for my tests because I don't have it set up locally.

The output of `mvn help:effective-pom -Prelease` suggests it will be run after the `buildinfo` goal of the `maven-artifact-plugin`, as desired. And if I understand it correctly the GPG plugin does not modify the JARs but just attaches additional `.asc` signature files, so this should not be a problem (I hope).

eamonnmcmanus (approved):
This all looks great. Thanks!

I'm not too concerned about including the build environment information. It will likely be Google's internal version of Linux in actual releases, which means nobody outside will be able to reproduce it exactly. But I don't think there's anything secret there.

We haven't had a release since January and there has been quite a lot of activity since then, so we will probably have the opportunity to put this in action soon.

Let me know if this is ready to merge or if you still have changes in mind.

Marcono1234:
> It will likely be Google's internal version of Linux in actual releases, which means nobody outside will be able to reproduce it exactly. But I don't think there's anything secret there.

Ok thanks for the confirmation; I wasn't completely sure how you performed the previous releases since you were shown as Git committer. Even if it is an internal version of Linux, I assume it might still be reproducible. The main concern the Maven guide listed was line endings (`\n` for Linux, `\r\n` for Windows) which could have an effect; there should hopefully not be any relevant differences between Linux versions.

> We haven't had a release since January and there has been quite a lot of activity since then, so we will probably have the opportunity to put this in action soon.

I was thinking about the same some weeks ago, but I think it would be good to include a few more changes:
- #2364
- #2498 (I will try to resolve the conflicts in the next days)
- Some minor code clean-up changes I am currently working on, but haven't created a PR yet
- The nested JSON handling we have been writing about; I will have to resolve conflicts there as well

> Let me know if this is ready to merge or if you still have changes in mind.

I think this should be ready; my main concern is the GPG plugin, but in the worst case it will fail on release and we can revert the changes of this PR again temporarily.

PR #2544, opened 2023-11-18
Perform minor code clean-up

Marcono1234:
### Purpose
Performs some minor code clean-up, which should not noticeably (or at all) change functionality

### Description
Notable changes:
- Replace most usages of `<code>` with `{@code ...}` in Javadoc
- Add proper summary sentence to `GsonBuilder.enableComplexMapKeySerialization`
- Extend documentation for `GsonBuilder.setDateFormat` methods
- Fix `TypeToken.isAssignableFrom(Type)` throwing `AssertionError` in some cases
Maybe the method should not throw an exception in the first place, but it might not matter much since it is deprecated already anyway.
- Remove outdated `throws NumberFormatException` from internal `JsonReader.nextQuotedValue`; the behavior had been changed by 85ebaf7c3553a5d5c058fd6067824a7c5258d07f
- Fix incorrect documentation on `JsonScope` fields
- Fix unit tests having 'expected' and 'actual' switched
For that I removed the `.getType()` call from `TypeToken` to use `<T> Gson.fromJson(..., TypeToken<T>)` instead of `<T> Gson.fromJson(..., Type)` because otherwise the call to `assertThat(...)` would have been ambiguous.
- Use dedicated Truth methods instead of `isTrue()` / `isFalse()`
- Use common helper methods in `JsonWriterTest` to avoid duplication
- Implement `toString()` for `ReflectionAccessFilter` constants

### Checklist


gson/src/test/java/com/google/gson/JsonPrimitiveTest.java
```diff
@@ -297,7 +297,7 @@ public void testEqualsAcrossTypes() {
         new JsonPrimitive(new BigInteger("0")), new JsonPrimitive(0));
     MoreAsserts.assertEqualsAndHashCode(new JsonPrimitive(0), new JsonPrimitive(0L));
     MoreAsserts.assertEqualsAndHashCode(
-        new JsonPrimitive(new BigInteger("0")), new JsonPrimitive(0));
+        new JsonPrimitive(new BigDecimal("0")), new JsonPrimitive(0));
```
Marcono1234, line 300: Changed this because the same test (`new BigInteger("0"), new JsonPrimitive(0)`) already exists three lines above.

gson/src/test/java/com/google/gson/functional/ObjectTest.java
```diff
@@ -141,9 +141,9 @@ public void testClassWithTransientFieldsDeserialization() {
 
   @Test
   public void testClassWithTransientFieldsDeserializationTransientFieldsPassedInJsonAreIgnored() {
-    String json = "{\"transientLongValue\":1,\"longValue\":[1]}";
+    String json = "{\"transientLongValue\":5,\"longValue\":[1]}";
```
Marcono1234, line 144: Changed this because otherwise the test is extremely misleading: The point is that the value in JSON is ignored. However, the `ClassWithTransientFields` constructor also initializes the field to 1, so if the JSON contains 1 as well it is not possible to tell where the value came from.

examples/android-proguard-example/src/com/google/gson/examples/android/GsonProguardExampleActivity.java
```diff
@@ -41,13 +41,13 @@ public void onCreate(Bundle icicle) {
     Cart cart = buildCart();
     StringBuilder sb = new StringBuilder();
     sb.append("Gson.toJson() example: \n");
-    sb.append("  Cart Object: ").append(cart).append("\n");
-    sb.append("  Cart JSON: ").append(gson.toJson(cart)).append("\n");
+    sb.append("  Cart Object: ").append(cart).append('\n');
```
eamonnmcmanus, line 44: Not sure I would have bothered with these, but why not.

gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -309,21 +309,21 @@ private static boolean typeEquals(
     return false;
   }
 
-  private static AssertionError buildUnexpectedTypeError(Type token, Class<?>... expected) {
+  private static IllegalArgumentException buildUnsupportedTypeError(
```
eamonnmcmanus, line 312: I agree with the change from `AssertionError` to `IllegalArgumentException`, especially since this can happen quite easily if someone has the idea of passing a `WildcardType` or a `TypeVariable`. As a very minor thing, perhaps change the name of the method so it ends with `Exception` rather than `Error`.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1056,7 +1056,6 @@ public long nextLong() throws IOException {
    * consumes the closing quote, but does not include it in the returned string.
    *
    * @param quote either ' or ".
-   * @throws NumberFormatException if any unicode escape sequences are malformed.
```
eamonnmcmanus, line 1059: Perhaps mention `MalformedJsonException` instead? Though for a private method it's probably not worth it. On the other hand, I think we could change the `throws` clause to `MalformedJsonException`?

gson/src/test/java/com/google/gson/MockExclusionStrategy.java
eamonnmcmanus, line 1: Indeed! :-)

gson/src/test/java/com/google/gson/functional/MapTest.java
```diff
@@ -589,8 +589,7 @@ public void testStringKeyDeserialization() {
     Map<String, String> map = new LinkedHashMap<>();
     map.put("2,3", "a");
     map.put("5,7", "b");
-    assertThat(map)
-        .isEqualTo(gson.fromJson(json, new TypeToken<Map<String, String>>() {}.getType()));
+    assertThat(gson.fromJson(json, new TypeToken<Map<String, String>>() {})).isEqualTo(map);
```
eamonnmcmanus, line 592: While you are here, you might as well use the assertions from [`MapSubject`](https://truth.dev/api/latest/com/google/common/truth/MapSubject.html)? So

```java
assertThat(gson.fromJson(...)).containsExactly("2,3", "a", "5,7", "b");
```

Same thing in the next two methods, though not necessarily in `testDeeralizeMapOfMaps`. You could fix the name of that last one, though.

gson/src/test/java/com/google/gson/functional/ObjectTest.java
```diff
@@ -141,9 +141,9 @@ public void testClassWithTransientFieldsDeserialization() {
...
     ClassWithTransientFields<?> target = gson.fromJson(json, ClassWithTransientFields.class);
-    assertThat(target.transientLongValue != 1).isFalse();
```
eamonnmcmanus, line 146: Very odd! Seems like it was this way [from the beginning](https://github.com/google/gson/blob/18b301dfeb91ca4468e0955d4cafcb32186efe1e/gson/src/test/java/com/google/gson/functional/ObjectTest.java#L134).

eamonnmcmanus (commented):
Very nice! Just a few comments.

gson/src/main/java/com/google/gson/stream/JsonReader.java
Marcono1234: `MalformedJsonException` is a subclass of `IOException`, but this method (like most of the other `JsonReader` methods) can throw both:
- `MalformedJsonException` for invalid JSON
- `IOException` when reading from the underlying `Reader` fails

I think I am not going to add `throws MalformedJsonException` to the method signature because then for consistency (but not because the compiler requires it) it would have to be done for all other methods as well, and that is probably not worth it.

However, I have adjusted two internal methods now which always throw `MalformedJsonException` (`checkLenient` and `syntaxError`).

gson/src/test/java/com/google/gson/functional/MapTest.java
Marcono1234: Would it be ok if I leave the assertions as they are and not use `MapSubject`? I am currently not completely convinced by `MapSubject.containsExactly` because it makes it a bit difficult to differentiate keys from values, especially here where the keys looks like they might be values (though it is not so bad since this are only 2 entries).

eamonnmcmanus: Yeah, that's fair.

gson/src/main/java/com/google/gson/stream/JsonReader.java
eamonnmcmanus: OK, thanks!

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
```diff
@@ -117,6 +117,11 @@ public FilterResult check(Class<?> rawClass) {
               ? FilterResult.BLOCK_INACCESSIBLE
               : FilterResult.INDECISIVE;
         }
+
+        @Override
```
eamonnmcmanus, line 121: Can you mention these in the PR description?

eamonnmcmanus (changes requested):
Looking good! Just one small thing...

gson/src/main/java/com/google/gson/ReflectionAccessFilter.java
Marcono1234: Have included it now; sorry that I added this in a follow-up commit and not in the original changes of this PR. I just remembered that this change was something I wanted to do for quite some time because the default `toString()` during debugging makes it difficult to understand which filter is used.

eamonnmcmanus (approved):
Thanks again. This is a big improvement!

PR #2545, opened 2023-11-18
Validate `GsonBuilder.setDateFormat` style arguments & extend tests

Marcono1234:
### Purpose
Validate `GsonBuilder.setDateFormat` style arguments & extend tests

### Description
As side note: `DateFormat` validates these arguments as well, but it is currently not documented, see https://bugs.openjdk.org/browse/JDK-8319628.

Also moves the other tests for `setDateFormat(String)` from `DefaultTypeAdaptersTest` to `GsonBuilderTest` because they just test functionality of `GsonBuilder.setDateFormat` would testing any type adapter functionality.
(Acts a bit as follow-up for #2538.)

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -593,8 +593,10 @@ public final class GsonBuilder {
    * class. See the documentation in {@link java.text.SimpleDateFormat} for more information on
    * valid date and time patterns.
    *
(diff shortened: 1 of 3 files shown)


eamonnmcmanus (approved):
Looks great, thanks!

unicornsavior42:
Unicornsavior42

Marcono1234:
As side note: Even though this extends the tests, it does not cover all use cases; #1529 is most likely still unresolved.

PR #2546, opened 2023-11-19
Remove `DateTypeAdapter` to avoid code duplication

Marcono1234:
### Purpose
Remove the internal `DateTypeAdapter` class to avoid code duplication

### Description
The same functionality exists in `DefaultDateTypeAdapter` (with some slight overhead though), and there is currently code duplication between these classes.

The code in `DefaultDateTypeAdapter` and the way it is used now as replacement should behave the same way `DateTypeAdapter` did, but feel free to double-check in case I overlooked something.

### Checklist


eamonnmcmanus (commented):
This is great, thanks! I think I've said before that the way it handles `Date` is probably Gson's biggest mistake. We can't easily fix that, but we can at least make it easier to improve the situation.

The changes look good to me. I'm running this past Google's tests in case that turns anything up.

eamonnmcmanus (approved):
Google's internal tests all passed.

PR #2548, opened 2023-11-19
Fix `DateFormat` time zone is not restored.

Carpe-Wang:
### Purpose
Fix `DateFormat` time zone is not restored after parsing. About issue is  #2547
### Description

Before attempting to parse the date string, save the current timezone for each DateFormat object. Use a `finally` block to ensure that the original timezone is restored for each DateFormat object.

### Checklist


gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java
```diff
@@ -31,10 +31,7 @@
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.ParsePosition;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
```
eamonnmcmanus, line 34: Your IDE probably did this for you, but please undo it. :-)
This project follows the Google Java Style Guide, notably the bit [excluding wildcard imports](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports).

eamonnmcmanus (changes requested):
This looks good, but can you add a test that fails without this change and passes with it?

Marcono1234:
Just as forewarning: This change conflicts with #2546, so one of the PRs will cause conflicts for the other one.

Also this affects more classes:
- `DefaultDateTypeAdapter`
- SQL type adapters:
  - `SqlDateTypeAdapter`
  - `SqlTimeTypeAdapter`
  
   (they might not actually be affected because their pattern does not include the time zone, but might be good nonetheless to restore the `DateFormat` time zone, just to be safe)

If you don't want to change these yet, I can do that in a follow-up PR.

Some additional notes:
- Please adjust the title of this PR; "fix issue##2547" is not really expressive when someone at a glance tries to understand what a PR or eventually the commit on `main` did; what about this?
    > Fix \`DateFormat\` time zone is not restored after parsing
- Please edit the description and write "Fix #2547" somewhere in it; GitHub [understands certain keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) and then automatically links issue and PR; otherwise it does not understand that this PR fixes the GitHub issue

Carpe-Wang:
> Just as forewarning: This change conflicts with #2546, so one of the PRs will cause conflicts for the other one.
> 
> Also this affects more classes:
> 
> * `DefaultDateTypeAdapter`
> * SQL type adapters:
>   
>   * `SqlDateTypeAdapter`
>   * `SqlTimeTypeAdapter`
>   
>   (they might not actually be affected because their pattern does not include the time zone, but might be good nonetheless to restore the `DateFormat` time zone, just to be safe)
> 
> If you don't want to change these yet, I can do that in a follow-up PR.
> 
> Some additional notes:
> 
> * Please adjust the title of this PR; "fix issue##2547" is not really expressive when someone at a glance tries to understand what the PR or eventually the commit on `main` did; what about this?
>   > Fix `DateFormat` time zone is not restored after parsing
> * Please edit the description and write "Fix #2547" somewhere in it; GitHub [understands certain keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) and then automatically links issue and PR; otherwise it does not understand that this PR fixes the GitHub issue

ok, I change title and pr's description. About other classes need to be changed, I will do it in next  week.

gson/src/main/java/com/google/gson/internal/bind/DateTypeAdapter.java
Carpe-Wang: Oh, Idea change it automatically.

Carpe-Wang:
> This looks good, but can you add a test that fails without this change and passes with it?

yes, I add the test, and find my change can not fix the bug. I will debug to find how to fix.

Carpe-Wang:
there are some issue in my local Idea, I can’t fetch new branch. I will close the issue and make a new pr when I finished the development.

PR #2550, opened 2023-11-20
Spotless: Fail for wildcard imports

Marcono1234:
### Purpose
Fail for wildcard `import` statements because they violate the Google Java Style Guide

### Description
Neither Spotless nor google-java-format can currently detect or fix wildcard imports. This might be a bit frustrating for new contributors because for example IntelliJ by default uses wildcard imports, and currently the formatting check passes. So the contributor would only learn during the PR review that wildcard imports are not allowed.

This change causes Spotless to add a comment behind the import statement and turn the import invalid so that the user has to fix it:
```diff
-import·java.util.*;
+import·java.util.DISALLOWED.*;·//·wildcard·imports·are·not·allowed
```

This is a slightly hacky approach, so no worries if you don't want to merge this. Or in case you merge this but in the future this causes cryptic google-java-format exceptions because this replacement somehow made the Java source code invalid, or if this does not turn out to be useful, then we can also revert the change again.

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,7 @@
               <exclude>src/test/java/com/google/gson/functional/Java17RecordTest.java</exclude>
               <exclude>src/test/java/com/google/gson/native_test/Java17RecordReflectionTest.java</exclude>
             </excludes>
+
             <googleJavaFormat>
               <style>GOOGLE</style>
               <reflowLongStrings>true</reflowLongStrings>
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Shouldn't wildcard imports trigger an [Error Prone check](https://errorprone.info/bugpattern/WildcardImport)?

Marcono1234:
Obsolete now due to #2561

PR #2552, opened 2023-11-22
Add CI build using JDK 21

Marcono1234:
### Purpose
Add CI build using JDK 21

### Description
Depends on #2551

This is mainly to make sure that besides the known JDK 21 build incompatibilities Gson can be built and run using JDK 21. However, the restriction that JDK 21 must not be used for building enforced by #2551 still remains, the GitHub workflow just disables this restriction, and performs some adjustments to the build to work around these issues.

Since a JDK 21 build still does not work out of the box and differs from the desired configuration for building a release (e.g. release Java version is increased to Java 8), I think it does not make sense to allow or encourage users to use JDK 21 for building yet.

I have split this into a separate PR from #2551 because otherwise it would be a bit weird to have the commit message say on one hand that builds must use JDK <= 17, but also say that CI now builds with JDK 21 as well.

### Checklist

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -191,6 +191,19 @@
             </goals>
           </execution>
         </executions>
+        <!-- Upgrades ProGuard to version newer than the one included by plugin by default -->
+        <dependencies>
+          <dependency>
+            <groupId>com.guardsquare</groupId>
+            <artifactId>proguard-base</artifactId>
+            <version>7.4.0</version>
+          </dependency>
+          <dependency>
+            <groupId>com.guardsquare</groupId>
+            <artifactId>proguard-core</artifactId>
+            <version>9.1.0</version>
+          </dependency>
+        </dependencies>
         <configuration>
           <obfuscate>true</obfuscate>
           <injar>test-classes-obfuscated-injar</injar>
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -196,7 +196,7 @@
           <dependency>
             <groupId>com.guardsquare</groupId>
             <artifactId>proguard-base</artifactId>
-            <version>7.4.0</version>
+            <version>7.4.1</version>
           </dependency>
           <dependency>
             <groupId>com.guardsquare</groupId>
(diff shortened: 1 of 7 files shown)


gson/pom.xml
```diff
@@ -191,6 +191,19 @@
             </goals>
           </execution>
         </executions>
+        <!-- Upgrades ProGuard to version newer than the one included by plugin by default -->
+        <dependencies>
+          <dependency>
+            <groupId>com.guardsquare</groupId>
+            <artifactId>proguard-base</artifactId>
+            <version>7.4.0</version>
```
eamonnmcmanus, line 199: Looks like 7.4.1 was just released a few days ago.

pom.xml
```diff
@@ -461,6 +463,32 @@
...
+        <!-- JDK 21 does not support Java 7 as release, must use at least Java 8 -->
+        <maven.compiler.release>8</maven.compiler.release>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <!-- Don't fail on warning because javac reports lint warnings for generated code of 'proto' module -->
```
eamonnmcmanus, line 484: Can we fix those warnings? Not necessarily now, but perhaps an issue could track this.

eamonnmcmanus (changes requested):
This is great, thanks! Just a couple of very minor things.

pom.xml
Marcono1234: Ah nevermind, this is actually fixed in newer Protobuf versions: https://github.com/protocolbuffers/protobuf/issues/9673

The problem is that the Protobuf version we are using is outdated. Thanks for your question, which prompted me to properly investigate this!

Marcono1234:
Thanks for the review!

If you want you can merge #2551 separately to make sure they appear separate on `main`; or if you don't think that this is important, you can also just merge this PR here and then #2551 becomes obsolete.

eamonnmcmanus (approved):

PR #2558, opened 2023-11-28
Reformat the `GsonBuildConfig.java` template.

eamonnmcmanus:
Otherwise the spotless-maven-plugin checker doesn't like the code generated from it.


eamonnmcmanus:
I think we're going to have to disable Spotless for generated code in the `proto` module, but at least here there is a simple fix.

PR #2559, opened 2023-11-28
Disable Spotless format checking for some generated files.

eamonnmcmanus:
Files generated by the protobuf compiler are not going to conform to the expected format. It's not useful to check them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -168,11 +168,13 @@
           </formats>
 
           <java>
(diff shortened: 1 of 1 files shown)


Marcono1234:
This was necessary due to the plugin version update in #2557, right?

Apparently that is a known issue in version 2.41.0: https://github.com/diffplug/spotless/issues/1914

eamonnmcmanus:
Yes. We may be able to revert this exclusion if that is fixed in a future update.

PR #2588, opened 2023-12-27
Add nesting limit for `JsonReader`

Marcono1234:
### Purpose
Add nesting limit for `JsonReader`

### Description
For now don't expose this as additional GsonBuilder method assuming that the default nesting limit is high enough for most users. Otherwise users can first obtain a JsonReader from `Gson.newJsonReader` and then set a custom nesting limit.

The reasons why this pull request makes the nesting limit configurable at all are
- that some users might want to even further decrease the limit, for example if their call stack is already quite large before Gson is called
- and because `JsonReader` can be used directly in a way which does not involve recursion, so for these advanced use cases there might be no limit needed.
It appears there are rare cases where users need to handle more deeply nested JSON data; here are examples from other programming languages: https://github.com/valyala/fastjson/issues/65, https://github.com/serde-rs/json/issues/162#issuecomment-713662554

### Checklist


Marcono1234:
@eamonnmcmanus regarding https://github.com/google/gson/discussions/2632#discussioncomment-9481105: Is there some general problem with the changes or should I change something?

eamonnmcmanus:
> @eamonnmcmanus regarding [#2632 (reply in thread)](https://github.com/google/gson/discussions/2632#discussioncomment-9481105): Is there some general problem with this concept of a nesting limit or should I change something?

Last time I looked, this triggered a test failure in a Google-internal test. I think what I might do is patch Google's internal copy of the Gson source code to increase the threshold so that test passes. I think the limit of 255 is reasonable for the public version. I'm looking into this now.

eamonnmcmanus (approved):

PR #2598, opened 2024-01-14
Upgrade GitHub actions & pin to commit hash

Marcono1234:
### Purpose
- Upgrade GitHub actions & pin to commit hash
- Enable Dependabot for GitHub actions

### Description
See the commit messages for additional information.

Pinning the GitHub actions to a commit hash is [recommended for increased security](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), in case a malicious user gains write access to the repository of an action and changes the Git tags. Though for most of the actions we use the risk of this seems rather low since the actions are maintained by GitHub.

Using Dependabot for the GitHub actions makes sure no deprecated versions are used by accident (such as the recently [deprecated CodeQL Action v2](https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/)). We will have to make sure though that Dependabot version updates don't contain any breaking changes, but hopefully the maintainers follow SemVer, so this should hopefully be easy to notice.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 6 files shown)


eamonnmcmanus (approved):
Thanks!

PR #2599, opened 2024-01-14
Fix $Gson$Types equals method for TypeVariable when its generic declaration is not a Class

d-william:
### Purpose
When $Gson$Types equals method check two `TypeVariable`, it can return false while `Objects#equals(Object, Object)` returns true.

### Description
```java
public class Main {

    private static class MyClass {

        public <T> T method() { return null; }

    }

    public static void main(String[] args) throws NoSuchMethodException {
        Method m1 = MyClass.class.getMethod("method");
        Method m2 = MyClass.class.getMethod("method");

        Type rt1 = m1.getGenericReturnType();
        Type rt2 = m2.getGenericReturnType();

        TypeToken<?> t1 = TypeToken.get(rt1);
        TypeToken<?> t2 = TypeToken.get(rt2);

        System.out.println(Objects.equals(rt1, rt2)); // true
        System.out.println(Objects.equals(t1, t2));  // false
    }

}
```
The root cause is [here](https://github.com/google/gson/blob/51cce39e0064a755d32f820670770f82519039de/gson/src/main/java/com/google/gson/internal/%24Gson%24Types.java#L220).
For `Class` generic declaration, reference equality is enough because there is one reference of each class.
But for `Method`generic declaration, it is not, `Class#getMethod(String, Class<?>...)` returns a new instance of the `Method` per call.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/$Gson$Types.java
+++ b/gson/src/main/java/com/google/gson/internal/$Gson$Types.java
@@ -217,7 +217,7 @@ public final class $Gson$Types {
       }
       TypeVariable<?> va = (TypeVariable<?>) a;
       TypeVariable<?> vb = (TypeVariable<?>) b;
-      return va.getGenericDeclaration() == vb.getGenericDeclaration()
+      return Objects.equals(va.getGenericDeclaration(), vb.getGenericDeclaration())
           && va.getName().equals(vb.getName());
 
     } else {
--- a/gson/src/test/java/com/google/gson/internal/GsonTypesTest.java
+++ b/gson/src/test/java/com/google/gson/internal/GsonTypesTest.java
@@ -19,6 +19,8 @@ package com.google.gson.internal;
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertThrows;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


Marcono1234:
Thanks for this pull request, and for fixing this bug!

Could you please add a unit test similar to what you have shown in your example code (probably either to `GsonTypesTest` or `TypeTokenTest`)? I assume you can add this dummy class with generic method as local class within the test method, so that it does not have to be directly inside the enclosing test class.

Also, out of curiosity, how did you notice this bug? Creating `TypeToken`s which capture type variables is discouraged and will be disallowed in future Gson versions (see #2376). But obtaining it in the way shown in your example will still work (because it provides no false sense of type safety).

d-william:
I should be able to add the tests you request in a few days :)

To answer your question, I am implementing JSON objects representing method signature with the type information of a potentially declaring generic class.

```java
class SimpleClass {

    public void method1(String str) {...} 

} 
```

Will give :
```json
{
    "name" : "method1", 
    "parameters" : [
        {
            "name" : "str",
            "type" : "java.lang.String",
        } 
    ] 
```

```java
class GenericClass<T> {

    public void method2(T value) {...} 

} 
```

GenericClass<Number> will give :
```json
{
    "name" : "method2", 
    "parameters" : [
        {
            "name" : "value",
            "type" : "java.lang.Number",
        } 
    ] 
```

```java
class OtherClass {

    public <T> void method3(T param) {...} 

} 
```

Will give :
```json
{
    "name" : "method3", 
    "parameters" : [
        {
            "name" : "param",
            "type" : { "name" : "T", "declaring" : "package.OtherClass#method3..." } ,
        } 
    ] 
```

And with the POJO of this JSON objects, I am led to check equality between them.
I encountered this problem when trying with two objects from scenario 3, because internally, parameter types use TypeToken.

Marcono1234:
Thanks for the explanation. I guess Gson's `TypeToken` can be used for that use case, but it is mainly designed to be used with the Gson API. Maybe it would be better to use a more general "type token" class, such as Guava's [`com.google.common.reflect.TypeToken`](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/reflect/TypeToken.html) which also offers more functionality. What do you think?

Nonetheless, the bug here with the Gson `TypeToken` should still be fixed.

d-william:
Test added :)

Marcono1234 (approved):
Thanks a lot! The changes look good to me.

It might take some time though until this is merged because I think the direct members of this project are quite busy currently.

Marcono1234 (approved):

eamonnmcmanus (approved):

PR #2609, opened 2024-02-03
Improve location information for `JsonDeserializer` exceptions

### LINKED ISSUE #1135: Path $ is not provided when there is JsonSyntaxException with Custom Deserializer
With the latest version of the Gson library I do get detailed path to the json node that is causing the error.
However, this path $ information seems be absent when there is JsonSyntaxException in the custom deserializer

Following is the close representation of the deserializer i have implemented

```java
public class ScoresDeserializer implements JsonDeserializer<Event.ScoresWrapper> {

    Map<String, Type> mapScoreType;

    public MatchScoresDeserializer() {

        mapScoreType = new HashMap<>();
        mapScoreType.put("a", Event.A.class);
        mapScoreType.put("b", Event.B.class);
        mapScoreType.put("c", Event.C.class);

    }

    @Override
    public Event.ScoresWrapper deserialize(JsonElement jsonElement,
                                           Type type, JsonDeserializationContext jsonDeserializationContext) throws JsonParseException {

        JsonObject scoreJsonObj = jsonElement.getAsJsonObject();
        String scoreTypeValue = scoreJsonObj.getAsJsonPrimitive("scoreType").getAsString();
        JsonElement scoreData = scoreJsonObj.get("data");

        Type scoreType = mapScoreType.get(scoreTypeValue);
        if(scoreType == null || scoreData.isJsonNull()) {
            return null;
        }        
        Event.ScoreType scoreTypeEnum = jsonDeserializationContext.deserialize(scoreJsonObj.get("scoreType"), Event.ScoreType.class);
        Object scoreDataObj = jsonDeserializationContext.deserialize(scoreData, scoreType);
        return new Event.ScoresWrapper(scoreTypeEnum, scoreDataObj);
    }
}

//Exception thrown when the Json string doesnt match with the object being constructed
java.lang.IllegalStateException: Expected BEGIN_ARRAY but was BEGIN_OBJECT at path $
                                                            com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was BEGIN_OBJECT at path $
```

Gson Version: 2.8.+

Thanks in advance

Marcono1234:
### Purpose
Resolves #1135

### Description
When `JsonDeserializer.deserialize` throws an exception, wrap it and include the path of the `JsonElement` to make troubleshooting easier.

Risks of this approach:
- Troubleshooting might still be difficult / confusing because it is not obvious that the path reported by the wrapped exception is relative
- User code might expect the exception to be propagated without any wrapping
- If users use a lot `JsonDeserializer`s and a deeply nested one throws an exception, it would create a large exception chain, which is difficult to read and could be a performance issue

:warning: So I am not completely sure if this is worth it. Feel free to reject this pull request (or suggest another solution).

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
@@ -92,7 +92,15 @@ public final class TreeTypeAdapter<T> extends SerializationDelegatingTypeAdapter
     if (nullSafe && value.isJsonNull()) {
       return null;
     }
-    return deserializer.deserialize(value, typeToken.getType(), context);
+
+    try {
+      return deserializer.deserialize(value, typeToken.getType(), context);
+    } catch (Exception e) {
+      // Wrap and include path of `value`; otherwise exception would only include path relative
+      // to `value`, making troubleshooting difficult
+      String valuePath = in.getPreviousPath();
+      throw new JsonParseException("Deserialization failed at path " + valuePath, e);
+    }
   }
 
   @Override
--- a/gson/src/test/java/com/google/gson/functional/CustomDeserializerTest.java
+++ b/gson/src/test/java/com/google/gson/functional/CustomDeserializerTest.java
@@ -17,6 +17,7 @@
 package com.google.gson.functional;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -27,7 +28,10 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.common.TestTypes.Base;
 import com.google.gson.common.TestTypes.ClassWithBaseField;
+import com.google.gson.reflect.TypeToken;
 import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/internal/bind/TreeTypeAdapter.java
```diff
@@ -92,7 +92,15 @@ public T read(JsonReader in) throws IOException {
     if (nullSafe && value.isJsonNull()) {
       return null;
     }
-    return deserializer.deserialize(value, typeToken.getType(), context);
+
+    try {
+      return deserializer.deserialize(value, typeToken.getType(), context);
+    } catch (Exception e) {
```
eamonnmcmanus, line 98: My main concern here is that `JsonDeserializer` is an interface that users can and do implement themselves. That means we don't know what sort of unchecked exception might be thrown here, and by changing it to `JsonParseException` we might change the functioning of existing code. Perhaps if we only catch `JsonParseException`?

I'm going to run this change through all of Google's tests to see what happens.

Marcono1234: > Perhaps if we only catch `JsonParseException`?

Sounds reasonable, that will then also cover the subclasses `JsonIOException` and `JsonSyntaxException`.

And thanks for testing this!

However, my concern is still that users might only use `JsonDeserializer` and then could end up with for example 10 chained exceptions when deserialization fails for a nested value.
So I really wouldn't mind if this change is not integrated, if the inconvenience of incomplete JSON path information does not justify risking other issues.

eamonnmcmanus: So the test results are in, and it's not great. Over 30 failures spread across a number of different projects. Quite a lot of tests looking for the exact exception message you get now, which is arguably overspecific. But some others expecting `JsonSyntaxException` where now they get `JsonParseException`. In some cases it is just `assertThrows` or the like, but in others people are catching only `JsonSyntaxException` in production code, and that no longer works if we rethrow as `JsonParseException`. While we can imagine various workarounds, I think this change is probably more trouble than it's worth?

PR #2614, opened 2024-02-09
Fix compilation warnings & add test

Marcono1234:
### Purpose
Follow-up for #2364

### Description
Unfortunately #2561 had introduced additional checks after #2364 had been last updated; so when it was now merged to `main` these checks caused failures for the new code.


Marcono1234:
Note: The "Check API compatibility" workflow fails because it tries to build the old broken state from `main`. That will be resolved once this is merged.

eamonnmcmanus (approved):
Thanks!

PR #2641, opened 2024-02-29
Create grouped Dependabot pull requests

Marcono1234:
### Purpose
Group Dependabot pull requests instead of creating a separate one for each update, see #2639

I think the syntax should be correct, but I haven't tested it yet. We will probably see within the next month whether this works as desired.


eamonnmcmanus (approved):
Let's give it a try!

PR #2648, opened 2024-03-17
Support strict mode for `JsonParser`

Marcono1234:
### Purpose
Support strict parsing for `JsonParser`

### Description
Since the `Strictness` API (#2323) has not been released yet, we can fix the parsing of `JsonParser` in the same backward compatible way we fixed it for the `Gson` methods:
- If the strictness is `LEGACY_STRICT` (the default) we parse with `LENIENT`
- But if the user has explicitly set `STRICT` (which will only be possible now with the `Strictness` API) we respect this setting and parse in strict mode

This pull request also changes the `getStrictness() == Strictness.LEGACY_STRICT` checks in `Gson` to make the intention clearer that this is for backward compatibility. Behavior-wise they still act like before:
- `LENIENT` -> `LENIENT` (but is not explicitly overwritten anymore now)
- `LEGACY_STRICT` -> `LENIENT`
- `STRICT` -> `STRICT`

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -930,7 +930,9 @@ public final class Gson {
     Strictness oldStrictness = writer.getStrictness();
     if (this.strictness != null) {
       writer.setStrictness(this.strictness);
-    } else if (writer.getStrictness() != Strictness.STRICT) {
+    }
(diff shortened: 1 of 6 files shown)


eamonnmcmanus (changes requested):
Good catch! I'm glad you thought of this before we had frozen the new API.

Just one tiny formatting concern.

eamonnmcmanus (approved):

Issue #1972, opened 2021-09-23
No example in JsonParser/JsonElement specifications


Chealer:
I never used Gson, but it seems `JsonParser` and `JsonElement` would best mimic how `org.json` works. It would be clearer if the specifications for at least one of these could include an example demonstrating their use though.

An example of JsonParser is available at https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/extras/examples/rawcollections/RawCollectionsExample.java

Marcono1234:
It is unfortunately not very well documented, but `JsonParser` parses the JSON data in lenient mode (e.g. it allows single quotes and non-finite floating point values).

The safest way which also respects the lenient setting of the `JsonReader` is to obtain the TypeAdapter for JsonElement from a Gson instance and then use the TypeAdapter:
```java
TypeAdapter<JsonElement> adapter = new Gson().getAdapter(JsonElement.class);

JsonArray array = new JsonArray();
array.add(1);
String json = adapter.toJson(array);
System.out.println(json);

JsonElement deserialized = adapter.fromJson("{\"a\":1}");
System.out.println(deserialized);
```

eamonnmcmanus:
PRs to improve documentation, including javadoc comments, are always welcome!

Issue #591, opened 2015-03-19
JsonElement.toString() output is unspecified


GoogleCodeExporter:
```
JsonElement.toString() Javadoc:

  /**
   * Returns a String representation of this element.
   */

The actual format of JsonElement.toString() is unspecified, but some 
applications would break if the output format were changed to something that 
isn't a valid JSON representation of that object.

Suggested fix:

To me it seems reasonable to expect toString() to produce valid JSON, not just 
some arbitrary programmer-friendly string representation.

The toString() method should state that the return value is a valid JSON 
representation of the object, and that details such as indentation are left 
unspecified. 

Subclasses like JsonArray and JsonObject should override toString() just to 
provide improved documentation with example return values.
```

Original issue reported on code.google.com by `hussdl` on 22 Aug 2014 at 10:09

PR #2654, opened 2024-03-20
Extend `JsonElement` documentation

Marcono1234:
### Purpose
Extends the `JsonElement` documentation, describing various ways to convert from / to `JsonElement`

Relates to #1972

### Description
Extends the `JsonElement` documentation to describe how to convert `JsonElement` from / to JSON, but also how to use `JsonElement` as JSON data and deserializing from / serializing to it.

Maybe the documentation provides too much information now though, potentially confusing users?

A fix for #591 has been implemented in the separate pull request #2659.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # v4.1.1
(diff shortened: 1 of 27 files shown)


eamonnmcmanus:
This looks good to me. I think it's valuable, because some aspects of `JsonElement` usage are not apparent to newcomers.

eamonnmcmanus (approved):

Marcono1234:
The changes were included as part of #2659, so this pull request is obsolete now.

PR #2664, opened 2024-04-02
Refactorings

pompeiifreckles:
### Checklist


eamonnmcmanus:
Could you give some background on what you are trying to achieve here? We are definitely not going to accept a huge PR like this without explanation, even if it were correct.

PR #2668, opened 2024-04-09
Added FormattedDouble Class

### LINKED ISSUE #1769: 00 as decimal places when serialize Double
Hi team, 

For a specific reason, I want 2 decimal places for my double when serializing into JSON.
Something like:

9900          --> 9900.00
9900.0       --> 9900.00
9900.1234 --> 9900.12

My best best result so far:

9900          --> 9900.0
9900.0       --> 9900.0
9900.1234 --> 9900.12

My DoubleAdapter.java, please advice. Thanks a lot!

`class DoubleAdapter extends TypeAdapter<Double>  {

    @Override
    public void write(JsonWriter jsonWriter, Double aDouble) throws IOException {
        DecimalFormat df = new DecimalFormat("0.00");
        jsonWriter.value(Double.valueOf(df.format(aDouble)));
    }

    @Override
    public Double read(JsonReader jsonReader) throws IOException {
        return jsonReader.nextDouble();
    }
}`

capak07:
### Description

- Fixes #1769 
- Added FormattedDouble Class that'll help write double numbers using gson

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/FormattedDouble.java
+++ b/gson/src/main/java/com/google/gson/internal/FormattedDouble.java
@@ -1,0 +1,66 @@
+public class FormattedDouble extends Number {
+    private static final long serialVersionUID = 1L;
+
+    // ThreadLocal because DecimalFormat is not thread-safe
+    private static final ThreadLocal<DecimalFormat> format =
+            // Specify DecimalFormatSymbols to make code independent from default Locale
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
This is just copied from a comment in #1769. It's not connected to anything, there are no tests, and it does not build because of `-source 7`.

PR #2669, opened 2024-04-23
Add fall-back support for "timeless" java.util.Date deserialization in DefaultDateTypeAdapter

MuffinTheMan:
### Purpose
Adds support for a deserialization scenario where a "timeless" `java.util.Date` was serialized like `"Jan 1, 1970"` (Gson's default behavior) but cannot then be deserialized. It seems to make sense that Gson's deserializer should be able to handle deserializing what it serialized itself by default.

### Description
This example demonstrates the issue:
```
@Test
void testSqlDateToUtilDate() throws SQLException {
    try (
        Connection c = ConnectionFactory.getConnection();
        Statement stmt = c.createStatement();
        ResultSet rs = stmt.executeQuery("SELECT '1970-01-01';")
    ) {
        rs.next();
        // rs::getDate() returns java.sql.Date, but that *is a* java.util.Date
        final java.util.Date date = rs.getDate(1);
        final Gson gson = new Gson();
        // This will fail without the fix, because it's serialized as "Jan 1, 1970" but unable
        // to be deserialized from that format. Since adding support for this is trivial, it
        // seems worthwhile. 👇
        assertThat(gson.fromJson(gson.toJson(date), java.util.Date.class)).isEqualTo(date);

        // On the other hand, this does work; however the declared type of `date` is `java.util.Date`,
        // so it would be helpful if it could be deserialized with that same class. The "realistic" scenario
        // is that you have a java.util.Date field on a class that is being populated from a DATE database
        // column (ie using `ResultSet::getDate()` like above); if the class is serialized/deserialized, it
        // will fail deserialization without this fix. 👇
        assertThat(gson.fromJson(gson.toJson(date), java.sql.Date.class)).isEqualTo(date);
    }
}
```

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/bind/DefaultDateTypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/DefaultDateTypeAdapterTest.java
@@ -119,6 +119,7 @@ public class DefaultDateTypeAdapterTest {
     Locale.setDefault(Locale.US);
     try {
       assertParsed("Jan 1, 1970 0:00:00 AM", DefaultDateTypeAdapter.DEFAULT_STYLE_FACTORY);
+      assertParsed("Jan 1, 1970", DefaultDateTypeAdapter.DEFAULT_STYLE_FACTORY);
       assertParsed(
           "1/1/70 0:00 AM", DateType.DATE.createAdapterFactory(DateFormat.SHORT, DateFormat.SHORT));
       assertParsed(


gson/src/main/java/com/google/gson/internal/bind/DefaultDateTypeAdapter.java
```diff
@@ -134,6 +134,8 @@ private DefaultDateTypeAdapter(DateType<T> dateType, int dateStyle, int timeStyl
     if (JavaVersion.isJava9OrLater()) {
       dateFormats.add(PreJava9DateFormatProvider.getUsDateTimeFormat(dateStyle, timeStyle));
     }
+    // Include fall-back for "timeless" Date (could be from a java.sql.Date, for example)
+    dateFormats.add(DateFormat.getDateInstance(dateStyle, Locale.US));
```
Marcono1234, line 138: This seems a bit brittle for the use case you are describing because the `java.sql.Date` adapter actually uses a hardcoded pattern (with default locale) and not a date style:
https://github.com/google/gson/blob/18b08922537c16b0de1d503c82307d613aae0138/gson/src/main/java/com/google/gson/internal/sql/SqlDateTypeAdapter.java#L52

Marcono1234 (commented):
Thanks for this pull request!

At least personally I am not sure if this is a change which should be made (but I am not a direct member of this project). In general the whole human-readable date formatting in JSON is not ideal, see for example #2472.

Maybe here it would be better if you registered a custom adapter for `java.util.Date` and / or `java.sql.Date` in your code to handle this. But I am not completely sure.

gson/src/main/java/com/google/gson/internal/bind/DefaultDateTypeAdapter.java
Marcono1234: > I have to basically copy the exact `DefaultDateTypeAdapter` and make this one change

You can avoid this by writing a [`TypeAdapterFactory`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/TypeAdapterFactory.html) which first tries to deserialize as `java.util.Date` and if that fails tries `java.sql.Date` instead:
<details>
<summary><code>DateSqlFallbackFactory</code> (click to expand)</summary>

Note that I haven't tested this extensively, and this implementation is also quite verbose. And if parsing fails, the JSON path in the exception message will not be that helpful, saying `$` (root element) instead of the actual path.

```java
class DateSqlFallbackFactory implements TypeAdapterFactory {
  @Override
    if (type.getRawType() != java.util.Date.class) {
      return null;
    }

    TypeAdapter<JsonElement> jsonElementAdapter = gson.getAdapter(JsonElement.class);
    TypeAdapter<Date> utilDateAdapter = gson.getDelegateAdapter(this, TypeToken.get(Date.class));
    TypeAdapter<java.sql.Date> sqlDateAdapter = gson.getAdapter(java.sql.Date.class);

    TypeAdapter<T> adapter = (TypeAdapter<T>) new TypeAdapter<Date>() {
      @Override
      public Date read(JsonReader in) throws IOException {
        // First read as JsonElement tree to be able to parse it twice (once as java.util.Date, once as
        // java.sql.Date) if necessary
        JsonElement json = jsonElementAdapter.read(in);
        try {
          return utilDateAdapter.fromJsonTree(json);
        }
        // Note: This makes assumptions about the exception type thrown by Gson's built-in
        // adapter for java.util.Date
        catch (JsonParseException e) {
          try {
            return sqlDateAdapter.fromJsonTree(json);
          } catch (Exception suppressed) {
            e.addSuppressed(suppressed);
            throw e;
          }
        }
      }

      @Override
      public void write(JsonWriter out, Date value) throws IOException {
        utilDateAdapter.write(out, value);
      }
    };
    return adapter;
  }
}
```

</details>

----

If you haven't stored the JSON data for any `java.util.Date` and SQL subclasses persistently yet, then maybe a reliable alternative would be to use [`GsonBuilder.setDateFormat(String)`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/GsonBuilder.html#setDateFormat(java.lang.String)) to specify a locale insensitive pattern (if possible).

Marcono1234:
@eamonnmcmanus, do you think we should add the `java.sql.Date` pattern (`new SimpleDateFormat("MMM d, yyyy")`) as fallback for deserialization to `DefaultDateTypeAdapter`? Similar to what is proposed currently by this pull request.

That pattern would then have to be enabled always, even if `SqlTypesSupport.SUPPORTS_SQL_TYPES` is `false`, in case users start relying on this pattern for deserialization even when not using `java.sql.Date`, to avoid failures when those users run on a JDK without SQL types.

PR #2674, opened 2024-05-09
Improved Long-Double Number Policy

ctasada:
### Purpose
Fixes a performance issue while using the `ToNumberPolicy.LONG_OR_DOUBLE` with `Double` values

### Description
The Parsing of a Double value was always executing a `Long.parseLong(value)`, which generated a `NumberFormatException`.

Identifying that a Number is a Double or a Long can be easily achieve (in a naive way) looking for the decimal separator.

This simple change avoids the extra `NumberFormatException`

A simple JUnit test, parsing a `Long` or a `Double` 10K times shows the next values:

* Double (old parsing): ~42 ms
* Double (new parsing): ~6 ms
* Long (old parsing): ~7 ms
* Long (new parsing): ~7 ms

As we can see, the parsing for `Long` values stays the same (±1ms), while the parsing for `Double` is dramatically improved.

Reducing the number of exceptions also has a positive side effect in memory consumption.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/ToNumberPolicy.java
+++ b/gson/src/main/java/com/google/gson/ToNumberPolicy.java
@@ -68,20 +68,28 @@ public enum ToNumberPolicy implements ToNumberStrategy {
     @Override
     public Number readNumber(JsonReader in) throws IOException, JsonParseException {
       String value = in.nextString();
-      try {
-        return Long.parseLong(value);
-      } catch (NumberFormatException longE) {
+      if (value.contains(".")) {
+        return parseAsDouble(value, in);
+      } else {
         try {
-          Double d = Double.valueOf(value);
-          if ((d.isInfinite() || d.isNaN()) && !in.isLenient()) {
-            throw new MalformedJsonException(
-                "JSON forbids NaN and infinities: " + d + "; at path " + in.getPreviousPath());
-          }
-          return d;
-        } catch (NumberFormatException doubleE) {
-          throw new JsonParseException(
-              "Cannot parse " + value + "; at path " + in.getPreviousPath(), doubleE);
+          return Long.parseLong(value);
+        } catch (NumberFormatException longE) {
+          return parseAsDouble(value, in);
         }
+      }
+    }
+
+    private Number parseAsDouble(String value, JsonReader in) throws IOException {
+      try {
+        Double d = Double.valueOf(value);
+        if ((d.isInfinite() || d.isNaN()) && !in.isLenient()) {
+          throw new MalformedJsonException(
+              "JSON forbids NaN and infinities: " + d + "; at path " + in.getPreviousPath());
+        }
+        return d;
+      } catch (NumberFormatException doubleE) {
+        throw new JsonParseException(
+            "Cannot parse " + value + "; at path " + in.getPreviousPath(), doubleE);
       }
     }
   },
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -68,22 +68,30 @@ public Number readNumber(JsonReader in) throws IOException {
     @Override
     public Number readNumber(JsonReader in) throws IOException, JsonParseException {
       String value = in.nextString();
-      try {
-        return Long.parseLong(value);
-      } catch (NumberFormatException longE) {
+      if (value.contains(".")) {
```
Marcono1234, line 71: Could also use `indexOf('.') >= 0` here, which might be more efficient due to using a `char`. But I am not sure if that is really more efficient and if that is worth it.

Marcono1234 (commented):
Thanks for this suggested improvement! What you are saying sounds reasonable to me, but I have not measured it yet myself.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
ctasada: @Marcono1234 the difference between `contains(".")` and `indexOf('.') >= 0` is very small, and nearly negligible. But you're right, `indexOf` is slightly faster. I just pushed the tweak.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -68,22 +68,30 @@ public Number readNumber(JsonReader in) throws IOException {
     @Override
     public Number readNumber(JsonReader in) throws IOException, JsonParseException {
       String value = in.nextString();
-      try {
-        return Long.parseLong(value);
-      } catch (NumberFormatException longE) {
+      if (value.indexOf('.') >= 0) {
```
eamonnmcmanus, line 71: I think this makes sense. We could also imagine a more thorough check or even a regular expression like `[-+]?[0-9]+`, but I think looking for just `'.'` is an excellent choice.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
```diff
@@ -68,22 +68,30 @@ public Number readNumber(JsonReader in) throws IOException {
...
-          if ((d.isInfinite() || d.isNaN()) && !in.isLenient()) {
-            throw new MalformedJsonException(
-                "JSON forbids NaN and infinities: " + d + "; at path " + in.getPreviousPath());
-          }
-          return d;
-        } catch (NumberFormatException doubleE) {
-          throw new JsonParseException(
-              "Cannot parse " + value + "; at path " + in.getPreviousPath(), doubleE);
+          return Long.parseLong(value);
+        } catch (NumberFormatException longE) {
```
eamonnmcmanus, line 76: Since you've moved the double parsing into a separate method (which I think is a good change), you no longer need to distinguish between `longE` and `doubleE`. I would suggest just calling them both `e`.

eamonnmcmanus (changes requested):
Thanks! The old code uses exceptions for control flow which is indeed bad, especially for performance. With this change we will _usually_ not have to do that.

gson/src/main/java/com/google/gson/ToNumberPolicy.java
ctasada: @eamonnmcmanus good call. Done

gson/src/main/java/com/google/gson/ToNumberPolicy.java
ctasada: Agreed. Since I'm looking for performance, the search for `'.'` is faster than applying a regexp. The `Double.valueOf()` will take care of anything else, but we only call it if there are chances of being a double

eamonnmcmanus (approved):

PR #2678, opened 2024-05-20
Add more Javadoc cross-links between methods

Marcono1234:
### Purpose
Add more Javadoc cross-links between methods, mainly between getters and setters

### Description
This hopefully makes it easier to find the corresponding getter / setter. Though on the other hand it also makes the documentation more verbose, so please let me know if you don't think this is needed.

### Checklist


eamonnmcmanus (approved):
This seems like a fine idea, and is fairly conventional I think.

PR #2684, opened 2024-05-24
Fix gha smells:

ceddy4395:
### Purpose
Hey! 🙂
I've made the following changes to your workflow:

- Use fixed version for runs-on argument
  - Using an explicit version for runs-on makes the developers more conscious about on what OS version they are testing/supporting. Furthermore, if breaking changes are introduced in the next version, they will not automatically break the pipeline.  
- Avoid jobs without timeouts
  - Jobs without a timeout can cause runners to be occupied when some process falls into an infinite loop or has multiple retries to perform a certain task, which will inevitably fail. Furthermore, it is also useful to be notified when a test-suite all of a sudden takes a lot longer than it used to, considering keeping tests fast is a good practice.
- Steps should only perform a single command
  - Making sure every step has a single responsibility makes it easier to understand the workflow and, in case a workflow fails, it makes it clearer where and why it failed. 

### Checklist


.github/workflows/build.yml
```diff
@@ -65,11 +65,12 @@ jobs:
       - name: "Verify no plugin issues"
         run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress
 
-      - name: "Verify reproducible build"
+      - name: "Do clean install of dependencies"
         # See https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
         run: |
           mvn clean install --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests
```
Marcono1234, line 71: Could simplify this by moving the command in the same line:
```suggestion
        run: mvn clean install --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests
```

.github/workflows/build.yml
```diff
@@ -65,11 +65,12 @@ jobs:
...
           # Run with `-Dbuildinfo.attach=false`; otherwise `artifact:compare` fails because it creates a `.buildinfo` file which
           # erroneously references the existing `.buildinfo` file (respectively because it is overwriting it, a file with size 0)
           # See https://issues.apache.org/jira/browse/MARTIFACT-57
```
Marcono1234, line 74: Since these `# Run with ...` comments are only relevant for the `mvn` command below, it would be good to move them down (and adjust their indentation), so that they are between the `name` and the `run` below.

.github/workflows/build.yml
```diff
@@ -65,11 +65,12 @@ jobs:
       - name: "Verify no plugin issues"
         run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress
 
-      - name: "Verify reproducible build"
+      - name: "Do clean install of dependencies"
         # See https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
```
Marcono1234, line 69: Might be good to move this `# See ...` comment above the `name: "Do clean ...` since it applies to both steps now and not just the `mvn clean install`.

.github/workflows/build.yml
```diff
@@ -65,11 +65,12 @@ jobs:
       - name: "Verify no plugin issues"
         run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress
 
-      - name: "Verify reproducible build"
+      - name: "Do clean install of dependencies"
```
Marcono1234, line 68: This should probably rather say "Build project".

In Maven `install` is a lifecycle phase, and it includes all previous phases and therefore performs a full build (except for deployment), see https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-lifecycle-is-made-up-of-phases

Marcono1234 (commented):
Thanks!

To me this looks reasonable. I have added a few review comments but they are mainly about formatting.

A few general points though:
- The hardcoded Ubuntu version might add some (very small) maintenance overhead due to having to remember to update it in the future
But I assume that is acceptable.
- The timeouts might be a bit too low, especially since the run duration probably depends on external factors we cannot control
  - "GraalVM Native Image test" should maybe have 10 minutes as timeout; for example here is a run where it took 4 minutes: https://github.com/google/gson/actions/runs/8910979694/job/24471354114
  - the other jobs should probably have 5 minutes as timeout, to be safe
- It probably makes sense to apply these changes also to the other workflows under `.github/workflows`

.github/workflows/build.yml
ceddy4395: I've fixed this. Also for the other files I've made the same change to keep everything consistent.

.github/workflows/build.yml
ceddy4395: Ah yes moved them

.github/workflows/build.yml
ceddy4395: Ah oke yes makes sense. I've updated this.

ceddy4395:
@Marcono1234 Thank you for the review! I've made the changes you've suggested, including bumping the timeouts. 

I've also made the same fixes to the other workflow files!

.github/workflows/check-api-compatibility.yml
```diff
@@ -22,21 +22,19 @@ jobs:
...
           cache: 'maven'
 
-      - name: Build old version
-        run: |
-          cd gson-old-japicmp
-          # Set dummy version
-          mvn --batch-mode --no-transfer-progress org.codehaus.mojo:versions-maven-plugin:2.11.0:set -DnewVersion=JAPICMP-OLD
-          # Install artifacts with dummy version in local repository; used later by Maven plugin for comparison
-          mvn --batch-mode --no-transfer-progress install -DskipTests
+      - name: Set dummy version
```
Marcono1234, line 25: Maybe the "used later by Maven plugin for comparison" should be preserved in some way (e.g. as comment) to make it clearer why this is done.

What about this?
```suggestion
      # First sets a dummy project version, then installs the artifacts with that version in the local
      # Maven repository, and in a later step uses it as 'old version' for the API compatibility check.
      - name: Set dummy version
```

Marcono1234:
> Would Dependabot or something else notify us if the latest supported version changes?

No, that does not seem to be supported yet, see https://github.com/dependabot/dependabot-core/issues/7182.

eamonnmcmanus:
> > Would Dependabot or something else notify us if the latest supported version changes?
> 
> No, that does not seem to be supported yet, see [dependabot/dependabot-core#7182](https://github.com/dependabot/dependabot-core/issues/7182).

Thanks for that pointer. In that case, I'd suggest we leave this part as it is (with `ubuntu-latest`) until Dependabot does have that support.

PR #2685, opened 2024-05-26
Add JPMS module test

Marcono1234:
### Purpose
Adds a Maven module to test Gson's `module-info.class`

### Description
Previously the build would not have detected if `module-info.class` was missing, or incorrectly configured. Therefore this pull request adds a new Maven module to test using the `module-info.class`.

The new integration tests might not work when run from the IDE, but they will work when run with `mvn ...` from the command line.

Please let me know if you think it is too verbose, if the test should cover additional cases, or if there might be a better way to implement this. Any feedback is appreciated.

To simplify disabling certain Maven plugins (e.g. install, gpg and deploy) for the integration test modules I have instead added two properties in the parent project which can disable the plugins. See the second commit.
I have done a short test of the release workflow (but without gpg), and everything still seems to work as desired: The build succeeds and only `gson-parent` and `gson` are deployed.

### Checklist

### DIFF
--- a/jpms-test/src/main/java/module-info.java
+++ b/jpms-test/src/main/java/module-info.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2024 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Dummy module to prevent Maven Compiler Plugin from failing if {@code module-info.java} exists
+ * only in test sources:
+ *
+ * <blockquote>
+ *
+ * Can't compile test sources when main sources are missing a module descriptor
+ *
+ * </blockquote>
+ */
+module com.google.gson.dummy {}
--- a/jpms-test/src/test/java/com/google/gson/jpms_test/ModuleTest.java
+++ b/jpms-test/src/test/java/com/google/gson/jpms_test/ModuleTest.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2024 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.gson.jpms_test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.google.gson.Gson;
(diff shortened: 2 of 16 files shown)


jpms-test/src/main/java/module-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Dummy module to prevent Maven Compiler Plugin from failing if {@code module-info.java} exists
+ * only in test sources:
+ *
+ * <blockquote>
+ *
+ * Can't compile test sources when main sources are missing a module descriptor
```
Marcono1234, line 23: Maybe there is another way to work around this; the Maven Compiler Plugin [checks if the main output directory exists](https://github.com/apache/maven-compiler-plugin/blob/maven-compiler-plugin-3.13.0/src/main/java/org/apache/maven/plugin/compiler/TestCompilerMojo.java#L364-L367), but maybe that check is a bit brittle?

https://lists.apache.org/thread/c3f71v8tf3643dto6bwf3ctcos4gt5cq suggests it might even depend on the JUnit version being used (?).

So for now having this dummy module declaration might be the most reliable workaround.

pom.xml
```diff
@@ -28,6 +28,7 @@
 
   <modules>
     <module>gson</module>
+    <module>jpms-test</module>
     <module>graal-native-image-test</module>
     <module>shrinker-test</module>
```
Marcono1234, line 33: Maybe it would make sense to change their names to be `test-...` instead of `...-test` so that they are grouped in the file system, and it is more obvious that they are tests.

What do you think?

jpms-test/src/test/java/com/google/gson/jpms_test/ModuleTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    Set<Exports> packageExports = moduleDescriptor.exports();
+    assertThat(packageExports.stream().map(Exports::source))
+        .containsExactly(
+            "com.google.gson",
+            "com.google.gson.annotations",
+            "com.google.gson.reflect",
+            "com.google.gson.stream");
+    // Gson currently does not export packages to specific modules only
+    assertThat(packageExports.stream().filter(Exports::isQualified)).isEmpty();
```
eamonnmcmanus, line 78: I think `noneMatch` would be a bit more succinct.

pom.xml
eamonnmcmanus: I think that would be an excellent idea for a follow-up PR.

eamonnmcmanus (commented):
This is great, thanks very much for doing it!

Are you able to test dry-run releasing? No problem if not, we'll find out soon enough and I'll fix any issues that arise.

jpms-test/src/test/java/com/google/gson/jpms_test/ModuleTest.java
Marcono1234: Do you mean `assertThat(stream.noneMatch(...)).isTrue()`? (Or is there a Truth method I am overlooking?)

The disadvantage might be that if there are actually matching packages, the assertion failure message would just say that `true` was expected but it was `false`, whereas this approach should (hopefully) produce a more helpful assertion failure message.

Marcono1234:
> Are you able to test dry-run releasing? No problem if not, we'll find out soon enough and I'll fix any issues that arise.

I tried the approach described in https://github.com/google/gson/blob/main/ReleaseProcess.md#testing-maven-release-workflow-locally (except for GPG signing), and it looks like everything was still working. So I hope this shouldn't break anything.

But yes, in the worst case we notice it for the next release... (or hopefully earlier)

(Also sorry that I had missed that the GPG plugin was not disabled for the graal-native-image-test module, which you then fixed in #2675.)

jpms-test/src/test/java/com/google/gson/jpms_test/ModuleTest.java
eamonnmcmanus: Yes, good point. It's better the way it is.

eamonnmcmanus (approved):

pom.xml
Marcono1234: Have created #2691

PR #2687, opened 2024-05-26
Migrate all tests to Truth & use `assertThrows`

Marcono1234:
### Purpose
Migrate all tests to Truth & use `assertThrows`

### Description
- Migrates all remaining tests to use Truth instead of JUnit for assertions
- Refactors all test code from `try { ...; fail(); } catch (expected) { }` to `assertThrows`
- Improves some tests

### Checklist


gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
```diff
@@ -532,159 +528,167 @@ private static String extractElementFromArray(String json) {
...
+                + " use GsonBuilder.serializeSpecialFloatingPointValues() method.");
   }
 
   @Test
   public void testDoubleNaNSerialization() {
     Gson gson = new GsonBuilder().serializeSpecialFloatingPointValues().create();
-    double nan = Double.NaN;
-    assertThat(gson.toJson(nan)).isEqualTo("NaN");
     assertThat(gson.toJson(Double.NaN)).isEqualTo("NaN");
+    assertThat(gson.toJson((Double) Double.NaN)).isEqualTo("NaN");
```
Marcono1234, line 550: As side note: It seems all these `Float` and `Double` tests here were flawed before:
The floating point constants such as `Double.NaN` are of the primitive type (e.g. `double` instead of `Double`) so these tests were previously calling `toJson(...)` with a `double` twice, instead of once with `double` and once with `Double`.

Though even now the code is a bit redundant because there is only `toJson(Object)`, so even if called with primitive `double` it will be implicitly boxed to a `Double`.

eamonnmcmanus: Yeah, I would be inclined to delete the second one since they are indeed both calling the same method with the same value.

eamonnmcmanus (changes requested):
This is fantastic! Thanks for the many small improvements throughout, beyond just the mechanical transformation into `assertThat` and `assertThrows`.

Just a couple of very minor things.

gson/src/test/java/com/google/gson/functional/JsonParserTest.java
```diff
@@ -112,11 +104,8 @@ public void testBadFieldTypeForDeserializingCustomTree() {
     obj.add("primitive1", primitive1);
     obj.add("primitive2", array);
 
-    try {
-      gson.fromJson(obj, Nested.class);
-      fail("Nested has field BagOfPrimitives which is not an array");
-    } catch (JsonParseException expected) {
-    }
+    // Nested has field BagOfPrimitives which is not an array
```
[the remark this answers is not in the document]
Marcono1234: The test was previously only calling `fail(...)` with that message. So if the test unexpectedly did not throw an exception, it would have failed with that message.

But it did not perform any assertions on the message of the expected thrown exception.

Or am I misunderstanding you?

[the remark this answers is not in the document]
eamonnmcmanus: Dunno what I was thinking.

gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
Marcono1234: I have changed it now to use `Gson.toJson(Object, Class)` to cover both `toJson(..., double.class)` and `toJson(..., Double.class)` (which is equivalent to the existing `toJson(...)`), and same for `float`.

Marcono1234:
Sorry for the delay; I have added a few more test improvements. This pull request should now be ready.

eamonnmcmanus (approved):

PR #2690, opened 2024-05-29
Support building with JDK 21.

eamonnmcmanus:
I think the main reason we weren't doing this before is that the ProGuard plugin didn't support it. Apparently it does now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,7 @@ jobs:
     name: "Build on JDK ${{ matrix.java }}"
     strategy:
       matrix:
-        java: [ 11, 17 ]
-        # Custom JDK 21 configuration
-        include:
-          - java: 21
-            # Disable Enforcer check which (intentionally) prevents using JDK 21 for building
-            extra-mvn-args: -Denforcer.fail=false
+        java: [ 11, 17, 21 ]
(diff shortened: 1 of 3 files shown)


eamonnmcmanus:
Hmm, this is looking more complicated than I thought.

We are not planning to build using JDK 21 for releases any time soon, but it would be more convenient for testing if we supported building there. The current failures don't prevent testing but make for an awkward setup.

Marcono1234:
I think all plugins should support JDK 21 now. The main problem with JDK 21 is that it does not support Java 7 as target version.
That is why the `pom.xml` currently increases the target version to Java 8 when building with JDK 21. And it is also the reason why building with JDK 21 is artificially prevented with the Enforcer Plugin at the moment, to avoid that users accidentally build a Gson release version which targets Java 8 instead of 7. See also #2552.

Most likely the check-api-compatibility failure is unrelated; maybe #2685 introduced some issues I forgot to consider, I will have a look. The problem with the current check-api-compatibility setup is that these issues often only occur for the next pull request..., sorry.

Marcono1234:
Also to clarify, CI is already building with JDK 21, see #2552. It just disables the Enforcer Plugin check which intentionally prevents this by default.

eamonnmcmanus:
I think this is basically obsolete.

PR #2691, opened 2024-05-30
Rename Maven integration test modules

Marcono1234:
### Purpose
Renames all Maven integration test modules to have the name `test-...`

See also https://github.com/google/gson/pull/2685#discussion_r1614974142

### Description
The new naming scheme `test-...` groups the modules in the file system and the IDE and therefore hopefully makes it easier to quickly see which modules are integration tests.


eamonnmcmanus:
Is the "Check API compatibility" failure related to the changes here, or is it already failing?

eamonnmcmanus (approved):
Thanks!

PR #2692, opened 2024-05-30
Run japicmp against last release instead of previous commit

Marcono1234:
### Purpose
Run japicmp against last release instead of previous commit

### Description
This will also fix the recent build failures, e.g. in #2690, which are caused by the currently used `JAPICMP-OLD` version which seems to cause issues when the new JPMS module test (#2685) tries to use it as version for the `module-info.class`:
> Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project jpms-test: Fatal error compiling: error: bad value for --module-version option: 'JAPICMP-OLD'

Advantages:
- Simplifies setup (no need to build a 'JAPICMP-OLD' version anymore)
This also allows easily running it locally now; though it is currently not part of the normal Maven build to allow having a separate GitHub workflow which also separately attaches the API change report as artifact.
- No spurious incompatibility failures anymore when API is changed which was added in the current SNAPSHOT by a previous pull request (but had not been released yet)

Disadvantages:
- japicmp cannot be used for not published modules 'extras' and 'proto' anymore
- Minor: For pull requests the API diff report generated by the GitHub workflow and attached as artifact now contains the diff to the last release and not the changes introduced by the pull request (e.g. added methods, ...)
- If we ever intentionally decide to perform a breaking change (or if there is a japicmp bug), the workflow will keep failing for all subsequent pull requests until the next release

Here is also a small demo showing that this still works (performs incompatible change of `JsonArray#add` return type): https://github.com/Marcono1234/gson/actions/runs/9310443366

### DIFF
--- a/.github/workflows/check-api-compatibility.yml
+++ b/.github/workflows/check-api-compatibility.yml
@@ -9,11 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout old version
-        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b  # v4.1.4
-        with:
-          ref: ${{ github.event.pull_request.base.sha }}
-          path: 'gson-old-japicmp'
+      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b  # v4.1.4
 
       - name: Set up JDK 11
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9  # v4.2.1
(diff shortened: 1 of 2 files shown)


Marcono1234:
Note that this is most likely not the only solution to the current build failures. Another alternative would be to replace `JAPICMP-OLD` with a different dummy version number, for example something like `0.0.0-JAPICMP-OLD` might work.

Marcono1234:
Have marked this as Draft now again because I am not really sure whether the current or this approach here is better.

Marcono1234:
I am going to close this. Originally when I started creating this pull request, I thought the advantages outweighed the disadvantages. But afterwards I became aware of more potential disadvantages (which I added to the description of this pull request), and I am not sure anymore now if one approach is really better than the other.

Therefore it might be better to keep the current approach for now.

PR #2694, opened 2024-06-01
Fix API compatibility workflow failing when building previous version

Marcono1234:
### Purpose
Fixes the current API compatibility workflows failures

### Description
The workflow is currently failing for the jpms-test module:
> Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project jpms-test:
> Fatal error compiling: error: bad value for --module-version option: 'JAPICMP-OLD'

Therefore this pull request changes the dummy version number format, so that it is not rejected anymore.

### DIFF
--- a/.github/workflows/check-api-compatibility.yml
+++ b/.github/workflows/check-api-compatibility.yml
@@ -8,8 +8,17 @@ jobs:
   check-api-compatibility:
     runs-on: ubuntu-latest
 
+    # This setup tries to determine API incompatibility only for the changes introduced by the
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Thanks!

It seems like the API compatibility workflow is _not_ failing with this PR. Not sure if you want to update the description.

Marcono1234:
Oh you are right; I forgot to consider that since this changes the workflow, it applies directly (and not delayed only on the next pull request).

eamonnmcmanus (approved):

PR #2702, opened 2024-06-15
Improve `JsonWriter#value(Number)` performance

Marcono1234:
### Purpose
Improve `JsonWriter#value(Number)` performance for JDK number types

### Description
For JDK number types other than `Float` and `Double` there is no need to check if the number string is NaN or Infinity.

This was an oversight when I originally added the validation in #2060.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -646,9 +646,9 @@ public class JsonWriter implements Closeable, Flushable {
         if (strictness != Strictness.LENIENT) {
           throw new IllegalArgumentException("Numeric values must be finite, but was " + string);
         }
-      } else if (!(numberClass == Float.class
-          || numberClass == Double.class
-          || VALID_JSON_NUMBER_PATTERN.matcher(string).matches())) {
+      } else if (numberClass != Float.class
+          && numberClass != Double.class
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -638,15 +638,17 @@ public JsonWriter value(Number value) throws IOException {
...
+
+    if (!alwaysCreatesValidJsonNumber(numberClass)) {
       // Validate that string is valid before writing it directly to JSON output
-      if (!isTrustedNumberType(numberClass)
-          && !VALID_JSON_NUMBER_PATTERN.matcher(string).matches()) {
+      if (string.equals("-Infinity") || string.equals("Infinity") || string.equals("NaN")) {
+        if (strictness != Strictness.LENIENT) {
+          throw new IllegalArgumentException("Numeric values must be finite, but was " + string);
+        }
+      } else if (!(numberClass == Float.class
```
eamonnmcmanus, line 649: I think I would be inclined to apply De Morgan here, similar to the way the condition was before:
```java
if (numberClass != Float.class
    && numberClass != Double.class
    && !VALID_JSON_NUMBER_PATTERN.matcher(string).matches())
```

eamonnmcmanus (changes requested):
This looks like a good improvement. One very nitpicky comment.

eamonnmcmanus (approved):

PR #2706, opened 2024-06-25
Fix typeArguments array not being cloned when resolving ParameterizedType with changed owner

TBlueF:
### Purpose
This PR fixes an inconsistency in `$Gson$Types#resolve()` where the `ParameterizedType#getActualTypeArguments()` array is only cloned when the resolved owner-type has not changed from the original owner-type.

### Description
In most cases this is a non-issue, as usually all `ParameterizedType` implementations clone the array themselves when invoking the `getActualTypeArguments()` getter. However, this is not guaranteed. 
Cloning the array only when the owner-type didn't change seems like an inconsistency and an oversight, so this PR is trying to fix that.

### Checklist


RahulVijaykar (approved):

eamonnmcmanus (approved):
Thanks! I agree that the original code seems a bit suspect and this is the cleanest way to fix it.

Issue #1602, opened 2019-10-31
Gson OSGi metadata claims support for JavaSE-1.5 but target generation is 1.6


briandealwis:
Although the OSGi metadata claims Gson will work with Java 1.5 or greater:
https://github.com/google/gson/blob/47b28d99a17d492b37fcdd0fc41290419a010a0f/gson/bnd.bnd#L6

the compilation configuration sets the target as 1.5 and the classes themselves are Java 1.6:
https://github.com/google/gson/blob/47b28d99a17d492b37fcdd0fc41290419a010a0f/pom.xml#L89-L95

```sh
$ file gson/target/classes/com/google/gson/Gson.class
gson/target/classes/com/google/gson/Gson.class:                       compiled Java class data, version 50.0 (Java 1.6)
```

And that looks to have been a decision made a while ago #790.

elharo:
fixed by #1603

chanseokoh:
Please close this.

Issue #1601, opened 2019-10-29
OSGi metadata on 2.8.6 requires Java 9 or greater


briandealwis:
gson 2.8.6's `MANIFEST.MF` includes the following line:
```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9.0))"
```

And so it requires Java 9 (or greater, as `osgi.ee` is typically populated with all lesser versions too) and so refuses to install on Java 8 (or below).

briandealwis:
2.8.5's `MANIFEST.MF` has:
```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
```

PR #2711, opened 2024-07-04
Remove duplicated declaration of required OSGi execution environment

HannesWell:
### Description

The `bnd-maven-plugin` used to generate the OSGi metadata can determine the required EE automatically based on the version of the generated class-files. This avoids inconsistencies if the release target is raised in the future.

In general there is no need to declare a `Bundle-RequiredExecutionEnvironment` together with an `osgi.ee` requirement. At runtime the former is automatically converted to the later, since the former is deprecated.
Furthermore declaring multiple values for `Bundle-RequiredExecutionEnvironment` is also not useful since the driving requirement is the highest one.

With the change being applied bnd generates the following header:
```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
```
This matches the minimally required Java version of Java 7 for Gson 2.9 and later and https://github.com/google/gson/issues/1601 or https://github.com/google/gson/issues/1602 should not be re-introduced.

### Checklist

### DIFF
--- a/gson/bnd.bnd
+++ b/gson/bnd.bnd
@@ -3,8 +3,6 @@ Bundle-Name: ${project.name}
 Bundle-Description: ${project.description}
 Bundle-Vendor: Google Gson Project
 Bundle-ContactAddress: ${project.parent.url}
-Bundle-RequiredExecutionEnvironment: JavaSE-1.7, JavaSE-1.8
-Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
 
 # Optional dependency for JDK's sun.misc.Unsafe
 # https://bnd.bndtools.org/chapters/920-faq.html#remove-unwanted-imports-


HannesWell:
Thanks for submitting this.


> but it looks as if you have checked that this does the right thing

I have checked the result of a local mvn clean verify build and yes the result is as expected.
Do you deploy snapshots that can be checked again?

And regarding the `bnd-maven-plugin`:
- Instead of having a separated bnd.bnd file, you could also specify these instructions in a `bnd` configuration element of the bnd-maven-plugin:
https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-maven-plugin#bnd-process-goal
Would you be interested in having that?
- The bnd-maven-plugin could also be used to generate the `module-info.class`:
https://bnd.bndtools.org/chapters/330-jpms.html
This could be used to avoid the need for the `moditect-maven-plugin` to compile the `module-info.java`
Would you be interested in using that?

eamonnmcmanus:
We don't usually deploy snapshots, no. I think it's very likely that the updated configuration works given your local check, and if by chance the configuration _doesn't_ work we can fix it after the next release.

I like the idea of getting rid of `bnd.bnd` and using Maven configuration instead. It seems more logical for all configuration to be in `pom.xml` rather than having this one special file in the top-level `gson` directory.

On the other hand having `bnd-maven-plugin` generate `module-info.class` feels backwards to me. I think we do want to have an explicit `module-info.java` that users can consult.

HannesWell:
Sounds all good.
Please see https://github.com/google/gson/pull/2712 for the merge.

PR #2717, opened 2024-07-23
Remove old code style configs

Marcono1234:
### Description
Gson now uses Spotless (with google-java-format) for formatting, so these configs are not that useful anymore.
Additionally, users can obtain IDE formatting configs from https://github.com/google/styleguide, or use https://github.com/google/google-java-format integrations.

This is probably better than duplicating the IDE configurations here.


eamonnmcmanus (approved):
Thanks!

PR #2727, opened 2024-08-11
Move enum and `JsonElement` adapter classes to separate class files

Marcono1234:
### Purpose
Make internal code easier to read and maintain by moving some nested classes inside `TypeAdapters` to separate class files.

### Description
Their implementation is complex enough to probably justify having them in their own class files, to make maintaining them easier. For the users this should not be noticeable since these classes are still in the Gson internal package.

I have also added a few comments to the implementations.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -26,7 +26,9 @@ import com.google.gson.internal.Streams;
 import com.google.gson.internal.bind.ArrayTypeAdapter;
 import com.google.gson.internal.bind.CollectionTypeAdapterFactory;
 import com.google.gson.internal.bind.DefaultDateTypeAdapter;
+import com.google.gson.internal.bind.EnumTypeAdapter;
 import com.google.gson.internal.bind.JsonAdapterAnnotationTypeAdapterFactory;
+import com.google.gson.internal.bind.JsonElementTypeAdapter;
 import com.google.gson.internal.bind.JsonTreeReader;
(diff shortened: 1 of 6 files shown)


eamonnmcmanus (changes requested):
I like this a lot. I'm certainly not a fan of giant anonymous classes! However, even though people shouldn't be accessing the `.internal.` APIs, you just know they do. [Here](https://github.com/immutables/immutables/blob/4c737f7fcc3e00799b12e57f3f6abd6b3ebcd800/mongo/test/org/immutables/mongo/bson4gson/Jsons.java#L33) is an example. It doesn't cost much to keep the public fields in `TypeAdapters` and just change how they get their values. So I think we should do that.

Marcono1234:
Good point, I forgot to consider that. I added back `JSON_ELEMENT` and `ENUM_FACTORY`, but not `JSON_ELEMENT_FACTORY` because it [seems to be rarely used](https://github.com/search?q=%22JSON_ELEMENT_FACTORY%22+language%3AJava+NOT+path%3AGson.java+NOT+path%3ATypeAdapters.java&type=code). Is that ok?

eamonnmcmanus:
I think we might as well just keep all the public fields? It doesn't cost much.

Marcono1234:
Ok, I have added `JSON_ELEMENT_FACTORY` back as well. I was a bit worried that we keep accumulating `@Deprecated` fields and methods in Gson's internal code this way, but so far it is not that bad yet.

eamonnmcmanus (approved):

PR #2731, opened 2024-08-30
Add NullSafeTypeAdapter to prevent TypeAdapter.nullSafe() from returning nested null-safe type adapters (#2729)

### LINKED ISSUE #2729: Multiple invocation of the `TypeAdapter.nullSafe` method causes the given type adapter to be wrapped recursively.

Avoid recursive null-safe wrapping.

# Feature description

Currently the `TypeAdapter.nullSafe()` method is implemented like this:

```java
  public final TypeAdapter<T> nullSafe() {
    return new TypeAdapter<T>() {
      @Override
      public void write(JsonWriter out, T value) throws IOException {
        if (value == null) {
          out.nullValue();
        } else {
          TypeAdapter.this.write(out, value);
        }
      }

      @Override
      public T read(JsonReader reader) throws IOException {
        if (reader.peek() == JsonToken.NULL) {
          reader.nextNull();
          return null;
        }
        return TypeAdapter.this.read(reader);
      }
    };
  }
```

As seen, the `TypeAdapter.nullSafe` method is not guarded with its own type check, thus invoking the `TypeAdapter.nullSafe` method multiple times wraps the type adapter in the null-safe wrapper multiple times.

Currently it passes the following unit test(s):

```java
@ParameterizedTest
@ValueSource(ints = { 0, 1, 10 })
public void testNullSafe0OrMore(final int expectedSafeCount) {
	final TypeAdapter<Object> nullSafeTypeAdapter = failingTypeAdapter(expectedSafeCount);
	final RuntimeException ex = Assertions.assertThrows(RuntimeException.class, () -> nullSafeTypeAdapter.toJson(dummy));
	// probably bad check (is StackTraceElement.toString() guaranteed to return full.method.Name)
	final long actualNullSafeCount = Stream.of(ex.getStackTrace())
			.map(StackTraceElement::toString)
			.filter(toString -> toString.startsWith("com.google.gson.TypeAdapter$1.write("))
			.count();
	Assertions.assertEquals(expectedSafeCount, actualNullSafeCount);
}

private static final Object dummy = new Object();

private static TypeAdapter<Object> failingTypeAdapter(final int nullSafeCount) {
	if ( nullSafeCount < 0 ) {
		throw new IllegalArgumentException();
	}
	TypeAdapter<Object> typeAdapter = new TypeAdapter<>() {
		@Override public void write(final JsonWriter out, final Object value) { throw new RuntimeException("nullSafe count: " + nullSafeCount); }
		@Override public Object read(final JsonReader in) { throw new AssertionError("must never be invoked"); }
	};
	for ( int i = 0; i < nullSafeCount; i++ ) {
		typeAdapter = typeAdapter.nullSafe();
	}
	return typeAdapter;
}
```

# Alternatives / workarounds

The null-safe type adapter class would probably be extracted as a private named class, say `NullSafeTypeAdapter`. Then the null-safe wrapper method would be able to check if the given type adapter `instanceof`s against the `NullSafeTypeAdapter`, and if it's `true` simply return it as it is already wrapped in the null-safe type adapter. Otherwise the given type adapter instance would be wrapped in the `NullSafeTypeAdapter` class, thus once passing the following assertion:

```java
Assertions.assertEquals(expectedSafeCount > 0 ? 1 : 0, actualNullSafeCount);
```

This wouldn't probably be any significant improvement as the use site in most code, I believe, is controlled and uses `TypeAdapter.nullSafe()` once, so there are no recursive calls as it suggested with `expectedSafeCount` in the unit test above. But it would be probably be useful for more advanced uses of `TypeAdapter` (thus its chaining) if any. This change would be also backwards-compatible.

lyubomyr-shaydariv:
Add `NullSafeTypeAdapter` to prevent `TypeAdapter.nullSafe()` from returning nested null-safe type adapters (#2729)

### Purpose / Description

Resolves #2729.

### Checklist

Closes #2729.

### DIFF
--- a/gson/src/main/java/com/google/gson/TypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/TypeAdapter.java
@@ -289,24 +289,34 @@ public abstract class TypeAdapter<T> {
    * Note that we didn't need to check for nulls in our type adapter after we used nullSafe.
    */
   public final TypeAdapter<T> nullSafe() {
-    return new TypeAdapter<T>() {
-      @Override
-      public void write(JsonWriter out, T value) throws IOException {
-        if (value == null) {
-          out.nullValue();
-        } else {
-          TypeAdapter.this.write(out, value);
-        }
-      }
+    if (!(this instanceof TypeAdapter.NullSafeTypeAdapter)) {
+      return new NullSafeTypeAdapter();
+    }
+    return this;
+  }
 
-      @Override
-      public T read(JsonReader reader) throws IOException {
-        if (reader.peek() == JsonToken.NULL) {
-          reader.nextNull();
-          return null;
-        }
-        return TypeAdapter.this.read(reader);
+  private final class NullSafeTypeAdapter extends TypeAdapter<T> {
+    @Override
+    public void write(JsonWriter out, T value) throws IOException {
+      if (value == null) {
+        out.nullValue();
+      } else {
+        TypeAdapter.this.write(out, value);
       }
-    };
+    }
+
+    @Override
+    public T read(JsonReader reader) throws IOException {
+      if (reader.peek() == JsonToken.NULL) {
+        reader.nextNull();
+        return null;
+      }
+      return TypeAdapter.this.read(reader);
+    }
+
+    @Override
+    public String toString() {
+      return "NullSafeTypeAdapter[" + TypeAdapter.this + "]";
+    }
   }
 }
--- a/gson/src/test/java/com/google/gson/TypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/TypeAdapterTest.java
@@ -28,22 +28,49 @@ import org.junit.Test;
 public class TypeAdapterTest {
   @Test
   public void testNullSafe() throws IOException {
-    TypeAdapter<String> adapter =
-        new TypeAdapter<String>() {
-          @Override
-          public void write(JsonWriter out, String value) {
-            throw new AssertionError("unexpected call");
-          }
-
-          @Override
-          public String read(JsonReader in) {
-            throw new AssertionError("unexpected call");
-          }
-        }.nullSafe();
+    TypeAdapter<String> adapter = assertionErrorAdapter.nullSafe();
 
     assertThat(adapter.toJson(null)).isEqualTo("null");
     assertThat(adapter.fromJson("null")).isNull();
   }
+
+  @Test
+  public void testNullSafe_ReturningSameInstanceOnceNullSafe() {
+    TypeAdapter<?> nullSafeAdapter = assertionErrorAdapter.nullSafe();
+
+    assertThat(nullSafeAdapter.nullSafe()).isSameInstanceAs(nullSafeAdapter);
+    assertThat(nullSafeAdapter.nullSafe().nullSafe()).isSameInstanceAs(nullSafeAdapter);
+    assertThat(nullSafeAdapter.nullSafe().nullSafe().nullSafe()).isSameInstanceAs(nullSafeAdapter);
+  }
+
+  @Test
+  public void testNullSafe_ToString() {
+    TypeAdapter<?> adapter = assertionErrorAdapter;
+
+    assertThat(adapter.toString()).isEqualTo("assertionErrorAdapter");
+    assertThat(adapter.nullSafe().toString())
+        .isEqualTo("NullSafeTypeAdapter[assertionErrorAdapter]");
+    assertThat(adapter.nullSafe().nullSafe().toString())
+        .isEqualTo("NullSafeTypeAdapter[assertionErrorAdapter]");
+  }
+
+  private static final TypeAdapter<String> assertionErrorAdapter =
+      new TypeAdapter<>() {
+        @Override
+        public void write(JsonWriter out, String value) {
+          throw new AssertionError("unexpected call");
+        }
+
+        @Override
+        public String read(JsonReader in) {
+          throw new AssertionError("unexpected call");
+        }
+
+        @Override
+        public String toString() {
+          return "assertionErrorAdapter";
+        }
+      };
 
   /**
    * Tests behavior when {@link TypeAdapter#write(JsonWriter, Object)} manually throws {@link


gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -289,24 +289,30 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
   public final TypeAdapter<T> nullSafe() {
-    return new TypeAdapter<T>() {
-      @Override
-      public void write(JsonWriter out, T value) throws IOException {
-        if (value == null) {
-          out.nullValue();
-        } else {
-          TypeAdapter.this.write(out, value);
-        }
+    if (this.getClass() != NullSafeTypeAdapter.class) {
```
lyubomyr-shaydariv, line 292: Not using `instanceof` may look weird, but doesn't `instanceof` cause a compile time error, or am I missing something?

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -289,24 +289,30 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
+      return new NullSafeTypeAdapter();
+    }
+    return this;
+  }
+
+  private final class NullSafeTypeAdapter extends TypeAdapter<T> {
```
lyubomyr-shaydariv, line 298: * I prefer static nested classes, but I also guess that it might be an exception from the preferences, since `NullSafeTypeAdapter` captures `this` anyway.
* I'm not sure if it's worth adding `toString` here since the base class doesn't (?) override `Object.toString` that would be transparent anyway.

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -289,24 +289,29 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
   public final TypeAdapter<T> nullSafe() {
-    return new TypeAdapter<T>() {
-      @Override
-      public void write(JsonWriter out, T value) throws IOException {
-        if (value == null) {
-          out.nullValue();
-        } else {
-          TypeAdapter.this.write(out, value);
-        }
+    if (!NullSafeTypeAdapter.class.isInstance(this)) {
```
Marcono1234, line 292: Minor, but why not `!(this instanceof NullSafeTypeAdapter)`?

gson/src/test/java/com/google/gson/TypeAdapterTest.java
```diff
@@ -28,23 +28,34 @@
...
+  public void testNullSafe_ReturningSameInstanceOnceNullSafe() {
+    TypeAdapter<?> nullSafeAdapter = assertionErrorAdapter.nullSafe();
+
+    assertThat(nullSafeAdapter.nullSafe()).isSameInstanceAs(nullSafeAdapter);
+    assertThat(nullSafeAdapter.nullSafe().nullSafe()).isSameInstanceAs(nullSafeAdapter);
+    assertThat(nullSafeAdapter.nullSafe().nullSafe().nullSafe()).isSameInstanceAs(nullSafeAdapter);
+  }
+
+  private static final TypeAdapter<String> assertionErrorAdapter =
+      new TypeAdapter<String>() {
```
Marcono1234, line 47: Minor: `String` can probably be omitted here since it can be inferred
```suggestion
      new TypeAdapter<>() {
```

gson/src/test/java/com/google/gson/TypeAdapterTest.java
```diff
@@ -73,7 +84,7 @@ public Integer read(JsonReader in) {
   }
 
   private static final TypeAdapter<String> adapter =
-      new TypeAdapter<>() {
+      new TypeAdapter<String>() {
```
Marcono1234, line 87: I think this is not needed since we compile the test code with Java 11.

Though I think IDEs sometimes have problems detecting this, and erroneously assume Java 7 is used for test compilation as well.

gson/src/main/java/com/google/gson/TypeAdapter.java
```diff
@@ -289,24 +289,29 @@ public final T fromJsonTree(JsonElement jsonTree) {
...
+      return new NullSafeTypeAdapter();
+    }
+    return this;
+  }
+
+  private final class NullSafeTypeAdapter extends TypeAdapter<T> {
```
Marcono1234, line 298: As suggested in https://github.com/google/gson/issues/2729#issuecomment-2318722702, do you want to take the chance and implement `toString` as well. I think at least some other adapters implement it, and it would then make it easier to see which adapter is wrapped by this one.

Marcono1234 (commented):
Thanks a lot!

I have added a few comments, but feel free to consider them only suggestions. Hopefully they are useful.

Also, don't worry too much about the Git history of this PR branch; the commits will be squashed on merge anyway.

Marcono1234:
And could you please edit the description of the pull request to say "Closes #2729"? Otherwise GitHub does not understand this, and will keep the issue open after merge, unless they are linked manually.

gson/src/main/java/com/google/gson/TypeAdapter.java
lyubomyr-shaydariv: @Marcono1234 

Yeah, I had `instanceof` in mind, but not sure how it all works as of today. The rule `org.apache.maven.enforcer.rules.version.RequireJavaVersion` fails with

```
Detected JDK version 21 (JAVA_HOME=/opt/.bin/jdk-21) is not in the allowed range [11,18)
```

If the type check is changed to `!(this instanceof NullSafeTypeAdapter)`, then:

* for JDK 11, it fails with `/src/main/java/com/google/gson/TypeAdapter.java:[292,27] illegal generic type for instanceof`;
* for JDK 17, it fails with `/src/main/java/com/google/gson/TypeAdapter.java:[292,16] reifiable types in instanceof are not supported in -source 7`.

The `Class.instanceOf` check works for all. However, according to the source code, it can work with `TypeAdapter<?>` at least in `GsonBuilder`.

 Please suggest.

gson/src/test/java/com/google/gson/TypeAdapterTest.java
lyubomyr-shaydariv: I merely moved the `TypeAdapter` instance from the method to the field leaving the type parameter as it was previously in method. My rationale is that I should follow the convention and do not touch code I don't own and that can be a subject of refactoring in the future.

If you're really fine with changing from `TypeAdapter<String>`  to `TypeAdapter<>`, I'll change `<String>` to `<>`.

gson/src/main/java/com/google/gson/TypeAdapter.java
lyubomyr-shaydariv: @Marcono1234 

Oh, I seem to have misterstood your `toString()` suggestion earlier leaving it out of scope (I mistakenly assumed it leads to a stack overflow error, but it doesn't).

Could you please suggest the `NullSafeTypeAdapter.toString()` pattern you'd like to see?

lyubomyr-shaydariv:
> And could you please edit the description of the pull request to say "Closes #2729"? Otherwise GitHub does not understand this, and will keep the issue open after merge, unless they are linked manually.

Do you mean to add `Closes #2729` literally, so that GitHub detects the ticked #-id following the `Closes `? I was not aware of this behavior, and thought that `... type adapters (#2729)` would be sufficient.

Maybe I'm misusing git but `git rev-list --perl-regexp --count --grep='\bClose:?\b' fix-2729` returns no commmits. (My local `fix-2729` branch is currently pointing to `91246bef7631533658eea2af0125566cbc9331b4`)

gson/src/test/java/com/google/gson/TypeAdapterTest.java
lyubomyr-shaydariv: > Minor: `String` can probably be omitted here since it can be inferred

My bad: I seem to have missed the line change I made accidentally. Also I removed the `assertionErrorAdapter` `TypeAdapter` type from `<String>` to `<>` -- I saw the diamond operator has been in use at least in this test class.

gson/src/main/java/com/google/gson/TypeAdapter.java
lyubomyr-shaydariv: > As suggested in [#2729 (comment)](https://github.com/google/gson/issues/2729#issuecomment-2318722702), do you want to take the chance and implement `toString` as well. I think at least some other adapters implement it, and it would then make it easier to see which adapter is wrapped by this one.

I applied the `NullSafeTypeAdapter[]` pattern you suggested in https://github.com/google/gson/issues/2729#issuecomment-2318722702.

gson/src/main/java/com/google/gson/TypeAdapter.java
Marcono1234: Ah sorry, it seems `!(this instanceof NullSafeTypeAdapter)` is not possible because `NullSafeTypeAdapter` is a non-static inner class (and the enclosing class `TypeAdapter<T>` is generic?).

The compiler error message is really not helpful, but in Eclipse IDE the error message included the suggested fix, qualify with the enclosing type: `TypeAdapter.NullSafeTypeAdapter`. Can you please try this instead?
```suggestion
    if (!(this instanceof TypeAdapter.NullSafeTypeAdapter)) {
```

Marcono1234:
> Do you mean to add `Closes #2729` literally, so that GitHub detects the ticked #-id following the `Closes `?

Yes, just edit the description of your pull request text and add that anywhere. See for example [#2704](https://redirect.github.com/google/gson/pull/2704) which used "Resolves #...".

GitHub recognizes certain [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) and in that case will automatically close the linked issue.
It will also show this information then on the right sidebar of
- the issue\
issue screenshot
- the pull request
pull request screenshot

You can also link issue and pull request manually[^1], but it is easier if the creator of the pull request directly does this themselves through the keywords. Otherwise it is easy to forget manually linking the issue and pull request.

That works as well if you write it in the commit message (and the commit is then merged), but personally I avoid this (unless the contribution guide of a project requests it) since for each commit this already shows in the GitHub UI for the issue a "added a commit to ... that referenced this issue" link, which can be rather distracting. So I prefer to only add it to the pull request description.

> I was not aware of this behavior, and thought that `... type adapters (#2729)` would be sufficient.

It seems that just creates a regular "mention" without having any other effect. See for example #2729 which you referenced:
issue screenshot


(No worries though about the multiple commit mentions there.)

[^1]: If you are a member of the project (?). But I am not a direct member of the Gson project, so I cannot do it here.

lyubomyr-shaydariv:
> > Do you mean to add `Closes #2729` literally, so that GitHub detects the ticked #-id following the `Closes `?
> 
> Yes, just edit the description of your pull request text and add that anywhere.

The funny thing is that I thought you meant the git commit message, not the PR description, and this is why I referenced the git-rev-list command output that produced empty result. And now I also see why it is not a good idea for GitHub, kind of shame on GitHub from this perspective. Thank you for the comprehensive explanation!

gson/src/main/java/com/google/gson/TypeAdapter.java
lyubomyr-shaydariv: I've never encountered this error before preferring static inner classes to capture as few as possible, and now I see there are still "good-to-knows". But no doubt, `isAssignable` looked really weird and unnatural. Thank you!

Marcono1234 (approved):

eamonnmcmanus (approved):
Thanks for doing this!

PR #2734, opened 2024-09-03
EnumTypeAdapter constructor optimization

esaulpaugh:
### Purpose
Improve the runtime performance of `EnumTypeAdapter(Class)`.

### Description
Eliminate the temporary `ArrayList` in favor of array truncation via `<T> java.util.Arrays.copyOf(T[],int)`.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
@@ -67,7 +67,8 @@ public class EnumTypeAdapter<T extends Enum<T>> extends TypeAdapter<T> {
         }
       }
 
-      // Trim the array to the new length
+      // Trim the array to the new length. Every enum type can be expected to have at least
+      // one declared field which is not an enum constant, namely the implicit $VALUES array
       fields = Arrays.copyOf(fields, constantCount);
 
       AccessibleObject.setAccessible(fields, true);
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -59,17 +59,20 @@ private EnumTypeAdapter(final Class<T> classOfT) {
...
+        // Filter out non-constant fields, replacing elements as we go
         if (f.isEnumConstant()) {
-          constantFieldsList.add(f);
+          fields[constantCount++] = f;
         }
       }
 
-      Field[] constantFields = constantFieldsList.toArray(new Field[0]);
-      AccessibleObject.setAccessible(constantFields, true);
+      // Trim the array to the new length
```
eamonnmcmanus, line 70: My initial thought was that most of the time you won't need to do this. Every call to `getDeclaredFields()` returns a new array and if there are no fields that are _not_ enum constants then you can just use that array as is. However, with javac at least, it turns out that there is always a private `$VALUES` field which is an array that is cloned every time `values()` is called on the enum. So perhaps add a further note like "This will always be needed because of the $VALUES field."

esaulpaugh: Good call; comment added.

eamonnmcmanus (approved):

PR #2738, opened 2024-09-07
Extend test for OSGi manifest attributes

Marcono1234:
### Purpose
Perform more extensive tests on generated OSGi attributes in `MANIFEST.MF`

### Description
The test is converted to a Maven Failsafe integration test (`...IT.java`) to run on the final JAR, and additionally performs more assertions on the attributes.

However, there are the following disadvantages:
- Maybe this single `...IT.java` is confusing, compared to the regular `...Test.java`
- Executing this test in the IDE is not possible
However, the test tries to detect this and produce helpful assertion failures in that case.
- The test must be run with `mvn clean ...` due to https://github.com/bndtools/bnd/issues/6221
(on the other hand this is maybe an advantage since this would probably also fail the release when for some reason `clean` is not performed, and the `MANIFEST.MF` would contain unexpected / not-reproducible content)

Therefore, I would not mind if you reject this pull request if you think there are too many disadvantages with this or if you think the new test is too complex or too difficult to maintain.

### Checklist


gson/pom.xml
```diff
@@ -69,6 +69,12 @@
       <version>33.3.0-jre</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>33.2.1-jre</version>
```
Marcono1234, line 75: This direct dependency also fixes a different issue where previously `truth` and `guava-testlib` pulled in Guava transitively, but apparently `truth`'s Guava `-android` version was chosen by Maven.

So maybe that could have lead to issues for `guava-testlib` which expected the `-jre` version.

gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
```diff
@@ -0,0 +1,237 @@
...
+    // Should not define deprecated "Bundle-RequiredExecutionEnvironment"
+    assertThat(getAttribute("Bundle-RequiredExecutionEnvironment")).isNull();
+  }
+
+  private ManifestData findManifest(String bundleName) throws IOException {
+    List<URL> manifestResources =
+        Collections.list(getClass().getClassLoader().getResources("META-INF/MANIFEST.MF"));
+
+    for (URL manifestResource : manifestResources) {
+      InputStream is = manifestResource.openStream();
```
eamonnmcmanus, line 163: Can you use try-with-resources here?

eamonnmcmanus (changes requested):
This looks fine (based on my limited knowledge). Just one small suggestion.

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks for doing this!

PR #2740, opened 2024-09-13
Fix a class loading deadlock involving `TypeAdapters`.

### LINKED ISSUE #2739: Class loading deadlock with `JsonElementTypeAdapter`
At Google we import Gson from HEAD periodically. After the last such import, we had reports of a server appearing to get stuck, which we were able to track down to a class loading deadlock introduced by https://github.com/google/gson/pull/2727. In short, what happens is that Thread 1 calls `GsonBuilder.registerTypeHierarchyAdapter` which triggers a load of the `TypeAdapters` class [here](https://github.com/google/gson/blob/48889dba12f20c7e828469f2617e1bb49b6cfe30/gson/src/main/java/com/google/gson/GsonBuilder.java#L791). That in turn requires `JsonElementTypeAdapter` to be loaded for a static final field [here](https://github.com/google/gson/blob/48889dba12f20c7e828469f2617e1bb49b6cfe30/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java#L823). Meanwhile, Thread 2 calls `GsonBuilder.create`, which triggers a load of `JsonElementTypeAdapter` [here](https://github.com/google/gson/blob/48889dba12f20c7e828469f2617e1bb49b6cfe30/gson/src/main/java/com/google/gson/Gson.java#L328). That involves calling `TypeAdapters.newTypeHierarchyFactory` [here](https://github.com/google/gson/blob/48889dba12f20c7e828469f2617e1bb49b6cfe30/gson/src/main/java/com/google/gson/internal/bind/JsonElementTypeAdapter.java#L39), which requires `TypeAdapters` to be loaded. So Thread 1 is loading `TypeAdapters` and waiting for `JsonElementTypeAdapter` to be loaded, while Thread 2 is loading `JsonElementTypeAdapter` and waiting for `TypeAdapters` to be loaded.

The deadlock might be due to my [request](https://github.com/google/gson/pull/2727#pullrequestreview-2231845896) on that PR that the static final fields in `TypeAdapters` be preserved, because some existing projects are depending on them.

The fix is straightforward, in any case. Instead of `JsonElementTypeAdapter` and `EnumTypeAdapter` being public classes that are referenced directly from the `Gson` class, we should go back to having that class reference static final fields in `TypeAdapters`. Then every classloading path goes through `TypeAdapters` and the deadlock is averted.

Fortunately this bug didn't make it into any release, so most users will be unaffected.

eamonnmcmanus:
Fixes #2739.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -806,12 +806,6 @@ public final class TypeAdapters {
 
   public static final TypeAdapterFactory LOCALE_FACTORY = newFactory(Locale.class, LOCALE);
 
-  /*
-   * The following adapter and factory fields have not been removed yet and are only deprecated
-   * for now because external projects might be using them, despite being part of Gson's internal
-   * implementation.
-   */
-
   /**
    * @deprecated {@code TypeAdapters} is an internal Gson class. To obtain the adapter for {@link
    *     JsonElement} and subclasses use instead:
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -26,9 +26,7 @@ import com.google.gson.internal.Streams;
 import com.google.gson.internal.bind.ArrayTypeAdapter;
 import com.google.gson.internal.bind.CollectionTypeAdapterFactory;
 import com.google.gson.internal.bind.DefaultDateTypeAdapter;
-import com.google.gson.internal.bind.EnumTypeAdapter;
 import com.google.gson.internal.bind.JsonAdapterAnnotationTypeAdapterFactory;
-import com.google.gson.internal.bind.JsonElementTypeAdapter;
 import com.google.gson.internal.bind.JsonTreeReader;
 import com.google.gson.internal.bind.JsonTreeWriter;
 import com.google.gson.internal.bind.MapTypeAdapterFactory;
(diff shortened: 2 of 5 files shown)


Marcono1234:
Thanks a lot for noticing and troubleshooting this! It was also my fault for not paying attention to this during the refactoring.

In addition to your changes here, we could probably make it even safer by removing the `JsonElementTypeAdapter.FACTORY` field and instead using the `newTypeHierarchyFactory` call directly as initializer for `TypeAdapters.JSON_ELEMENT_FACTORY`, e.g.:
```java
public static final TypeAdapterFactory JSON_ELEMENT_FACTORY =
    newTypeHierarchyFactory(JsonElement.class, JSON_ELEMENT);
```

That would then completely remove the cyclic dependency between `JsonElementTypeAdapter` and `TypeAdapters`.

(Also, at least in my opinion, we could also un-deprecate the fields in `TypeAdapters` again and remove the Javadoc. I think it was only useful in the situation where those fields were not used internally by Gson anymore.)

eamonnmcmanus:
Thanks @Marcono1234, those changes seem reasonable.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -806,40 +806,12 @@ public void write(JsonWriter out, Locale value) throws IOException {
...
-   * @deprecated {@code TypeAdapters} is an internal Gson class. To obtain the adapter for {@link
-   *     JsonElement} and subclasses use instead:
-   *     <pre>{@code
-   * TypeAdapter<JsonElement> adapter = gson.getAdapter(JsonElement.class);
-   * }</pre>
-   */
-  @Deprecated
-  public static final TypeAdapterFactory JSON_ELEMENT_FACTORY = JsonElementTypeAdapter.FACTORY;
+  public static final TypeAdapterFactory JSON_ELEMENT_FACTORY =
+      TypeAdapters.newTypeHierarchyFactory(JsonElement.class, JSON_ELEMENT);
```
Marcono1234, line 812: Really minor: You can omit the `TypeAdapters.` qualifier here (unless you added it intentionally) since this is already within `TypeAdapters`.
```suggestion
      newTypeHierarchyFactory(JsonElement.class, JSON_ELEMENT);
```

cpovirk (approved):
Sorry, this wasn't actually that complicated, but I needed to find a minute when I wasn't half-thinking about something else at the time.

Issue #2501, opened 2023-09-29
Build failure with Java 21


ebourg:
Gson fails to build with Java 21, for two reasons:

1. The source/target level 7 is no longer supported:

        [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project gson: Fatal error compiling: error: release version 7 not supported -> [Help 1]

2. If the level is changed to 8, compiler warnings appear and break the build:

        [INFO] -------------------------------------------------------------
        [WARNING] COMPILATION WARNING :
        [INFO] -------------------------------------------------------------
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[500,24] non-transient instance field of a serializable class declared with a non-serializable type
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[501,24] non-transient instance field of a serializable class declared with a non-serializable type
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[502,26] non-transient instance field of a serializable class declared with an array having a non-serializable base component type java.lang.reflect.Type
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[567,24] non-transient instance field of a serializable class declared with a non-serializable type
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[601,24] non-transient instance field of a serializable class declared with a non-serializable type
        [WARNING] gson/src/main/java/com/google/gson/internal/$Gson$Types.java:[602,24] non-transient instance field of a serializable class declared with a non-serializable type
        [INFO] 6 warnings
        [INFO] -------------------------------------------------------------
        [INFO] -------------------------------------------------------------
        [ERROR] COMPILATION ERROR :
        [INFO] -------------------------------------------------------------
        [ERROR] gson/src/main/java/com/google/gson/internal/$Gson$Types.java: warnings found and -Werror specified
        [INFO] 1 error
        [INFO] -------------------------------------------------------------

eamonnmcmanus:
Thanks for letting us know!

[...]

PR #2744, opened 2024-09-18
Build for Java 8 rather than Java 7.

### LINKED ISSUE #2743: Make Java 8 the minimum supported version
Java 8 has been around for 10 years now and I think it is reasonable to drop support for earlier versions. With Java 8, we can simplify some code by using lambdas and method references. We can also support serializing and deserializing types like `Optional` and `Instant`. And we can add `default` and `static` methods to interfaces in the API.

We already have the [java8](https://github.com/google/gson/labels/java8) label for issues that can be solved once Java 8 is required. This issue just tracks making that happen.

eamonnmcmanus:
Closes #2743.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ Maven:
 
 ### Requirements
 #### Minimum Java version
-- Gson 2.9.0 and newer: Java 7
+- Gson 2.12.0 and newer: Java 8
+- Gson 2.9.0 to 2.11.0: Java 7
 - Gson 2.8.9 and older: Java 6
 
 Despite supporting older Java versions, Gson also provides a JPMS module descriptor (module name `com.google.gson`) for users of Java 9 or newer.
--- a/gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
+++ b/gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
@@ -144,8 +144,7 @@ public class OSGiManifestIT {
 
   @Test
   public void testRequireCapability() {
-    // When building with JDK >= 21, the minimum target version is Java 8
-    String expectedJavaVersion = Runtime.version().feature() < 21 ? "1.7" : "1.8";
+    String expectedJavaVersion = "1.8";
 
     // Defines the minimum required Java version
     assertThat(getAttribute("Require-Capability"))
--- a/gson/src/test/java/com/google/gson/internal/JavaVersionTest.java
+++ b/gson/src/test/java/com/google/gson/internal/JavaVersionTest.java
@@ -30,8 +30,8 @@ public class JavaVersionTest {
 
   @Test
   public void testGetMajorJavaVersion() {
-    // Gson currently requires at least Java 7
-    assertThat(JavaVersion.getMajorJavaVersion()).isAtLeast(7);
+    // Gson currently requires at least Java 8
+    assertThat(JavaVersion.getMajorJavaVersion()).isAtLeast(8);
   }
 
   @Test
--- a/pom.xml
+++ b/pom.xml
@@ -556,20 +556,6 @@
       </build>
     </profile>
 
-    <!-- Slightly adjust build to support building with JDK >= 21
-      However, by default this will intentionally be blocked by the Maven Enforcer Plugin defined above
-      and must be explicitly circumvented to avoid building non-release Gson artifacts by accident -->
-    <profile>
-      <id>jdk21+</id>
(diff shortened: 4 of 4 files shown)


cpovirk (approved):
A few notes, mostly on the Android front:

- I see that you already have [an Animal Sniffer check for Android compatibility](https://github.com/google/gson/blob/3ce6eb87983dd71bf6b47ff64645398e9caf8d14/pom.xml#L504). We seem to be getting good results [in Guava](https://github.com/google/guava/blob/e5b58e2fc8561192338dfe0c68c0bce27066a6aa/android/pom.xml#L287) from the https://github.com/open-toast/gummy-bears signatures that you mention, the ones that assume only [the always-on desugaring of a smallish set of methods](https://github.com/open-toast/gummy-bears?tab=readme-ov-file#android). Those are safe to use even if your users don't opt in to the more extensive feature that we normally refer to as "library desugaring," so you might have a look if javac starts generating usages of those methods automatically(?) or if you just want a smattering of Java 8 goodies.
- I don't see any issues with conditional support for `Optional` and `Instant` and friends. Once you've checked that the classes exist (see sub-bullet), you'll now be able to write normal Java code to operate on those classes. Android doesn't even mind if you have random methods in a class with `Optional` and `Instant` in the signature (in contrast to Java, which would, IIUC) as long as you don't call them.
   - I imagine the best way to guard that support will be with Java reflection that looks up those classes. I would _hope_ that that would support both the case of Android versions new enough to support the classes directly and the case of desugaring (which I _hope_ is smart enough to rewrite strings used in reflection). I have made the mistake before of just trying to use the classes and then catching the resulting error, but I have learned [one](https://github.com/google/truth/issues/333#issuecomment-765616282) or [two](https://github.com/google/guava/commit/fb109b051c9ec48d5e77bf73b33a228ab5d57462) (see also internal thread "Android and java7 Futures.getChecked()" about log spam) lessons about that, at least one of which is unlikely to be news to you :)).
- I have seen a handful of problems when experimenting with adding `default` methods in Guava inside Google. (See bug 229266760.) One of them is a result of a system that ill-advisedly tries to use a mix of the internal Guava and the external Guava (with the external Guava still lacking the `default` method), but I'm not sure I ever had a theory for the other. Anyway, adding the `default` methods would _probably_ work fine, but I'd advise factoring in at least a little possibility that it turns out not to.
- Lambda and method references should be entirely fine :) Maybe also remove some unnecessary `final` keywords from effectively final variables if you have any of those hanging around.
- Kudos on actually using `--release` to avoid [`ByteBuffer` issues](https://github.com/google/guava/issues/3990) (not that GSON appears to use the class) and other horrors.

Marcono1234 (approved):
If you want you can also adjust the build setup to properly support JDK 21, since the only blocking thing was using Java 7 as target. So basically your changes from #2690, except:
- Omit the `JAPICMP-OLD` changes (that has already been solved)
- In the README, adjust
  > JDK 17 is recommended. Newer JDKs are currently not supported for building

This would then close #2501

Or alternatively I can also do that in a separate PR if you prefer.

----

It seems there are also a few comments in the code suggesting refactoring for Java 8:
- These Error Prone patterns can probably be enabled now, but might introduce a lot of new warnings; so maybe worth doing in a separate PR?
https://github.com/google/gson/blob/3ce6eb87983dd71bf6b47ff64645398e9caf8d14/pom.xml#L283-L286
- Using `Executable` might simplify the code, but it seems Android only has it [since API level 26](https://developer.android.com/reference/kotlin/java/lang/reflect/Executable) so we cannot use it yet.
https://github.com/google/gson/blob/3ce6eb87983dd71bf6b47ff64645398e9caf8d14/gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java#L131-L132
- https://github.com/google/gson/blob/3ce6eb87983dd71bf6b47ff64645398e9caf8d14/gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java#L130

eamonnmcmanus:
OK, I'm proposing to merge this change as is and address any comments in followups. @Marcono1234 if you want to send PRs for the items you mention, I would certainly appreciate that.

cpovirk:
(I forgot to mention that Animal Sniffer will be upset about any appearance of `Optional` or `Instant` in a signature or method body. We suppress similar problems in Guava. Another alternative would be to investigate using multi-release jars, except for the part that that wouldn't actually help on Android.)

PR #2746, opened 2024-09-23
Use Guava testlib for collection tests

Marcono1234:
### Purpose
Use the Guava collection testlib for Gson's own collection classes

### Description
Use the Guava collection testlib for Gson's own collection classes `NonNullElementWrapperList` and `LinkedTreeMap`, covered through the tests for `JsonArray#asList` and `JsonObject#asMap`, as well as a test for `LinkedTreeMap`.

I don't have that much experience with the Guava testlib yet, so any feedback or recommendations are appreciated!

### Checklist

### DIFF
--- a/gson/src/test/java/com/google/gson/JsonArrayAsListSuiteTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayAsListSuiteTest.java
@@ -1,0 +1,68 @@
+package com.google.gson;
+
+import com.google.common.collect.testing.ListTestSuiteBuilder;
+import com.google.common.collect.testing.SampleElements;
+import com.google.common.collect.testing.TestListGenerator;
+import com.google.common.collect.testing.features.CollectionFeature;
+import com.google.common.collect.testing.features.CollectionSize;
+import com.google.common.collect.testing.features.ListFeature;
+import java.util.List;
+import junit.framework.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.AllTests;
+
+/**
+ * Dynamic {@link ListTestSuiteBuilder List test suite} for {@link JsonArray#asList()}. This
+ * complements {@link JsonArrayAsListTest}, which can cover some cases which are not covered here,
+ * e.g. making sure changes in the {@code List} view are visible in the {@code JsonArray}.
+ */
+@RunWith(AllTests.class)
+public class JsonArrayAsListSuiteTest {
+  private static class ListGenerator implements TestListGenerator<JsonElement> {
+    @Override
+    public SampleElements<JsonElement> samples() {
+      return new SampleElements<JsonElement>(
+          JsonNull.INSTANCE,
+          new JsonPrimitive(true),
+          new JsonPrimitive("test"),
+          new JsonArray(),
+          new JsonObject());
+    }
+
+    @Override
+    public JsonElement[] createArray(int length) {
+      return new JsonElement[length];
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


gson/src/test/java/com/google/gson/JsonArrayAsListSuiteTest.java
```diff
@@ -0,0 +1,68 @@
...
+  public static Test suite() {
+    return ListTestSuiteBuilder.using(new ListGenerator())
+        .withFeatures(
+            CollectionSize.ANY,
+            // Note: There is current a Guava bug which causes 'null additions' to not be tested if
+            // 'null queries' is enabled, see https://github.com/google/guava/issues/7401
+            CollectionFeature.ALLOWS_NULL_QUERIES,
+            CollectionFeature.RESTRICTS_ELEMENTS, // List only allows JsonElement
+            CollectionFeature.SUPPORTS_ADD,
+            ListFeature.REMOVE_OPERATIONS,
```
cpovirk, line 62: We typically use the `GENERAL_PURPOSE` features when applicable. There's such a constant [for `List`](https://github.com/google/guava/blob/d9dbb1594e6d412973e99119bd01ecdca1690baa/guava-testlib/src/com/google/common/collect/testing/features/ListFeature.java#L39) and also [for `Map`](https://github.com/google/guava/blob/d9dbb1594e6d412973e99119bd01ecdca1690baa/guava-testlib/src/com/google/common/collect/testing/features/MapFeature.java#L73). I haven't confirmed whether it applies here, and it's fine to use an explicit list instead even if it _would_ apply, but I wanted to at least point it out.

gson/src/test/java/com/google/gson/JsonObjectAsMapSuiteTest.java
```diff
@@ -0,0 +1,87 @@
...
+          Map.entry("three", new JsonPrimitive("test")),
+          Map.entry("four", new JsonArray()),
+          Map.entry("five", new JsonObject()));
+    }
+
+    @Override
+    public Map<String, JsonElement> create(Object... elements) {
+      JsonObject object = new JsonObject();
+      for (Object element : elements) {
+        @SuppressWarnings("unchecked")
```
cpovirk, line 38: nit: I would tend to cast to `Entry<?, ?>` and then cast the results of `getKey()` and `getValue()` so that the casts can be clearly checked at runtime, avoiding the need for a suppression. (They probably are checked at runtime anyway below? I'd have to check the bytecode.) I have no doubt that we do it the way you do here in a ton of existing code, though. And you'll always be stuck with suppressions elsewhere because of our use of arrays....

cpovirk (approved):
Thanks also for the Guava bug report, which I might not look at in detail for a while but which sounds concerning.

gson/src/test/java/com/google/gson/JsonArrayAsListSuiteTest.java
Marcono1234: Thanks for the hint! I wasn't completely sure whether they were suitable because `CollectionFeature.GENERAL_PURPOSE` [says](https://github.com/google/guava/blob/3c7c173e9c6ac93f154bfe40876f0c792d849f6e/guava-testlib/src/com/google/common/collect/testing/features/CollectionFeature.java#L86-L91):
> Features supported by general-purpose collections - everything but `RESTRICTS_ELEMENTS`.
>
> See java.util.Collection the definition of general-purpose collections.

And the `List` and `Map` here are not 'general-purpose' in the sense of the `java.util.Collection` definition.

PR #2747, opened 2024-09-23
Java 8 follow-up changes

### LINKED ISSUE #2501: Build failure with Java 21

Marcono1234:
### Purpose
Follow-up for #2744
Resolves #2501 (building with JDK 21 is now possible without any custom configuration)

### Description
Regarding the Error Prone warnings:
For the `final` variable warnings, it seems some of those variables were made `final` intentionally, without the compiler requiring it. But for consistency I removed the `final` from them as well.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,7 @@ jobs:
     name: "Build on JDK ${{ matrix.java }}"
     strategy:
       matrix:
-        java: [ 11, 17 ]
-        # Custom JDK 21 configuration
-        include:
-          - java: 21
-            # Disable Enforcer check which (intentionally) prevents using JDK 21 for building
-            extra-mvn-args: -Denforcer.fail=false
+        java: [ 11, 17, 21 ]
     runs-on: ubuntu-latest
 
     steps:
@@ -28,7 +23,7 @@ jobs:
           cache: 'maven'
       - name: Build with Maven
         # This also runs javadoc:jar to detect any issues with the Javadoc generated during release
-        run: mvn --batch-mode --no-transfer-progress verify javadoc:jar ${{ matrix.extra-mvn-args || '' }}
+        run: mvn --batch-mode --no-transfer-progress verify javadoc:jar
 
   native-image-test:
     name: "GraalVM Native Image test"
--- a/gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java
+++ b/gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java
@@ -43,7 +43,13 @@ import java.util.Set;
 @SuppressWarnings("serial") // ignore warning about missing serialVersionUID
 public final class LinkedTreeMap<K, V> extends AbstractMap<K, V> implements Serializable {
   @SuppressWarnings({"unchecked", "rawtypes"}) // to avoid Comparable<Comparable<Comparable<...>>>
-  private static final Comparator<Comparable> NATURAL_ORDER = Comparator.naturalOrder();
+  private static final Comparator<Comparable> NATURAL_ORDER =
+      new Comparator<Comparable>() {
+        @Override
+        public int compare(Comparable a, Comparable b) {
+          return a.compareTo(b);
+        }
+      };
 
   private final Comparator<? super K> comparator;
   private final boolean allowNullValues;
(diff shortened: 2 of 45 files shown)


gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java
```diff
@@ -43,13 +43,7 @@
...
 public final class LinkedTreeMap<K, V> extends AbstractMap<K, V> implements Serializable {
   @SuppressWarnings({"unchecked", "rawtypes"}) // to avoid Comparable<Comparable<Comparable<...>>>
-  private static final Comparator<Comparable> NATURAL_ORDER =
-      new Comparator<Comparable>() {
-        @Override
-        public int compare(Comparable a, Comparable b) {
-          return a.compareTo(b);
-        }
-      };
+  private static final Comparator<Comparable> NATURAL_ORDER = Comparator.naturalOrder();
```
Marcono1234, line 46: The code below performs identity checks (`== NATURAL_ORDER`) on this, but I hope this should be fine.

gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java
```diff
@@ -107,6 +109,16 @@ public int lastIndexOf(Object o) {
     return delegate.lastIndexOf(o);
   }
 
+  @Override
+  public Spliterator<E> spliterator() {
+    return delegate.spliterator();
+  }
```
Marcono1234, line 115: I have also overridden `spliterator` since the `ArrayList` implementation might be more efficient than the default `List` one.

Though I am not completely sure if this is really necessary; it seems the default `List#spliterator` uses a specialized implementation if the list if `RandomAccess` (which is the case here), and that implementation looks quite similar to the `ArrayList` one.

Marcono1234: Obsolete: Had to remove these overrides again because the minimum Android API level we support does not have them yet.

.github/workflows/build.yml
```diff
@@ -10,12 +10,7 @@ jobs:
     name: "Build on JDK ${{ matrix.java }}"
     strategy:
       matrix:
-        java: [ 11, 17 ]
-        # Custom JDK 21 configuration
-        include:
-          - java: 21
-            # Disable Enforcer check which (intentionally) prevents using JDK 21 for building
-            extra-mvn-args: -Denforcer.fail=false
+        java: [ 11, 17, 21 ]
```
eamonnmcmanus, line 13: Yay!

eamonnmcmanus (approved):
This is great, thanks! Look at all those `final` modifiers deleted.

PR #2753, opened 2024-10-01
Fix some new Error Prone warnings.

eamonnmcmanus:
These break the build because we compile with `-Werror`.

The new warnings about class names are intrusive and should perhaps be refined.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
+++ b/gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
@@ -31,6 +31,7 @@ import java.util.Objects;
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
+@SuppressWarnings("MemberName") // legacy class name
 public final class $Gson$Preconditions {
   private $Gson$Preconditions() {
     throw new UnsupportedOperationException();
--- a/gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
+++ b/gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
@@ -18,7 +18,6 @@ package com.google.gson.functional;
 
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertThrows;
-import static org.junit.Assume.assumeNotNull;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -81,12 +80,12 @@ public class ReflectionAccessFilterTest {
 
     // But serialization should succeed for classes with only public fields.
     // Not many JDK classes have mutable public fields, thank goodness, but java.awt.Point does.
-    Class<?> pointClass = null;
+    Class<?> pointClass;
     try {
       pointClass = Class.forName("java.awt.Point");
     } catch (ClassNotFoundException ignored) {
+      return; // If not found then we don't have AWT and the rest of the test can be skipped.
     }
-    assumeNotNull(pointClass);
     Constructor<?> pointConstructor = pointClass.getConstructor(int.class, int.class);
     Object point = pointConstructor.newInstance(1, 2);
     String json = gson.toJson(point);
--- a/gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
+++ b/gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
@@ -45,7 +45,7 @@ import org.junit.Test;
  * Running Maven's {@code clean} phase is necessary due to a <a
  * href="https://github.com/bndtools/bnd/issues/6221">bnd-maven-plugin bug</a>.
  */
-@SuppressWarnings("MemberName") // test name does not conform to Google Style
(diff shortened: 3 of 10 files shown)


gson/src/main/java/com/google/gson/internal/$Gson$Preconditions.java
```diff
@@ -31,6 +31,7 @@
  * @author Inderjeet Singh
  * @author Joel Leitch
  */
+@SuppressWarnings("MemberName") // legacy class name
 public final class $Gson$Preconditions {
```
Marcono1234, line 35: Slightly related: #1744

At least Eclipse IDE has problems with these class names sometimes. On the other hand they make accidental usage by external projects less likely.

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
```diff
@@ -81,12 +80,12 @@ public void testBlockInaccessibleJava() throws ReflectiveOperationException {
...
     // But serialization should succeed for classes with only public fields.
     // Not many JDK classes have mutable public fields, thank goodness, but java.awt.Point does.
-    Class<?> pointClass = null;
+    Class<?> pointClass;
     try {
       pointClass = Class.forName("java.awt.Point");
     } catch (ClassNotFoundException ignored) {
+      return; // If not found then we don't have AWT and the rest of the test can be skipped.
     }
-    assumeNotNull(pointClass);
```
Marcono1234, line 89: What was the Error Prone warning here?

`return;` makes it look as if this test passed, whereas JUnit's `assume...` will mark it as skipped. I guess the latter is rather desired here.

gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
```diff
@@ -45,6 +45,7 @@
  * Running Maven's {@code clean} phase is necessary due to a <a
  * href="https://github.com/bndtools/bnd/issues/6221">bnd-maven-plugin bug</a>.
  */
+@SuppressWarnings("MemberName") // test name does not conform to Google Style
 public class OSGiManifestIT {
```
Marcono1234, line 49: Can you please make it more clear that the test name must end with `IT` (which I assume is part of the Error Prone warning)?
```suggestion
@SuppressWarnings("MemberName") // class name must end with 'IT' for Maven Failsafe Plugin
public class OSGiManifestIT {
```

test-shrinker/src/test/java/com/google/gson/it/ShrinkingIT.java
```diff
@@ -40,6 +40,7 @@
 import org.junit.runners.Parameterized.Parameters;
 
 /** Integration test verifying behavior of shrunken and obfuscated JARs. */
+@SuppressWarnings("MemberName") // class name
```
Marcono1234, line 43:
```suggestion
@SuppressWarnings("MemberName") // class name must end with 'IT' for Maven Failsafe Plugin
```

cpovirk (approved):
I'm sorry, I seem to have marked this as read without looking at it. 

I have nothing to add to what @Marcono1234 has said.

gson/src/test/java/com/google/gson/functional/ReflectionAccessFilterTest.java
cpovirk: The warning is probably from https://errorprone.info/bugpattern/EmptyCatch, which (along with the Google Style Guide, though that update hasn't been published externally yet) recently got stricter: https://github.com/google/error-prone/pull/4576. You'll eventually see the "Exception" removed from https://google.github.io/styleguide/javaguide.html#s6.2-caught-exceptions. (That's also what finally got me to go through with https://github.com/google/guava/pull/7395 and some similar PRs to come.)

I agree that an assumption makes more conceptual sense. (But I say this knowing that we consistently just `return` in Guava's tests :))

Marcono1234: JUnit 4 has `assumeNoException(Throwable)`, but using it in the `catch` seems a bit weird since the exception will then always be non-null, so the assumption always fails.

Maybe it would be cleaner to directly throw an `AssumptionViolatedException`?
```java
Class<?> pointClass;
try {
  pointClass = Class.forName("java.awt.Point");
} catch (ClassNotFoundException e) {
  // If not found then we don't have AWT and the rest of the test can be skipped.
  throw new AssumptionViolatedException("AWT class not present", e);
}
```

eamonnmcmanus: I still feel that "assume" is not right given that this is happening half way through the test method. The right thing to do is to split out the remainder of the test into a separate method. I will look into doing that in a separate PR.

eamonnmcmanus (commented):
I'm realizing that I wrote some draft comments a couple of days ago but left them unsent. Sorry about that.

PR #2755, opened 2024-10-06
Change the use of try to try-with-resources in ParseBenchmark resourceToString()

panic08:
Here we will change the use of `try` to use `try-with-resources` in the `resourceToString()` method of the `ParseBenchmark` class.

Since the current version of Gson uses Java 7 as the minimum version, and `try-with-resources` has been around since Java 7, this will be a nice and even desirable change


eamonnmcmanus (approved):
Thanks! This is a good improvement. Incidentally the minimum version is Java 8, as of a couple of weeks ago.

PR #2756, opened 2024-10-07
Refactor if-else chains to switch in benchmark methods

panic08:
For better readability and maintainability, I think we should change the if-else chain to a switch


eamonnmcmanus (approved):
This makes sense to me. I'd also expect it to be slightly more performant, which is good because that means it interferes less with what the code is actually trying to measure.

I'll note that I don't think anybody actually uses these metrics.

PR #2757, opened 2024-10-09
Split out the `java.awt.Point` check into its own test method.

eamonnmcmanus:


Marcono1234 (approved):
Thanks! This makes the test cleaner indeed.

cpovirk (approved):

PR #2758, opened 2024-10-11
Support Properties subclasses in GsonTypes.getMapKeyAndValueTypes

panic08:
There is a `TODO` in this method that calls to test it with the `Properties` subclasses. They are supposed to return `new Type[] {String.class, String.class}`, just like if you cast the `Properties` class itself. But still when you pass a subclass of `Properties`, such as `CustomProperties`, the method returns `new Type[] {Object.class, Object.class}`, which is incorrect. Here I tried to fix this problem by using `isAssignableFrom()` and also wrote tests which indicate that the problem is solved

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/$Gson$Types.java
+++ b/gson/src/main/java/com/google/gson/internal/$Gson$Types.java
@@ -328,8 +328,8 @@ public final class $Gson$Types {
      * class should extend Hashtable<String, String>, but it's declared to
      * extend Hashtable<Object, Object>.
      */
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Thanks! This does seem like the sort of thing where existing code might conceivably be depending on the current behaviour. I'll run this against all of Google's internal tests and if nothing fails we can go ahead with the change.

eamonnmcmanus (approved):
I ran this against all of Google's internal tests and there were no failures, so I think it is pretty safe to merge.

PR #2759, opened 2024-10-12
Enforce rawType to be a Class in ParameterizedTypeImpl

panic08:
There is a [JDK-6255169](https://bugs.openjdk.org/browse/JDK-6255169) issue that states that `getRawType()` in `ParameterizedType `should return `Class` and not `Type`. However, this issue is closed, it was explained by the fact that there is no way to fix it due to backward compatibility. Our TODO specifies a related issue [JDK-8250659](https://bugs.openjdk.org/browse/JDK-8250659) which calls to comment out that `getRawType()` always returns `Class` as `Type`

Accordingly, looking at all this, we implement `ParameterizedType` and assume that `rawType` will always be `Class`. However, in our implementation class in the constructor we can throw under `rawType` more than just `Class`, which is not the right thing to do

Here I tried to solve this problem by throwing the `IllegalArgumentException(“rawType must be a Class type”)` exception


eamonnmcmanus:
I think we should probably just change the type of the `rawType` parameter of `newParameterizedTypeWithOwner` and of the `ParameterizedTypeImpl` constructor from `Type` to `Class<?>`. That then requires two casts in `$Gson$Types` itself and one change from `rawType` to the already-existing `rawClass` in `TypeToken`. I think it's fine to get a `ClassCastException` in the supposedly-impossible case that these `Type` instances are not actually instances of `Class`.

Ordinarily we wouldn't change parameter types in a public API, but in this case the API is _doubly_ marked as internal: it has `.internal.` in its package name _and_ it has those funky `$` signs in its name. I do see a few references in GitHub search that might conceivably be broken by this change but I'm relatively unsympathetic. If they are indeed affected then the fix is just inserting a cast. I tried making a change along these lines and running it against all of Google's internal tests (including lots of third-party code) and did not discover any problems.

panic08:
I agree with you. I did what you suggested, in this case in `$Gson$Types` itself I needed to make casts to `Class<?>`, and also changed the description of the rawType field in `TypeToken` from `Type` to `Class`.

I also directly removed a small piece of code in `TypeTokenTest` that checks for an error if Type is not passed in `Class`.

Also, am I correct that we can ignore the error `“Check API compatibility / check-api-compatibility (pull_request) Failed after 1m”`?

gson/src/main/java/com/google/gson/reflect/TypeToken.java
```diff
@@ -379,27 +379,19 @@ public static <T> TypeToken<T> get(Class<T> type) {
    * <p>If {@code rawType} is a non-generic class and no type arguments are provided, this method
    * simply delegates to {@link #get(Class)} and creates a {@code TypeToken(Class)}.
    *
-   * @throws IllegalArgumentException If {@code rawType} is not of type {@code Class}, or if the
-   *     type arguments are invalid for the raw type
+   * @throws IllegalArgumentException If the type arguments are invalid for the raw type
    */
-  public static TypeToken<?> getParameterized(Type rawType, Type... typeArguments) {
+  public static TypeToken<?> getParameterized(Class<?> rawType, Type... typeArguments) {
```
eamonnmcmanus, line 384: Oh, we can't change the signature of this method. It is part of the public API and it is _not_ internal. So I think you want to undo the changes in this method, and just change `rawType` to `rawClass` in the `return` statement.

gson/src/test/java/com/google/gson/reflect/TypeTokenTest.java
```diff
@@ -186,16 +185,7 @@ public void testParameterizedFactory_Invalid() {
         NullPointerException.class,
         () -> TypeToken.getParameterized(List.class, new Type[] {null}));
 
-    GenericArrayType arrayType = (GenericArrayType) TypeToken.getArray(String.class).getType();
```
eamonnmcmanus, line 189: I would undo these changes too.

eamonnmcmanus (changes requested):

eamonnmcmanus (approved):
Thanks, this is great!

PR #2760, opened 2024-10-13
Refactor switch-case to if-else in UtcDateTypeAdapter

panic08:
The current implementation of the read() method uses a `switch-case` construct that is not needed to handle only two cases. To improve the readability and maintainability of the code, I suggest refactoring into a simpler `if-else` structure


eamonnmcmanus (approved):
Seems fair.

PR #2762, opened 2024-10-15
Update some internal classes to use Java 8 java.time

panic08:
As of late, we have gotten [support](https://github.com/google/gson/issues/2743) for java 8. In this PR, I added java.time package support for `ISO8601Utils, SqlDateTypeAdapter, SqlTimeTypeAdapter, UtcDateTypeAdapter` classes.

I also added a TODO to `DefaultDateTypeAdapter`. Since I encountered some problems, such as that with `SimpleDateFormat` the date becomes longer and passes under tests, and additional logic needs to be written to support java.time, I decided to install this TODO. In the future it will be possible to figure out and add support for java.time and this class.

### DIFF
--- a/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
@@ -22,264 +22,40 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import java.io.IOException;
-import java.text.ParseException;
-import java.text.ParsePosition;
-import java.util.Calendar;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Date;
-import java.util.GregorianCalendar;
-import java.util.Locale;
-import java.util.TimeZone;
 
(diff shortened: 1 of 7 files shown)


panic08:
It might be better to put the java.time support for each class in a separate commit

Marcono1234:
Thanks! I think for `UtcDateTypeAdapter` and `ISO8601Utils` it might be best for now to keep them as they are. It is quite likely that their behavior differs slightly from `java.time`, so this risks introducing compatibility issues. And (unlike Java's `SimpleDateFormat`) they are already thread-safe, so from a performance perspective these classes are probably fine.

The main problem is Android compatibility (see also "Check Android compatibility" workflow): While we have Java 8 now as minimum version, Gson still targets Android API level 21 as minimum (see README). However, the `java.time` package was only added in API level 26, see https://developer.android.com/reference/java/time/package-summary. The Android Gradle Plugin [supports desugaring of these classes](https://developer.android.com/studio/write/java8-support-table), but not all users might have enabled this.


So I think it is unlikely that we will use the `java.time` classes in any of the existing adapters in the near future.

There are vague plans to add built-in adapters for `java.time` classes, but that would probably require reflection to check if these classes are available (see also https://github.com/google/gson/pull/2744#pullrequestreview-2316630809), and requires experimentation to see if that approach actually works properly for Android builds (and probably also for a GraalVM Native Image build).

(This is only my personal view on this, I am not a direct member of this project.)

panic08:
I agree, I think then it will be worth thinking about this update in the future when (if) gson will support android 27

PR #2763, opened 2024-10-16
Replace anonymous class creation with lambdas in ConstructorConstructor

panic08:
I think we should change the complete creation of anonymous classes to lambdas, which will reduce the amount of code and make it more readable

Since we recently got Java 8 support, we can afford to use lambdas

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -103,24 +103,14 @@ public final class ConstructorConstructor {
     @SuppressWarnings("unchecked") // types must agree
     InstanceCreator<T> typeCreator = (InstanceCreator<T>) instanceCreators.get(type);
     if (typeCreator != null) {
-      return new ObjectConstructor<T>() {
-        @Override
-        public T construct() {
-          return typeCreator.createInstance(type);
-        }
-      };
+      return () -> typeCreator.createInstance(type);
     }
 
     // Next try raw type match for instance creators
     @SuppressWarnings("unchecked") // types must agree
     InstanceCreator<T> rawTypeCreator = (InstanceCreator<T>) instanceCreators.get(rawType);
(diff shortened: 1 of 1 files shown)


panic08:
Hmm, I noticed odd that the `Build / GraalVM Native Image test (pull_request)` job crashes with text:
```
Run graalvm/setup-graalvm@22cc13fe88ef133134b3798e128fb208df55e1f5
Error: Failed to download graalvm-jdk-17_linux-x64_bin.tar.gz. Are you sure java-version: '17' is correct?
```
I assume it's not related to the change?

Marcono1234:
> Hmm, I noticed odd that the `Build / GraalVM Native Image test (pull_request)` job crashes with text

Seems to be a bug with that GitHub action; I have created #2764 to update it a fixed version. Once that is merged you can merge `main` into your branch and the build should succeed again.

Marcono1234 (commented):
Thanks for the changes! I have one small suggestion, see review comment.

Though @eamonnmcmanus might have additional suggestions or requests for changes.

gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
```diff
@@ -42,7 +42,7 @@ private interface Interface {}
   public void testGet_AbstractClassNoArgConstructor() {
     ObjectConstructor<AbstractClass> constructor =
         constructorConstructor.get(TypeToken.get(AbstractClass.class));
-    var e = assertThrows(RuntimeException.class, () -> constructor.construct());
+    var e = assertThrows(RuntimeException.class, constructor::construct);
```
eamonnmcmanus, line 45: Unless Error Prone is making you do this, I think this code should probably be left the way it is. It's asserting that the operation `constructor.construct()` throws an exception, and I think `() -> constructor.construct()` is a clearer way to express that.

eamonnmcmanus (approved):

PR #2765, opened 2024-10-16
Use GraalVM for JDK 21 in GitHub workflow

Marcono1234:
### Purpose
Upgrade to GraalVM for JDK 21 to avoid being stuck on old JDK 17 version

### Description
GraalVM for JDK 17 does not receive any updates anymore (at least not under the "GraalVM Free Terms and Conditions")

See https://github.com/graalvm/setup-graalvm/tree/v1.2.4?tab=readme-ov-file#notes-on-oracle-graalvm-for-jdk-17

As mentioned there in the README and in https://github.com/graalvm/setup-graalvm/issues/107#issuecomment-2417563249, we will always be using Oracle GraalVM under the "GraalVM Free Terms and Conditions". After that expires (which for JDK 21 will be in September 2026, see also [GraalVM licensing FAQ](https://www.oracle.com/java/technologies/javase/jdk-faqs.html#GraalVM-licensing)) we will simply be stuck on an unmaintained GraalVM version (unless we already migrated to GraalVM for JDK 25 in the meantime), but we won't accidentally use GraalVM under the Oracle Technology Network License Agreement and violate it by accident.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       - name: "Set up GraalVM"
         uses: graalvm/setup-graalvm@6f327093bb6a42fe5eac053d21b168c46aa46f22  # v1.2.4
         with:
-          java-version: '17'
+          java-version: '21'
           distribution: 'graalvm'
           # According to documentation in graalvm/setup-graalvm this is used to avoid rate-limiting issues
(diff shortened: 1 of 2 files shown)


eamonnmcmanus (approved):
Thanks!

PR #2766, opened 2024-10-17
Replace anonymous classes creation with lambdas in tests

panic08:
This change will reduce the amount of code in the tests and simplify it in places where the method implementation is too small, such as:
```
        new FieldNamingStrategy() {
          @Override
          public String translateName(Field f) {
            return "test";
          }
        }
```
can be replaced by:
`f -> “test”`


eamonnmcmanus (approved):
Thanks, that's a big improvement in readability!

PR #2767, opened 2024-10-18
Simplify equality checks using Objects.equals in PostConstructAdapterFactoryTest

panic08:
Here we replace manual checks by using `Objects.equals()`, which reduces the amount of code and makes it more readable


eamonnmcmanus (approved):
Thanks!

PR #2768, opened 2024-10-19
Change the use of == operator to isEmpty() in JsonObject

panic08:
Here we replace the use of `size == 0` with the corresponding `isEmpty()`


eamonnmcmanus (approved):
Looks good.

PR #2770, opened 2024-10-20
Use better names for lambda parameters in tests

Marcono1234:
### Purpose
Follow-up for #2766

### Description
Without the parameter type it is difficult to understand what some of the lambda parameters represent, therefore I have renamed them.


eamonnmcmanus (approved):
This is fair. The short names are the same as the ones that were used before this code was converted to use lambdas, though there we did at least have their types.

Issue #2359, opened 2023-03-27
Create proper unchecked exception hierarchy


Marcono1234:
# Problem solved by the feature
Currently Gson's unchecked exception hierarchy looks like this:
- `JsonParseException`
  - `JsonIOException`
  - `JsonSyntaxException`

This has the following issues:
- `JsonIOException` is a subclass of `JsonParseException`, however `JsonIOException` should also be usable for serialization
- Multiple use cases are not covered be these, leading to misuse in Gson's code base, usage of `RuntimeException` or other `java.lang` exceptions (slightly related: #548)

This mainly or only affects `Gson` and built-in adapters. Other API such as `JsonReader`, `JsonWriter` and `JsonElement` (and subclasses) are quite consistent with the exceptions they throw. For checked exceptions Gson only has `MalformedJsonException` which is only used by `JsonReader` so there is probably no need for changes there (and changes would most likely also be source incompatible due to it being a checked exception).

# Feature description
Define a proper hierarchy for unchecked exceptions. Maybe something like the following would be useful, but I have not thought that much about it yet:

- `GsonException`
  - `JsonIOException`
  - `DeserializationException` (or keep using existing `JsonParseException` for this)
    - `JsonSyntaxException`
  - `SerializationException` (or `JsonSerializionException` or similar to match `JsonParseException`)
  - `TypeAdapterException` (maybe?) for all issues which occur during creation of type adapters and are therefore neither `DeserializationException` nor `SerializationException`

Then adjust the Gson code to throw proper and consistent exceptions, mainly / only `Gson` and built-in adapters. Existing exception classes could be kept but their superclasses might have to be adjusted in some cases. While this might be behavior-wise a breaking change it hopefully makes usage easier and more consistent in the future.

`GsonException` and all subclasses can be used by user code as well. Since it is likely that user-written adapters throw arbitrary runtime exceptions though, and that for malformed JSON built-in and user-written adapters might throw other runtime exceptions as well, the documentation for `GsonException` should probably recommend that users should not only catch `GsonException` but any `RuntimeException` to be safe.

# Alternatives / workarounds
- Only extend the existing exception hierarchy with additional classes, but don't change any existing ones.
This would keep the problem that `JsonIOException` can also occur during serialization and therefore `JsonParseException` is not an ideal subclass.
- Do nothing, and maybe just recommend that users catch `RuntimeException` when calling Gson methods

PR #2771, opened 2024-10-29
Use JsonIOException for unchecked exceptions instead of RuntimeException in ConstructorConstructor

panic08:
In the `newDefaultConstructor` method in `ConstructorConstructor` we throw `RuntimeException` when _Failed to invoke constructor_. However, as noted in the TODO, we should throw something specific like `JsonIOException` when an unchecked exception occurs. In this method we already throw `JsonIOException` on similar exceptions. I believe this change will give more readability, but more importantly, more semantic precision

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -270,15 +270,13 @@ public final class ConstructorConstructor {
       // Note: InstantiationException should be impossible because check at start of method made
       // sure that class is not abstract
       catch (InstantiationException e) {
-        throw new RuntimeException(
+        throw new JsonIOException(
             "Failed to invoke constructor '"
                 + ReflectionHelper.constructorToString(constructor)
                 + "' with no args",
             e);
       } catch (InvocationTargetException e) {
-        // TODO: don't wrap if cause is unchecked?
-        // TODO: JsonParseException ?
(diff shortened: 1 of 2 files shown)


Marcono1234 (commented):
> but more importantly, more semantic precision

But only due to using a Gson-specific exception instead of a generic `RuntimeException`. `JsonIOException` is still kind of 'wrong' for this because these issues are not IO related.

So I am a bit indecisive on this change. The underlying issue is that Gson has no proper exception hierarchy (#2359), and I am not sure if that can even be solved without breaking backward compatibility.

@eamonnmcmanus, what is your opinion?

----

Could you please adjust this test though to expect a `JsonIOException` now and remove the TODO comment there (I hope that is the only case):
https://github.com/google/gson/blob/44c93364c9347404e93cd6a3323201ceae6b2a43/gson/src/test/java/com/google/gson/functional/ObjectTest.java#L733

panic08:
I agree that `JsonIOException` doesn't quite fit since it's typically used for IO related issues, but in my opinion it provides a bit more context than RuntimeException while we expect a more refined exception hierarchy in Gson (as discussed in [#2359](https://github.com/google/gson/issues/2359))

eamonnmcmanus:
Thanks for trying to clean this exceptions up a bit! But I think the small improvement to usability is probably not worth the small risk of incompatibility. Some existing code might be depending on the exceptions the way they are today.

Issue #1002, opened 2017-01-25
Change Return type of FieldNamingStrategy.translateName from String to List<String> for alternate like case insensitive.


KimSeongJun:
To serialize/Deserialize case insensitive fieldnames,

I use annotation SerializedName like.. 

`@SerializedName(value="fieldName",alternate={"fieldname"})`

But too many @SerializedName .... I don't wanna check typo any more.

Just all i want is to case insensitive FieldNamingPolicy.. but.. no implementations..

So, reviewed some gson source, and I found ReflectiveTypeAdapterFactory is adding alternate...

but by FieldNamingStragy, just one fieldname is added to BoundField.

If FieldNamingStrategy.translateName returns List<String>, and TypeAdapterFactory support it.

I can add case insensitive FieldNamePolicy for my own. :)

Or how about add case insensitive options?

Case insensitive fieldname policy is bad ideas? :(

Some json parser has restricted naming policy, i'ts hard to make multiple language RESTful API SDK.

PR #2776, opened 2024-11-15
Give FieldNamingStrategy the ability to return multiple String names

mfriesen:
### Purpose

This pull request enhances the `FieldNamingStrategy` in Gson to support returning multiple strings for a field name. This behavior aligns with the alternate names functionality provided by the `@SerializedName` annotation, allowing for greater flexibility in field mapping.

### Description

The `FieldNamingStrategy` interface has been updated to allow implementations to return a list of potential field name mappings. This allows deserialization to succeed if any of the alternate names are present in the JSON data. 

The changes include is to modifying the `FieldNamingStrategy` interface to support returning a collection of names. This allows the `FieldNamingStrategy` interface to behave like the `@SerializedName` annotation.

### Checklist

- [ X] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)\
- [X ] If necessary, new public API validates arguments, for example rejects `null`
- [X ] New public API has Javadoc
    - [ X] Javadoc uses `@since $next-version$`  
- [ X] If necessary, new unit tests have been added  
  - [ X] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [X ] No JUnit 3 features are used (such as extending class `TestCase`)
  - [X ] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [X ] `mvn clean verify javadoc:jar` passes without errors


Marcono1234:
Thanks for this suggestion! It seems that would resolve #1002

However, we cannot change the return type of `translateName` from `String` to `List<String>` since that would be an incompatible change.
But what might be possible, similar to `@SerializedName#alternate`, is adding an _additional_ method to `FieldNamingStrategy`, for example `translateToAlternateNames(Field): List<String>` (or a better name). The next Gson version will target Java 8, so we could provide a `default` implementation of this method which returns an empty list. This should preserve backward compatibility.

This is just an idea though, maybe there are other possible approaches.

mfriesen:
@Marcono1234 Great idea. I wasn't sure what the best approach would be. I'll review your suggestion and see if I can get it to work.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -86,8 +86,10 @@ private boolean includeField(Field f, boolean serialize) {
   private List<String> getFieldNames(Field f) {
     SerializedName annotation = f.getAnnotation(SerializedName.class);
     if (annotation == null) {
-      String name = fieldNamingPolicy.translateName(f);
-      return Collections.singletonList(name);
+      String fieldName = fieldNamingPolicy.translateName(f);
+      List<String> fieldNames = new ArrayList<>(fieldNamingPolicy.translateToAlternateNames(f));
+      fieldNames.add(fieldName);
```
Marcono1234, line 91: For consistency with `@SerializedName` should probably first add `fieldName` (so it will be used for serialization and deserialization by default), and then the 'alternate names' so that they are only used as fallback for deserialization.

And then maybe also add a test which verifies that when 'alternate names' are configured they don't affect serialization, and only affect deserialization.

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -33,8 +35,18 @@
...
    */
   public String translateName(Field f);
+
+  /**
+   * Translates the field name into its JSON field alternative names representation.
+   *
+   * @param f the field object that we are translating
+   * @return the list of possible translated field names.
+   */
+  default List<String> translateToAlternateNames(Field f) {
```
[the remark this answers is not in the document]
Marcono1234: This CodeQL finding can probably be ignored.

Though might be useful to also add a test which uses the `Field`, e.g. to convert its name to uppercase. (But that would likely not affect this CodeQL finding here, since CodeQL here is configured to not run for test code.)

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -33,8 +35,18 @@ public interface FieldNamingStrategy {
...
    *
    * @param f the field object that we are translating
-   * @return the translated field name.
+   * @return the list of possible translated field names.
    * @since 1.3
    */
   public String translateName(Field f);
+
+  /**
+   * Translates the field name into its JSON field alternative names representation.
```
Marcono1234, line 44: Would probably be good to mention that this only affects deserialization, and maybe also mention `@SerializedName#alternate`, for example something like this:
```suggestion
   * Translates the field name into its JSON field alternative names representation,
   * used for deserialization only. This is similar to {@link SerializedName#alternate()}.
```

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -33,8 +35,18 @@ public interface FieldNamingStrategy {
    * Translates the field name into its JSON field name representation.
    *
    * @param f the field object that we are translating
-   * @return the translated field name.
+   * @return the list of possible translated field names.
```
Marcono1234, line 38: Could you please revert this?

Marcono1234 (commented):
Thanks for the changes!

The build currently fails due to formatting violations. You can solve this by running `mvn spotless:apply`.

Also note that the commits are squashed on merge here, so you don't have to force-push / rewrite commits, but instead can incrementally make changes to this pull request.

----

@eamonnmcmanus what do you think about these proposed changes?

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -37,4 +40,15 @@ public interface FieldNamingStrategy {
...
    */
   public String translateName(Field f);
+
+  /**
+   * Translates the field name into its JSON field alternative names representation. used for
+   * deserialization only. This is similar to {@link SerializedName#alternate()}.
+   *
+   * @param f the field object that we are translating
+   * @return the list of possible translated field names.
+   */
```
Marcono1234, line 50: Sorry, forgot to mention that this should have a `@since $next-version$`
(`$next-version$` is a special placeholder which is replaced on next release)

Marcono1234 (commented):
Thanks for your changes! A few more small points, but looks good besides that.

We will however have to wait for the opinion and review from a direct Gson maintainer then.

Marcono1234 (approved):

mfriesen:
@Marcono1234 Is there anything else I need to do for this PR?

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -86,8 +86,12 @@ private boolean includeField(Field f, boolean serialize) {
   private List<String> getFieldNames(Field f) {
     SerializedName annotation = f.getAnnotation(SerializedName.class);
     if (annotation == null) {
-      String name = fieldNamingPolicy.translateName(f);
-      return Collections.singletonList(name);
+      String fieldName = fieldNamingPolicy.translateName(f);
```
eamonnmcmanus, line 89: I wonder if the logic for the two cases (with and without the annotation) could share more code. Would it be possible to have a `mainName` and `alternates`, which are set to `annotation.value()` and `Arrays.asList(annotation.alternate())` when `@SerializedName` is present; or `fieldNamingPolicy.translateName(f)` and `fieldNamingPolicy.translateToAlternateNames(f)` when it isn't? Then the code on lines 99–106 could be shared, with the necessary changes for `alternates` to be a list rather than an array.

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -37,4 +40,16 @@ public interface FieldNamingStrategy {
...
+
+  /**
+   * Translates the field name into its JSON field alternative names representation. Used for
+   * deserialization only. This is similar to {@link SerializedName#alternate()}.
+   *
+   * @param f the field object that we are translating
+   * @return the list of possible translated field names.
+   * @since $next-version$
+   */
+  default List<String> translateToAlternateNames(Field f) {
```
eamonnmcmanus, line 52: I think the existing name `translateName` is maybe not ideal. It's not really about translating the name, it's about deriving a name for the `Field`, which is _probably_ based in some way on the name of the field but could for example come from an annotation on the field. So perhaps just `alternateNames` would be a better name here.

eamonnmcmanus (changes requested):
Sorry, I lost sight of this. I think it makes sense. Just a couple of suggestions.

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
mfriesen: I have changed the method name to alternateNames

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
mfriesen: I have merged the similar code together and simplified the method.

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
```diff
@@ -37,4 +40,16 @@ public interface FieldNamingStrategy {
    * @since 1.3
    */
   public String translateName(Field f);
+
+  /**
+   * Translates the field name into its JSON field alternative names representation. Used for
```
eamonnmcmanus, line 45: Perhaps write something like this:

> Returns alternative names for this field when it is being deserialized. This is similar...

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -84,21 +84,21 @@ private boolean includeField(Field f, boolean serialize) {
...
+      alternates = fieldNamingPolicy.alternateNames(f);
+    } else {
+      fieldName = annotation.value();
+      alternates = Arrays.asList(annotation.alternate());
     }
 
-    List<String> fieldNames = new ArrayList<>(alternates.length + 1);
-    fieldNames.add(serializedName);
-    Collections.addAll(fieldNames, alternates);
+    List<String> fieldNames = new ArrayList<>(alternates.size() + 1);
```
eamonnmcmanus, line 99: Although it isn't strictly necessary, I think the special case for `Collections.singletonList` should be preserved, because it is also the usual case.

eamonnmcmanus (changes requested):
Thanks! Just a couple of small things...

gson/src/main/java/com/google/gson/FieldNamingStrategy.java
mfriesen: Javadoc has been updated as per the suggestion.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
mfriesen: The special case with Collections.singletonList has been added back.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -84,21 +84,24 @@ private boolean includeField(Field f, boolean serialize) {
...
-    String serializedName = annotation.value();
-    String[] alternates = annotation.alternate();
-    if (alternates.length == 0) {
-      return Collections.singletonList(serializedName);
+      fieldName = fieldNamingPolicy.translateName(f);
+      alternates = fieldNamingPolicy.alternateNames(f);
+    } else {
+      fieldName = annotation.value();
+      alternates = Arrays.asList(annotation.alternate());
+      if (alternates.size() == 0) {
```
eamonnmcmanus, line 97: I think this `if` could be moved outside the containing `if` since it would apply to both cases.

eamonnmcmanus (changes requested):

gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
```diff
@@ -237,6 +238,78 @@ public void testAtSignInSerializedName() {
...
+                    return "primary-name";
+                  }
+
+                  @Override
+                  public List<String> alternateNames(Field f) {
+                    return List.of("SomeConstantStringInstanceField");
+                  }
+                })
+            .create();
+    String target = "{\"SomeConstantStringInstanceField\":\"someValue\"}";
```
Marcono1234, line 282: Could you please change this name here to a more custom name? I am a bit worried that `SomeConstantStringInstanceField` is too close to the real field name. So it could look like the field naming strategy is not actually used.
```suggestion
                    return List.of("alternate-name");
                  }
                })
            .create();
    String target = "{\"alternate-name\":\"someValue\"}";
```

gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
```diff
@@ -237,6 +238,78 @@ public void testAtSignInSerializedName() {
...
+                    return "SomeConstantStringInstanceField";
+                  }
+
+                  @Override
+                  public List<String> alternateNames(Field f) {
+                    return List.of("alternate-name");
+                  }
+                })
+            .create();
+    String target = "{\"SomeConstantStringInstanceField\":\"someValue\"}";
```
Marcono1234, line 259: Similar to the other comment, use a custom name which differs more from the real field name.
```suggestion
                    return "primary-name";
                  }

                  @Override
                  public List<String> alternateNames(Field f) {
                    return List.of("alternate-name");
                  }
                })
            .create();
    String target = "{\"primary-name\":\"someValue\"}";
```

gson/src/test/java/com/google/gson/functional/NamingPolicyTest.java
```diff
@@ -237,6 +238,78 @@ public void testAtSignInSerializedName() {
...
+                new FieldNamingStrategy() {
+
+                  @Override
+                  public String translateName(Field f) {
+                    return "some-constant-string-instance-field";
+                  }
+
+                  @Override
+                  public List<String> alternateNames(Field f) {
+                    return List.of("SomeConstantStringInstanceField");
```
Marcono1234, line 301:
```suggestion
                    return List.of("alternate-name");
```

Marcono1234 (commented):
Thanks a lot for updating this pull request! I have also added some small suggestion comments, hopefully that is ok for you.

eamonnmcmanus (approved):
This looks good to me, but I will wait for @Marcono1234 before merging.

Marcono1234 (approved):

Marcono1234:
@eamonnmcmanus, I think if we strictly followed SemVer, the latest version should have been 2.14.0 (instead of 2.13.1) since this PR here added a new method. But hopefully that does not matter that much.

eamonnmcmanus:
> @eamonnmcmanus, I think if we strictly followed SemVer, the latest version should have been 2.14.0 (instead of 2.13.1) since this PR here added a new method. But hopefully that does not matter that much.

Yes, I realized that when it came time to draft the release notes, but I'd already published the release at that point. I agree that it doesn't matter very much.

PR #2779, opened 2024-12-02
bug fix issue #2769

### LINKED ISSUE #2769: JsonReader with strictness set to LENIENT,
# Gson version
2.11.0

# Java / Android version
any version

# Used tools
none

# Description
With a JsonReader with strictness = LENIENT, all values after a null are read as a single string begining by "null".

## Expected behavior
Values are correctly read.

# Reproduction steps
```java
public static void test(Stream<String> stream) throws IOException {
    Iterator<String> iterator = stream.iterator();
    StringWriter str = new StringWriter();

    try (JsonWriter writer = new JsonWriter(str)) {
        while (iterator.hasNext()) {
            TypeAdapters.STRING.write(writer, iterator.next());
        }
        writer.flush();
    }

    JsonReader reader = new JsonReader(new StringReader(str.toString()));
    reader.setStrictness(Strictness.LENIENT);

    System.out.println(TypeAdapters.STRING.read(reader));
    System.out.println(TypeAdapters.STRING.read(reader));
  }
```  

  

With `stream = Stream.of("value1", "value2");` -> OK
```
value1
value2
```   

  

With `stream = Stream.of("value1", null);`, -> OK
```
value1
null
```
(the printed null is really a null value, not "null" String -> OK)  

  
With `stream = Stream.of(null, "value1");` -> KO
```
null"value1"
java.lang.IllegalStateException: Expected a string but was END_DOCUMENT at line 1 column 13 path $
```  

  
  
With `stream = Stream.of(null, "value1", "value2");` -> KO  (and so on)
```
null"value1""value2"
Exception in thread "main" java.lang.IllegalStateException: Expected a string but was END_DOCUMENT at line 1 column 21 path $
```  

  

With `stream = Stream.of("value1", null, "value2");` -> KO
```
value1
null"value2"
```  

  
With `stream = Stream.of("value1", null, "value2", "value3");` -> KO (and so on)
```
value1
null"value2""value3"
```  

  

Same behavior happens with other json primitive type but not with json object or json array.
Save behavior happens in previous gson version with `#setLenient(true);`.

IlCommittatore:
### Purpose
Hello! I think this pull request could fix the issue with the id : #2769.
Closes #2769


Marcono1234:
Thanks a lot for these changes!

I am a bit concerned that tweaking lenient parsing like this could have undesired side-effects for other use cases. As mentioned in https://github.com/google/gson/issues/2769#issuecomment-2425011564, the proper solution might be if users explicitly added a separator between values when writing them to avoid ambiguities during parsing (see also #2783).

However, I will leave this pull request open and let the direct project members decide since they might have a different opinion on this.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1172,6 +1172,7 @@ private String nextUnquotedValue() throws IOException {
     while (true) {
       for (; pos + i < limit; i++) {
         switch (buffer[pos + i]) {
+          case '"':
```
Marcono1234, line 1175: Maybe for consistency it would also be necessary to change `JsonReader#skipUnquotedValue` so that reading a value and skipping a value is consistent, and adding a test for this.

Not sure if `JsonReader#isLiteral` would have to be changed as well, maybe not.

Issue #1733, opened 2020-07-07
Making Gson.fromJson(Reader, Class) return after one JSON value is read


TaylanUB:
When you pass a Reader to `Gson.fromJson()`, it seems to keep reading data until EOF or a syntax error. Perhaps this is desired in some scenarios, like strict error checking to ensure there's no superfluous data after the JSON value, but what if you want to read a JSON value that's just part of a longer-running stream?

It would be great if, as an optional setting, one could instruct Gson to stop reading from a reader after the end of a value is reached.

For all value types except Number, the end of the value is unambiguous: in case of an Array the closing square bracket, in case of an Object the closing curly brace, in case of a string the closing double quote, in case of booleans and null the end of the corresponding character string.

Numbers don't have an obvious end. Theoretically one could stop reading when the addition of the next digit would cause an overflow, but that might mask some actual errors. I have no opinion on how exactly the behavior should be because I don't think I'll ever need to read bare JSON numbers over a continuous stream; one can always wrap them in a one-element array or a trivial object like `{"value":12345}`.

Ensuring proper behavior would probably require the use of `mark()` and `reset()` on the reader. Once Gson has read a whole chunk of characters, it might realize that it's already read beyond the end of the JSON value. In that case it would need to count the number of characters that belong to the JSON value, `reset()` to before the last `read()` operation, and re-read only as many characters as it knows there exist until the end of the JSON value.

Thoughts?

lyubomyr-shaydariv:
Not sure but are you probably talking about something like:

```java
final Gson gson = new Gson();
final JsonReader jsonReader = new JsonReader(new StringReader("1 2 3 4"));
jsonReader.setLenient(true);
while ( jsonReader.peek() != JsonToken.END_DOCUMENT && jsonReader.hasNext() ) {
	final int i = gson.fromJson(jsonReader, int.class);
	System.out.println(i);
}
```

? This seems to fit

> what if you want to read a JSON value that's just part of a longer-running stream

> instruct Gson to stop reading from a reader after the end of a value is reached

>  read bare JSON numbers over a continuous stream; one can always wrap them in a one-element array or a trivial object like `{"value":12345}`

if combined, but does not seem to fit `mark()` and `reset()` though.

TaylanUB:
@lyubomyr-shaydariv I don't think this does what I want. The main problem is that `.fromJson()` as well as `.peek()` might block even if a whole JSON value is available in the stream, because they will wait for EOF or another token.

[...]

PR #2783, opened 2024-12-26
Add better support for multiple top-level values

### LINKED ISSUE #2769: JsonReader with strictness set to LENIENT,

Marcono1234:
### Purpose
Add better support for reading and writing multiple top-level values

Resolves #2769
Related to #1733
Obsoletes #2779

### Description
Currently reading and writing multiple top-level JSON values is coupled with lenient mode. This might be undesired for users who normally want to use strict mode.
Having multiple top-level values is a common use case when streaming JSON values, see for example [JSON Lines](https://jsonlines.org/). But another problem is that `JsonWriter` currently concatenates multiple JSON values without any separating whitespace, and without any way to configure it.

Other JSON libraries seem to have dedicated classes or configuration for multiple top-level value support, e.g. [Jackson's `MappingIterator` and `SequenceWriter`](https://cowtowncoder.medium.com/line-delimited-json-with-jackson-69c9e4cb6c00) and [kotlinx-serialization's `DecodeSequenceMode.WHITESPACE_SEPARATED`](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-decode-sequence-mode/-w-h-i-t-e-s-p-a-c-e_-s-e-p-a-r-a-t-e-d/).
However, because Gson's `JsonReader` and `JsonWriter` already had (limited) functionality for multiple top-level values, I added new API to those classes instead of adding completely new classes.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -70,8 +70,9 @@ import java.util.Objects;
  * The behavior of this reader can be customized with the following methods:
  *
  * <ul>
- *   <li>{@link #setStrictness(Strictness)}, the default is {@link Strictness#LEGACY_STRICT}
+ *   <li>{@link #setMultiTopLevelValuesAllowed(boolean)}, the default is {@code false}
  *   <li>{@link #setNestingLimit(int)}, the default is {@value #DEFAULT_NESTING_LIMIT}
+ *   <li>{@link #setStrictness(Strictness)}, the default is {@link Strictness#LEGACY_STRICT}
  * </ul>
  *
  * The default configuration of {@code JsonReader} instances used internally by the {@link Gson}
@@ -257,6 +258,8 @@ public class JsonReader implements Closeable {
   static final int DEFAULT_NESTING_LIMIT = 255;
   private int nestingLimit = DEFAULT_NESTING_LIMIT;
 
+  private boolean multiTopLevelValuesEnabled = false;
+
   static final int BUFFER_SIZE = 1024;
 
   /**
@@ -380,7 +383,8 @@ public class JsonReader implements Closeable {
    *         <li>Streams that start with the <a href="#nonexecuteprefix">non-execute prefix</a>,
    *             {@code ")]}'\n"}
    *         <li>Streams that include multiple top-level values. With legacy strict or strict
-   *             parsing, each stream must contain exactly one top-level value.
+   *             parsing, each stream must contain exactly one top-level value. Can be enabled
+   *             independently using {@link #setMultiTopLevelValuesAllowed(boolean)}.
    *         <li>Numbers may be {@link Double#isNaN() NaNs} or {@link Double#isInfinite()
    *             infinities} represented by {@code NaN} and {@code (-)Infinity} respectively.
    *         <li>End of line comments starting with {@code //} or {@code #} and ending with a
@@ -402,8 +406,8 @@ public class JsonReader implements Closeable {
    * @since 2.11.0
    */
   public final void setStrictness(Strictness strictness) {
-    Objects.requireNonNull(strictness);
-    this.strictness = strictness;
+    this.strictness = Objects.requireNonNull(strictness);
+    setMultiTopLevelValuesAllowed(strictness == Strictness.LENIENT);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -416,6 +420,31 @@ public final Strictness getStrictness() {
...
+
+  /**
+   * Returns whether multiple top-level values are allowed.
+   *
+   * @see #setMultiTopLevelValuesAllowed(boolean)
+   * @since $next-version$
+   */
+  public final boolean isMultiTopLevelValuesAllowed() {
+    return multiTopLevelValuesEnabled;
+  }
```
Marcono1234, line 446: I am not completely happy about these method names `setMultiTopLevelValuesAllowed` and `isMultiTopLevelValuesAllowed` yet. And feedback and suggestions are welcome.

gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -661,10 +690,13 @@ int doPeek() throws IOException {
...
-      } else {
-        checkLenient();
-        pos--;
+      } else if (!multiTopLevelValuesEnabled) {
+        throw new MalformedJsonException(
+            "Multiple top-level values support has not been enabled, use"
+                + " `JsonReader.setMultiTopLevelValuesAllowed(true)`,"
+                + locationString());
       }
+      pos--;
```
Marcono1234, line 699: Moved this `pos--;` outside the if-else since the `if { ... }` performs a `return`.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -807,10 +845,14 @@ private void beforeName() throws IOException {
...
     switch (peek()) {
       case NONEMPTY_DOCUMENT:
-        if (strictness != Strictness.LENIENT) {
-          throw new IllegalStateException("JSON must have only one top-level value.");
+        if (topLevelSeparator == null) {
+          throw new IllegalStateException(
+              "Multiple top-level values support has not been enabled, use"
+                  + " `JsonWriter.setTopLevelSeparator(String)`");
         }
-        // fall-through
```
Marcono1234, line 813: Looks like this "fall-through" was previously redundant: The next case is `EMPTY_DOCUMENT` which replaces the top of the stack with `NONEMPTY_DOCUMENT`, but here the top is already `NONEMPTY_DOCUMENT`.

Marcono1234:
An alternative would be to always enable support for multiple top-level values, regardless of strictness, and then only have `JsonWriter#setTopLevelSeparator(String)` to allow customizing the separator when writing.

Though since there are multiple formats / standards for encoding multiple top-level values, maybe this is not desired? And could cause confusion when users accidentally write or read multiple top-level values? (though a bit unlikely?)

gson/src/main/java/com/google/gson/stream/JsonReader.java
eamonnmcmanus: What about `setAllowingMultipleValues` and `isAllowingMultipleValues`? "Multiple values" isn't quite as precise as "multi top-level values", but the latter is kind of verbose. I'm proposing "allowing X" rather than "X allowed" to avoid the grammatical issue with "is top-level values allowed".

eamonnmcmanus (changes requested):
If you are on for reviving this, I have run across some use cases for it and would be interested in following through, ideally before the next release.

gson/src/main/java/com/google/gson/stream/JsonReader.java
Marcono1234: Ok, have renamed the methods. I have left the corresponding test methods unchanged (`enabled`/`disabled`) to make it more obvious that they are about explicitly enabling / disabling this.

gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -356,6 +367,33 @@ public final Strictness getStrictness() {
...
+   * Using {@code null} disables support for multiple top-level values (the default).
+   *
+   * <p>This setting overwrites and is overwritten by whether {@link #setStrictness(Strictness)}
+   * enabled support for multiple top-level values.
+   *
+   * @param separator separator between top-level values, or {@code null} to disable.
+   * @see #getTopLevelSeparator()
+   * @since $next-version$
+   */
+  public final void setTopLevelSeparator(String separator) {
```
Marcono1234, line 383: Not sure if we should also support an `boolean addTrailing` parameter, so that the JsonWriter emits the separator also after the last element.

It seems [JSON Lines](https://jsonlines.org/) recommends adding a trailing line break, though does not require it. Not sure about other multi-value JSON standards.

PR #2786, opened 2025-01-01
Fix collection tests for latest guava-testlib update

Marcono1234:
### Purpose
Fixes build failure for #2784

**Important:** Targets the `dependabot/...` branch instead of `main`.

### Description
With the fix in the latest guava-testlib version (see https://github.com/google/guava/issues/7401) null support is now properly checked. But it expects that if a collection does not permit null, then creating it with null should fail as well, see guava-testlib's:
- `MapCreationTester.testCreateWithNullValueUnsupported()`
- `CollectionCreationTester.testCreateWithNull_unsupported()`

However, the previous implementation for the JsonArray and JsonObject test was using methods which implicitly converted null to JsonNull.

### Checklist

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/LinkedTreeMapSuiteTest.java
+++ b/gson/src/test/java/com/google/gson/internal/LinkedTreeMapSuiteTest.java
@@ -31,6 +31,8 @@ public class LinkedTreeMapSuiteTest {
 
     @Override
     protected Map<String, String> create(Entry<String, String>[] entries) {
+      // This is not completely accurate: Because LinkedTreeMap has no constructor which accepts
+      // existing entries, this has to add the entries individually with `Map#put`
       var map = new LinkedTreeMap<String, String>(allowNullValues);
       for (var entry : entries) {
         map.put(entry.getKey(), entry.getValue());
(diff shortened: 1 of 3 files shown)


gson/src/test/java/com/google/gson/internal/LinkedTreeMapSuiteTest.java
```diff
@@ -74,7 +74,8 @@ public static Test suite() {
             .named("nullValues=false")
             .createTestSuite();
 
-    TestSuite testSuite = new TestSuite("LinkedTreeMap");
+    // Use qualified class name to make it easier to find this test class in the IDE
+    TestSuite testSuite = new TestSuite(LinkedTreeMapSuiteTest.class.getName());
```
Marcono1234, line 78: For Eclipse this makes a difference: Otherwise it just shows the test suite name in the UI, and assertion stack traces are just Guava code, so there is no way to easily get to this test class.

Maven and IntelliJ automatically include the `LinkedTreeMapSuiteTest` class name.

The other test suite classes here are affected as well, but changing their name (e.g. "JsonObject#asMap") to the qualified class name would make the names derived for nested tests by Guava quite verbose. Could alternatively wrap the test suites in an additional `TestSuite` with the class as name (similar to here), but not sure if worth it.

eamonnmcmanus:
Thanks!

PR #2789, opened 2025-01-27
Allow registering adapters for `JsonElement` again

### LINKED ISSUE #2787: GsonBuilder backwards compatibility broken, throws "Cannot override built-in adapter" error
# Gson version
2.11

# Java / Android version
Semeru JDK 21.0.5

# Used tools

- [ ] Maven; version: 
- [x] Gradle; version: 8.12
- [ ] ...

# Description
We are currently maintaining a legacy application in an ongoing support mode. In order to address security vulnerabilities, we needed to update several third-party dependencies, which required upgrading to Gson version 2.11. However, after the update, another dependency, com.cloudant:cloudant-client:2.20.1 (https://github.com/cloudant/java-cloudant), began to fail. This dependency worked correctly with Gson 2.10.1 but now fails with Gson 2.11. The error encountered is as follows:
```
java.lang.IllegalArgumentException: Cannot override built-in adapter for class com.google.gson.JsonObject
 	at com.google.gson.GsonBuilder.registerTypeAdapter(GsonBuilder.java:714) ~[gson-2.11.0.jar:?]
 	at com.cloudant.client.org.lightcouch.internal.GsonHelper.initGson(GsonHelper.java:38) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.org.lightcouch.CouchDbClient.<init>(CouchDbClient.java:95) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.org.lightcouch.CouchDbClient.<init>(CouchDbClient.java:145) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.api.CloudantClient.<init>(CloudantClient.java:141) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.api.ClientBuilder.build(ClientBuilder.java:396) ~[cloudant-client-2.20.1.jar:?]
```
The issue appears to be related to how GsonBuilder is used in the method defined in https://github.com/cloudant/java-cloudant/blob/master/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/GsonHelper.java
```
    public static GsonBuilder initGson(GsonBuilder gsonBuilder) {
        gsonBuilder.registerTypeAdapter(JsonObject.class, new JsonDeserializer<JsonObject>() {
            public JsonObject deserialize(JsonElement json,
                                          Type typeOfT, JsonDeserializationContext context)
                    throws JsonParseException {
                return json.getAsJsonObject();
            }
        });
        gsonBuilder.registerTypeAdapter(JsonObject.class, new JsonSerializer<JsonObject>() {
            public JsonElement serialize(JsonObject src, Type typeOfSrc,
                                         JsonSerializationContext context) {
                return src.getAsJsonObject();
            }

        });

        return gsonBuilder;
    }
```
It seems that something has changed in GsonBuilder between versions 2.10.1 and 2.11 that is now causing a conflict and breaking backward compatibility. Could you clarify whether this is a bug, or is it the intended behavior in the new Gson version?

## Expected behavior
Considering this was minor version update, I would expect that GsonBuilder would continue to work as it was.

## Actual behavior
It seems that something has changed in GsonBuilder between versions 2.10.1 and 2.11 that is now causing a conflict and breaking backward compatibility. 

# Reproduction steps
https://github.com/cloudant/java-cloudant/blob/master/cloudant-client/src/main/java/com/cloudant/client/org/lightcouch/internal/GsonHelper.java should still be able to register TypeAdapter in GsonBuilder.

_NOTE: I can't change com.cloudant:cloudant-client:2.20.1 code. I know that it's deprecated, sunset and no longer supported by any team - but to replace it with some other library is involved change. Therefore I am trying to rather ask if backward compatibility of GsonBuilder could be kept._

# Exception stack trace
```
java.lang.IllegalArgumentException: Cannot override built-in adapter for class com.google.gson.JsonObject
 	at com.google.gson.GsonBuilder.registerTypeAdapter(GsonBuilder.java:714) ~[gson-2.11.0.jar:?]
 	at com.cloudant.client.org.lightcouch.internal.GsonHelper.initGson(GsonHelper.java:38) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.org.lightcouch.CouchDbClient.<init>(CouchDbClient.java:95) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.org.lightcouch.CouchDbClient.<init>(CouchDbClient.java:145) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.api.CloudantClient.<init>(CloudantClient.java:141) ~[cloudant-client-2.20.1.jar:?]
 	at com.cloudant.client.api.ClientBuilder.build(ClientBuilder.java:396) ~[cloudant-client-2.20.1.jar:?]
```

Marcono1234:
### Purpose
Fixes #2787

### Description
This is done for backward compatibility, however registering the adapter still has no effect because the built-in adapter has higher precedence (like it is the case for multiple years already).

The Javadoc for the GsonBuilder methods has not been adjusted, to discourage users from trying to register adapters for JsonElement.

Registering an adapter for `Object` is still not allowed, it is less likely there that users have done this _and_ have not noticed that the code has no effect for multiple years since unlike for `JsonElement` there might not be a canonical implementation for such an adapter, or at least the implementation wouldn't be as simple as for `JsonElement`.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -710,7 +710,7 @@ public final class GsonBuilder {
             || typeAdapter instanceof InstanceCreator<?>
             || typeAdapter instanceof TypeAdapter<?>);
 
-    if (isTypeObjectOrJsonElement(type)) {
+    if (hasNonOverridableAdapter(type)) {
       throw new IllegalArgumentException("Cannot override built-in adapter for " + type);
     }
 
@@ -730,9 +730,14 @@ public final class GsonBuilder {
     return this;
   }
 
-  private static boolean isTypeObjectOrJsonElement(Type type) {
-    return type instanceof Class
-        && (type == Object.class || JsonElement.class.isAssignableFrom((Class<?>) type));
+  /** Whether the type has a built-in adapter which cannot be overridden. */
+  private static boolean hasNonOverridableAdapter(Type type) {
+    return type == Object.class;
+    // This should also cover `JsonElement.class.isAssignableFrom(type)`, however for backward
+    // compatibility this is not covered here because really old Gson versions had no built-in
+    // adapter for JsonElement so users registered custom adapters. These adapters don't have any
+    // effect in recent Gson versions. See
+    // https://github.com/google/gson/issues/2787#issuecomment-2581568157
   }
 
   /**
@@ -777,10 +782,6 @@ public final class GsonBuilder {
         typeAdapter instanceof JsonSerializer<?>
             || typeAdapter instanceof JsonDeserializer<?>
             || typeAdapter instanceof TypeAdapter<?>);
-
-    if (JsonElement.class.isAssignableFrom(baseType)) {
-      throw new IllegalArgumentException("Cannot override built-in adapter for " + baseType);
-    }
 
     if (typeAdapter instanceof JsonDeserializer || typeAdapter instanceof JsonSerializer) {
       hierarchyFactories.add(TreeTypeAdapter.newTypeHierarchyFactory(baseType, typeAdapter));
--- a/gson/src/test/java/com/google/gson/GsonBuilderTest.java
+++ b/gson/src/test/java/com/google/gson/GsonBuilderTest.java
@@ -28,6 +28,7 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.text.DateFormat;
 import java.util.Date;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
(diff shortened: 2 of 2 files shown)


eamonnmcmanus:
Thanks! I'll merge this as soon as the tests pass (assuming they do). Then I think it is about time we had another release.

Marcono1234:
Should this be backported and released as Gson 2.11.1 without the Java 8 and API changes, to make it easier for users to switch to that version?

CC @jirijanak would migration from Java 7 to Java 8 as minimum version be a problem for you?

PR #2793, opened 2025-01-29
Add adapters for Guava's immutable ccollections.

eamonnmcmanus:
These adapters are added automatically, but only if Guava is on the classpath. They correctly deserialize immutable collections and maps, by creating a builder and adding the incoming elements to it, before finally building the immutable object.

The principal motivation for this change is that #2068, if applied, will trigger a large number of failures from Google's internal tests. Code is incorrectly deserializing `TypeToken<ImmutableList<String>>` and the like, and getting away with it because it doesn't happen to perform any actions that would notice that the deserialized object is actually an `ArrayList`.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -145,9 +145,10 @@
                 Bundle-Vendor: Google Gson Project
                 Bundle-ContactAddress: ${project.parent.url}
 
-                # Optional dependency for JDK's sun.misc.Unsafe
+                # Optional dependencies for JDK's sun.misc.Unsafe and Guava's collections.
+                # Both are only accessed through reflection and the code works without them.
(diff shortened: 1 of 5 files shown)


eamonnmcmanus:
I think the approach here is not really viable. With ProGuard/R8 and also with GraalVM, you can end up having only a subset of Guava methods be available. That makes a reflective solution like the one here seem very fragile. Instead, for Google's tests that fail with #2068, I am planning to update the affected code so it explicitly registers a `TypeAdapterFactory` that understands Guava's immutable collection types. We might consider including that in Gson, _without_ turning it on by default. But it would imply a new optional dependency on Guava, and I think that might be more trouble than it's worth.

Issue #2794, opened 2025-01-30
Only import errorprone.annotations with optional (if at all)


laeubi:
# Gson version
2.12.0

# Java / Android version
any java version

# Used tools

- [x] Maven; version: 
- [ ] Gradle; version: 
- [ ] ProGuard (attach the configuration file please); version: 
- [ ] ...

# Description
Currently the gson release requires in its manifest:

`com.google.errorprone.annotations;version="[2.36,3)"`

this requires to pull in the error prone annotations to the OSGi runtime.

## Expected behavior
The annotations should only be imported with optional

`com.google.errorprone.annotations;version="[2.36,3)";resolution:=optional`

because they are not interesting/required at runtime, see also

- https://github.com/google/error-prone/issues/4797

or not imported at all.

eamonnmcmanus:
See [this comment](https://github.com/google/gson/blob/b2e26fa97b7ccba080a082a9bf9741e24d5c523d/gson/pom.xml#L45-L54) and the discussion it links to. Unless you have reason to think the situation has evolved, this doesn't seem like something we want to change.

chrisrueger:
@eamonnmcmanus I support @laeubi in this regard and would encourage to reopen. 

This topic is similar to my work done in https://github.com/google/gson/pull/2735 and is about runtime and resolving bundles in an OSGi context. Without `resolution:=optional` all OSGi applications now need to add error-prone as a runtime depedency to their OSGi containers.

laeubi:
Especially the last version of the artifact don't seem to use that header at all 

![image](https://github.com/user-attachments/assets/30147cb9-febf-4810-a7cc-fc4e5e80b70d)

eamonnmcmanus:
OK. To be honest, OSGi support is something of an annoyance to the project maintainers, since none of us has expertise in it or uses it in any way. As recently as yesterday I was forced to spend quite some time figuring out how to adjust the OSGi configuration, and I seriously contemplated just removing it from the project. However, apparently some people do need it, presumably because of Eclipse, and this change is probably harmless.

iloveeclipse:
Is there a chance to release 2.12.1 ASAP with this fix?

eamonnmcmanus:
> Is there a chance to release 2.12.1 ASAP with this fix?

Has something changed to make this urgent? Was the situation not already the same with 2.11.0, released last May?

iloveeclipse:
[...]

PR #2795, opened 2025-01-30
Make the import of com.google.errorprone optional

laeubi:
Fix https://github.com/google/gson/issues/2794


eamonnmcmanus:
Closing for the same reason as the corresponding issue.

PR #2800, opened 2025-02-01
Fix TODO and add proper documentation for GraphAdapterBuilder

panic08:
I have described a detailed java doc for extras/GraphAdapterBuilder and thus TODO corrected

### DIFF
--- a/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
+++ b/extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
@@ -38,8 +38,38 @@ import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Queue;
 
-/** Writes a graph of objects as a list of named nodes. */
-// TODO: proper documentation
+/**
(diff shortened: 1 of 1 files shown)


extras/src/main/java/com/google/gson/graph/GraphAdapterBuilder.java
```diff
@@ -51,6 +81,13 @@ public GraphAdapterBuilder() {
             instanceCreators, true, Collections.<ReflectionAccessFilter>emptyList());
   }
 
+  /**
+   * Registers the specified type with a default instance creator. The default instance creator is
+   * obtained using the {@link ConstructorConstructor} for the given type.
```
eamonnmcmanus, line 86: I think I would delete this sentence, since `ConstructorConstructor` is an internal type.

eamonnmcmanus (changes requested):
This is really excellent, thank you! Just one very small thing.

Incidentally this code isn't published to Maven Central. It _is_ available internally to code at Google, but nothing is currently using it. Perhaps the better documentation will change that.

eamonnmcmanus (approved):

PR #2809, opened 2025-02-20
Refactore unreachable default case in `JsonReader#peek`

aldenbro:
### Purpose

Refactored unreachable default case in `JsonReader#peek` so that it can be reached while keeping existing functionality.

### Checklist

- [X] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)\
- [X] If necessary, new public API validates arguments, for example rejects `null`
- [X] New public API has Javadoc
    - [X] Javadoc uses `@since $next-version$`  
- [X] If necessary, new unit tests have been added  
  - [X] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [X] No JUnit 3 features are used (such as extending class `TestCase`)
  - [X] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [X] `mvn clean verify javadoc:jar` passes without errors

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -573,10 +573,9 @@ public class JsonReader implements Closeable {
       case PEEKED_LONG:
       case PEEKED_NUMBER:
         return JsonToken.NUMBER;
-      case PEEKED_EOF:
-        return JsonToken.END_DOCUMENT;
       default:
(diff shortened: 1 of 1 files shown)


aldenbro:
The reason for making this change is to increase code coverage. 

As `p` is effectively operating as an enum, no invalid values should ever be present, and thus the assertion should never fail.

eamonnmcmanus:
Closing for lack of response.

PR #2810, opened 2025-02-20
Remove redundant generic type arguments from method calls

Marcono1234:
### Description
The type arguments can be inferred now (maybe due to our Java 8 migration).

In the test code there is some remaining usage of `gson.<TypeArg>fromJson(..., Type)`, but it seems that is needed to prevent compilation errors due to ambiguous `assertThat` calls.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -254,7 +254,7 @@ public final class Gson {
     this(
         Excluder.DEFAULT,
         DEFAULT_FIELD_NAMING_STRATEGY,
-        Collections.<Type, InstanceCreator<?>>emptyMap(),
+        Collections.emptyMap(),
         DEFAULT_SERIALIZE_NULLS,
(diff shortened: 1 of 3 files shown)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -254,7 +254,7 @@ public Gson() {
     this(
         Excluder.DEFAULT,
         DEFAULT_FIELD_NAMING_STRATEGY,
-        Collections.<Type, InstanceCreator<?>>emptyMap(),
+        Collections.emptyMap(),
```
Marcono1234, line 257: Not sure if the changes here for the `Gson` constructor are good, or if without type arguments this makes the constructor call more difficult to understand.

Might depend on whether your IDE shows parameter name hints.

eamonnmcmanus (approved):

gson/src/main/java/com/google/gson/Gson.java
eamonnmcmanus: I think it's fine.

PR #2812, opened 2025-02-20
Fix duplicate resource warning for shrinker test

Marcono1234:
### Description
Currently the `test-shrinker` Maven module produces this warning:
```
[WARNING] error_prone_annotations-2.36.0.jar, gson-2.12.2-SNAPSHOT.jar define 1 overlapping classes:
[WARNING]   - META-INF.versions.9.module-info
[WARNING] maven-shade-plugin has detected that some files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the file is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See https://maven.apache.org/plugins/maven-shade-plugin/
```

Excluding `module-info.class` should probably not affect shrinker behavior.

### DIFF
--- a/test-shrinker/pom.xml
+++ b/test-shrinker/pom.xml
@@ -104,6 +104,7 @@
             <lib>${java.home}/jmods/java.compiler.jmod</lib>
           </libs>
           <!-- Include dependencies in the final JAR -->
+          <!-- Note: Currently causes warnings due to duplicate resource files, see https://github.com/wvengen/proguard-maven-plugin/issues/388 -->
           <includeDependencyInjar>true</includeDependencyInjar>
           <outjar>proguard-output.jar</outjar>
         </configuration>
@@ -132,6 +133,7 @@
                   <excludes>
                     <!-- Ignore duplicate files in dependencies -->
                     <exclude>META-INF/MANIFEST.MF</exclude>
+                    <exclude>META-INF/versions/9/module-info.class</exclude>
                   </excludes>
                 </filter>
               </filters>


test-shrinker/pom.xml
```diff
@@ -104,6 +104,7 @@
             <lib>${java.home}/jmods/java.compiler.jmod</lib>
           </libs>
           <!-- Include dependencies in the final JAR -->
+          <!-- Note: Currently causes warnings due to duplicate resource files, see https://github.com/wvengen/proguard-maven-plugin/issues/388 -->
```
eamonnmcmanus, line 107: I'm a bit confused. Is this PR reducing the number of warnings, but not to zero?

Marcono1234: There are two separate warnings, one is by the Maven Shade Plugin which prepares the JAR for R8 (that warning is fixed by this PR).

The other warning is by ProGuard (this comment refers to it):
> [proguard] Note: duplicate definition of resource file [META-INF/MANIFEST.MF]
> [proguard] Note: duplicate definition of resource file [META-INF/MANIFEST.MF]
> [proguard] Note: there were 2 duplicate class definitions.
> [proguard]       (https://www.guardsquare.com/proguard/manual/troubleshooting#duplicateclass)
> ...
> [proguard] Warning: can't write resource [META-INF/MANIFEST.MF] (Duplicate jar entry [META-INF/MANIFEST.MF])
> [proguard] Warning: can't write resource [META-INF/MANIFEST.MF] (Duplicate jar entry [META-INF/MANIFEST.MF])

Which I think will be resolved once that proguard-maven-plugin issue is fixed.

----

These warnings likely don't affect the shrinker test, regardless of whether they are addressed or not. But they might be a bit distracting, which is why I tried to resolve them.

eamonnmcmanus: OK, thanks!

eamonnmcmanus (approved):

PR #2814, opened 2025-02-22
Fix numeric conversion to Long when serializing numbers as JsonElement tree

### LINKED ISSUE #2680: TypeAdapterFactory INTEGER_FACTORY will change INTEGER type to LONG type
# Gson version
2.10.1

# Java / Android version
21

# Description
I try to get a JsonElement by toJsonTree method. but i found that when i use integer object as input, i will get a long object with method ```JsonElement.getAsNumber()```

I think this commit caused this problem：https://github.com/google/gson/commit/3e3266cf48f132928225e1561a6ae4cb5503d08f

Is this a bug or a feature?

## Expected behavior
use integer type as input, get integer type as output

## Actual behavior
use integer type as input, get long type as output

# Reproduction steps
```
    @Test
    public void test(){
        Integer testNum = 1;
        JsonElement jsonTree = new Gson().toJsonTree(testNum);
        System.out.println(jsonTree.getAsNumber().getClass());
    }
```

Marcono1234:
### Purpose
Fixes #2680

### Description
As mentioned in #2680, an unintended side-effect of https://github.com/google/gson/commit/3e3266cf48f132928225e1561a6ae4cb5503d08f was that serializing to a `JsonElement` tree is now converting all integral numbers to `Long` instances.

This pull request here tries to solve this by using the `JsonWriter#value(Number)` overload again, boxing the converted number. If the type is already matching (e.g. serializing a `Byte` as `Byte`) then this also has the advantage that it avoids unwrapping and having `JsonTreeWriter` wrap the value again.
But on the other hand when using `JsonWriter` and not `JsonTreeWriter` there might be a bit overhead from `JsonWriter#value(Number)` compared to `JsonWriter#value(long)`.

An alternative to this PR could be to instead revert https://github.com/google/gson/commit/3e3266cf48f132928225e1561a6ae4cb5503d08f partially or completely. After all the underlying issue #2156 had been reported by me (and was not blocking me in any way), so maybe other users would not have cared about the missing numeric conversion.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -210,7 +210,9 @@ public final class TypeAdapters {
           if (value == null) {
             out.nullValue();
           } else {
-            out.value(value.byteValue());
+            // Always wrap as Number; don't use `JsonWriter#value(long)`, which converts number
+            Number byteNumber = value instanceof Byte ? value : value.byteValue();
+            out.value(byteNumber);
           }
         }
       };
@@ -245,7 +247,9 @@ public final class TypeAdapters {
           if (value == null) {
             out.nullValue();
           } else {
-            out.value(value.shortValue());
+            // Always wrap as Number; don't use `JsonWriter#value(long)`, which converts number
+            Number shortNumber = value instanceof Short ? value : value.shortValue();
+            out.value(shortNumber);
           }
         }
       };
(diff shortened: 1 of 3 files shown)


Marcono1234:
Closed and reopend this PR to fix an intermittent test failure, see #2815.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -368,7 +374,10 @@ public Number read(JsonReader in) throws IOException {
         public void write(JsonWriter out, Number value) throws IOException {
           if (value == null) {
             out.nullValue();
+          } else if (value instanceof Long) {
+            out.value(value);
           } else {
+            // Only perform unwrapping if numeric conversion is necessary
             out.value(value.longValue());
```
Marcono1234, line 381: For correctness this `value instanceof Long` check is not needed; the previous code which always called `out.value(value.longValue())` was fine as well. However, for `JsonTreeWriter` it would have unwrapped and then wrapped the value again. But on the other hand for `JsonWriter` `value(long)` might be more efficient. I am not really sure which approach is better for performance, or if it even matters at all.

(same applies to `Double` below)

Issue #2816, opened 2025-02-24
Method fromJson throws JsonParseException instead of declared JsonSyntaxException


DGorokhov123:
# Gson version

I found a bug in GSON library. It's actual fo version 2.12.1

# Description

Bug description:

Method fromJson signature looks like that:
```
public <T> T fromJson(String json, Class<T> classOfT) throws JsonSyntaxException {
```

so, it's obvious to coders to catch JsonSyntaxException:
```
try {
    goodGson.fromJson(badJson, User.class);
} catch (JsonSyntaxException e) {
    System.out.println("Good deserializer caught JsonSyntaxException");
}
```
Such construction works until we add Deserializer:
```
Gson badGson = new GsonBuilder()
          .registerTypeAdapter(User.class, new UserBadDeserializer())
          .create();

class UserBadDeserializer implements JsonDeserializer<User> {
    @Override
    public User deserialize(JsonElement json, Type typeOfT, 
                         JsonDeserializationContext context) throws JsonParseException {
        JsonObject jsonObject = json.getAsJsonObject();
        if (!jsonObject.has("name")) 
                     throw new JsonParseException("json should contain the name field!");
        String name = jsonObject.get("name").getAsString();
        return new User(name);
    }
}
```
The signature of deserialize method we should override, throws another 
exception - JsonParseException, which is parent to JsonSyntaxException 
and isn't caught by try-catch construction above
It flies further and may cause unexpected behaviour. 

```
        try {
            try {
                badGson.fromJson(badJson, User.class);
            } catch (JsonSyntaxException e) {
                System.out.println("Good deserializer caught JsonSyntaxException");
            }
        } catch (RuntimeException e) {
            System.out.println("Bad deserializer didn't catch JsonSyntaxException, and " 
                        + e.getClass().getSimpleName() + " flew further");
        }
```
Solution 1 : Change signature of deserialize method to throw JsonSyntaxException.

Solution 2 : Change signature of fromJson method (and other methods calling
deserialize) to throw JsonParseException.

I've made a simple example to show this bug:
https://github.com/DGorokhov123/gson-bug

(С) Dmitriy Gorokhov dgorokhov123@gmail.com, tg:@DGorokhov

Shershah03:
Its seems changing it T _com.google.gson.JsonDeserializer.deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws **JsonSyntaxException_** is the best option.

Marcono1234:
[...]

PR #2820, opened 2025-03-05
[Issue-2816] - Update exception type in JsonDeserializer to JsonSyntaxException

Shershah03:
…xException

### Checklist


Shershah03:
#2820

Marcono1234:
Thanks for this pull request! I am however not sure if it might be better to instead change the signature of the `Gson#fromJson` methods to mention that they can throw a `JsonParseException`, see https://github.com/google/gson/issues/2816#issuecomment-2727065426.

PR #2824, opened 2025-03-16
Don't run GitHub workflows twice for Dependabot PRs

Marcono1234:
### Purpose
Don't run GitHub workflows twice for Dependabot PRs

### Description
Currently when Dependabot creates a pull request, some of the workflows run twice: One time for the `dependabot/...` branch within the repository and a second time for the PR update.

This is redundant and possibly also leads to confusion when the same workflow fails twice.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,12 @@
 name: Build
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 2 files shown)


eamonnmcmanus (approved):
Thanks!

PR #2825, opened 2025-03-16
Fix Native Image test execution

Marcono1234:
### Purpose
Fix Native Image test execution, see https://github.com/google/gson/pull/2818#pullrequestreview-2652489437

### Description
Looks like the "Build / GraalVM Native Image test" workflow failed because there was an incompatibility between the new JUnit version and the Native Maven Plugin. See also https://github.com/graalvm/native-build-tools/issues/706.


test-graal-native-image/pom.xml
```diff
@@ -118,7 +137,7 @@
           <plugin>
             <groupId>org.graalvm.buildtools</groupId>
             <artifactId>native-maven-plugin</artifactId>
-            <version>0.10.5</version>
+            <version>0.10.6</version>
```
Marcono1234, line 140: Increasing the version was not strictly necessary, and is not directly related to this issue.

eamonnmcmanus: LGTM

eamonnmcmanus:
Thanks!

PR #2829, opened 2025-03-23
[Issue-2816] - Update exception type in Gson#fromJson to JsonParseException

Shershah03:
…eption

### Purpose
Closes https://github.com/google/gson/issues/2816

### Description
As mentioned in https://github.com/google/gson/issues/2816 , If _com.google.gson.Gson.fromJson(String, Class<User>)_ throws JsonSyntaxException but in case any one imeplements custom JsonDeserializer then it by default throws JsonParseException.
For any com.google.gson.Gson.fromJson(String, Class<User>) calls , its normal to catch JsonSyntaxException, But in case use of JsonDeserializer, this exception will not be caught.
Ideally fromJson and JsonDeserializer interface should throw same exceptions.
**Updating exception type in Gson#fromJson to JsonParseException**

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1098,12 +1098,12 @@ public final class Gson {
    * @param classOfT the class of T
    * @return an object of type T from the string. Returns {@code null} if {@code json} is {@code
    *     null} or if {@code json} is empty.
-   * @throws JsonSyntaxException if json is not a valid representation for an object of type
+   * @throws JsonParseException if json is not a valid representation for an object of type
    *     classOfT
    * @see #fromJson(Reader, Class)
    * @see #fromJson(String, TypeToken)
    */
-  public <T> T fromJson(String json, Class<T> classOfT) throws JsonSyntaxException {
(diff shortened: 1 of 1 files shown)


Shershah03:
@Marcono1234 Modified the Exception as per suggestion

Marcono1234:
Thanks! I think it would be good to adjust all the other `Gson#fromJson` methods in the same way, since it affects all of them in the same way.

For some of them it would then list `JsonIOException` and `JsonParseException` (`throws JsonIOException, JsonParseException`). In theory that is redundant because `JsonIOException` is a subclass. But I think that is not an issue, and only makes it more explicit that users might want to handle `JsonIOException` separately.

----

But note that I am not a direct member of this project, so this still needs approval from a project member. So feel free to wait with further changes until further feedback from a project member.

eamonnmcmanus:
I agree with @Marcono1234 that we should make this documentation change everywhere it applies. I would go further and say that we should also _remove_ all unchecked exceptions from `throws` clauses. That's certainly the norm at Google, since declaring unchecked exceptions there has basically zero effect on compilation or execution. Of course we still do want `@throws` clauses in the javadoc, and if you can improve the accuracy of those then that it is a good service to our users.

kathleenmotley49-maker (approved):

PR #2835, opened 2025-04-03
Remove obsolete comment in pom.xml

Marcono1234:
Obsolete since the proguard-maven-plugin update to 2.7.0

See https://github.com/google/gson/pull/2831#discussion_r2026722684


eamonnmcmanus (approved):
Thanks!

PR #2841, opened 2025-04-08
Add multi-catch support to the code base

panic08:
Multi-catch appeared in java 7 and I think we can add it to the code base. It's a small fix, but I think it will kind of freshen up those code sections and make some code sections shorter


eamonnmcmanus:
Thanks, yes, that does make the code nicer.

Issue #2658, opened 2024-03-29
Edge case where R8 removes the TypeToken's generic type (while consuming the library's recently updated proguard rules)


alipov:
(Originally reported [here](https://github.com/google/gson/issues/2401#issuecomment-2009588140)).

# Gson version
2.10.1 + [those](https://github.com/google/gson/blob/main/gson/src/main/resources/META-INF/proguard/gson.pro) proguard rules.

# Java / Android version
I have tested on Android 34.

# Used tools
- [x] Gradle; version: 8.2
- [x] R8; version: 8.2.47

# Description
@Marcono1234 First of all thanks a lot for introducing consumer proguard rules!  

I have added the rules to one of my projects. Everything seemed to work fine, but I ran into one issue.  
Here's a sample code that makes use of Gson's TypeToken:  
```
private List<? extends MyInterface> getData() {
    String input = "[{\"value\": \"data\"}]";
    Type type = new TypeToken<List<MyImplementation>>() { }.getType();
    return new Gson().fromJson(input, type);
}
```
`MyImplementation` class implements the `MyInterface` interface, and this is its only usage in code.  

Upon minification, R8 removes the `MyImplementation` class completely, while `TypeToken` is being left with `java.lang.Object` as its generic argument:  
```
.class Linfo/osom/typetokenminify/a;
.super Lh/a;
.source "SourceFile"

# annotations
.annotation system Ldalvik/annotation/Signature;
    value = {
        "Lh/a<",
        "Ljava/util/List<",
        "Ljava/lang/Object;",
        ">;>;"
    }
.end annotation
```
This, in turn, causes a runtime exception:
```
Caused by: java.lang.ClassCastException
    at c.a.a(SourceFile:1)
    at info.osom.typetokenminify.MainActivity.onCreate(SourceFile:163)
```

As a workaround, I currently have to explicitly keep the class:  
```
-keep,allowobfuscation,allowoptimization class info.osom.typetokenminify.MyImplementation {
  @com.google.gson.annotations.SerializedName <fields>;
}
```
I'm wondering if more general rule can be used in such cases.

## Expected behavior
Run the application without errors.

## Actual behavior
The application crash at runtime.

# Reproduction steps
I have uploaded a [sample project](https://github.com/alipov/typetoken-minify) for demonstration.

Goooler:
I guess this should be fixed after updating to Gson 2.11 or above.

Marcono1234:
[...]

PR #2843, opened 2025-04-12
Remove outdated android-proguard-example

### LINKED ISSUE #2718: Remove `examples/android-proguard-example`
This repository currently contains an Android sample project in `examples/android-proguard-example`. The README and Troubleshooting Guide currently refer to it, and as seen by https://github.com/google/gson/discussions/2715#discussioncomment-10119789 users are still looking at it often.

The problem is that this sample project is quite old, and might not represent current best practices:
- It has no build config (no Gradle scripts), and the structure does not follow the `src/main/java`, ... conventions
- It uses a PNG `icon.png` which is probably not best practice; nowadays Android apps normally use vector drawables
- The `Cart#toString` implementation seems to [contain debug code](https://github.com/google/gson/blob/gson-parent-2.11.0/examples/android-proguard-example/src/com/google/gson/examples/android/model/Cart.java#L56-L106) which is not relevant to users
- Issues with `proguard.cfg`
  - The default file name nowadays seems to be `proguard-rules.pro`, see https://developer.android.com/build/shrink-code
  - It contains a `-keep class com.google.gson.examples.android.model.**` entry which is only relevant for that example
  - It contains a confusing `-keep class com.google.gson.stream.**`, see #1571
  - It is possibly incomplete and redundant nowadays since there is now `META-INF/proguard/gson.pro` (which users can also use as reference if they are using Gson < 2.11.0, which does not include that file)
- It contains the file `default.properties` which is probably redundant nowadays
- `AndroidManifest.xml` contains `package="..."`, which is ignored by recent Android Gradle Plugin versions, see https://developer.android.com/guide/topics/manifest/manifest-element#package

It also contains little configuration or code which is Android specific; the most important part is probably `proguard.cfg`. `GsonProguardExampleActivity.java` just performs basic conversion from and to JSON and displays the result as text; it does not use `SharedPreferences`, Room, or communicate with a REST API or anything similar.

Side note: After creating a new empty project in Android Studio, adding the files from the example and making some small adjustments the project still works though.

# Feature description
- Remove `examples/android-proguard-example`
- Optionally consider adding Android (or generally R8 / ProGuard) information to the User Guide instead
- Replace references to `examples/android-proguard-example`

# Alternatives / workarounds
An alternative would be to fix the issues above, or even set up a new sample Android project following the current best practices. However, the risk is that this becomes outdated again in the next years.

Alternatively could also just do nothing (or maybe only fix some of the issues), but the question is then how much value `examples/android-proguard-example` in its current form provides to users.

Goooler:
### Purpose

Closes #2718.

### Checklist


Troubleshooting.md
```diff
@@ -393,3 +396,11 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
 
 - This does not solve any of the type-safety problems mentioned above; in the long term you should prefer one of the other solutions listed above. This system property might be removed in future Gson versions.
 - You should only ever set the property to `"true"`, but never to any other value or manually clear it. Otherwise this might counteract any libraries you are using which might have deliberately set the system property because they rely on its behavior.
+
+## R8 / ProGuard
```
Goooler, line 400: I didn't apply `## <a id="r8-proguard"></a> R8 / ProGuard` for this header like the others, as the index of it (`#a-idr8-proguarda-r8--proguard`) can't be honored by GitHub web correctly.

Marcono1234: Is that specific to this header? Because for the others GitHub UI seems to ignore the `<a>` when deriving the default anchor name (or it only seems to prepend a `-`).

If possible I think a stable anchor with `<a>` would be better.

README.md
```diff
@@ -68,6 +68,10 @@ see [`GsonBuilder.disableJdkUnsafe()`](https://javadoc.io/doc/com.google.code.gs
 
 Older Gson versions may also support lower API levels, however this has not been verified.
 
+### R8 / ProGuard
+
+See the details in the related section in [Troubleshooting guide](Troubleshooting.md#r8--proguard).
```
Marcono1234, line 73:
```suggestion
See the details in the related section in the [Troubleshooting guide](Troubleshooting.md#r8--proguard).
```

Maybe it would also be better to remove this below the "Documentation" section, since it is more specific?

Troubleshooting.md
```diff
@@ -22,7 +22,8 @@ This guide describes how to troubleshoot common issues when using Gson.
...
 
-If you are using a code shrinking tool such as ProGuard or R8 (for example when building an Android app), make sure it is correctly configured to keep generic signatures and to keep Gson's `TypeToken` class. See the [Android example](examples/android-proguard-example/README.md) for more information.
+If you are using a code shrinking tool such as Proguard / R8 (for example when building an Android app), make sure it is correctly configured to keep generic signatures and to keep Gson's `TypeToken` class.
+See the [Proguard / R8](#r8--proguard) section for more information.
```
Marcono1234, line 26:
```suggestion
If you are using a code shrinking tool such as ProGuard / R8 (for example when building an Android app), make sure it is correctly configured to keep generic signatures and to keep Gson's `TypeToken` class.
See the [ProGuard / R8](#r8--proguard) section for more information.
```

Troubleshooting.md
```diff
@@ -67,15 +68,17 @@ Or in case this occurs for a field in one of your classes which you did not actu
 
 **Reason:** You probably have not configured ProGuard / R8 correctly
 
-**Solution:** Make sure you have configured ProGuard / R8 correctly to preserve the names of your fields. See the [Android example](examples/android-proguard-example/README.md) for more information.
+**Solution:** Make sure you have configured ProGuard / R8 correctly to preserve the names of your fields.
+See the section below, it's related to this issue.
```
Marcono1234, line 72: Directly link to the section as well?
```suggestion
See the [ProGuard / R8](#r8--proguard) section for more information.
```

Troubleshooting.md
```diff
@@ -67,15 +68,17 @@ Or in case this occurs for a field in one of your classes which you did not actu
...
 
 ## <a id="android-app-broken-after-app-update"></a> Android app unable to parse JSON after app update
 
 **Symptom:** You released a new version of your Android app and it fails to parse JSON data created by the previous version of your app
 
 **Reason:** You probably have not configured ProGuard / R8 correctly; probably the field names are being obfuscated and their naming changed between the versions of your app
 
-**Solution:** Make sure you have configured ProGuard / R8 correctly to preserve the names of your fields. See the [Android example](examples/android-proguard-example/README.md) for more information.
+**Solution:** Make sure you have configured ProGuard / R8 correctly to preserve the names of your fields.
+See the [Proguard / R8](#r8--proguard) section for more information.
```
Marcono1234, line 81:
```suggestion
See the [ProGuard / R8](#r8--proguard) section for more information.
```

Troubleshooting.md
```diff
@@ -346,7 +349,7 @@ For older Gson versions a `RuntimeException` with message 'Missing type paramete
 -keep class * extends com.google.gson.reflect.TypeToken
 ```
 
-See also the [Android example](examples/android-proguard-example/README.md) for more information.
+See the [Proguard / R8](#r8--proguard) section for more information.
```
Marcono1234, line 352:
```suggestion
See the [ProGuard / R8](#r8--proguard) section for more information.
```

Troubleshooting.md
```diff
@@ -393,3 +396,11 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
 
 - This does not solve any of the type-safety problems mentioned above; in the long term you should prefer one of the other solutions listed above. This system property might be removed in future Gson versions.
 - You should only ever set the property to `"true"`, but never to any other value or manually clear it. Otherwise this might counteract any libraries you are using which might have deliberately set the system property because they rely on its behavior.
+
+## R8 / ProGuard
+
+If you use Gson as a dependency in an Android project which uses R8 as a default compiler you don’t have to do anything.
+The specific rules are [already bundled](gson/src/main/resources/META-INF/proguard/gson.pro) (from Gson 2.10.1) into
```
Marcono1234, line 403: Wrong version number? See https://github.com/google/gson/releases/tag/gson-parent-2.11.0
```suggestion
The specific rules are [already bundled](gson/src/main/resources/META-INF/proguard/gson.pro) (from Gson 2.11.0) into
```

Troubleshooting.md
```diff
@@ -393,3 +396,11 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
 
 - This does not solve any of the type-safety problems mentioned above; in the long term you should prefer one of the other solutions listed above. This system property might be removed in future Gson versions.
 - You should only ever set the property to `"true"`, but never to any other value or manually clear it. Otherwise this might counteract any libraries you are using which might have deliberately set the system property because they rely on its behavior.
+
+## R8 / ProGuard
+
+If you use Gson as a dependency in an Android project which uses R8 as a default compiler you don’t have to do anything.
```
Marcono1234, line 402: Minor: Maybe use a regular `'` for consistency with the remainder of this document.
```suggestion
If you use Gson as a dependency in an Android project which uses R8 as a default compiler you don't have to do anything.
```

Marcono1234, line 402:
```suggestion
If you use Gson as a dependency in an Android project which uses R8 or ProGuard as code shrinking and obfuscation tool you don’t have to do anything.
```

Troubleshooting.md
```diff
@@ -393,3 +396,11 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
...
+The specific rules are [already bundled](gson/src/main/resources/META-INF/proguard/gson.pro) (from Gson 2.10.1) into
+the JAR which can be interpreted by R8 automatically. For users who still use the older Gson versions, you may need to
+copy the rules from the [`gson.pro`](gson/src/main/resources/META-INF/proguard/gson.pro) into your own ProGuard
+configuration file.
```
Marcono1234, line 406:
```suggestion
the JAR which can be interpreted by R8 automatically. For users who still use older Gson versions, you may need to
copy the rules from the [`gson.pro`](gson/src/main/resources/META-INF/proguard/gson.pro) file into your own ProGuard
configuration file.
```

Marcono1234 (commented):
Thanks a lot for this! I have added some small comments.

Maybe the R8 / ProGuard section needs to be expanded:
- it should only say no "additional ProGuard configuration is needed"
  - because the Java code of the users needs to adhere to a specific pattern (see below)
  - and only "in most cases", there are some corner cases ([example](https://github.com/google/gson/issues/2658))
- it should document requires for user classes
  - classes must have a no-args constructor, and must be either top-level or `static` (to avoid additional implicit constructor args)
  - fields must be annotated with `@SerializedName`
- maybe it should mention that ProGuard for non-Android projects has to configure the `gson.pro` file manually, see https://github.com/Guardsquare/proguard/issues/337

If you want I can do that in a follow-up pull request instead though.

Troubleshooting.md
Goooler: Seems all existing ones are invalid.

Troubleshooting.md
Goooler: This is intended. As I see it, the "random property names" and "field names are being obfuscated" are the same issue; I just wanted to link them together.

README.md
Goooler: I document the R8 / ProGuard section explicitly as I want to warn users like https://github.com/square/retrofit?tab=readme-ov-file#r8--proguard and OkHttp did.

Troubleshooting.md
Goooler: View https://github.com/Goooler/gson/blob/e8f52fb490dc7b58d95c01447701b986e655a214/Troubleshooting.md#a-idproguard-r8a-proguard--r8.

The ref id works in IDEA's markdown render but GitHub website.

Marcono1234: That rather looks like an issue with IntelliJ. In GitHub both anchors work:
- GitHub derived one (also used for the "🔗" icon next to the title):
[`#-proguard--r8`](https://github.com/Goooler/gson/blob/e8f52fb490dc7b58d95c01447701b986e655a214/Troubleshooting.md#-proguard--r8)
- The one assigned with `<a>` (which should be referred to within the guide)
[`#proguard-r8`](https://github.com/Goooler/gson/blob/e8f52fb490dc7b58d95c01447701b986e655a214/Troubleshooting.md#proguard-r8)

(though maybe for GitHub this also relies on implementation details to some extent)

README.md
Marcono1234: Yes that is fine and a good thing. I was just wondering whether it would be better to have the (existing) "Documentation" section first in the README:
> ...
>
> ## Documentation
> - API Javadoc: Documentation for the current release
> - ...
>
> ## ProGuard / R8
> See the details in the related ...

Because this is a more specific aspect of the documentation.

Goooler: Reasonable! 495aa1b19e0f6526eaf323d23232361989961293

Troubleshooting.md
Goooler: We probably should merge these 2 sections.

Troubleshooting.md
Goooler: I prefer to keep the header without id, it works for both IDEA and GitHub. But I can update the style if you want.

Goooler:
> it should only say "no additional ProGuard configuration is needed"

- The current desc is copy-pasted from [OkHttps](https://github.com/square/okhttp/blob/4984568367caaf359b82c452bd28b5e192824d1c/docs/r8_proguard.md?plain=1#L4-L9), which is good enough from my side.
- #2658 looks like happening under Gson 2.10.1, I guess it has been fixed by https://github.com/google/gson/blob/4e65e6ab368d92638b4ca04521958dbf1d3753e7/gson/src/main/resources/META-INF/proguard/gson.pro#L63-L72
which has already been bundled in 2.11.0

> it should document requirements for user classes

This point looks unrelated to this PR.

> fields must be annotated with @SerializedName

Yeah, I see this point has been noted times like 

https://github.com/google/gson/blob/a656395e3ac3e07c7740a5dfcb8046a7577de7a2/Troubleshooting.md#L83

https://github.com/google/gson/blob/a656395e3ac3e07c7740a5dfcb8046a7577de7a2/Troubleshooting.md#L282

Actually, users can keep full the classes as instead, it's not **fields must be**.

> https://github.com/Guardsquare/proguard/issues/337

Good to know this! We can note it. And may invert all `ProGuard / R8` stuff to `R8 / ProGuard` to prefer the R8 side.

Troubleshooting.md
Marcono1234: I would keep them separate for now. The symptoms are different (but similar) and if I recall correctly users had experienced and asked about both problems already in the past.

Marcono1234 (approved):
Thanks for the changes, looks good to me. I might do some small follow-up changes, but thanks for taking up this task in the first place!

eamonnmcmanus (approved):
Thanks for doing this!

Troubleshooting.md
Marcono1234: IntelliJ (IntelliJ IDEA 2024.3.5) seems to be a bit inconsistent. Code completion for Markdown links suggests these `#a-idr8-proguarda-r8--proguard` anchors you mentioned. However, for the Markdown preview the custom `<a>` anchor links seem to work fine within the same document. But they do not work between different files, e.g. the link from the README to the Troubleshooting guide section.

Goooler: Yeah, the same behavior as IDEA 2025.1

PR #2844, opened 2025-04-14
Adjust Troubleshooting Guide ProGuard / R8 section

Marcono1234:
Follow-up for #2843


Troubleshooting.md
```diff
@@ -397,10 +398,18 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
 - This does not solve any of the type-safety problems mentioned above; in the long term you should prefer one of the other solutions listed above. This system property might be removed in future Gson versions.
 - You should only ever set the property to `"true"`, but never to any other value or manually clear it. Otherwise this might counteract any libraries you are using which might have deliberately set the system property because they rely on its behavior.
 
-## ProGuard / R8
+## <a id="proguard-r8"></a> ProGuard / R8
```
Marcono1234, line 401: Used a custom anchor here for consistency with the other ones. If we really want to change something about these anchors, then it would be better to do it for all and not just for a single one.

Within GitHub these custom anchors seem to work fine, and also within IntelliJ they work to some extent. See https://github.com/google/gson/pull/2843#discussion_r2042917087.

eamonnmcmanus (approved):

PR #2845, opened 2025-04-15
Add shrinker test for interface implementation being deserialized

Marcono1234:
Adds a shrinker test which tries to replicate the issue described in #2658.
@alipov, could you please check whether this test, specifically `InterfaceWithImplementation` and the usage in `testDeserializingInterfaceImpl` represents the setup you are using?

For R8 this currently fails, see TODOs in the new code. Not sure if there is something Gson can do at the moment, without making `gson.pro` unconditionally keep classes with `@SerializedName` regardless of whether they are used (which is probably something we want to avoid?). Also note that this might be an uncommon corner case.
Maybe this is actually an R8 bug or limitation.

### DIFF
--- a/test-shrinker/src/main/java/com/example/InterfaceWithImplementation.java
+++ b/test-shrinker/src/main/java/com/example/InterfaceWithImplementation.java
@@ -1,0 +1,21 @@
+package com.example;
+
+import com.google.gson.annotations.SerializedName;
+
+/** Interface whose implementation class is only referenced for deserialization */
+public interface InterfaceWithImplementation {
+  String getValue();
+
+  // Implementation class which is only referenced in `TypeToken`, but nowhere else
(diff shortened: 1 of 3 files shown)


alipov:
@Marcono1234 Yep, that setup matches my usage.  
Would you like me to open an R8 issue, perhaps @sgjesse might have an idea for a rule that handles this specific case?

Marcono1234:
> Would you like me to open an R8 issue

Yes that would be great, thanks! Contact details are here: https://r8.googlesource.com/r8#getting-help (in case you didn't know that already).
Please link to the issue you created in your GitHub issue #2658 then so that others can follow it.

Marcono1234:
@eamonnmcmanus, do you think we can merge this here already to have a test case for it? (even if it currently has special handling for the R8 failure)

PR #2853, opened 2025-04-30
Run CodeQL for GitHub Actions & improve workflows

Marcono1234:
CodeQL support for GitHub Actions is now generally available, see [changelog](https://github.blog/changelog/2025-04-22-github-actions-workflow-security-analysis-with-codeql-is-now-generally-available/).

This pull request enables scanning of GitHub Actions and improves some of the other workflows. The CIFuzz workflow has been adjusted based on
https://github.com/google/oss-fuzz/blob/cafd7a0eb8ecb4e007c56897996a9b65c49c972f/docs/getting-started/continuous_integration.md, to also upload SARIF results (I don't know exactly how the results, if any, will be shown in the GitHub UI though).

**Notes:**
- The CodeQL workflow still uses `queries: +security-and-quality`, which runs additional queries which are not run by default (see [documentation](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#using-queries-in-ql-packs))
If that causes too many irrelevant alerts, we could consider changing it (either for `actions` only, or also for `java`).
- An alternative to this CodeQL setup with a dedicated workflow would be [default setup](https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning)
In that case the `codeql-analysis.yml` workflow has to be removed and CodeQL scanning has to be enabled in the repository settings.
However, 'default setup' seems to use [build mode 'none'](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages#about-build-mode-none-for-codeql) for Java, and that might not be as accurate, for example because we have the `GsonBuildConfig.java` class which is processed during build.
So maybe for now keeping the 'advanced setup' (with explicit workflow) is better?


eamonnmcmanus (approved):
Thanks! We'll see how this works out...

Marcono1234:
Thanks!

It seems my change to include `category: "/language:${{ matrix.language }}"`[^1] changed the name of the configuration, so GitHub is now reporting a warning. For example for https://github.com/google/gson/pull/2876/checks?check_run_id=46137593013:
> ## 1 configuration not found
>
> **Warning:** Code scanning cannot determine the alerts introduced by this pull request, because 1 configuration present on `refs/heads/main` was not found:
> ### Actions workflow (`codeql-analysis.yml`)
>
>    ❓  `.github/workflows/codeql-analysis.yml:analyze/language:java`


Apparently the configuration name was previously implicitly generated as `.github/workflows/codeql-analysis.yml:analyze/language:java` but now due to the `category: "/language:${{ matrix.language }}"` uses the explicit name `/language:java`[^2].
I hope this did not cause previous alerts which you had suppressed to reappear.

Maybe the warning about the not found configuration can be safely ignored. But it can possibly also be solved by [deleting the configuration](https://docs.github.com/en/code-security/code-scanning/managing-your-code-scanning-configuration/about-the-tool-status-page#removing-configurations) (the top of that page describes how to get to the list of configurations).
There should probably be two configurations shown as `language:java`, but if you click on them the full name is shown. The configuration `.github/workflows/codeql-analysis.yml:analyze/language:java ` can probably be deleted.

Sorry for the inconveniences.

[^1]: I had based this change on what the GitHub UI is suggesting as workflow file when newly setting up CodeQL with the advanced setup.
[^2]: Maybe the explicit name has the advantage though that it is independent of the workflow file name and the job name, so it won't cause similar "not found configuration" warnings in case we ever rename the workflow or the job.

eamonnmcmanus:
OK, I've deleted the old configuration. Thanks for the detailed explanation!

PR #2856, opened 2025-05-14
Add warning about usage of Gson on Android

ChrisCraik:
### Purpose
Goal of this PR is to clarify that Gson should generally not be used for Android projects since it will often make optimization harder, especially now with [R8 full mode](https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md#r8-full-mode) being the default.

### Description
When trying to encourage Android developers to optimize their apps, Gson is one of the primary causes of crashes due to conflicts between the open ended reflection of Gson and the efforts of optimizations from R8 (or Proguard before it).

### Checklist
Not applicable, markdown-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 > When using languages other than Java, prefer a JSON library with explicit support for that language.
 
 > [!IMPORTANT]\
-> GSON is not a recommended library for interacting with JSON on Android. The open ended reflection in the GSON runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
-> If your app or library may be running on Android, instead look at [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) which uses code generation instead of reflection. This avoids GSON's runtime crashes when optimizations are applied, and results in faster performance on Android devices. If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
+> Gson is not a recommended library for interacting with JSON on Android. The open ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
+> If your app or library may be running on Android, instead look at [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) which uses code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied, and results in faster performance on Android devices. If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
 
 ### Goals
   * Provide simple `toJson()` and `fromJson()` methods to convert Java objects to JSON and vice-versa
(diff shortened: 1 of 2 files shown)


Troubleshooting.md
```diff
@@ -398,18 +398,15 @@ For backward compatibility it is possible to restore Gson's old behavior of allo
...
-However, this might completely disable some optimizations and obfuscation for these classes.
+### Constrain reflected classes
+- Ensure all of your objects have a no-arg constructor (and be a top-level or static class to avoid implicit constructor arguments)
+- Annotate all fields with `@SerializedName()`
 
-If you still use a Gson version older than 2.11.0 or if you are using ProGuard for a non-Android project ([related ProGuard issue](https://github.com/Guardsquare/proguard/issues/337)),
-you may need to copy the rules from the [`gson.pro`](gson/src/main/resources/META-INF/proguard/gson.pro) file into your own ProGuard configuration file.
+### Avoid Reflection
+Use [addReflectionAccessFilter()](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/GsonBuilder.html#addReflectionAccessFilter(com.google.gson.ReflectionAccessFilter)) to add a filter which always returns `BLOCK_ALL` to block all reflection into custom classes. This will mean you need to use explicit objects such as JsonObject and JsonArray.
```
eamonnmcmanus, line 412: I know it was my suggestion that led to the last sentence here, but I think what we really want to say instead is this:

> This will mean you will need to have a `TypeAdapter` or `TypeAdapterFactory` for every type you might want to serialize or deserialize, or that you are only using Gson through its explicit JSON API via classes like `JsonObject` and `JsonArray`.

eamonnmcmanus (changes requested):
Thanks! I think this is very worthwhile. Just one small thing...

Troubleshooting.md
ChrisCraik: I tried to incorporate this - I still wanted to reference the `BLOCK_ALL` step since it sounds as if it would give you the most safety, but let me know if it needs more clarification.

eamonnmcmanus (approved):
That looks great, better than the wording I proposed!

README.md
```diff
@@ -12,6 +12,10 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 > Gson's main focus is on Java. Using it with other JVM languages such as Kotlin or Scala might work fine in many cases, but language-specific features such as Kotlin's non-`null` types or constructors with default arguments are not supported. This can lead to confusing and incorrect behavior.\
 > When using languages other than Java, prefer a JSON library with explicit support for that language.
 
+> [!IMPORTANT]\
+> Gson is not a recommended library for interacting with JSON on Android. The open ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
+> If your app or library may be running on Android, instead look at [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) which uses code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied, and results in faster performance on Android devices. If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
```
Goooler, line 17: Seems we can also mention Moshi here? 

Instead look at [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi](https://github.com/square/moshi), that use code generation instead of reflection.

Moshi could handle Java classes.

Goooler: Addressing it to #2857.

PR #2857, opened 2025-05-18
Document using Kotlin Serialization or Moshi instead

Goooler:
### Description

Refs https://github.com/google/gson/pull/2856#discussion_r2092209522.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 > Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
 > If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
 > which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or being obfuscated), and results in faster performance on Android devices.
-> The Moshi APIs may be more familiar to users who already know Gson, making it more seamless for users to migrate between the two.
+> The Moshi APIs may be more familiar to users who already know Gson.
 > If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#proguard-r8).
 
 ### Goals
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 > Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
 > If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
 > which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or obfuscating), and results in faster performance on Android devices.
-> If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
+> If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#proguard-r8).
 
 ### Goals
   * Provide simple `toJson()` and `fromJson()` methods to convert Java objects to JSON and vice-versa
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 > [!IMPORTANT]\
 > Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
 > If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
...
-> If your app or library may be running on Android, instead look at [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) which uses code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied, and results in faster performance on Android devices. If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
+> If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi](https://github.com/square/moshi), which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied, and results in faster performance on Android devices. If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
```
eamonnmcmanus, line 17: I don't think we can recommend Moshi unreservedly. By default it uses reflection, which is just as problematic on Android as with Gson, for the same reasons. So at least I think the text would have to say "Moshi codegen" and link to the [codegen section](https://github.com/square/moshi?tab=readme-ov-file#codegen) of the documentation.

I'd also like to have something to help readers decide which solution is more appropriate for them. What would lead someone to choose Moshi over `kotlinx.serialization`?

eamonnmcmanus (changes requested):

README.md
Goooler: https://github.com/square/moshi?tab=readme-ov-file#codegen is only for KSP for Kotlin, I didn't paste it as Moshi also supports APT for Java.

eamonnmcmanus: Could you write something like this?

> consider a solution that uses code generation instead of reflection, such as [Kotlin Serialization](...) or Moshi with codegen (for [Java](...) or [Kotlin](...)).

I don't see where Java codegen is documented on the Moshi site, though I do see [this separate project](https://github.com/nename0/moshi-java-codegen). Is that what you're referring to? If we're talking about how people can assemble other projects in order to avoid reflection, we might also reference [`auto-value-gson`](https://github.com/rharter/auto-value-gson). That could be a separate change, though.

I still think that we would need to tell people why they might consider using Moshi codegen with Kotlin instead of the solution that is built in to the Kotlin platform.

Goooler: Ooops, I misremembered that...

Moshi's official code generation has only been available for Kotlin from the start, initially via KAPT (which is based on APT). I used it in my projects through KAPT instead of APT.

Goooler: Updated the doc!

README.md
```diff
@@ -13,8 +13,10 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
...
+> Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
+> If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
+> which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or obfuscating), and results in faster performance on Android devices.
+> If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#-proguard--r8).
```
Marcono1234, line 19: The anchor is incorrect. This was already a problem in the previous pull request though.
```suggestion
> If you still want to use Gson and attempt to avoid these crashes, you can see how to do so [here](Troubleshooting.md#proguard-r8).
```

README.md
```diff
@@ -13,8 +13,10 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
...
+> Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
+> If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
+> which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or obfuscating), and results in faster performance on Android devices.
```
Marcono1234, line 18: I think this should probably say "obfuscation" or "being obfuscated"?
```suggestion
> which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or being obfuscated), and results in faster performance on Android devices.
```

Goooler: Either one should be right 😄

eamonnmcmanus (changes requested):
I am still reluctant to make this change unless we can say why someone would want to use Moshi-codegen with Kotlin instead of the solution that is proposed and supported by JetBrains. I can believe there would be reasons, but can we say what they are?

Goooler:
Moshi's APIs share [similarities with Gson](https://github.com/square/moshi?tab=readme-ov-file#borrows-from-gson), making it easier and more seamless for users to migrate between the two.

Goooler:
Done.

README.md
```diff
@@ -13,8 +13,11 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
...
+> Gson is not a recommended library for interacting with JSON on Android. The open-ended reflection in the Gson runtime doesn't play nicely with shrinking/optimization/obfuscation passes that Android release apps should perform.\
+> If your app or library may be running on Android, consider using [Kotlin Serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#basics) or [Moshi's Codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen),
+> which use code generation instead of reflection. This avoids Gson's runtime crashes when optimizations are applied (usually due to the fields missing or being obfuscated), and results in faster performance on Android devices.
+> The Moshi APIs may be more familiar to users who already know Gson, making it more seamless for users to migrate between the two.
```
eamonnmcmanus, line 19: Thanks! But I don't think the "making it more seamless..." clause is necessary.

Goooler, line 19:
```suggestion
> The Moshi APIs may be more familiar to users who already know Gson.
```

eamonnmcmanus (approved):

PR #2858, opened 2025-05-20
Tweak R8 / ProGuard Troubleshooting Guide entry

Marcono1234:
Follow-up for #2856

- Added back the mention of Gson's `gson.pro` file for users of older Gson versions, and for non-Android ProGuard users
- Mentioned manual JSON data processing using `JsonReader` and `JsonWriter` as safe
(not completely sure if mentioning this is worth it though)


eamonnmcmanus (approved):
Thanks, very helpful!

PR #2864, opened 2025-06-23
add Gson.DEFAULT

MukjepScarlet:
*This is a draft PR so the deprecated usages and documentations are not updated yet*


kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

eamonnmcmanus:
I'm not really convinced here. It might have been better if this had been the API from the outset, but now we have 17 years of client code that uses `new Gson()`. So I think it would be very disruptive to deprecate that, and if we don't then we have two different ways of doing the same thing.

However, it's true that every invocation of `new Gson()` ends up doing [quite a lot of work](https://github.com/google/gson/blob/72537db9a6095720aca3a67abe77ec5a16200081/gson/src/main/java/com/google/gson/Gson.java#L322-L397) in order to construct an exactly equivalent list of factories every time. I believe we could improve that situation as follows:

* Replace the package-private `Gson` constructor that has 21 parameters by a constructor that takes a `GsonBuilder` parameter.
* Have a `static final GsonBuilder` that the public `Gson()` constructor passes to the new `Gson(GsonBuilder)`.
* Somehow cache the `List<TypeAdapterFactory>` corresponding to this builder, and reuse it when the new constructor sees that it is being called a second time from the `Gson()` constructor.

The last item is potentially controversial, since it does mean that some memory would end up being retained even when there are no active `Gson` instances.

Whether or not we do the last item, I think the first two would be worthwhile, if only because it would eliminate the duplication whereby the default value for every Gson property is specified in both the `Gson()` constructor and the `GsonBuilder` instance initializers.

MukjepScarlet:
I will drop `Gson.DEFAULT` and try to make the Default constructor share more contents.

MukjepScarlet:
I need help about Java 9 modules...

Marcono1234:
> I need help about Java 9 modules...

It looks like there is something wrong with how the adapters are looked up. It is trying to use the reflective adapter for types such as `java.lang.Double`.

I have made some changes to this code here and proposed them as https://github.com/MukjepScarlet/gson/pull/1; that also fixes the test failures. If you are fine with them, you can merge them and they will directly appear in this PR here then.

But @MukjepScarlet, could you please edit the title of this PR here (click the "Edit" button at the top right) and rename it to something like "Reduce overhead of default `Gson` instances" or similar? The current title ("add Gson.DEFAULT") is misleading because the PR is not actually doing this anymore.

Marcono1234:
@eamonnmcmanus, what do you think about the current state of this PR?

eamonnmcmanus:
> @eamonnmcmanus, what do you think about the current state of this PR?

The outline looks correct, but I think there is a subtle change in behaviour. When I ran this against Google's internal tests, I saw failures that looked like this:
```
Error occurred during initialization of VM
java.lang.BootstrapMethodError: java.lang.IllegalStateException: getSystemClassLoader cannot be called during the system class loader instantiation
        at com.google.gson.internal.ConstructorConstructor.newMapConstructor(ConstructorConstructor.java:381)
        at com.google.gson.internal.ConstructorConstructor.newDefaultImplementationConstructor(ConstructorConstructor.java:328)
        at com.google.gson.internal.ConstructorConstructor.get(ConstructorConstructor.java:137)
        at com.google.gson.internal.bind.MapTypeAdapterFactory.create(MapTypeAdapterFactory.java:146)
        at com.google.gson.Gson.getAdapter(Gson.java:349)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:205)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:396)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:165)
        at com.google.gson.Gson.getAdapter(Gson.java:349)
        at com.google.gson.Gson.fromJson(Gson.java:1079)
        at com.google.gson.Gson.fromJson(Gson.java:981)
        at com.google.gson.Gson.fromJson(Gson.java:891)
        at com.google.gson.Gson.fromJson(Gson.java:828)
        at [... Google-internal stuff ...]
        at java.lang.reflect.Constructor.newInstanceWithCaller(java.base@25/Constructor.java:499]
        at java.lang.reflect.Constructor.newInstance(java.base@25/Constructor.java:483]
        at java.lang.ClassLoader.initSystemClassLoader(java.base@25/ClassLoader.java:1889]
        at java.lang.System.initPhase3(java.base@25/System.java:1993]
Caused by: java.lang.IllegalStateException: getSystemClassLoader cannot be called during the system class loader instantiation
        at java.lang.ClassLoader.getSystemClassLoader(java.base@25/ClassLoader.java:1849]
        at java.lang.invoke.MethodHandles$Lookup.canBeCached(java.base@25/MethodHandles.java:4022]
        at java.lang.invoke.MethodHandles$Lookup.linkMethodHandleConstant(java.base@25/MethodHandles.java:3999]
        at java.lang.invoke.MethodHandleNatives.linkMethodHandleConstant(java.base@25/MethodHandleNatives.java:603]
        at com.google.gson.internal.ConstructorConstructor.newMapConstructor(ConstructorConstructor.java:381)
        at com.google.gson.internal.ConstructorConstructor.newDefaultImplementationConstructor(ConstructorConstructor.java:328)
        at com.google.gson.internal.ConstructorConstructor.get(ConstructorConstructor.java:137)
        at com.google.gson.internal.bind.MapTypeAdapterFactory.create(MapTypeAdapterFactory.java:146)
        at com.google.gson.Gson.getAdapter(Gson.java:349)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:205)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:396)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:165)
        at com.google.gson.Gson.getAdapter(Gson.java:349)
        at com.google.gson.Gson.fromJson(Gson.java:1079)
        at com.google.gson.Gson.fromJson(Gson.java:981)
        at com.google.gson.Gson.fromJson(Gson.java:891)
        at com.google.gson.Gson.fromJson(Gson.java:828)
        at [... more Google-internal stuff ...]
```

I guess this code is running with `-Djava.system.class.loader=something`, but in any case it is clearly triggering a code path that was not triggered before. I haven't had time to investigate further.

Marcono1234:
This PR does change `() -> new C()` to `C::new` in `ConstructorConstructor.java` (which is not actually needed for this PR).

If you revert those `ConstructorConstructor.java` changes, do the tests pass then? Would be quite interesting though that lambda vs. method reference makes a difference.

MukjepScarlet:
I'm interesting about that too. If you think the change of `ConstructorConstructor` is not needed (or should be in another PR), I'll revert it here.

MukjepScarlet:
I think this is the reason because stuffs like `TreeMap::new` will access `getSystemClassLoader`, but in the test it's not ready. This might need further test.

Marcono1234:
I think I am able to reproduce this locally:
1. For testing (but don't include it in this PR), add `gson/src/main/java/com/google/gson/SystemClassLoader.java` with this content:
    ```java
    package com.google.gson;
    
    import java.util.Map;
    import java.util.concurrent.atomic.AtomicReference;
    
    public class SystemClassLoader extends ClassLoader {
        private static final AtomicReference<Object> value = new AtomicReference<>();
    
        public SystemClassLoader(ClassLoader parent) {
            super(parent);
            Object object = new Gson().fromJson("{}", Map.class);
            value.set(object);
        }
    
        @SuppressWarnings("SystemOut")
            System.out.println(value.get());
            System.out.println("done");
        }
    }
    ```
2. Build the Gson JAR
   ```
   mvn clean package --projects gson "-Dmaven.test.skip"
   ```
3. Run with the custom system class loader
   ```
   java -cp ./gson/target/gson-2.13.3-SNAPSHOT.jar "-Djava.system.class.loader=com.google.gson.SystemClassLoader" com.google.gson.SystemClassLoader
   ```

Fails with the same stack trace as above. Replacing the method references in `ConstructorConstructor` fixes it.
Though to be sure, would be good @eamonnmcmanus if you could double-check if that is indeed the same issue.

In that case maybe we could indeed replace the method references in `ConstructorConstructor` with lambdas again. Though this seems somewhat brittle and I am wondering if there are other not-yet discovered issues which could occur when using Gson from within a custom system class loader.

Interestingly this only seems to affect `LinkedTreeMap::new`, maybe because it is a reference to a non-JDK class?
You can reproduce this with this custom class loader (and same build / run steps as above):
```java
public class SystemClassLoader extends ClassLoader {
    public SystemClassLoader(ClassLoader parent) {
        super(parent);
        List<Supplier<?>> unused = Arrays.asList(
            ArrayList::new,
            LinkedHashSet::new,
            TreeSet::new,
            ArrayDeque::new,
            // LinkedTreeMap::new,
            () -> new LinkedTreeMap<>(),
            LinkedHashMap::new,
            TreeMap::new,
            ConcurrentHashMap::new,
            ConcurrentSkipListMap::new
        );
    }

    @SuppressWarnings("SystemOut")
        System.out.println("done");
    }
}
```

Edit: It seems the reason why this issue only occurs for ...
- Gson's `LinkedTreeMap`: the `MethodHandles` code only checks the system class loader [if the class loader of the class is non-`null`](https://github.com/openjdk/jdk/blob/6c48f4ed707bf0b15f9b6098de30db8aae6fa40f/src/java.base/share/classes/java/lang/invoke/MethodHandles.java#L4020-L4021); for JDK classes it is `null` because they are loaded by the bootstrap class loader
- a method reference (`::new`) but not a lambda: for the lambda the compiler seems to generate a private snythetic method, and the method handle then refers to that, but the [caching logic in `MethodHandles` ignores non-`public` members](https://github.com/openjdk/jdk/blob/6c48f4ed707bf0b15f9b6098de30db8aae6fa40f/src/java.base/share/classes/java/lang/invoke/MethodHandles.java#L4016)

At least that is my understanding of it. But this also seems to highly depend on JDK and compiler implementation details. But given that usage of lambdas here worked before apparently, that seems to be a reasonable solution, albeit brittle. Though using Gson in general from within the constructor of a custom system class loader seems to be brittle.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -370,29 +370,29 @@ private static boolean hasStringKeyType(Type mapType) {
...
+  private static ObjectConstructor<? extends Map<?, Object>> newMapConstructor(
       Type type, Class<?> rawType) {
     // First try Map implementation
     /*
      * Legacy special casing for Map<String, ...> to avoid DoS from colliding String hashCode
      * values for older JDKs; use own LinkedTreeMap<String, Object> instead
      */
     if (rawType.isAssignableFrom(LinkedTreeMap.class) && hasStringKeyType(type)) {
-      return () -> new LinkedTreeMap<>();
+      return LinkedTreeMap::new;
```
Marcono1234:
```suggestion
      // Must use lambda instead of method reference (`LinkedTreeMap::new`) here, otherwise this
      // causes an exception when Gson is used by a custom system class loader, see
      // https://github.com/google/gson/pull/2864#issuecomment-3528623716
      return () -> new LinkedTreeMap<>();
```

eamonnmcmanus (changes requested):
Thanks! This is basically what I had in mind. But it looks like there are a lot of changes in here that are not really related to the important change, which is eliminating duplication between `Gson` and `GsonBuilder`. Would it possible to make it so that that is the only thing that changes?

Marcono1234:
> But it looks like there are a lot of changes in here that are not really related to the important change

Sorry, might also have been my fault to some extent due to the changes I had proposed to @MukjepScarlet.
I guess the changes to `EnumTypeAdapter.java` and `ConstructorConstructor.java` are unrelated and could be done in a separate PR indeed.

The changes to `SqlTypesSupport.java` and `LongSerializationPolicy.java` are related and probably nice to have since they slightly simplify creation of the default factories, but maybe could be omitted / be performed in a subsequent PR?

The change to `TypeAdapters.java` is closely related:
These adapters and factories are currently created within `Gson.java`, but with these changes, `GsonBuilder` instead of `Gson` now creates the default so it would make sense to move those adapters and factories out of `Gson.java` either way, and `TypeAdapters.java` seems suitable?
But maybe that could be done in a predecessor PR then? So first move the adapters and adapter factories to `TypeAdapters` in one PR, and then base this PR here on it.

@eamonnmcmanus, is that what you had in mind?

And @MukjepScarlet what do you think about this? Do you think this is a reasonable way to split this PR, and would you be able to do that?

MukjepScarlet:
Splitting the PR is reasonable, as certain changes within it were deliberately intended to serve as separate prerequisite PR. I have no objection. @Marcono1234

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -50,11 +50,15 @@ public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> typeToken) {
...
-  private final Map<T, String> constantToName = new HashMap<>();
+  private final Map<String, T> nameToConstant;
+  private final Map<String, T> stringToConstant;
+  private final Map<T, String> constantToName;
 
   private EnumTypeAdapter(Class<T> classOfT) {
+    nameToConstant = new HashMap<>();
+    stringToConstant = new HashMap<>();
+    // Don't use `EnumMap` here; that can break when using ProGuard or R8
+    constantToName = new HashMap<>();
```
MukjepScarlet: Can we use `ordinal` here? Or the `EnumMap`'s issue is due to using ordinal @Marcono1234

Marcono1234: > Or the `EnumMap`'s issue is due to using ordinal

The previous issue was because ProGuard and R8 seem to remove the `values()` method from the enum, and then [this code in the JDK](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/java.base/share/classes/java/lang/Class.java#L4004-L4021) encounters a `NoSuchMethodException` and returns `null`, despite the class being an enum (but a malformed one).
You can reproduce this by switching back to `EnumMap` and running `mvn clean verify` in the project root, which will run the test-shrinker module as well. You can even debug this if you afterwards run `ShrinkingIT` from the IDE (without running `mvn clean` in between though).


Do you want to use a `String[]` (or similar) and then look up the name based on the `ordinal`?

It seems the `test-shrinker` Maven module still builds successfully when using `ordinal`. However, maybe the tests there are simply not exhaustive enough. Not sure if ProGuard or R8 can remove the `ordinal()` method as well, or if they can omit enum constants, which could lead to a gap between `ordinal` values.

So I am not sure if it would be safe to use `ordinal`. Might be better to keep the current `HashMap` approach to be safe?

MukjepScarlet: Okay. I think we can add default size for the maps here.

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -77,6 +73,11 @@ private EnumTypeAdapter(Class<T> classOfT) {
 
       AccessibleObject.setAccessible(fields, true);
 
+      nameToConstant = new HashMap<>(constantCount);
+      stringToConstant = new HashMap<>(constantCount);
+      // Don't use `EnumMap` here; that can break when using ProGuard or R8
+      constantToName = new HashMap<>(constantCount);
```
Marcono1234: Note that these maps will now all be effectively too small and will have to be resized, because the behavior of the `initialCapacity` parameter is misleading, see also [JDK-8284377](https://bugs.openjdk.org/browse/JDK-8284377).

Could solve this by duplicating what the JDK does with its internal `HashMap#calculateHashMapCapacity` method in newer versions, that is, `(int) Math.ceil(numMappings / (double) DEFAULT_LOAD_FACTOR)`. But not sure if that is worth it.

The idea to pre-size the maps here was good nonetheless.

eamonnmcmanus:
(I'm still looking for a PR that has no changes other than the ones needed to remove the duplication between `Gson` and `GsonBuilder`. Not that those other changes are necessarily bad, they just shouldn't be here.)

Marcono1234:
@eamonnmcmanus, most of the changes have been moved by @MukjepScarlet into #2951, which is kind of a prerequisite for this PR here now I think. If possible could you please have a look at that PR?

Once that is (hopefully) merged and this PR here is updated, the remaining changes here should be only[^1] about the improvements for the Gson default instance.

[^1]: @MukjepScarlet, the `EnumTypeAdapter` change here is still unrelated to the Gson default instance changes though. Maybe it would be best to omit it from this PR here?

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
MukjepScarlet: This method looks like Guava and Kotlin-stdlib. (`ceil(n / DEFAULT_LOAD_FACTOR)`). But I think we can directly set load factor as 1.0 to avoid this. Waiting for your suggestion. I will revert this change in current PR.

MukjepScarlet:
@eamonnmcmanus I think this is ready for review now.

gson/src/test/java/com/google/gson/GsonTest.java
```diff
@@ -53,36 +49,25 @@ public final class GsonTest {
   private static final ToNumberStrategy CUSTOM_NUMBER_TO_NUMBER_STRATEGY =
       ToNumberPolicy.LAZILY_PARSED_NUMBER;
 
+  private static Gson createGson() {
+    GsonBuilder gsonBuilder = new GsonBuilder();
+    gsonBuilder.excluder = CUSTOM_EXCLUDER;
```
eamonnmcmanus, line 54: I think it would be a bit cleaner to set these values through the public API of `GsonBuilder`.

MukjepScarlet: Yes, better readability, just like how we use it outside of this package.

eamonnmcmanus:
Thanks again, @MukjepScarlet, for seeing this PR through to completion! I think it's a great simplification of the code.

PR #2869, opened 2025-07-02
Create android.yml

CathleenTico-jpg:
### Checklist


Marcono1234:
@CathleenTico-jpg, similar to #2866 please explain what you are trying to achieve with this pull request. The `android.yml` file you added cannot work because this project here is a Maven project, not a Gradle project.

PR #2878, opened 2025-07-15
Remove `GsonPreconditions#checkNotNull`

Marcono1234:
The method is not used within the Gson code and was only kept for backward compatibility in case external projects used it, see https://github.com/google/gson/pull/2180#discussion_r950869693.
However, #2838 renamed the enclosing class, which already broke backward compatibility. Therefore it should be safe to remove the method now.


Marcono1234:
Have created #2879 to remove the `GsonPreconditions` class

PR #2879, opened 2025-07-16
Remove internal class `GsonPreconditions`

Marcono1234:
Instead manually throw exceptions, also with exception messages providing more context

Follow-up for https://github.com/google/gson/pull/2878#pullrequestreview-3022433654


eamonnmcmanus (approved):
This is great, thanks! I like the improved exception messages a lot.

PR #2887, opened 2025-08-01
Update sources to satisfy a new Error Prone check.

eamonnmcmanus:
The latest Error Prone release includes a new check `EffectivelyPrivate`, which flags class members that are public or protected when the enclosing class isn't. This is only a warning in Error Prone, but because we treat warnings as errors it currently breaks the build.

Nearly all cases were in private nested classes in tests. For the most part the members didn't need to be public. In a couple of places they were accessed through reflection in the test, and it was enough to change `getField` to `getDeclaredField` (etc). (`getField` only retrieves public fields.) I don't believe any of these tests depended on the public members for the correctness of what they were testing. Although `getField` works when the target field is public, regardless of the visibility of the containing class, actually reading or writing a field requires both the field and its containing class to be public unless `Field.setAccessible(true)` is called, and likewise for constructors and methods.

### DIFF
--- a/gson/src/test/java/com/google/gson/FieldAttributesTest.java
+++ b/gson/src/test/java/com/google/gson/FieldAttributesTest.java
@@ -75,7 +75,7 @@ public class FieldAttributesTest {
   }
 
   private static class Foo {
-    @SuppressWarnings("unused")
+    @SuppressWarnings({"unused", "EffectivelyPrivate"})
     public transient List<String> bar;
   }
 }
--- a/extras/src/test/java/com/google/gson/interceptors/InterceptorTest.java
+++ b/extras/src/test/java/com/google/gson/interceptors/InterceptorTest.java
@@ -157,7 +157,7 @@ public final class InterceptorTest {
     String email;
     Address address;
 
-    public User(String name, String password) {
+    User(String name, String password) {
       this.name = name;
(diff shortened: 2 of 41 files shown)


gson/src/test/java/com/google/gson/FieldAttributesTest.java
```diff
@@ -75,7 +75,7 @@ public void testDeclaredTypeAndClass() {
   }
 
   private static class Foo {
-    @SuppressWarnings("unused")
+    @SuppressWarnings({"unused", "EffectivelyPrivate"})
```
cpovirk, line 78: Another option is `@Keep`, but I don't have much reason to prefer one over the other:

- `@Keep` would also satisfy Error Prone's unused-field checking, though AFAIK not any other tools' similar checking. (So it's unclear whether we'd be wise to remove `"unused"` at that point.) It also prevents Error Prone from suggesting things like making fields `final`, which may be good or bad in various cases.
- `@Keep` has runtime retention, which is likely irrelevant here.
- `@Keep` isn't applicable to method parameters, but that is likely irrelevant here, too.
- `@Keep` can be used as a meta-annotation, but that's not relevant here.

So this is fine as it is as far as I'm concerned, as are the other suppressions.

PR #2353, opened 2023-03-24
Create .github/workflows/scorecard.yml

eamonnmcmanus:
### Checklist

- [ ] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)
- [ ] If necessary, new public API validates arguments, for example rejects `null`
- [ ] New public API has Javadoc
    - [ ] Javadoc uses `@since $next-version$`  
      (`$next-version$` is a special placeholder which is automatically replaced during release)
- [ ] If necessary, new unit tests have been added  
  - [ ] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [ ] No JUnit 3 features are used (such as extending class `TestCase`)
  - [ ] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [ ] `mvn clean verify javadoc:jar` passes without errors

PR #2888, opened 2025-08-13
Create scorecard.yml

eamonnmcmanus:
Suggested by Google's internal GitHub Security Recommendations.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -73,6 +73,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
+        uses: github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
         with:
           sarif_file: results.sarif
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -73,6 +73,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@v3
+        github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
         with:
           sarif_file: results.sarif
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,78 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard (optional).
+      # Commenting out will disable upload of results to your repo's Code Scanning dashboard
+      - name: "Upload to code-scanning"
+        uses: github/codeql-action/upload-sarif@v3
```
cpovirk:
```suggestion
        github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
```

Prompted by some tool at some point, we've been hard-coding commit IDs, as you've done elsewhere in the file. This edit suggests [the version we're using in Guava](https://github.com/google/guava/blob/9fb2a4accc9717938176d2036293bf53eb1c18c7/.github/workflows/scorecard.yml#L70).

cpovirk (approved):
As you might already have surmised, this new file should be covered by [your existing Dependabot setup for GitHub Actions](https://github.com/google/gson/blob/f7de5c2c22dae98f2f98eefcb70c920ff71e437b/.github/dependabot.yml#L14), which matches [Guava's setup](https://github.com/google/guava/blob/9fb2a4accc9717938176d2036293bf53eb1c18c7/.github/dependabot.yml#L23), which has successfully led to PRs like https://github.com/google/guava/pull/7912/files.

.github/workflows/scorecard.yml
Marcono1234: The `uses:` is missing, causing a syntax error

cpovirk: Thanks. Sorry about that.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,78 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
```
Marcono1234, line 4: Do you think it makes sense to remove this disclaimer? It probably does not add much value.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard (optional).
+      # Commenting out will disable upload of results to your repo's Code Scanning dashboard
+      - name: "Upload to code-scanning"
+        github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
+        with:
```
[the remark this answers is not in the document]
eamonnmcmanus: Should be fixed by latest commit.

Marcono1234:
There is also an old PR (#2353) for adding `scorecard.yml`, which is probably obsolete now.

.github/workflows/scorecard.yml
eamonnmcmanus: I think I prefer to stick as closely as possible to the canned file that GitHub spat out. Apart from the version change suggested by @cpovirk, this file is exactly that.

eamonnmcmanus:
> There is also an old PR (#2353) for adding `scorecard.yml`, which is probably obsolete now.

That's funny! I have zero recollection of that, even though I apparently created it.

Issue #2892, opened 2025-08-25
Migrate from OSSRH to Central


cpovirk:
See https://central.sonatype.org/pages/ossrh-eol/ and https://central.sonatype.org/publish/publish-portal-maven/.

I had told @eamonnmcmanus that I would make an attempt at this, but I am emerging defeated... :)

The source-code changes for most of our own migrations have looked like https://github.com/google/truth/pull/1544. I started working on a similar change to GSON, removing `<distributionManagement>`, removing `maven-deploy-plugin` from `dependencyManagement`, and adding `central-publishing-maven-plugin` to `dependencyManagement` with `<skipPublishing>${gson.isInternalModule}</skipPublishing>` inside the `configuration` section. I also tried listing `central-publishing-maven-plugin` inside the `release` profile.

On top of that, we had to create a new user token, following the instructions at https://central.sonatype.org/publish/generate-portal-token/, and plug it into `~/.m2/settings.xml`. (I've been using `<id>central</id>`, but I'm not sure if that exact name is significant.)

Anyway, I was working through the quite detailed instructions at https://github.com/google/gson/blob/main/UserGuide.md (even knowing that I wouldn't ultimately have the permissions to deploy), and I got the impression that `maven-release-plugin` (which I hadn't used before) is running `maven-deploy-plugin` whether we want it to or not.

It is _probably_ possible to continue using `maven-deploy-plugin` (instead of migrating to `central-publishing-maven-plugin`) by switching to the legacy shim API documented at https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/. However, I and others have had trouble with missing files in our resulting deployments. (Supposedly that is an issue only if you upload different files from different IPs, but that seems highly unlikely in the cases I'm aware of. I think I may at least have heard that there were some potential fixes to remaining issues.) But as a more general point, it's a legacy shim.

The right thing to do is probably to either migrate off `maven-release-plugin` (at least for the actual deployment) or to find a way to force it to use `central-publishing-maven-plugin` instead of `maven-deploy-plugin` (if possible).

cpovirk:
> On top of that, we had to create a new user token, following the instructions at https://central.sonatype.org/publish/generate-portal-token/, and plug it into `~/.m2/settings.xml`. (I've been using `<id>central</id>`, but I'm not sure if that exact name is significant.)

[...]

PR #2899, opened 2025-09-09
Publish attempt [DO NOT MERGE]

eamonnmcmanus:
This is a snapshot of my attempts to make publishing with `central-publishing-maven-plugin` work. As detailed in #2892, this currently fails; it never actually stages anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ There are a few open-source projects that can convert Java objects to JSON. Howe
 Gradle:
 ```gradle
 dependencies {
-  implementation 'com.google.code.gson:gson:2.13.1'
+  implementation 'com.google.code.gson:gson:2.13.2-rc1'
 }
 ```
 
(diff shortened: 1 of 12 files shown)


Marcono1234:
Did you push the Git tag https://github.com/google/gson/releases/tag/gson-parent-2.13.2 by accident during your testing? Or was it intentional and you are planning to delete it again before the "real" release?

eamonnmcmanus:
> Did you push the Git tag https://github.com/google/gson/releases/tag/gson-parent-2.13.2 by accident during your testing? Or was it intentional and you are planning to delete it again before the "real" release?

Oh, thanks for catching that. It was indeed a mistake.

Marcono1234:
Might be good to delete the test branches now (unless you think they are still useful):
- `publish`
(this one contains the accidental `-rc1` version; though maybe the Git tag is enough and there is no need to keep the branch)
- `publish-attempt`

There are also two other stale branches:
- `ossf-scorecard-action` (was superseded by your recent scorecard PR)
- `strictness` (see https://redirect.github.com/google/gson/pull/2323#issuecomment-1575258558)

eamonnmcmanus:
Yes, thanks, we can certainly clean up all those branches.

Issue #2679, opened 2024-05-22
The module declaration is not visible in Eclipse IDE 


eitan-rosenberg:
The module declaration in version 2.11.0 is not visible, in contrast to version 2.10.1

Cannot use "requires com.google.gson;" in existing code like I used to.

Please help,

Marcono1234:
The module declaration is still under `META-INF/versions/9/module-info.class`, so nothing has changed regarding this between version 2.10.1 and 2.11.0.

Is only your IDE showing this error, or do you get this error also when you run the build tool (probably Maven or Gradle) from command line?

If this also occurs with Maven or Gradle, could you please try to create a small self-contained demo repository which showcases this issue? Please mention:
- which exact build command you using
- JDK version
- Maven / Gradle version
- operating system

eitan-rosenberg:
I am aware that you are using multi-release jar and I noticed META-INF/versions/9/module-info.class.(this is also the case in 2.10.1)
My environment:
Eclipse 2024-03 (4.31).
M2E Maven Integration for Eclipse Core 2.6.0.20240220-1109
Java openjdk-22.0.1_windows-x64_bin.zip
Maven 3.9.6.

I created a small Maven project. (included)

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

	<modelVersion>4.0.0</modelVersion>

	<groupId>ModuleTest</groupId>
	<artifactId>ModuleTest</artifactId>
	<version>0.0.1-SNAPSHOT</version>

	<properties>
		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
		<maven.compiler.source>22</maven.compiler.source>
		<maven.compiler.target>22</maven.compiler.target>
	</properties>

	<dependencies>

		<dependency>
			<groupId>com.google.code.gson</groupId>
			<artifactId>gson</artifactId>
			<!-- <version>2.11.0</version> -->
			<version>2.10.1</version>
		</dependency>

	</dependencies>

</project>
```
When using <version>2.10.1</version> every thing is ok.

![2024-05-23_08-51-31](https://github.com/google/gson/assets/48506054/b5d611f2-0c90-4d6a-ad19-f9706d8e9065)

[...]

PR #2900, opened 2025-09-10
Switch to using central-publishing-maven-plugin.

eamonnmcmanus:
I tested this by making a gson-parent-2.13.2 release, which however I dropped rather than publishing it. I intend to make a full release based on this change, once it is in the main branch.

There is a rather verbose incantation repeated in every module other than `gson` and `gson-parent`, to exclude those other modules from publication. There is probably a simpler way to do this, with profiles, but for now I'm going with simplicity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,13 @@
     <tag>HEAD</tag>
   </scm>
 
+  <distributionManagement>
+    <snapshotRepository>
+      <id>sonatype-nexus-snapshots</id>
+      <url>https://central.sonatype.com/repository/maven-snapshots/</url>
+    </snapshotRepository>
+  </distributionManagement>
+
   <developers>
     <developer>
       <id>google</id>
@@ -80,14 +87,6 @@
       <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
     </license>
   </licenses>
-
-  <distributionManagement>
-    <repository>
-      <id>sonatype-nexus-staging</id>
-      <name>Nexus Release Repository</name>
-      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
-    </repository>
-  </distributionManagement>
 
   <dependencyManagement>
     <dependencies>
@@ -377,11 +376,12 @@
           </configuration>
         </plugin>
         <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-deploy-plugin</artifactId>
-          <version>3.1.4</version>
+          <groupId>org.sonatype.central</groupId>
+          <artifactId>central-publishing-maven-plugin</artifactId>
+          <version>0.8.0</version>
+          <extensions>true</extensions>
           <configuration>
-            <skip>${gson.isInternalModule}</skip>
+            <publishingServerId>central</publishingServerId>
           </configuration>
         </plugin>
         <plugin>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -377,11 +376,12 @@
...
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-deploy-plugin</artifactId>
-          <version>3.1.4</version>
+          <groupId>org.sonatype.central</groupId>
+          <artifactId>central-publishing-maven-plugin</artifactId>
+          <version>0.8.0</version>
+          <extensions>true</extensions>
           <configuration>
-            <skip>${gson.isInternalModule}</skip>
+            <publishingServerId>central</publishingServerId>
```
cpovirk, line 384: I eventually concluded that [we did not need this](https://github.com/google/auto/pull/1953/files), and I've successfully released projects in that state.

But if you want to merge as-is because you know that things are working, I'm fine with that :)

pom.xml
```diff
@@ -377,11 +376,12 @@
...
         <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-deploy-plugin</artifactId>
-          <version>3.1.4</version>
+          <groupId>org.sonatype.central</groupId>
+          <artifactId>central-publishing-maven-plugin</artifactId>
+          <version>0.8.0</version>
+          <extensions>true</extensions>
           <configuration>
-            <skip>${gson.isInternalModule}</skip>
```
cpovirk, line 384: I suspect that we could now remove the places where we _set_ `isInternalModule`, which doesn't appear to have been used from places other than here.

But if you want to merge as-is because you know that things are working, I'm fine with that.

cpovirk (approved):
Hooray, and thank you!

pom.xml
eamonnmcmanus: Yes, that's exactly right. I want to get a thing in that works and is at least moderately clean. We can tweak later.

Marcono1234:
@eamonnmcmanus, for the release notes do you want to explicitly mention that this should fix the Eclipse and VS Code module declaration issue? (#2679)

eamonnmcmanus:
> @eamonnmcmanus, for the release notes do you want to explicitly mention that this should fix the Eclipse and VS Code module declaration issue? (#2679)

Thanks, I was trying to find that but couldn't in the moment. Do you want to suggest a wording for this note? (Or if you are able to edit them directly that would also be fine.)

Marcono1234:
Maybe something like this (but feel free to change it as you like):
> ### What's Changed
> - Improved packaging of JPMS module declaration in Gson jar\
> This fixes an issue where Eclipse and VS Code users could not refer to the Gson module name `com.google.gson`. See issue https://github.com/google/gson/issues/2679.
> - ...

Markdown source:
```md
- Improved packaging of JPMS module declaration in Gson jar\
  This fixes an issue where Eclipse and VS Code users could not refer to the Gson module name `com.google.gson`. See issue https://github.com/google/gson/issues/2679.
```

(I don't have permissions to edit the release notes myself.)

eamonnmcmanus:
> Maybe something like this (but feel free to change it as you like):

Thanks! Updated.

Marcono1234:
Thanks!

I have created #2902 to try simplifying the release configuration a bit (and updating the Release Process guide). If you have time the next days or weeks please let me know what you think.
Though it is not urgent; as we saw here the current release config works fine.

PR #2902, opened 2025-09-10
Simplify Central deployment & update Release Process guide

Marcono1234:
Follow-up for #2900, trying to simplify the configuration and updating the Release Process guide.

The general approach that I used here is to declared the central-publishing-maven-plugin for the parent `pom.xml`, but with `<inherited>false</inherited>`. Then only the submodules which should be deployed as well (currently only `gson`) need to declare the plugin.

That seems to work as desired and avoids having to apply central-publishing-maven-plugin configuration for all projects which should _not_ be deployed.
However, it does require explicitly setting `skip=false` for the maven-deploy-plugin. It seems otherwise for those modules which are not deployed (and don't have central-publishing-maven-plugin configured) the maven-deploy-plugin is not implicitly disabled.

I have marked this as Draft for now to discuss some of the changes first, because they might have downsides. Also, feel free to do any other changes / simplifications independently; I will resolve the merge conflicts in this PR then.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,6 +203,15 @@
           </execution>
         </executions>
       </plugin>
+
+      <!-- Publish gson-parent to Maven Central on `deploy` (common configuration is defined in `pluginManagement`) -->
+      <plugin>
+        <groupId>org.sonatype.central</groupId>
+        <artifactId>central-publishing-maven-plugin</artifactId>
+        <!-- Set `inherited=false` to only apply this execution to gson-parent;
+          if modules should be deployed, it has to be configured explicitly for them -->
+        <inherited>false</inherited>
+      </plugin>
     </plugins>
 
     <pluginManagement>
@@ -600,13 +609,6 @@
                 </goals>
               </execution>
             </executions>
-          </plugin>
-          <plugin>
-            <groupId>org.sonatype.central</groupId>
-            <artifactId>central-publishing-maven-plugin</artifactId>
-            <!-- Set `inherited=false` to only apply this execution to the parent project;
-              if modules should be deployed, it has to be configured explicitly for them -->
-            <inherited>false</inherited>
           </plugin>
         </plugins>
       </build>


pom.xml
```diff
@@ -48,8 +48,6 @@
     <!-- These properties are to be overwritten by the Maven modules -->
     <!-- Whether this module is an integration test module -->
     <gson.isTestModule>false</gson.isTestModule>
-    <!-- Whether this module is internal and currently not deployed -->
-    <gson.isInternalModule>${gson.isTestModule}</gson.isInternalModule>
```
Marcono1234, line 52: As mentioned in https://github.com/google/gson/pull/2900#discussion_r2337833744, `gson.isInternalModule` is not used anymore.

pom.xml
```diff
@@ -61,13 +59,6 @@
     <tag>HEAD</tag>
   </scm>
 
-  <distributionManagement>
-    <snapshotRepository>
-      <id>sonatype-nexus-snapshots</id>
-      <url>https://central.sonatype.com/repository/maven-snapshots/</url>
-    </snapshotRepository>
```
Marcono1234, line 68: I think currently Gson publishes no snapshots, and if I understand it correctly that URL is also the default for central-publishing-maven-plugin.

Edit: See `Constants.java` in the plugin source ([unofficial mirror code](https://github.com/mavenplugins/central-publishing-maven-plugin/blob/76ebfe6616d41c398884a100abb451b5b5710f12/src/main/java/org/sonatype/central/publisher/plugin/Constants.java#L77-L79); this seems to match the official plugin code in the [`-sources.jar`](https://repo1.maven.org/maven2/org/sonatype/central/central-publishing-maven-plugin/0.8.0/)).

ReleaseProcess.md
```diff
@@ -4,16 +4,20 @@ The following is a step-by-step procedure for releasing a new version of Google-
...
-1. Release the staging repository for Gson. Gson will now get synced to Maven Central with-in the next hour. For issues consult [Sonatype Guide](https://central.sonatype.org/publish/release/).
-1. Create a [GitHub release](https://github.com/google/gson/releases) for the new version. You can let GitHub [automatically generate the description for the release](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes), but you should edit it manually to point out the most important changes and potentially incompatible changes.
+1. [Log in to the Central Portal](https://central.sonatype.com/)
+1. Download and sanity check all files.\
+   Do not skip this step! Once you release the staging repository, there is no going back. It will get synced with Maven Central and you will not be able to update or delete anything. Your only recourse will be to release a new version of Gson and hope that no one uses the old one.
+1. Publish the new Gson version in the Central Portal.\
+   Gson will now get synced to Maven Central within the next hour.
```
Marcono1234, line 18: I am not sure if these steps are accurate since I haven't published any artifacts on Maven Central myself so far.

Any feedback here is appreciated.

ReleaseProcess.md
```diff
@@ -24,62 +28,49 @@ The following is a step-by-step procedure for releasing a new version of Google-
...
 Important: When aborting a release / rolling back release preparations, make sure to also revert all changes to files which were done during the release (e.g. automatic replacement of version references).
 
-## Configuring a machine for deployment to Sonatype Repository
-
-This section was borrowed heavily from [Doclava release process](https://code.google.com/archive/p/doclava/wikis/ProcessRelease.wiki).
-
-1. Install/Configure GPG following this [guide](https://blog.sonatype.com/2010/01/how-to-generate-pgp-signatures-with-maven/).
-1. [Create encrypted passwords](https://maven.apache.org/guides/mini/guide-encryption.html).
-1. Create `~/.m2/settings.xml` similar to as described in [Doclava release process](https://code.google.com/p/doclava/wiki/ProcessRelease).
-1. Now for deploying a snapshot repository, use `mvn deploy`.
```
Marcono1234, line 34: Removed this because it is most likely pretty outdated / there is probably better or more up to date information available.

ReleaseProcess.md
eamonnmcmanus: I'm not 100% sure that you can download the files from the portal, but I can check at the next release and update the text if necessary.

pom.xml
eamonnmcmanus: OK, yes.

ReleaseProcess.md
eamonnmcmanus: OK

Marcono1234:
> I think we won't know 100% for sure whether it works until the next release, but we can adjust as necessary then.

I tried with the "Testing Maven release workflow locally" steps in the Release Process guide, which is probably pretty close to how it behaves for a real release. But yes you are right, we can't be completely sure. Especially also in case the central-publishing-maven-plugin is updated in the meantime.

eamonnmcmanus:
Thanks again for doing this!

PR #2916, opened 2025-10-03
Fix outdated version comment for `github/codeql-action`

Marcono1234:
In https://github.com/google/gson/pull/2894 Dependabot did not update the version comments.

See also https://github.com/dependabot/dependabot-core/issues/13197. The assumption there is that this happened because `github/codeql-action` used two different tags ([v3.29.5](https://github.com/github/codeql-action/releases/tag/v3.29.5) and [v3.29.7](https://github.com/github/codeql-action/releases/tag/v3.29.7)) for the same commit. So when Dependabot updated the version, it thought it was updating from `v3.29.7`. When updating the version comment it checks for text matching the previous version number. Thus the old `v3.29.5` did not match and was not updated.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -32,7 +32,7 @@ jobs:
         path: ./out/artifacts
     - name: Upload Sarif
       if: always() && steps.build.outcome == 'success'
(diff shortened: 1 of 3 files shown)


eamonnmcmanus (approved):
Thanks for fixing this!

PR #2917, opened 2025-10-03
Update exec-maven-plugin to 3.6.0

Marcono1234:
Plugin is now using the configured regular repositories for downloading dependencies. See https://redirect.github.com/mojohaus/exec-maven-plugin/pull/480

See also https://github.com/google/gson/pull/2914#issuecomment-3363683962

### DIFF
--- a/test-shrinker/pom.xml
+++ b/test-shrinker/pom.xml
@@ -32,13 +32,13 @@
     <gson.isTestModule>true</gson.isTestModule>
   </properties>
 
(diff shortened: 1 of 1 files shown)


Marcono1234:
Going to close this; seems like it does not work in version 3.6.0 of that plugin and the maintainers will revert to the previous behavior, see https://github.com/mojohaus/exec-maven-plugin/pull/496.

PR #2919, opened 2025-10-06
Improve skipping of tests

Marcono1234:
### Purpose
Improve skipping of tests and avoid redundant CI builds

### Description
Uses the standard property `${maven.test.skip}` for skipping tests (which also includes skipping compilation of test classes, unlike `skipTests`), and configures test related plugin executions to use `${maven.test.skip}` as well.
As part of this the copy-rename-maven-plugin usage is replaced with maven-resources-plugin, which is probably also good because that former plugin is apparently not maintained anymore, see https://github.com/coderplus/copy-rename-maven-plugin.

Also configures the CI workflows to skip redundant execution of some Maven modules: For reproducible build, Android and API compatibility check it is irrelevant to check the metrics and test modules.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,13 +71,13 @@ jobs:
           cache: 'maven'
 
       - name: "Verify no plugin issues"
-        run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress
+        run: mvn artifact:check-buildplan --batch-mode --no-transfer-progress --projects '!metrics,!test-graal-native-image,!test-jpms,!test-shrinker'
 
       - name: "Verify reproducible build"
         # See https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
         run: |
-          mvn clean install --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests
+          mvn clean install --batch-mode --no-transfer-progress -Dmaven.test.skip --projects '!metrics,!test-graal-native-image,!test-jpms,!test-shrinker'
           # Run with `-Dbuildinfo.attach=false`; otherwise `artifact:compare` fails because it creates a `.buildinfo` file which
           # erroneously references the existing `.buildinfo` file (respectively because it is overwriting it, a file with size 0)
           # See https://issues.apache.org/jira/browse/MARTIFACT-57
-          mvn clean verify artifact:compare --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests -Dbuildinfo.attach=false
+          mvn clean verify artifact:compare --batch-mode --no-transfer-progress -Dmaven.test.skip --projects '!metrics,!test-graal-native-image,!test-jpms,!test-shrinker' -Dbuildinfo.attach=false
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -194,32 +194,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>com.coderplus.maven.plugins</groupId>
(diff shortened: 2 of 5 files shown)


gson/pom.xml
```diff
@@ -262,14 +237,47 @@
...
+            <!--
+              Note: This has to run after test compilation is already done; ideally would use `process-test-classes`
+              phase here, but the problem is that within this phase maven-resources-plugin has to run twice:
+              1. maven-resources-plugin: copy original classes
+              2. proguard-maven-plugin: obfuscate classes
+              3. maven-resources-plugin: copy obfuscated classes back
+
+              But Maven does not permit declaring the same plugin at different positions in the POM. Therefore have
+              to use a phase before `process-test-classes` here, and rely on the compiler plugin already being done.
+            -->
```
Marcono1234, line 251: This is not ideal here, but it seems to work currently. And if it failed, we would notice it because the unit test `EnumWithObfuscatedTest` verifies that the class was successfully obfuscated.

An alternative might be to run ProGuard in place on the `.class` files (not sure if that is possible), but that might be a bit risky since it could by accident affect unrelated classes as well. So the current approach of using separate temp dirs (`test-classes-obfuscated-injar` and `test-classes-obfuscated-outjar`) might be cleaner.
Though possibly we could avoid the `test-classes-obfuscated-injar` and let ProGuard directly use the classes from `test-classes` as input (but not as output), so that we would only need one copy-resources invocation.

eamonnmcmanus:
Thanks! But it seems there's an issue with the "Check API compatibility" step in the CI?

.github/workflows/build.yml
```diff
@@ -71,13 +71,13 @@ jobs:
...
       - name: "Verify reproducible build"
         # See https://maven.apache.org/guides/mini/guide-reproducible-builds.html#how-to-test-my-maven-build-reproducibility
         run: |
-          mvn clean install --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests
+          mvn clean install --batch-mode --no-transfer-progress -Dmaven.test.skip --projects '!metrics,!test-graal-native-image,!test-jpms,!test-shrinker'
           # Run with `-Dbuildinfo.attach=false`; otherwise `artifact:compare` fails because it creates a `.buildinfo` file which
           # erroneously references the existing `.buildinfo` file (respectively because it is overwriting it, a file with size 0)
           # See https://issues.apache.org/jira/browse/MARTIFACT-57
-          mvn clean verify artifact:compare --batch-mode --no-transfer-progress -Dproguard.skip -DskipTests -Dbuildinfo.attach=false
+          mvn clean verify artifact:compare --batch-mode --no-transfer-progress -Dmaven.test.skip --projects '!metrics,!test-graal-native-image,!test-jpms,!test-shrinker' -Dbuildinfo.attach=false
```
Marcono1234, line 83: Note: For the reproducible build check, in theory a broken test plugin or incorrect test setup here could affect the final artifact, so running a full build including tests (for all modules even?) could be useful.

But maybe that is quite unlikely and not worth it to cover.

Marcono1234:
Ah yes, check-api-compatibility fails because it always compares (and therefore builds) against the current `main`, but that still has the old Maven build config.

So I think this should solve itself for future PRs once this PR is merged.

eamonnmcmanus:
> Ah yes, check-api-compatibility fails because it always compares (and therefore builds) against the current `main`, but that still has the old Maven build config.
> 
> So I think this should solve itself for future PRs once this PR is merged.

OK, let's give that a try. We can always revert if it doesn't work.

eamonnmcmanus (approved):

PR #2920, opened 2025-10-06
Migrate to `io.github.ascopes:protobuf-maven-plugin`

Marcono1234:
### Purpose
Switch Maven plugin used for `proto` test execution

### Description
The previous plugin is not maintained anymore, and explicitly recommends that new plugin as successor, see
https://github.com/xolstice/protobuf-maven-plugin

The `*.proto` test resources have been moved from `/proto` to `/protobuf` since that is the directory recommended by the new plugin, see https://ascopes.github.io/protobuf-maven-plugin/generate-test-mojo.html


proto/pom.xml
```diff
@@ -74,27 +74,21 @@
...
-        <groupId>org.xolstice.maven.plugins</groupId>
+        <groupId>io.github.ascopes</groupId>
         <artifactId>protobuf-maven-plugin</artifactId>
-        <version>0.6.1</version>
+        <version>3.1.0</version>
         <configuration>
-          <protocArtifact>com.google.protobuf:protoc:${protobufVersion}:exe:${os.detected.classifier}</protocArtifact>
+          <skip>${maven.test.skip}</skip>
+          <protocVersion>${protobufVersion}</protocVersion>
+          <fatalWarnings>true</fatalWarnings>
```
Marcono1234, line 86: `fatalWarnings` is not strictly needed, but might be nice to notice potential future issues with the `*.proto` files early.

PR #2847, opened 2025-04-16
Add Java 24 into the test matrix

Goooler:
### Description

The latest JDK version should be checked on.

### Checklist

- [ ] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)\
  This is automatically checked by `mvn verify`, but can also be checked on its own using `mvn spotless:check`.\
  Style violations can be fixed using `mvn spotless:apply`; this can be done in a separate commit to verify that it did not cause undesired changes.
- [ ] If necessary, new public API validates arguments, for example rejects `null`
- [ ] New public API has Javadoc
    - [ ] Javadoc uses `@since $next-version$`  
      (`$next-version$` is a special placeholder which is automatically replaced during release)
- [ ] If necessary, new unit tests have been added  
  - [ ] Assertions in unit tests use [Truth](https://truth.dev/), see existing tests
  - [ ] No JUnit 3 features are used (such as extending class `TestCase`)
  - [ ] If this pull request fixes a bug, a new test was added for a situation which failed previously and is now fixed
- [x] `mvn clean verify javadoc:jar` passes without errors

PR #2921, opened 2025-10-06
Add CI build using JDK 25

Marcono1234:
### Purpose
Add CI build using JDK 25, so that we can be sure Gson works with JDK 25

Supersedes #2847

### Description
However, because building with JDK 25 does not fully work yet, mainly due to issues with ProGuard, still disallow building with JDK 25 by default. This way it fails fast for users locally instead of causing confusing failures later on.

Note that there are a few warnings related to `Unsafe` usage by Maven (respectively its dependencies) and by protobuf-java.

And also one test is causing warnings:
```
[INFO] Running com.google.gson.typeadapters.UtcDateTypeAdapterTest
WARNING: Use of the three-letter time zone ID "NST" is deprecated and it will be removed in a future release
WARNING: Use of the three-letter time zone ID "AST" is deprecated and it will be removed in a future release
WARNING: Use of the three-letter time zone ID "MST" is deprecated and it will be removed in a future release
```
(that test uses all time zones provided by `TimeZone.getAvailableIDs()`)

But besides that the build and Gson itself seems to work fine with JDK 25.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -365,5 +365,27 @@
         <excludeTestCompilation />
       </properties>
     </profile>
+    <!-- For JDK 25+ skip tests which rely on ProGuard because it does not support JDK 25 yet -->
+    <!-- TODO: Remove this once ProGuard supports JDK 25, and there is a solution for https://github.com/Guardsquare/proguard/issues/473 -->
+    <profile>
+      <id>JDK25</id>
+      <activation>
+        <jdk>[25,)</jdk>
+      </activation>
+      <properties>
+        <excludeTestCompilation>com/google/gson/functional/EnumWithObfuscatedTest.java</excludeTestCompilation>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.github.wvengen</groupId>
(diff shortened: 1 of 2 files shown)


gson/pom.xml
```diff
@@ -365,5 +365,27 @@
...
     </profile>
+    <!-- For JDK 25+ skip tests which rely on ProGuard because it does not support JDK 25 yet -->
+    <!-- TODO: Remove this once ProGuard supports JDK 25, and there is a solution for https://github.com/Guardsquare/proguard/issues/473 -->
+    <profile>
+      <id>JDK25</id>
+      <activation>
+        <jdk>[25,)</jdk>
+      </activation>
+      <properties>
+        <excludeTestCompilation>com/google/gson/functional/EnumWithObfuscatedTest.java</excludeTestCompilation>
```
Marcono1234, line 376: For simplicity I reused the existing `excludeTestCompilation` property which we are already using for excluding JDK 17+ tests on JDK 11.

A slightly cleaner solution might be to adjust the maven-surefire-plugin configuration instead, to exclude that test. But I am not sure if that is worth it, given that hopefully soon in the future ProGuard supports JDK 25.

PR #2925, opened 2025-10-23
Make AppendableWriter do flush and close if delegation object supports

MukjepScarlet:
### Purpose

`Streams.writerForAppendable` allows us to wrap an `Appendable` into a `Writer`. In the old implementation, the wrapped Writer does nothing on invocation of `flush` and `close`. Imagine we wrapped an object which implements both `Appendable` and `Closeable`, this might confuse the user.

### Description

If the wrapped object implements `Flushable` or `Closeable`, the corresponding function will be delegated to the Writer.

### Checklist

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/StreamsTest.java
+++ b/gson/src/test/java/com/google/gson/internal/StreamsTest.java
@@ -19,15 +19,59 @@ package com.google.gson.internal;
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertThrows;
 
+import java.io.Closeable;
+import java.io.Flushable;
 import java.io.IOException;
 import java.io.Writer;
 import org.junit.Test;
 
 public class StreamsTest {
+  private static class TestAppendable implements Appendable, Flushable, Closeable {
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
In Java I rarely see a class which implements both Appendable and one of Flushable/Closable while it's not a `Writer`. Maybe this case happens in pure Kotlin libs more commonly.

Marcono1234:
> In Java I rarely see a class which implements both Appendable and one of Flushable/Closable while it's not a `Writer`

Yes, you are probably right. One such case is `java.io.PrintStream` though, so this change here could be a problem if a user provides `System.out` or `System.err` and the stream is now closed due to these changes.

However, it seems for all cases where `Streams#writerForAppendable` is used internally in Gson, the `AppendableWriter` is not actually 'leaked' to the user[^1], and Gson itself apparently does not call `close()` or `flush()` on it either. So maybe this change is not actually noticeable by users and does not have any negative side effect.

And I agree that this is a reasonable change.

[^1]: Unless they implement a custom type adapter which calls `jsonWriter.close()`, though that is rather unlikely.

PR #2926, opened 2025-10-23
Migrate Writer APIs to Appendable parameter

### LINKED ISSUE #2924: Use `Appendable` instead of `Writer` as the parameter of `JsonWriter`
`new JsonWriter(new StringBuilder))`

# Feature description
I noticed Gson uses Appenable for other APIs but Writer for this. It might be wrapped easily.

# Alternatives / workarounds
-

MukjepScarlet:
### Purpose

Currently some of APIs use Appendable as parameter and some other use Writer as parameter. This PR migrates all Writer parameters into Appendable. No effect for normal usages.

### Description

closes #2924
Also adjusts some documentations.

### Checklist

- [] New public API has Javadoc (I don't know if these are "new" APIs)

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonWriter.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonWriter.java
@@ -229,8 +229,6 @@ public class JsonWriter implements Closeable, Flushable {
    * Creates a new instance that writes a JSON-encoded stream to {@code out}. For best performance,
    * ensure {@link Writer} is buffered; wrapping in {@link java.io.BufferedWriter BufferedWriter} if
    * necessary.
-   *
-   * @deprecated Use {@link #JsonWriter(Appendable)} instead. For compatibility only!
    */
   @InlineMe(replacement = "this((Appendable) out)")
   public JsonWriter(Writer out) {
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1034,10 +1034,11 @@ public final class Gson {
   }
 
   /**
-   * @deprecated Use {@link #newJsonWriter(Appendable)} instead. For compatibility only!
+   * For compatibility only!
+   *
+   * @see Gson#newJsonWriter(Appendable)
    */
   @InlineMe(replacement = "this.newJsonWriter((Appendable) writer)")
-  @Deprecated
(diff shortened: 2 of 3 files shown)


gson/src/main/java/com/google/gson/stream/JsonWriter.java
```diff
@@ -701,24 +701,32 @@ public JsonWriter jsonValue(String value) throws IOException {
   }
 
   /**
-   * Ensures all buffered data is written to the underlying {@link Writer} and flushes that writer.
+   * Ensures all buffered data is written to the underlying {@link Appendable} if it is an instance
+   * of {@link Flushable} and flushes that writer.
```
Marcono1234: I think the first part refers to data written by this `JsonWriter` (though the current implementation does not actually buffer any data it seems). So maybe should be reworded like this?
```suggestion
   * Ensures all buffered data is written to the underlying {@link Appendable} and flushes it if it is an instance of {@link Flushable}.
```
(might have to be formatted to match the code style)

MukjepScarlet:
> This only applies to _source_ compatibility.

I understand your meaning. My `no effect` means users can upgrade the lib without changing their code. I will add the old ones back and add comment for them.

MukjepScarlet:
@Marcono1234 This is because the JsonWriter don't depend on Writer API. About the reason, `Writer` has builtin synchronized, but I think 99% of JSON use cases don't need it. `Streams` is more likely an internal API.
Also, `StringReader` has synchronized too. I might replace it with a custom one in another PR.

Marcono1234:
I don't think the existing constructors and methods with `Writer` parameter should be deprecated. There is nothing broken or error-prone with them, so the deprecations just cause noise for users. Additionally using the new `Appendable` overload would require the users to do an `(Appendable) ...` cast just to avoid the deprecation warning, which is rather verbose.

---

> Also, `StringReader` has synchronized too. I might replace it with a custom one in another PR.

For this one, definitely create an issue first please to discuss this. Especially because there is no equivalent base interface (as it is the case for `Writer` and `Appendable`) and because [JDK 24 added `Reader.of(CharSequence)`](https://bugs.openjdk.org/browse/JDK-8341596). So I don't think this is necessary.

MukjepScarlet:
I think so. The `deprecation` will make it annoying, as a `Writer` param will auto use the old constructor.

PR #2927, opened 2025-10-23
JsonArray.of & JsonObject.of

### LINKED ISSUE #2923: JsonArray.of & JsonObject.of
No actual `problem`.

# Feature description
JsonArray.of(...Object/JsonElement)

JsonObject.of(String, JsonElement, String, JsonElement...)
JsonObject.of(String, Object,....)
JsonObject.of(Object...)

# Alternatives / workarounds
Java 9. List.of, Map.of
Guava. ImmutableList.of ImmutableMap.of

MukjepScarlet:
### Purpose

Like Java 9 and Guava.

### Description

closes #2923

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -21,6 +21,7 @@ import com.google.gson.internal.NonNullElementWrapperList;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 
@@ -40,12 +41,17 @@ import java.util.List;
  * @author Joel Leitch
  */
 public final class JsonArray extends JsonElement implements Iterable<JsonElement> {
+  /** All elements in this list are non-null. */
   private final ArrayList<JsonElement> elements;
 
+  @SuppressWarnings({"deprecation", "NonApiType"}) // superclass constructor
+  private JsonArray(ArrayList<JsonElement> elements) {
+    this.elements = elements;
+  }
+
   /** Creates an empty JsonArray. */
-  @SuppressWarnings("deprecation") // superclass constructor
   public JsonArray() {
-    elements = new ArrayList<>();
+    this(new ArrayList<>());
   }
 
   /**
@@ -55,9 +61,35 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @throws IllegalArgumentException if the {@code capacity} is negative
    * @since 2.8.1
    */
-  @SuppressWarnings("deprecation") // superclass constructor
   public JsonArray(int capacity) {
-    elements = new ArrayList<>(capacity);
+    this(new ArrayList<>(capacity));
+  }
+
+  /**
+   * Creates a JsonArray with the specified elements.
+   *
+   * @return a new JsonArray.
+   * @since $next-version$
+   */
+  public static JsonArray of(JsonElement... elements) {
+    JsonArray array = new JsonArray(elements.length);
+    array.addAll(elements);
+    return array;
+  }
+
+  /**
+   * Creates a JsonArray with the specified elements.
+   *
+   * @return a new JsonArray.
+   * @since $next-version$
+   */
+  public static JsonArray of(Iterable<? extends JsonElement> elements) {
+    JsonArray array =
+        elements instanceof Collection
+            ? new JsonArray(((Collection<?>) elements).size())
+            : new JsonArray();
+    array.addAll(elements);
+    return array;
   }
 
   /**
@@ -123,10 +155,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * @param element the element that needs to be added to the array.
    */
   public void add(JsonElement element) {
-    if (element == null) {
-      element = JsonNull.INSTANCE;
-    }
-    elements.add(element);
+    elements.add(element == null ? JsonNull.INSTANCE : element);
   }
 
   /**
@@ -136,6 +165,38 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    */
   public void addAll(JsonArray array) {
     elements.addAll(array.elements);
+  }
+
+  /**
+   * Adds all the elements to self.
+   *
+   * @param elements the elements to be added.
+   * @since $next-version$
+   */
+  public void addAll(JsonElement... elements) {
+    this.elements.ensureCapacity(elements.length + this.elements.size());
+    for (JsonElement element : elements) {
+      add(element);
+    }
+  }
+
+  /**
+   * Adds all the elements of the specified iterable to self.
+   *
+   * @param iterable the iterable whose elements need to be added to the array.
+   * @since $next-version$
+   */
+  public void addAll(Iterable<? extends JsonElement> iterable) {
+    if (iterable instanceof JsonArray) {
+      addAll((JsonArray) iterable);
+      return;
+    } else if (iterable instanceof Collection) {
+      elements.ensureCapacity(((Collection<?>) iterable).size() + elements.size());
+    }
+
+    for (JsonElement element : iterable) {
+      add(element);
+    }
   }
 
   /**
--- a/gson/src/main/java/com/google/gson/JsonObject.java
+++ b/gson/src/main/java/com/google/gson/JsonObject.java
@@ -45,6 +45,316 @@ public final class JsonObject extends JsonElement {
   public JsonObject() {}
 
   /**
+   * Creates a JsonObject with the specified member.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(String key, JsonElement value) {
+    JsonObject object = new JsonObject();
+    object.add(key, value);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(String key1, JsonElement value1, String key2, JsonElement value2) {
+    JsonObject object = new JsonObject();
+    object.add(key1, value1);
+    object.add(key2, value2);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
+      String key1,
+      JsonElement value1,
+      String key2,
+      JsonElement value2,
+      String key3,
+      JsonElement value3) {
+    JsonObject object = new JsonObject();
+    object.add(key1, value1);
+    object.add(key2, value2);
+    object.add(key3, value3);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
+      String key1,
+      JsonElement value1,
+      String key2,
+      JsonElement value2,
+      String key3,
+      JsonElement value3,
+      String key4,
+      JsonElement value4) {
+    JsonObject object = new JsonObject();
+    object.add(key1, value1);
+    object.add(key2, value2);
+    object.add(key3, value3);
+    object.add(key4, value4);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
+      String key1,
+      JsonElement value1,
+      String key2,
+      JsonElement value2,
+      String key3,
+      JsonElement value3,
+      String key4,
+      JsonElement value4,
+      String key5,
+      JsonElement value5) {
+    JsonObject object = new JsonObject();
+    object.add(key1, value1);
+    object.add(key2, value2);
+    object.add(key3, value3);
+    object.add(key4, value4);
+    object.add(key5, value5);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,241 @@ public final class JsonObject extends JsonElement {
...
+   * @since $next-version$
+   */
+  public static JsonObject of(String key, JsonElement value) {
+    JsonObject object = new JsonObject();
+    object.add(key, value);
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified member.
```
Marcono1234: Here and for the other `of` overloads taking more than one pair:

```suggestion
   * Creates a JsonObject with the specified members.
```

gson/src/main/java/com/google/gson/JsonObject.java
Marcono1234, line 1: Maybe something like `JsonObject.of(Map<String, JsonElement>)` would be useful as well? (or rather `copyOf` to make the behavior clearer?)

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,241 @@ public final class JsonObject extends JsonElement {
...
+      String key7,
+      JsonElement value7,
+      String key8,
+      JsonElement value8,
+      String key9,
+      JsonElement value9,
+      String key10,
+      JsonElement value10) {
+    JsonObject object =
+        of(
```
Marcono1234: Not sure if this delegation to the other `of` methods (10 -> 9 -> 8 -> ...) causes a bit too much nesting for the call stack.

Maybe it would be better to avoid this and either have duplicated code for the `object.add(key1, value1)` ... calls, or create a private helper method `private static JsonObject ofKeyValues(Object... keyValues)` which then loops over the array and does `add((String) keyValues[i], (JsonElement) keyValues[i + 1])`.

What do you think?

But I am not that sure about this. Would be good to wait for the opinion of one of the project members.

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,241 @@ public final class JsonObject extends JsonElement {
...
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified member.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
```
Marcono1234, line 281: These `of` overloads make it a bit verbose, though I guess there is no better way at the moment and Guava and the JDK do it like this as well.

The other alternative would be to only have maybe `of` 1-3 and for the rest require the user to use a different approach, e.g. the above mentioned `of(Map)`.

Would be good to wait for the opinion of one of the project members.

Marcono1234 (commented):
Thanks for the pull request! To me this looks quite reasonable and the implementation looks good[^1], but let's see what the project members think about this.

[^1]: It will need unit tests though, but might be good to wait for the feedback from the project members, to avoid wasted work.

gson/src/main/java/com/google/gson/JsonObject.java
MukjepScarlet: Basically I referred `Map.of`. I know Guava has `ImmutableMap.of(Object...)` but I'm not sure if it's capable here.

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -138,6 +148,38 @@ public void addAll(JsonArray array) {
...
+    }
+  }
+
+  /**
+   * Adds all the elements of the specified iterable to self.
+   *
+   * @param iterable the iterable whose elements need to be added to the array.
+   * @since $next-version$
+   */
+  public void addAll(Iterable<JsonElement> iterable) {
```
Marcono1234:
```suggestion
  public void addAll(Iterable<? extends JsonElement> iterable) {
```

Otherwise this doesn't compile:
```java
List<JsonPrimitive> list = List.of();
array.addAll(list);
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,313 @@ public final class JsonObject extends JsonElement {
...
+  }
+
+  /**
+   * Creates a JsonObject with the specified entries.
+   *
+   * @return a new JsonObject with given entries.
+   * @since $next-version$
+   */
+  @SafeVarargs
+  public static JsonObject ofEntries(Map.Entry<String, JsonElement>... entries) {
```
Marcono1234:
```suggestion
  public static JsonObject ofEntries(Map.Entry<String, ? extends JsonElement>... entries) {
```

Otherwise this doesn't compile:
```java
Map.Entry<String, JsonPrimitive> e = Map.entry("a", new JsonPrimitive(1));
JsonObject.ofEntries(e);
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,313 @@ public final class JsonObject extends JsonElement {
...
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified entries.
+   *
+   * @return a new JsonObject with given entries.
+   * @since $next-version$
+   */
+  public static JsonObject ofEntries(Iterable<Map.Entry<String, JsonElement>> entries) {
```
Marcono1234: I think this has to be:

```suggestion
  public static JsonObject ofEntries(Iterable<? extends Map.Entry<String, ? extends JsonElement>> entries) {
```

Otherwise this doesn't compile:
```java
var list = List.of(new AbstractMap.SimpleEntry<>("a", new JsonPrimitive(1));
JsonObject.ofEntries(list);
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,313 @@ public final class JsonObject extends JsonElement {
...
+  }
+
+  /**
+   * Creates a JsonObject with the specified map. All entries of the map will be shallow copied into
+   * the new JsonObject.
+   *
+   * @return a new JsonObject with all entries of given map.
+   * @since $next-version$
+   */
+  public static JsonObject copyOf(Map<String, JsonElement> map) {
```
Marcono1234:
```suggestion
  public static JsonObject copyOf(Map<String, ? extends JsonElement> map) {
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,313 @@ public final class JsonObject extends JsonElement {
...
+      object.add(entry.getKey(), entry.getValue());
+    }
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified map. All entries of the map will be shallow copied into
+   * the new JsonObject.
+   *
+   * @return a new JsonObject with all entries of given map.
```
Marcono1234: Maybe "from" is more correct here? "with" could maybe be misunderstood as being used as internal backing map.
```suggestion
   * Creates a JsonObject from the specified map. All entries of the map will be shallow copied into
   * the new JsonObject.
   *
   * @return a new JsonObject with all entries of the given map.
```

MukjepScarlet:
@Marcono1234 Can we use some codegen technique to generate factory methods for them? I don't know if we can modify an existing class, for example, adding a static method.

Marcono1234:
Which factory methods are you referring to?

Sorry if my previous comment was a bit confusing; I just meant that the methods you added should use `? extends JsonElement` (as shown in the individual comments), but nothing else.

Or do you want to use code generation to generate the `JsonObject#of` methods automatically to avoid the code duplication? In my opinion the code duplication there is ok since it is quite simple (it is not ideal but there is maybe no better solution). Setting up code generation just for that does not seem to justify the additional complexity.

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -318,9 +318,9 @@ public static JsonObject of(
    * @since $next-version$
    */
   @SafeVarargs
-  public static JsonObject ofEntries(Map.Entry<String, JsonElement>... entries) {
+  public static JsonObject ofEntries(Map.Entry<String, ? extends JsonElement>... entries) {
     JsonObject object = new JsonObject();
-    for (Map.Entry<String, JsonElement> entry : entries) {
+    for (Map.Entry<? extends String, ? extends JsonElement> entry : entries) {
```
Marcono1234: `? extends String` here is redundant, because `String` is `final`.

```suggestion
    for (Map.Entry<String, ? extends JsonElement> entry : entries) {
```

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -332,22 +332,22 @@ public static JsonObject ofEntries(Map.Entry<String, JsonElement>... entries) {
    * @return a new JsonObject with given entries.
    * @since $next-version$
    */
-  public static JsonObject ofEntries(Iterable<Map.Entry<String, JsonElement>> entries) {
+  public static JsonObject ofEntries(Iterable<? extends Map.Entry<String, ? extends JsonElement>> entries) {
     JsonObject object = new JsonObject();
-    for (Map.Entry<String, JsonElement> entry : entries) {
+    for (Map.Entry<? extends String, ? extends JsonElement> entry : entries) {
```
Marcono1234: `? extends String` here is redundant, because `String` is `final`.

```suggestion
    for (Map.Entry<String, ? extends JsonElement> entry : entries) {
```

Marcono1234:

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -41,12 +41,17 @@
  * @author Joel Leitch
  */
 public final class JsonArray extends JsonElement implements Iterable<JsonElement> {
+  /** All elements in this list are nonnull. */
   private final ArrayList<JsonElement> elements;
 
-  /** Creates an empty JsonArray. */
   @SuppressWarnings("deprecation") // superclass constructor
```
Marcono1234: You probably need this to suppress the Error Prone warning:

```suggestion
  @SuppressWarnings({"deprecation", "NonApiType"}) // superclass constructor
```

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -41,12 +41,17 @@
  * @author Joel Leitch
  */
 public final class JsonArray extends JsonElement implements Iterable<JsonElement> {
+  /** All elements in this list are nonnull. */
```
Marcono1234:
```suggestion
  /** All elements in this list are non-null. */
```

MukjepScarlet:
I will probably split this PR into 2 or more for easier review & merging.

eamonnmcmanus:
Sorry for not responding earlier. I completely agree that the `JsonArray` and `JsonObject` APIs are somewhat awkward, and in particular it is unnecessarily inconvenient to build up a tree of `JsonElement` objects using those APIs. However I'm not sure that just adding these factory methods is the right way to improve things. My feeling is that we might be better off with a general API that translates an arbitrary tree of Java objects (using `List`, `Map`, `Number` and so on) into the corresponding `JsonElement` tree. We would need to take the time to specify such an API carefully before sending any PRs. I hope to create a Gson issue shortly where we can discuss approaches.

MukjepScarlet:
I think extensions about `addAll` can be a simpler PR, like `addAll(JsonElement...)`
Also I noticed Gson is going to be Java 8, so we can have collectors for JsonArray and JsonObject too.

MukjepScarlet:
> My feeling is that we might be better off with a general API that translates an arbitrary tree of Java objects (using List, Map, Number and so on) into the corresponding JsonElement tree.

Generally we have something like DTO list/map which can't be simply parsed into JSON tree without serializer. If we only allow Map to JsonObject, this API might be not so useful.

eamonnmcmanus:
> > My feeling is that we might be better off with a general API that translates an arbitrary tree of Java objects (using List, Map, Number and so on) into the corresponding JsonElement tree.
> 
> Generally we have something like DTO list/map which can't be simply parsed into JSON tree without serializer. If we only allow Map to JsonObject, this API might be not so useful.

I think what you are describing is what Gson already does with its `TypeAdapter` API. What I'm describing is a way to build JSON trees more conveniently, typically with `List.of(...)` and `Map.of(...)`, where the only types that appear in the compound expression are `List`, `Map`, and primitives (in the `JsonPrimitive` sense). So I shouldn't have said "an arbitrary tree of Java objects". But I hope to spell all that out in a separate issue to make it clearer.

gson/src/main/java/com/google/gson/JsonObject.java
```diff
@@ -44,6 +44,316 @@ public final class JsonObject extends JsonElement {
...
+    return object;
+  }
+
+  /**
+   * Creates a JsonObject with the specified members.
+   *
+   * @return a new JsonObject.
+   * @since $next-version$
+   */
+  public static JsonObject of(
```
softpapers, line 176: 1. All the overloaded methods for "of" are not needed.
2. Consider using 
```
      public static JsonObject of(String... keysAndValues){
           int size = keysAndValues.lenght();
           if(size % 2 != 0) throw YourExceptionForInvalidParam("******");
               JsonObject object = new JsonObject();
               for(int i = 0; i < size; i+2;){
                   object.add(keysAndValues[i], keysAndValues[i+1]);
               }
             return object;
     }
```

softpapers (changes requested):

gson/src/main/java/com/google/gson/JsonObject.java
MukjepScarlet: This can only be a latest overload, like ImmutableMap.

softpapers: Why is such a requirement in the first instance? 
What problems are you solving with that?

MukjepScarlet: var it = JsonObject.of("key", JsonArray.of());
it.toString() -> {"key":[]}

instead of

var it = new JsonObject()
it.add("key", new JsonArray());

PR #2931, opened 2025-11-02
Update to latest Error Prone, and limit its use to Java 21+.

eamonnmcmanus:


pom.xml
Marcono1234, line 582: It seems you need this as well:

```suggestion
              <annotationProcessorPaths combine.self="override" />
            </configuration>
```

Plugins specified with `-Xplugin:...` are loaded from the annotation processors path (see [documentation](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.compiler/com/sun/source/util/Plugin.html)). The current configuration here only removes Error Prone from `<compilerArgs>` but not from `<annotationProcessorPaths>`. So the JDK still tries to load it.

Now as for why this was not an issue for JDK 11: There was a check [which returned fast if no plugins were specified](https://github.com/openjdk/jdk11u/blob/d3b1c2be9e87aad07cac29d94679130fe5807c17/src/jdk.compiler/share/classes/com/sun/tools/javac/api/BasicJavacTask.java#L204-L205). In newer JDK versions this check was intentionally removed, see the [JDK-8234211](https://bugs.openjdk.org/browse/JDK-8234211) and [corresponding commit](https://github.com/openjdk/jdk17u/commit/45b8d09e1bc2e41ddda79ac7881d840eb6e2d49a#diff-8dcb9e3145cee4b3f8669b7f437080dfe69041192d75faacfa9cc05a2af0c624L211-L212).

At least that is my understanding of it.

eamonnmcmanus: Thanks! Indeed, I couldn't figure out how this had worked before.

There's a later failure in building the `proto` tests but I'm hoping that one will be more tractable.

eamonnmcmanus: > There's a later failure in building the `proto` tests but I'm hoping that one will be more tractable.

(That went away when I did `mvn clean`.)

cpovirk: Thanks, I get confused every time I see merging of configuration across `pom.xml` files, and I don't want to think about how long it would have taken me to figure out that that was even the right area to look.

cpovirk (approved):

PR #2936, opened 2025-11-03
Simplify common Maven args for GitHub workflows

Marcono1234:
Additionally include `--show-version` to make troubleshooting easier.

The `MAVEN_ARGS` environment variable exists since Maven 3.9.0, see https://maven.apache.org/configure.html


eamonnmcmanus:
Thanks, that's a great improvement!

PR #2937, opened 2025-11-03
Add cooldown to Dependabot config

Marcono1234:
### Purpose
Add cooldown to the Dependabot config, to not use new dependency versions right after they were released

### Description
GitHub added a "cooldown" feature a few months ago to Dependabot ([changelog entry](https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/), [documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-)). This cooldown causes Dependabot to only update a version once it has been available for a specified number of days.
This can reduce the risk of running into not-yet-discovered bugs and can reduce the risk for supply chain security issues (assuming they are discovered in the meantime).

See also https://docs.zizmor.sh/audits/#dependabot-cooldown

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,9 +190,6 @@
         <executions>
           <execution>
             <goals>
-              <!-- This logs a warning about `source.scm.tag=HEAD`, but this can be ignored;
-                during release Maven Release Plugin temporarily changes the `source.scm.tag`
-                value to the actual Git tag, which will then not cause a warning -->
               <goal>buildinfo</goal>
             </goals>
           </execution>


Marcono1234:
What do you think? One motivation is also that so far version updates have not been time critical (and if necessary could be applied manually anyway), so waiting a few days or at worst one month (until the next Dependabot run) should not be an issue.

Side note: I haven't used the cooldown feature yet, but I think the syntax I used here is correct.

eamonnmcmanus (approved):
Thanks, this seems like a good idea.

PR #2944, opened 2025-11-20
Clean up interface method declarations by removing redundant `public` modifiers

ThirdGoddess:
### Purpose
This pull request makes a small stylistic cleanup of nested interface method declarations.  
Some methods currently include the `public` modifier while others rely on Java’s implicit visibility.  
This inconsistency is purely visual and does not affect functionality.  
Removing the redundant modifiers improves readability and enforces a consistent code style.
﻿
### Description
In Java, interface methods are implicitly `public`, so explicitly declaring `public` is unnecessary.  
This PR removes the redundant `public` keywords from nested interface methods throughout the codebase.  
﻿
No behavior or public API changes are introduced.  
The change is intended solely to make the code easier to read, review, and maintain.
﻿
### Checklist
- [x] No API behavior or signature changes
- [x] No new tests required
﻿

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/Streams.java
+++ b/gson/src/main/java/com/google/gson/internal/Streams.java
@@ -21,7 +21,7 @@ import com.google.gson.JsonIOException;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonParseException;
 import com.google.gson.JsonSyntaxException;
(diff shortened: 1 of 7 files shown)


eamonnmcmanus:
This makes sense. Google's internal "Java Practices" guide calls for omitting redundant modifiers like these.

eamonnmcmanus (approved):

PR #2947, opened 2025-11-22
Add build config for Gson subset

Marcono1234:
### Purpose
Follow-up for #2946

Adds a build config which builds the Gson API subset, so that any changes which break this can be directly noticed.

### Description
The Maven config is unfortunately a bit verbose because:
- it has to make sure only the explicitly listed source files are compiled
- it should not use the already compiled class files from `target/classes`

See also the comments in the changed code. Any hints on how this can be simplified are welcome.

As requested in https://github.com/google/gson/pull/2946#issuecomment-3566984829, the new GitHub workflow job will cause a workflow failure when building the Gson API subset fails.

### DIFF
--- a/gson/pom.xml
+++ b/gson/pom.xml
@@ -413,7 +413,6 @@
                     <resource>
                       <directory>${project.build.sourceDirectory}</directory>
                       <includes>
-                        <include>com/google/gson/stream/*</include>
                         <include>com/google/gson/FormattingStyle.java</include>
                         <include>com/google/gson/JsonArray.java</include>
                         <include>com/google/gson/JsonElement.java</include>
@@ -421,14 +420,20 @@
                         <include>com/google/gson/JsonNull.java</include>
                         <include>com/google/gson/JsonObject.java</include>
                         <include>com/google/gson/JsonParseException.java</include>
+                        <include>com/google/gson/JsonParser.java</include>
                         <include>com/google/gson/JsonPrimitive.java</include>
+                        <include>com/google/gson/JsonStreamParser.java</include>
                         <include>com/google/gson/JsonSyntaxException.java</include>
                         <include>com/google/gson/Strictness.java</include>
                         <include>com/google/gson/TypeAdapter.java</include>
+                        <include>com/google/gson/package-info.java</include>
+                        <include>com/google/gson/stream/JsonReader.java</include>
+                        <include>com/google/gson/stream/JsonScope.java</include>
+                        <include>com/google/gson/stream/JsonToken.java</include>
+                        <include>com/google/gson/stream/JsonWriter.java</include>
+                        <include>com/google/gson/stream/MalformedJsonException.java</include>
+                        <include>com/google/gson/stream/package-info.java</include>
                         <!-- Internal classes used by the classes above -->
-                        <include>com/google/gson/internal/bind/JsonElementTypeAdapter.java</include>
-                        <include>com/google/gson/internal/bind/JsonTreeReader.java</include>
-                        <include>com/google/gson/internal/bind/JsonTreeWriter.java</include>
                         <include>com/google/gson/internal/JsonReaderInternalAccess.java</include>
                         <include>com/google/gson/internal/LazilyParsedNumber.java</include>
                         <include>com/google/gson/internal/LinkedTreeMap.java</include>
@@ -436,6 +441,9 @@
                         <include>com/google/gson/internal/NumberLimits.java</include>
                         <include>com/google/gson/internal/Streams.java</include>
                         <include>com/google/gson/internal/TroubleshootingGuide.java</include>
+                        <include>com/google/gson/internal/bind/JsonElementTypeAdapter.java</include>
+                        <include>com/google/gson/internal/bind/JsonTreeReader.java</include>
+                        <include>com/google/gson/internal/bind/JsonTreeWriter.java</include>
                       </includes>
                     </resource>
                   </resources>
@@ -485,7 +493,9 @@
                     <testInclude>com/google/gson/JsonObjectAsMapSuiteTest.java</testInclude>
                     <testInclude>com/google/gson/JsonObjectAsMapTest.java</testInclude>
                     <testInclude>com/google/gson/JsonObjectTest.java</testInclude>
+                    <testInclude>com/google/gson/JsonParserParameterizedTest.java</testInclude>
                     <testInclude>com/google/gson/JsonPrimitiveTest.java</testInclude>
+                    <testInclude>com/google/gson/JsonStreamParserTest.java</testInclude>
                     <testInclude>com/google/gson/SubsetTest.java</testInclude>
                     <testInclude>com/google/gson/TypeAdapterTest.java</testInclude>
                   </testIncludes>
--- a/gson/src/test/java/com/google/gson/JsonArrayTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayTest.java
@@ -21,8 +21,6 @@ import static org.junit.Assert.assertThrows;
 
 import com.google.common.testing.EqualsTester;
 import com.google.gson.common.MoreAsserts;
-import com.google.gson.internal.bind.JsonElementTypeAdapter;
-import java.io.IOException;
 import java.math.BigInteger;
 import org.junit.Test;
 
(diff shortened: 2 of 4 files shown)


gson/src/main/java/com/google/gson/JsonElement.java
```diff
@@ -45,7 +44,7 @@
  *       <pre>
  * JsonObject jsonObject = JsonParser.parseString("{}").getAsJsonObject();
  * </pre>
- *   <li>{@link Gson#fromJson(Reader, Class) Gson.fromJson(..., JsonElement.class)}<br>
+ *   <li>{@link Gson#fromJson(java.io.Reader, Class) Gson.fromJson(..., JsonElement.class)}<br>
```
Marcono1234, line 47: Changed this (removed the `import` and used the qualified name here instead) because Error Prone was reporting an unused import for `java.io.Reader` when compiling the subset.

gson/pom.xml
```diff
@@ -387,5 +387,92 @@
...
+                <goals>
+                  <goal>copy-resources</goal>
+                </goals>
+                <configuration>
+                  <outputDirectory>${gsonSubsetSrcDir}</outputDirectory>
+                  <overwrite>true</overwrite>
+                  <resources>
+                    <resource>
+                      <directory>${project.build.sourceDirectory}</directory>
+                      <includes>
```
Marcono1234, line 415: I tried to group the includes here a bit, but if you prefer them to not be grouped and instead only be sorted lexicographically, please let me know.

eamonnmcmanus: Yes, I would prefer a lexicographical grouping, because it makes it easier to see if a given source file is included or not.

gson/pom.xml
```diff
@@ -387,5 +387,92 @@
...
+                        <include>com/google/gson/stream/*</include>
+                        <include>com/google/gson/FormattingStyle.java</include>
+                        <include>com/google/gson/Strictness.java</include>
+                        <!-- Exception types -->
+                        <include>com/google/gson/JsonParseException.java</include>
+                        <include>com/google/gson/JsonIOException.java</include>
+                        <include>com/google/gson/JsonSyntaxException.java</include>
+                        <!-- Type adapter, serializer and deserializer -->
+                        <include>com/google/gson/TypeAdapter.java</include>
+                        <include>com/google/gson/JsonSerializer.java</include>
```
eamonnmcmanus: I did not find I needed to include `JsonSerializer` and `JsonSerializationContext` in the subset we're using internally at Google. Are they needed here for some reason?

eamonnmcmanus (changes requested):
Thanks for doing this, it is great!

Just a couple of things.

gson/pom.xml
Marcono1234: No, I just guessed the subset. I have also removed `JsonDeserializer` and `JsonDeserializationContext` now. I hope the list now matches what you are using.

gson/pom.xml
```diff
@@ -458,12 +458,49 @@
                   <!-- Use custom source root dir; using the default `src/main/java` would otherwise implicitly
                     compile needed classes, even if they are not explicitly listed in `<includes>` -->
                   <compileSourceRoots>${gsonSubsetSrcDir}</compileSourceRoots>
-                  <!-- Use custom output directory to avoid interfering with normal Gson build -->
-                  <outputDirectory>${project.build.directory}/gson-subset-classes</outputDirectory>
```
Marcono1234: Have removed this again because otherwise the test compilation cannot see these classes, and it looks like at least maven-compiler-plugin itself does not support extending the classpath? (could maybe be done through external plugins, but not sure if that is worth it)

I guess this is not an issue as long as you always run `mvn clean ...` to be safe, respectively don't cache the `target/` dir in CI.

Marcono1234:
> Would it be possible too to run `SubsetTest` against the copied subset of sources?

Have adjusted the PR with one possible solution for this, and have included some of the other tests as well. I hope that is ok and useful; if not I can limit it to `SubsetTest`.

I also made some small changes because `JsonParser` is currently not part of the subset. Should it be part of the subset?

I am not that sure though how useful it is to run the tests for the subset, given that the regular Gson build runs these tests as well. I guess most issues would already pop up during the compilation of the subset.
These tests would only help if the subset referred to any classes through reflection only (something which would not be noticed as compilation error).

This subset build config feels rather brittle (I think it would also not notice non-existent `<includes>` or `<testIncludes>`), so maybe it would be good to use a less brittle setup in the long term.

gson/pom.xml
eamonnmcmanus: Weird, I was sure I had commented with the exact list of source files that we're including internally, but I don't see my comment. Here it is again:
```
gson/src/main/java/com/google/gson/FormattingStyle.java
gson/src/main/java/com/google/gson/JsonArray.java
gson/src/main/java/com/google/gson/JsonElement.java
gson/src/main/java/com/google/gson/JsonIOException.java
gson/src/main/java/com/google/gson/JsonNull.java
gson/src/main/java/com/google/gson/JsonObject.java
gson/src/main/java/com/google/gson/JsonParseException.java
gson/src/main/java/com/google/gson/JsonParser.java
gson/src/main/java/com/google/gson/JsonPrimitive.java
gson/src/main/java/com/google/gson/JsonStreamParser.java
gson/src/main/java/com/google/gson/JsonSyntaxException.java
gson/src/main/java/com/google/gson/LongSerializationPolicy.java
gson/src/main/java/com/google/gson/Strictness.java
gson/src/main/java/com/google/gson/TypeAdapter.java
gson/src/main/java/com/google/gson/internal/JsonReaderInternalAccess.java
gson/src/main/java/com/google/gson/internal/LazilyParsedNumber.java
gson/src/main/java/com/google/gson/internal/LinkedTreeMap.java
gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java
gson/src/main/java/com/google/gson/internal/NumberLimits.java
gson/src/main/java/com/google/gson/internal/Streams.java
gson/src/main/java/com/google/gson/internal/TroubleshootingGuide.java
gson/src/main/java/com/google/gson/internal/bind/JsonElementTypeAdapter.java
gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
gson/src/main/java/com/google/gson/internal/bind/JsonTreeWriter.java
gson/src/main/java/com/google/gson/package-info.java
gson/src/main/java/com/google/gson/stream/JsonReader.java
gson/src/main/java/com/google/gson/stream/JsonScope.java
gson/src/main/java/com/google/gson/stream/JsonToken.java
gson/src/main/java/com/google/gson/stream/JsonWriter.java
gson/src/main/java/com/google/gson/stream/MalformedJsonException.java
```

In particular, yes, it does include `JsonParser`.

eamonnmcmanus: Also, in the comment that I remember writing but don't see, I mentioned that I wasn't sure that `LongSerializationPolicy` belonged in the list, since I don't think you can use it without also using the `Gson` class.

eamonnmcmanus:
> Have adjusted the PR with one possible solution for this, and have included some of the other tests as well. I hope that is ok and useful; if not I can limit it to `SubsetTest`.

Yes, we also do that internally, with a fairly random selection of tests. The one that matters is `SubsetTest`, though, because we want to make sure that it builds and runs against the subset.

gson/pom.xml
Marcono1234: Have adjusted the list of classes. But `LongSerializationPolicy` indeed only seems to be relevant for `Gson` / `GsonBuilder`, so I have omitted it.

gson/pom.xml
```diff
@@ -485,7 +493,9 @@
                     <testInclude>com/google/gson/JsonObjectAsMapSuiteTest.java</testInclude>
                     <testInclude>com/google/gson/JsonObjectAsMapTest.java</testInclude>
                     <testInclude>com/google/gson/JsonObjectTest.java</testInclude>
+                    <testInclude>com/google/gson/JsonParserParameterizedTest.java</testInclude>
                     <testInclude>com/google/gson/JsonPrimitiveTest.java</testInclude>
+                    <testInclude>com/google/gson/JsonStreamParserTest.java</testInclude>
```
Marcono1234, line 498: Can't include `JsonParserTest` because its `testReadWriteTwoObjects` method uses `Gson` and refers to `TestTypes`, which uses a lot of Gson classes not included in the subset.

Marcono1234:
Yes I think so, if the list of classes and tests is fine for you like this.

As side note: For the `<testIncludes>` I think it might also compile (and eventually run) other test classes not explicitly listed, in case they are referenced by one of the classes listed as `<testIncludes>`[^1]. But that is probably not an issue? And if they referenced classes not part of the Gson subset, it seems compilation would still fail (as desired).

[^1]: Otherwise the same approach as for the regular compilation might be needed, that is, copy the test classes to a separate directory and use that as `<compileSourceRoots>`.

eamonnmcmanus:
> As side note: For the `<testIncludes>` I think it might also compile (and eventually run) other test classes not explicitly listed, in case they are referenced by one of the classes listed as `<testIncludes>`. But that is probably not an issue?

Yeah, that doesn't seem problematic to me.

Issue #739, opened 2015-11-20
Could you please add support of JodaTime as is part of Java8 ?


CyborTronik:
Hi,
Could you please make builtin support of JodaTime types as is included in Java and no needs for separate library?

JakeWharton:
Note: JodaTime and JSR310 are very different things.

inder123:
minimum supported version of Java in Gson is Java 5. The next version will bump it up to Java 6.
We don't intend to require Java 8 any time soon.

Is there a way this can be supported without bumping up the minimum supported version?

JakeWharton:
I would guess a peer library built against Java 8 which just provides a
TypeAdapterFactory for javax.time.

kang0921ok:
Hi I hava a question.
Not supported yet?
What is recommended method use a jsr310 in gson?

Mario-Eis:
According to Oracle (https://www.oracle.com/java/technologies/java-se-support-roadmap.html), even the extended support for Java 7 ended in July 2022.

RockyMM:
@Mario-Eis I guess this does not matter. There are many clients of this library still using Java 7. I would guess that there are a lot or some very important internal Google projects that use Java 7. Until the gson in whole drops Java 7 support, there won't be any progress here.

joannakotula:
Until gson supports java.time properly it will not fully support java 17. We are currently migrating to java 17, and it seems that we will have to replace all gson calls, as we use java.time a lot:
```
com.google.gson.JsonIOException: Failed making field 'java.time.Duration#seconds' accessible; either increase its visibility or write a custom TypeAdapter for its declaring type.
```

ca-chan-ka:
Workaround: Create a gson instance with custom adapters like this

```java
import java.lang.reflect.Type;
import java.time.*;
import java.util.Map;
import java.util.function.Function;
import com.google.gson.*;
​
public class GsonWithJavaTime {
    public static Gson getInstance() {
        GsonBuilder builder = new GsonBuilder();
        Map<Type, Function<String, Object>> parsers = Map.of(
                ZonedDateTime.class, ZonedDateTime::parse,
                LocalDate.class, LocalDate::parse,
                LocalDateTime.class, LocalDateTime::parse
        );
        parsers.entrySet().forEach(entry -> builder.registerTypeAdapter(entry.getKey(), adapter(entry.getValue())));
        return builder.create();
    }
​
    private static <T> JsonDeserializer<T> adapter(Function<String, T> function) {
[...]

PR #2948, opened 2025-11-25
Add type adapters for `java.time` classes

eamonnmcmanus:
Add type adapters for `java.time` classes.

These adapters essentially freeze the JSON representation that`ReflectiveTypeAdapterFactory` established by default, based on the private fields of `java.time` classes. That's not a great representation, but it is understandable. Changing it to anything else would break compatibility with systems that are expecting the current format.

If Gson had been updated with `java.time` adapters at the time the `java.time` API was added, the representation would surely have been something else, probably ISO standard formats. We can potentially supply non-default adapters for that, but we'll still need to have these legacy adapters by default.

I've been meaning to make this change for a while, but the need to do so becomes more urgent with [this JDK
commit](https://github.com/openjdk/jdk/commit/cc05530b813564a40c233eaaa80b906795c6d752) which makes a number of `java.time` fields `transient`. That means that, in JDK 26, the reflective-based adapter will no longer work with the classes that were changed.

This addresses #1059 and #739. All relevant classes in the `java.time` package are covered.

(Google internal bug reference: b/463237567.)

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -16,6 +16,8 @@
 
 package com.google.gson.internal.bind;
 
+import static java.lang.Math.toIntExact;
+
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonIOException;
@@ -770,15 +772,6 @@ public final class TypeAdapters {
           };
         }
       };
-
-  // This is Math.toIntExact, but works on Android level 23.
-  private static int toIntExact(long value) {
-    int castValue = (int) value;
-    if (castValue != value) {
-      throw new ArithmeticException("Too big for an int: " + value);
-    }
-    return castValue;
-  }
 
   public static final TypeAdapterFactory CALENDAR_FACTORY =
       newFactoryForMultipleTypes(Calendar.class, GregorianCalendar.class, CALENDAR);


gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+
+  public static final TypeAdapter<Duration> DURATION =
+      new IntegerFieldsTypeAdapter<Duration>("seconds", "nanos") {
+        @Override
+        Duration create(long[] values) {
+          return Duration.ofSeconds(values[0], values[1]);
+        }
+
+        @Override
+        @SuppressWarnings("JavaDurationGetSecondsGetNano")
```
cpovirk: Huh, I would have hoped that the `getSeconds` and `getNano` calls below were close enough to avoid this warning :(

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+      this.fields = Arrays.asList(fields);
+    }
+
+    abstract T create(long[] values);
+
+    abstract long[] integerValues(T t);
+
+    @Override
+    public T read(JsonReader in) throws IOException {
+      if (in.peek() == JsonToken.NULL) {
```
cpovirk, line 713: Not any sort of blocker for this PR, but: Has there been discussion of a common `TypeAdapter` subclass (even if only a private one for our own use) that passes through nulls, leaving subclasses free to implement `readNonNull` and `writeNonNull` methods without the null-handling ceremony? Maybe the common class could even call `beginObject` and `endObject` (if it's not dangerous that that delays the call to `endObject` until after the user subclass has done its thing)? (I mean, it's always going to be possible to go further down this road, like with automatic exception handling and even a callback-style API that invokes your callback with a name and `JsonReader` inside the `while (!end)` loop..... And if only we could assume that records are available, we could declare a record class for each as a sort of serialization proxy, maybe? Anyway, I know we're not in the mode of adding features, so I'm just acting in my role as a reviewer who sees the same pattern a few times in this file :))

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+        in.nextNull();
+        return null;
+      }
+      in.beginObject();
+      long[] values = new long[fields.size()];
+      while (in.peek() != JsonToken.END_OBJECT) {
+        String name = in.nextName();
+        int index = fields.indexOf(name);
+        if (index >= 0) {
+          values[index] = in.nextLong();
```
cpovirk, line 723: In general, if we get an unrecognized field name, we're going to blow up, just probably when we request the _next name_, right (since we'll have a value left over, rather than a name)? I don't think it would be worth going far out of our way to fail moments earlier; I just didn't pick up on the behavior at first read.

update: And now I see a "Ignore unknown JSON property" comment in the original code below. I think that's what it _was_ doing (well, at least for integral-valued properties) but not what it's doing anymore. We could preserve the old behavior easily enough, but I'm not sure if we care to? Plus, GitHub is matching that comment up with your new "Ignore other fields" ZDT comment, and my guess is that we don't actually ignore other fields in that implementation.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+            return;
+          }
+          out.beginObject();
+          out.name("id");
+          out.value(value.getId());
+          out.endObject();
+        }
+      };
+
+  // TODO: this does not currently work correctly.
```
cpovirk: I can't say I've figured out why, but I'm flagging this in case you still need to fix anything here or you want to leave it `private` (not that it "should" matter in `internal` code).

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
       };
 
-  public static final TypeAdapterFactory CALENDAR_FACTORY =
-      newFactoryForMultipleTypes(Calendar.class, GregorianCalendar.class, CALENDAR);
+  public static final TypeAdapterFactory JAVA_TIME_FACTORY =
+      new TypeAdapterFactory() {
+        @Override
+        public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> typeToken) {
+          Class<? super T> rawType = typeToken.getRawType();
+          if (!rawType.getName().startsWith("java.time.")) {
```
cpovirk: If you want this to work with Android library desugaring, you might need to recognize `j$.time`—or maybe even obfuscated versions, which would be even more of a pain? (I recently ran into that in unsubmitted `guava-testlib` CL 827531714.)

Maybe we don't care unless the desugared versions of `java.time` start using `transient` internally or something? I bring this all up mostly as something to watch out for. I could try to think more about it if you'd like. (Maybe you could also get some information from testing this internally if you haven't already. My guess is that it would work fine for at least a while, especially with desugaring enabled in our internal builds with minSdkVersion 23.)

pom.xml
```diff
@@ -527,8 +527,8 @@
                     future, could consider switching to https://github.com/open-toast/gummy-bears
                     which accounts for Android desugaring and might allow usage of more Java classes -->
                   <groupId>net.sf.androidscents.signature</groupId>
-                  <artifactId>android-api-level-21</artifactId>
-                  <version>5.0.1_r2</version>
+                  <artifactId>android-api-level-26</artifactId>
```
cpovirk: I saw that your commit description mentions that you were thinking about putting this in a separate PR.

Alternatively, you can introduce your own annotation for Animal Sniffer suppressions (likely copying the standard name, "`IgnoreJRERequirement`"), [configure your build to respect it](https://github.com/google/truth/blob/55d3fa76c482565afde5ea84d4021999705c19f2/pom.xml#L290), and then suppress at the narrow scope so that we continue to test that we're compatible with 21 (or 23, since that's the typical Google default nowadays) except in code that we're careful to run only under newer versions.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+      long[] values = new long[fields.size()];
+      while (in.peek() != JsonToken.END_OBJECT) {
+        String name = in.nextName();
+        int index = fields.indexOf(name);
+        if (index >= 0) {
+          values[index] = in.nextLong();
+        }
+      }
+      in.endObject();
+      return create(values);
```
Marcono1234, line 729: For better troubleshooting, would it make sense to wrap the `create` call in a `try-catch(RuntimeException)` and wrap the exceptions in a `JsonSyntaxException` which includes `in.getPreviousPath()` (and includes the original exception as cause)?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
-                break;
-              case HOUR_OF_DAY:
-                hourOfDay = value;
+              default:
+                // Ignore other fields.
+            }
+          }
+          in.endObject();
+          return requireNonNull(localTime, "Missing time field")
+              .atDate(requireNonNull(localDate, "Missing date field"));
```
Marcono1234: Here and below for `ZONED_DATE_TIME`, would it make sense to have a custom `requireNonNull` method like this?
```java
private static <T> T requireNonNullField(T obj, String fieldName, JsonReader jsonReader) {
  if (obj == null) {
    throw new JsonSyntaxException("Missing " + fieldName + " field; at path " + jsonReader.getPreviousPath());
  }
  return obj;
}
```

Could also use it for the `JsonSyntaxException` thrown by `ZONE_REGION` then.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -812,6 +824,119 @@ public void testStringBufferDeserialization() {
...
+    roundTrip(zonedDateTime, json);
+  }
+
+  private static final boolean JAVA_TIME_FIELDS_ARE_ACCESSIBLE;
+
+  static {
+    boolean accessible = false;
+    try {
+      Instant.class.getDeclaredField("seconds").setAccessible(true);
+      accessible = true;
```
Marcono1234, line 1014: Can this ever succeed with the current Maven build config?

I think it could only succeed for the JDK 11 CI builds, but even there it won't work because we have `--illegal-access=deny` in `gson/pom.xml`.

So maybe we need a custom maven-surefire-plugin execution which just runs `DefaultTypeAdaptersTest` and has `--add-opens`.

(Side note: Would it make sense to move these `java.time` tests to a dedicated test class?)

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -812,6 +824,119 @@ public void testStringBufferDeserialization() {
...
+  @Test
+  public void testJavaZonedDateTime() {
+    ZonedDateTime zonedDateTime =
+        ZonedDateTime.of(
+            LocalDate.of(2021, 12, 2), LocalTime.of(12, 34, 56, 789_012_345), ZoneOffset.UTC);
+    String json =
+        "{\"dateTime\":{\"date\":{\"year\":2021,\"month\":12,\"day\":2},"
+            + "\"time\":{\"hour\":12,\"minute\":34,\"second\":56,\"nano\":789012345}},"
+            + "\"offset\":{\"totalSeconds\":0},"
+            + "\"zone\":{\"id\":\"Z\"}}";
```
Marcono1234: Not sure if that is the reason why you consider this adapter to not be working correctly, but it seems for the reflection roundtrip this fails due to this diff:

```suggestion
            + "\"zone\":{"totalSeconds":0}}";
```

It seems the reason is that because `ZoneOffset.UTC` is a `ZoneOffset`, it is used as both `offset` and `zone`, see [JDK code](https://github.com/openjdk/jdk/blob/d350158e060c01acf49759dcbdd1f4d72530111b/src/java.base/share/classes/java/time/ZonedDateTime.java#L373-L375).

But with the above change the non-reflection-based adapter fails.

So maybe you need a dedicated `ZoneId` adapter which delegates to the `ZoneOffset` / `ZoneRegion`:
- for serialization check `zone instanceof ZoneOffset`
- for deserialization check which fields are present, and decide based on that as which type to deserialize

What do you think?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
       }.nullSafe();
   public static final TypeAdapterFactory CURRENCY_FACTORY = newFactory(Currency.class, CURRENCY);
 
+  /**
+   * An abstract {@link TypeAdapter} for classes whose JSON serialization consists of a fixed set of
+   * integer fields. That is the case for the legacy serialization of various {@code java.time}
+   * types.
+   *
+   * <p>This is a base class for {@link Calendar}, {@link Duration}, {@link Instant}, {@link
+   * LocalDate}, {@link LocalTime}, {@link LocalDateTime}, and {@link ZoneOffset}.
```
Marcono1234: Not sure if it is really worth it to list all these classes here; it risks that this becomes outdated in case one of the adapters is changed, and users can get this information also by looking at the subclasses of this adapter class.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+              case "time":
+                localTime = LOCAL_TIME.read(in);
                 break;
-              case DAY_OF_MONTH:
-                dayOfMonth = value;
-                break;
-              case HOUR_OF_DAY:
-                hourOfDay = value;
+              default:
+                // Ignore other fields.
```
Marcono1234: Similar to https://github.com/google/gson/pull/2948#discussion_r2560858259, this is missing a `in.skipValue()` call for unknown fields?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+          while (in.peek() != JsonToken.END_OBJECT) {
+            if (in.nextName().equals("id")) {
+              String id = in.nextString();
+              try {
+                zoneId = ZoneId.of(id);
+              } catch (DateTimeException e) {
+                throw new JsonSyntaxException(
+                    "Failed parsing '" + id + "' as ZoneId; at path " + in.getPreviousPath(), e);
+              }
+            }
```
Marcono1234: Similar to https://github.com/google/gson/pull/2948#discussion_r2560858259, this is missing a `in.skipValue()` call for unknown fields?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,83 +702,339 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+                zoneOffset = ZONE_OFFSET.read(in);
                 break;
-              case SECOND:
-                second = value;
+              case "zone":
+                zoneId = ZONE_REGION.read(in);
                 break;
               default:
-                // Ignore unknown JSON property
+                // Ignore other fields.
```
Marcono1234: Similar to https://github.com/google/gson/pull/2948#discussion_r2560858259, this is missing a `in.skipValue()` call for unknown fields?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: I will somewhat sheepishly admit that an AI wrote this paragraph and I allowed myself to be convinced to include it. I agree that it is not useful; saying that it is the base class for `Calendar` and for a number of `java.time` types is enough.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Actually I was forgetting that you just need to tack on `.nullSafe()` when constructing a `TypeAdapter` in order to get null checking. I think it makes sense to keep the null check here, rather than having to remember to add `nullSafe()` for all the subclasses, but for the other new `TypeAdapter` implementations we can do it.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Yes, @Marcono1234 also noticed some carelessness about unrecognized fields elsewhere. This needs to be fixed.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: I mean, I think we would have to be a lot more thorough if we wanted to allow troubleshooting like that. It would catch things like `{"month":2,"day":31}` for a `MonthDay`, but not `{"month":"foo","day":"bar"}`, or all the other places the input could be wrong. This sounds more like a general improvement we could make at a higher level, to catch exceptions from `JsonReader.read` in general and add wrap in a `JsonSyntaxException` as you describe.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: I'm not sure what you mean? The warning is supposed to prevent you from calling `getSeconds()` when you meant `toSeconds()`.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Yes, thanks.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: That's easy enough, OK, though again I feel like we should figure out how to do this in general for all exceptions when reading.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Yes.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: I think I have now figured it all out.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Yes.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: I'm going to hope that this isn't a problem, and deal with it if turns out to be. Internally, as you say, we have minSdkVersion 23, and I am just going to exclude the `java.time` handling from the internal Android version of Gson for now.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
eamonnmcmanus: That is indeed what I ended up doing. Thanks!

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
eamonnmcmanus: We could and probably should do that. Meanwhile, it _is_ currently true when testing internally to Google. That's because we've had to add an `--add-opens` for Gson to be able to access `java.time`, which in turn is because of the exact problem that this PR is addressing.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
cpovirk: https://errorprone.info/bugpattern/JavaDurationGetSecondsGetNano:

> If you call duration.getNano(), you must also call duration.getSeconds() in ‘nearby’ code.

I get the impression that the check would be happy with `foo(duration.getSeconds(), duration.getNano())` but not the array case. But I see no other examples in my depot search for `([{]|new.*\[)[^\042]*getSeconds.*getNano lang:java`, so I don't think I'll bother to file a feature request.

(I forget where we landed with `getSeconds` vs. `toSeconds`. I just know that I got sad about the various `get*` and `to*` methods, including the `to*Part` methods, in b/201676318.)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
cpovirk: Your strategy for excluding things from the Android version LGTM. That said, we _might_ have been able to just leave these in the Android version, both internally and externally: I wouldn't expect runtime errors, since we check for `java.time` with reflection above.

The concern I was expressing in my original comment is merely that this new adapter might not hande desugared classes because they live under `j$.time`, not the `java.time` package that we check for here. I think that's fine for now because the `j$.time` classes don't currently use `transient` on their fields, so the default reflective approach should continue to work—at least until you succeed at disabling _that_ :) But by then, we _might_ no longer need `j$.time` internally, so the problem might go away. (Naturally, your new reflective approach _also_ won't handle `j$.time` internally if we omit the adapters from the Android version entirely :) I guess the adapters will still be in place for Android users externally, but I stand by my claim that that's likely to be safe.)

Still, your new reflective approach _might_ avoid trouble internally in at least one case: If I were to guess at something that _could_ actually break with the `java.time` adapters in the Android version, it would be an overeager test for an internally built, externally shipped SDK that (a) intentionally avoids requiring desugaring and (b) uses Gson. Such a test might have noticed the references to `Duration`, etc. below and in the `TypeAdapter` fields above, and it might complain. (And actually, it still might, since tools like Proguard are likely to follow the reflection and still see the new adapter.) But I don't know if anyone is in the intersection of that Venn diagram. If it comes up, I'm sure we can navigate it.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -689,81 +690,96 @@ public void write(JsonWriter out, Currency value) throws IOException {
...
+            calendar.get(Calendar.MONTH),
+            calendar.get(Calendar.DAY_OF_MONTH),
+            calendar.get(Calendar.HOUR_OF_DAY),
+            calendar.get(Calendar.MINUTE),
+            calendar.get(Calendar.SECOND)
+          };
         }
       };
 
+  // This is Math.toIntExact, but works on Android level 23.
```
cpovirk: `toIntExact` should work under any Android version, even if the user doesn't opt in to library desugaring, because [it's one of the methods that is "backported"](https://r8.googlesource.com/r8/+/d684a72b325fd8b1f6ec0ec3a7be78a44032a0e4/src/main/java/com/android/tools/r8/ir/desugar/BackportedMethodRewriter.java#906) instead of just handled by opt-in desugaring.

https://github.com/open-toast/gummy-bears, already mentioned in your `pom.xml`, gets this right. Not that there is any harm in having our own helper method here; this sort of thing just may come up again in time.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -813,6 +829,22 @@ public void write(JsonWriter out, Locale value) throws IOException {
...
+  }
+
+  public static TypeAdapterFactory javaTimeTypeAdapterFactory() {
+    try {
+      Class<?> javaTimeTypeAdapterFactoryClass =
+          Class.forName("com.google.gson.internal.bind.JavaTimeTypeAdapters");
+      FactorySupplier supplier =
+          (FactorySupplier) javaTimeTypeAdapterFactoryClass.getDeclaredConstructor().newInstance();
+      return supplier.get();
+    } catch (ReflectiveOperationException e) {
```
cpovirk, line 836: After https://github.com/openjdk/jdk/pull/28018#issuecomment-3655087181, I have again lost faith in my ability to determine when and how a JVM will report missing classes :) Notably, even if tests pass in the various environments we test in, we have seen [failures in other environments](https://github.com/google/truth/issues/333#issuecomment-765652454).

So: I wouldn't 100% rule out that `Class.forName` and `newInstance` could succeed (so there's no `ReflectiveOperationException`) but that `get` could fail. Options include "Try what we have and see what happens" and "`catch (Throwable t)`," or maybe you have something else we could try. Maybe the constructor for `JavaTimeTypeAdapters` could instantiate a `Duration` instance for the sake of (presumably) triggering a failure that would still result in `ReflectiveOperationException`? Or, again, we could hope for the best with what we have now.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -812,6 +836,232 @@ public void testStringBufferDeserialization() {
...
+  private void roundTrip(Gson customGson, Object value, Class<?> valueClass, String expectedJson) {
+    assertThat(customGson.getAdapter(valueClass).getClass().getName()).doesNotContain("Reflective");
+    assertThat(customGson.toJson(value, valueClass)).isEqualTo(expectedJson);
+    assertThat(customGson.fromJson(expectedJson, valueClass)).isEqualTo(value);
+  }
+
+  // Assuming we have reflective access to the fields of java.time classes, check that
+  // ReflectiveTypeAdapterFactory would produce the same JSON. This ensures that we are preserving
+  // a compatible JSON format for those classes even though we no longer use reflection.
+  private void checkReflectiveTypeAdapterFactory(Object value, String expectedJson) {
```
cpovirk, line 1044: Do you want to skip the tests yet under Java 26(?)+, or is it too early to worry about that?

eamonnmcmanus: I think I'll address that when it comes up. There's already a lot going on in this PR. :-)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: "Try what we have and see what happens" SGTM. I'm uncomfortable speculating on possible failure modes until we know what they are.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Thanks! Updated.

pom.xml
eamonnmcmanus: I did end up doing that, thanks.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
eamonnmcmanus: Actually, I think we should at least also catch `LinkageError` here, which seems eminently possible. I'll put that in a followup.

PR #2949, opened 2025-11-25
Change the minimum supported Android API to 26.

eamonnmcmanus:
This is [Android 8.0](https://developer.android.com/tools/releases/platforms#8.0), released in August 2017. Only about 2% of devices have an older version.

Requiring API Level 26 means we can use `java.time` APIs without reflection.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -109,14 +109,14 @@ public final class ConstructorConstructor {
     @SuppressWarnings("unchecked") // types must agree
     InstanceCreator<T> typeCreator = (InstanceCreator<T>) instanceCreators.get(type);
     if (typeCreator != null) {
-      return () -> typeCreator.createInstance(type);
+      return new InstanceCreatorConstructor<>(typeCreator, type);
     }
 
     // Next try raw type match for instance creators
     @SuppressWarnings("unchecked") // types must agree
     InstanceCreator<T> rawTypeCreator = (InstanceCreator<T>) instanceCreators.get(rawType);
     if (rawTypeCreator != null) {
-      return () -> rawTypeCreator.createInstance(type);
+      return new InstanceCreatorConstructor<>(rawTypeCreator, type);
     }
 
     // First consider special constructors before checking for no-args constructors
@@ -143,9 +143,7 @@ public final class ConstructorConstructor {
     // of adjusting filter suggested below is irrelevant since it would not solve the problem
     String exceptionMessage = checkInstantiable(rawType);
     if (exceptionMessage != null) {
-      return () -> {
-        throw new JsonIOException(exceptionMessage);
-      };
+      return new ThrowingObjectConstructor<>(exceptionMessage);
     }
 
     if (!allowUnsafe) {
@@ -153,9 +151,7 @@ public final class ConstructorConstructor {
           "Unable to create instance of "
               + rawType
               + "; Register an InstanceCreator or a TypeAdapter for this type.";
-      return () -> {
-        throw new JsonIOException(message);
-      };
+      return new ThrowingObjectConstructor<>(message);
     }
 
     // Consider usage of Unsafe as reflection, so don't use if BLOCK_ALL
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
This also means we can add `Optional` adapters into standard. Crazy.

eamonnmcmanus:
> This also means we can add `Optional` adapters into standard. Crazy.

Yes, that's a good point. Similar compatibility considerations to the `java.time` classes would probably lead us to copy what the (terrible) reflection-based adapter does currently, so `Optional.of("foo")` becomes `{"value":"foo"}` and `Optional.empty()` becomes `{}`, or `{"value":null}` if `serializeNulls()` is set. I think probably putting in the `null` value regardless of `serializeNulls()` would make sense. Plain `{}` is pretty cryptic.

Similarly `OptionalInt` would become `{"isPresent":true,"value":23}` or `{"isPresent":false}`. The current serialization includes `"value":0` for the empty case but I don't think that matters.

This is probably not the serialization we would have chosen if there were no compatibility considerations. People can always register their own adapters (and probably often do already). I found several in Google's codebase, more than one of which serialized `Optional.of(foo)` the same as `foo`, with `Optional.empty()` being serialized as `null`.

Marcono1234:
Can you please adjust the README as well:
https://github.com/google/gson/blob/a271f487b5969601308ed438b2da9c8fc1a8e927/README.md?plain=1#L71-L74

There are also (at least) these two places where the code could be improved now:
- https://github.com/google/gson/blob/a271f487b5969601308ed438b2da9c8fc1a8e927/gson/src/main/java/com/google/gson/internal/reflect/ReflectionHelper.java#L131-L133 (using `Executable` avoids the `executable instanceof Method` check further below)

- https://github.com/google/gson/blob/a271f487b5969601308ed438b2da9c8fc1a8e927/gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java#L130-L131 (could additionally also override and delegate `removeIf`; I think for all other methods overridden by `ArrayList` delegation would not be directly possible because the methods could add `null` values)
(but might require extending the tests in `JsonArrayAsListTest` then)

Or if you prefer I can do that in a separate PR.

cpovirk:
I ended up leaving [a comment about this on a separate PR](https://github.com/google/gson/pull/2948#discussion_r2560913444). Based on what you've subsequently said elsewhere, I think we may end up having to maintain three sorts of compatibility:

- Gson works under minSdkVersion _x_.
  - In theory, we could have different values for _x_ for users who enable library desugaring than for users who don't.
- Animal Sniffer's checking against minSdkVersion _x_ passes. (Again, you get to make a decision about library desugaring.) One part of that picture could be suppressions, as I mentioned in [the comment linked above](https://github.com/google/gson/pull/2948#discussion_r2560913444).
   - Incidentally, I see the existing comment about using the "scents" from https://github.com/open-toast/gummy-bears. That's worked well for us [in Guava](https://github.com/google/guava/pull/6846), possibly better than `net.sf.androidscents.signature` in handling always-desugared ("backported?") APIs. But if it ain't broke....
- The internal test you uncovered with your CL 836391909 passes its minSdkVersion-23 checking (since that's the current minimum for most Google foundational libraries). Again, part of the story may be suppressions, in that case using `@RequiresApi`.
   - Now that I think about it, I'm surprised that we don't seem to have seen problems with those tests when code in _Guava_ refers to classes like `Optional` and `Stream`. Maybe they avoid Guava in that particular library?

When it comes to Google-internal usage of the class, I think you'll end up wanting to maintain compatibility with minSdkVersion 23, but you could get away with requiring library desugaring. I wouldn't expect you to be able to get away with minSdkVersion 26 yet, though I hope we're only a year or two away....

Then the question would be whether all _other_ users could meet a baseline of 23+desugaring or (if you can support it easily enough) an optional alternative baseline of 26 without need for desugaring. (It's possible to have [fancy Animal Sniffer checking to check both of those](https://github.com/google/guava/issues/7197).)

You might prefer to keep things simple and compatible for now. That could mean supporting minSdkVersion 23 even without desugaring (as Guava does unless you use APIs with "Java 8" types in them), in which case the main work to do is to suppress the errors you're seeing, since the "Java 8" code in #2948 already won't run except under a newer Android version. But as always, nothing will be as simple as it should be.

MukjepScarlet:
The functional interfaces can have Java 8 annotation `FunctionalInterface`, like JsonSerializer.

Goooler:
You can also mention https://developer.android.com/studio/write/java8-support in the docs if users still want to stay on API 26 or lower.

eamonnmcmanus:
It turns out that this is a little bit premature. Internally, some Google tests are still requiring API level 23. I was able to rework #2948 so that it can work even at that level.

We will _eventually_ update the API level to 26 or more, but not just yet.

PR #2950, opened 2025-11-25
refactor: slightly optimize ConstructorConstructor

MukjepScarlet:
### Purpose

Part of #2864

### Description

lambda -> method reference (reduce synthetic methods)
duplicated lambdas -> a class

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -143,7 +143,7 @@ public final class ConstructorConstructor {
     // of adjusting filter suggested below is irrelevant since it would not solve the problem
     String exceptionMessage = checkInstantiable(rawType);
     if (exceptionMessage != null) {
-      return new ExceptionObjectConstructor<>(exceptionMessage);
+      return new ThrowingObjectConstructor<>(exceptionMessage);
     }
 
     if (!allowUnsafe) {
@@ -151,7 +151,7 @@ public final class ConstructorConstructor {
           "Unable to create instance of "
               + rawType
               + "; Register an InstanceCreator or a TypeAdapter for this type.";
-      return new ExceptionObjectConstructor<>(message);
+      return new ThrowingObjectConstructor<>(message);
     }
 
     // Consider usage of Unsafe as reflection, so don't use if BLOCK_ALL
@@ -163,7 +163,7 @@ public final class ConstructorConstructor {
               + "; ReflectionAccessFilter does not permit using reflection or Unsafe. Register an"
               + " InstanceCreator or a TypeAdapter for this type or adjust the access filter to"
               + " allow using reflection.";
-      return new ExceptionObjectConstructor<>(message);
+      return new ThrowingObjectConstructor<>(message);
     }
 
     // finally try unsafe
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
@eamonnmcmanus Please review this. Because I'm not Android developer, I'm not sure if these are benificial with Proguard/R8.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -431,17 +413,51 @@ private <T> ObjectConstructor<T> newUnsafeAllocator(Class<? super T> rawType) {
...
   }
+
+  /*
+   * Create ObjectConstructor which throws exception.
+   * This keeps backward compatibility (compared to returning `null` which
+   * would then choose another way of creating object).
+   * And it supports types which are only serialized but not deserialized
+   * (compared to directly throwing exception here), e.g. when runtime type
+   * of object is inaccessible, but compile-time type is accessible.
+   */
```
Marcono1234: Can you please convert this to a proper Javadoc comment please, and adjust the parts which were specific to the previous call site. For example something like this:

```java
  /**
   * {@link ObjectConstructor} which always throws an exception.
   * 
   * <p>This keeps backward compatibility, compared to using a {@code null} {@code ObjectConstructor},
   * which would then choose another way of creating the object. And it supports types which are only
   * serialized but not deserialized (compared to directly throwing an exception when the
   * {@code ObjectConstructor} is requested), e.g. when the runtime type of an object is inaccessible,
   * but the compile-time type is accessible.
   */
```
(will need `mvn spotless:apply` to reflow the comment)

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -431,17 +413,51 @@ private <T> ObjectConstructor<T> newUnsafeAllocator(Class<? super T> rawType) {
...
+
+    ExceptionObjectConstructor(String exceptionMessage) {
+      this.exceptionMessage = exceptionMessage;
+    }
+
+    @Override
+    public T construct() {
+      // New exception is created every time to avoid keeping reference
+      // to exception with potentially long stack trace, causing a
+      // memory leak
```
Marcono1234: Can you please adjust this comment to make it a bit clearer?
(at its original location the meaning might have been more obvious from the context)

```suggestion
      // New exception is created every time to avoid keeping a reference to an exception with potentially long stack trace, causing a memory leak
      // (which would happen if the exception was already created when the `ExceptionObjectConstructor` is created)
```

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
```diff
@@ -431,17 +413,51 @@ private <T> ObjectConstructor<T> newUnsafeAllocator(Class<? super T> rawType) {
...
+
+  /*
+   * Create ObjectConstructor which throws exception.
+   * This keeps backward compatibility (compared to returning `null` which
+   * would then choose another way of creating object).
+   * And it supports types which are only serialized but not deserialized
+   * (compared to directly throwing exception here), e.g. when runtime type
+   * of object is inaccessible, but compile-time type is accessible.
+   */
+  private static final class ExceptionObjectConstructor<T> implements ObjectConstructor<T> {
```
Marcono1234: Maybe a name like `ThrowingObjectConstructor` would be better, to avoid any misunderstandings that this creates (not throws) `Exception` instances?

Marcono1234 (commented):
Thanks a lot!

The idea to add an `ExceptionObjectConstructor` is really good to reduce the code duplication. I have just a few small suggestions.

The project members might have additional change requests though.

gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
MukjepScarlet: Applied

PR #2951, opened 2025-11-25
refactor: move derived adapters from Gson to TypeAdapters

MukjepScarlet:
### Purpose

Part of #2864

### Description

Move constant adapters from Gson to TypeAdapters.

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -44,11 +44,14 @@ import java.util.Currency;
 import java.util.GregorianCalendar;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.StringTokenizer;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongArray;
 
 /**
  * Type adapters for basic types. More complex adapters exist as separate classes in the enclosing
@@ -300,6 +303,22 @@ public final class TypeAdapters {
   public static final TypeAdapterFactory ATOMIC_INTEGER_FACTORY =
       newFactory(AtomicInteger.class, TypeAdapters.ATOMIC_INTEGER);
 
+  public static TypeAdapter<AtomicLong> atomicLongAdapter(TypeAdapter<Number> longAdapter) {
+    Objects.requireNonNull(longAdapter);
+    return new TypeAdapter<AtomicLong>() {
+      @Override
+      public AtomicLong read(JsonReader in) throws IOException {
+        Number value = longAdapter.read(in);
+        return new AtomicLong(value.longValue());
+      }
+
+      @Override
+      public void write(JsonWriter out, AtomicLong value) throws IOException {
+        longAdapter.write(out, value.get());
+      }
+    }.nullSafe();
+  }
+
   public static final TypeAdapter<AtomicBoolean> ATOMIC_BOOLEAN =
       new TypeAdapter<AtomicBoolean>() {
         @Override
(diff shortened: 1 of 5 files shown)


gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -374,51 +425,112 @@ public void write(JsonWriter out, Number value) throws IOException {
...
+              + " behavior, use GsonBuilder.serializeSpecialFloatingPointValues() method.");
+    }
+  }
+
+  public static final TypeAdapter<Number> FLOAT = new FloatAdapter(false);
+  public static final TypeAdapter<Number> FLOAT_STRICT = new FloatAdapter(true);
+
+  public static TypeAdapter<Number> floatAdapter(boolean serializeSpecialFloatingPointValues) {
+    return serializeSpecialFloatingPointValues ? TypeAdapters.FLOAT : TypeAdapters.FLOAT_STRICT;
+  }
```
Marcono1234: Not sure if these `floatAdapter(boolean)` and `doubleAdapter(boolean)` methods are really that useful if the only user is the `Gson` (and soon the `GsonBuilder`) class.

Maybe it would be easier to just make these instance methods of `Gson` like here (in the original PR):
https://github.com/google/gson/blob/c62f6f4d97b79785158f9bc0419634f3b9a75930/gson/src/main/java/com/google/gson/GsonBuilder.java#L1015-L1021

And then later move these methods to `GsonBuilder`.

What do you think?

gson/src/main/java/com/google/gson/internal/sql/SqlTypesSupport.java
Marcono1234, line 1: Can you please apply the other `SqlTypesSupport` related changes from #2864 as well?
- small Javadoc change to mention `SQL_TYPE_FACTORIES`
- making the `DATE_FACTORY`, ... fields package-private
- extending the `SqlTypesSupportTest`

Marcono1234 (commented):
Thanks a lot!

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
MukjepScarlet: Reasonable I think.

gson/src/main/java/com/google/gson/internal/sql/SqlTypesSupport.java
MukjepScarlet: I'm afraid that package-private of existing factories will break some compatibility...

MukjepScarlet: Other changes are commited

gson/src/test/java/com/google/gson/internal/sql/SqlTypesSupportTest.java
```diff
@@ -31,5 +31,8 @@ public void testSupported() {
     assertThat(SqlTypesSupport.DATE_FACTORY).isNotNull();
     assertThat(SqlTypesSupport.TIME_FACTORY).isNotNull();
     assertThat(SqlTypesSupport.TIMESTAMP_FACTORY).isNotNull();
+
+    // DATE_FACTORY, TIME_FACTORY, TIMESTAMP_FACTORY
+    assertThat(SqlTypesSupport.SQL_TYPE_FACTORIES.size()).isEqualTo(3);
```
Marcono1234: I think you could use `assertThat(...).hasSize(3)` here instead, for a better assertion message in case of a failure.

But is there a reason why you are not using `containsExactly(...).inOrder()` as in the original PR? To verify that it indeed contains the expected factories.
https://github.com/google/gson/blob/5b04e3cc034bf4a1377bf6a98eee8eec6190c1af/gson/src/test/java/com/google/gson/internal/sql/SqlTypesSupportTest.java#L34-L39

MukjepScarlet: I just haven't seen that, lol
btw I don't know this test framework before.

MukjepScarlet: I just haven't seen that, lol
btw I don't know this test framework before.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -374,51 +425,104 @@ public void write(JsonWriter out, Number value) throws IOException {
...
+    }
+
+    @Override
+    public void write(JsonWriter out, Number value) throws IOException {
+      if (value == null) {
+        out.nullValue();
+        return;
+      }
+      double doubleValue = value.doubleValue();
+      if (strict) {
```
merola90, line 506: I just want to point out the minor performance loss. The branching statement here is now dealt every single time the method `write()` gets called. Which can happen quite frequently. Previously, there were two alternative implementations of the adapter where the two different branches were inlined.

MukjepScarlet: Add two different implementation? (Or even constant, because it's enumerable as true/false)

eamonnmcmanus: I think it's unlikely that any performance difference is observable. I'd expect the branch cost to be negligible compared to the cost of converting the `double` to a `String`.

PR #2953, opened 2025-11-27
Update copyright year and owner in LICENSE file

mrshaikhmuhammad:
### Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2025 Google
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


Marcono1234:
Thanks for this PR!

My understanding is that the current `LICENSE` file is correct. The part you changed describes how to apply the license to a specific file, and only then the placeholder should be replaced.
It seems this topic also came up in other projects before, e.g. https://github.com/CycloneDX/cyclonedx-node-module/pull/245 or https://github.com/swagger-api/swagger-ui/issues/6635 which reverted back to using `[yyyy] [name of copyright owner]`.

https://www.apache.org/licenses/LICENSE-2.0#apply also says:
> Include a copy of the Apache License, typically in a file called LICENSE, in your work
> [...]
> To apply the Apache License to specific files in your work, attach the following boilerplate declaration, replacing the fields enclosed by brackets "[]" with your own identifying information.

So an (unmodified) copy of `LICENSE` should be added, and only when applying the license to a file the placeholder should be replaced.

PR #2959, opened 2025-12-03
Add default capacity to EnumTypeAdapter maps

MukjepScarlet:
### Purpose

To avoid re-hashing and re-allocating.

### Description

Part of #2864 

### Checklist

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
@@ -50,11 +50,11 @@ class EnumTypeAdapter<T extends Enum<T>> extends TypeAdapter<T> {
         }
       };
 
-  private static final float LOAD_FACTOR_FOR_MAPS = 0.75F;
-
-  /** Taken from Java 19 method {@link HashMap.newHashMap}. */
-  private static int calculateHashMapCapacity(int numMappings, float loadFactor) {
-    return (int) Math.ceil(numMappings / loadFactor);
+  /**
+   * Taken from Java 19 method {@link HashMap.newHashMap}, using default load factor {@code 0.75F}.
+   */
+  private static int calculateHashMapCapacity(int numMappings) {
+    return (int) Math.ceil(numMappings / 0.75F);
   }
 
   private final Map<String, T> nameToConstant;
@@ -78,11 +78,11 @@ class EnumTypeAdapter<T extends Enum<T>> extends TypeAdapter<T> {
       // one declared field which is not an enum constant, namely the implicit $VALUES array
       fields = Arrays.copyOf(fields, constantCount);
 
-      int hashMapCapacity = calculateHashMapCapacity(constantCount, LOAD_FACTOR_FOR_MAPS);
-      nameToConstant = new HashMap<>(hashMapCapacity, LOAD_FACTOR_FOR_MAPS);
-      stringToConstant = new HashMap<>(hashMapCapacity, LOAD_FACTOR_FOR_MAPS);
+      int hashMapCapacity = calculateHashMapCapacity(constantCount);
+      nameToConstant = new HashMap<>(hashMapCapacity);
+      stringToConstant = new HashMap<>(hashMapCapacity);
       // Don't use `EnumMap` here; that can break when using ProGuard or R8
-      constantToName = new HashMap<>(hashMapCapacity, LOAD_FACTOR_FOR_MAPS);
+      constantToName = new HashMap<>(hashMapCapacity);
 
       AccessibleObject.setAccessible(fields, true);
 


MukjepScarlet:
@Marcono1234 Here

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -69,7 +69,12 @@ private EnumTypeAdapter(Class<T> classOfT) {
 
       // Trim the array to the new length. Every enum type can be expected to have at least
       // one declared field which is not an enum constant, namely the implicit $VALUES array
-      fields = Arrays.copyOf(fields, constantCount);
+      fields = (constantCount == fields.length) ? fields : Arrays.copyOf(fields, constantCount);
+
+      nameToConstant = new HashMap<>(constantCount, 1F);
+      stringToConstant = new HashMap<>(constantCount, 1F);
+      // Don't use `EnumMap` here; that can break when using ProGuard or R8
+      constantToName = new HashMap<>(constantCount, 1F);
```
Marcono1234: Not sure if 'load factor' `1F` here could decrease performance (and if that would be noticeable)? See `HashMap` documentation:
> As a general rule, the default load factor (.75) offers a good tradeoff between time and space costs. Higher values decrease the space overhead but increase the lookup cost (reflected in most of the operations of the HashMap class, including get and put).

If you want feel free to keep it like this for now and wait for the feedback from the Gson project members.

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -69,7 +69,12 @@ private EnumTypeAdapter(Class<T> classOfT) {
 
       // Trim the array to the new length. Every enum type can be expected to have at least
       // one declared field which is not an enum constant, namely the implicit $VALUES array
-      fields = Arrays.copyOf(fields, constantCount);
+      fields = (constantCount == fields.length) ? fields : Arrays.copyOf(fields, constantCount);
```
merola90: > `(constantCount == fields.length)`

This condition is never true, if I interpret the code comment correctly:

> Every enum type can be expected to have at least one declared field which is not an enum constant, namely the implicit $VALUES array

The above loop filters based on the following condition:

> `f.isEnumConstant()`

If I'm mistaken, consider updating the code comment.

Marcono1234: Good point, maybe the `constantCount == fields.length` check is really not worth it in that case.

Or if the check is kept, then maybe the comment should be reworded to say something like:
```java
// Trim the array to the new length, if necessary. Enum types may have declared fields which
// are not an enum constant, namely the implicit $VALUES array
```
(after all the creation of the `$VALUES` field is an implementation detail of the compiler)

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
eamonnmcmanus: Yes, if you want to presize the map to the known number of entries then you need to do what [`HashMap.newHashMap`](https://github.com/openjdk/jdk/blob/659b53fe33eaa531bca1951a26f357b51902311e/src/java.base/share/classes/java/util/HashMap.java#L2581) does:

```java
    static int calculateHashMapCapacity(int numMappings) {
        return (int) Math.ceil(numMappings / (double) DEFAULT_LOAD_FACTOR);
    }
```

You can't use `newHashMap` itself because that isn't available on the older JDK versions that we still target. Guava's `Maps.newHashMapWithExpectedSize` leads to an [extensive description](https://github.com/google/guava/blob/c0535f5299945d593ff84e592eb063570a2fd9a2/guava/src/com/google/common/collect/Maps.java#L267-L278) of the behaviour here.

I'm not totally convinced this is worthwhile, but perhaps it could make a difference when someone is serializing or deserializing a lot of enums.

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
eamonnmcmanus: I would be in favour of removing the check here and just calling `Arrays.copyOf` always.

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
eamonnmcmanus: (I realize now that I was commenting on an old version.)

MukjepScarlet:
The documentation still needs adjustment I think.

gson/src/main/java/com/google/gson/internal/bind/EnumTypeAdapter.java
```diff
@@ -50,9 +50,16 @@ public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> typeToken) {
         }
       };
 
-  private final Map<String, T> nameToConstant = new HashMap<>();
-  private final Map<String, T> stringToConstant = new HashMap<>();
-  private final Map<T, String> constantToName = new HashMap<>();
+  private static final float LOAD_FACTOR_FOR_MAPS = 0.75F;
+
+  /** Taken from Java 19 method {@link HashMap.newHashMap}. */
+  private static int calculateHashMapCapacity(int numMappings, float loadFactor) {
```
eamonnmcmanus: This seems unnecessarily general. We're only ever calling it with 0.75F, so I would be in favour of just inlining that value here rather than passing it as a parameter. I note that [that is what the JDK method does](https://github.com/openjdk/jdk/blob/336894857bfc9f610da55e6180dd7b668bf67752/src/java.base/share/classes/java/util/HashMap.java#L2565).

PR #2969, opened 2025-12-24
Update `test-graal-native-image`.

eamonnmcmanus:
With the latest dependencies, this was failing with this message:
```
Error: Classes that should be initialized at run time got initialized during image building:
 org.junit.jupiter.engine.discovery.MethodSegmentResolver was unintentionally initialized at build time.
```

I don't know enough about GraalVM to be sure of what the best fix is, but Gemini suggested the thing I'm doing here. It seems to work.

### DIFF
--- a/test-graal-native-image/pom.xml
+++ b/test-graal-native-image/pom.xml
@@ -39,7 +39,7 @@
       <dependency>
         <groupId>org.junit</groupId>
         <artifactId>junit-bom</artifactId>
-        <version>6.0.0</version>
+        <version>6.0.1</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -137,7 +137,7 @@
           <plugin>
             <groupId>org.graalvm.buildtools</groupId>
             <artifactId>native-maven-plugin</artifactId>
-            <version>0.11.2</version>
+            <version>0.11.3</version>
             <extensions>true</extensions>
             <executions>
               <execution>
@@ -150,6 +150,7 @@
                   <buildArgs>
                     <!-- Show stack traces to make troubleshooting build issues easier -->
                     <buildArg>-H:+ReportExceptionStackTraces</buildArg>
+                    <buildArg>--initialize-at-build-time=org.junit.jupiter.engine.discovery.MethodSegmentResolver</buildArg>
                   </buildArgs>
                 </configuration>
               </execution>


cpovirk (approved):
I was curious enough to determine that various JUnit files, such as `~/.m2/repository/org/junit/jupiter/junit-jupiter-engine/5.12.1/junit-jupiter-engine-5.12.1.jar`, have `META-INF/native-image/org.junit.jupiter/junit-jupiter-engine/native-image.properties` files that include a bunch of `--initialize-at-build-time` values. It's plausible that something here should actually be fixed on the JUnit side, or it's possible that Maven is mixing up our dependency versions somehow despite the BOM.

I did dig up this, which sure looks related: https://github.com/junit-team/junit-framework/commit/09d1788e8fa4217f8a15b557a17fc3bb2eb49ae9

But the lazy way to interpret that is that JUnit already knows what to do and that we're just following along, so let's go with that!

cpovirk:
(I see the comment reference to https://github.com/graalvm/native-build-tools/issues/706, which seems to be about duplicate classes, so maybe we're just getting an old GraalVM copy of a class and a new JUnit-proper copy or similar.)

eamonnmcmanus:
It looks like @Marcono1234 may have a better handle on what's going on here, and may be able to propose a better fix. I'll go with this one for now since it does allow us to update the dependencies.

Marcono1234:
I had mentioned it in https://github.com/google/gson/pull/2957#issuecomment-3596271679, but maybe you accidentally overlooked it. The proper fix has a pending open PR in graalvm/native-build-tools but this workaround here for Gson is also the official recommendation, see https://github.com/junit-team/junit-framework/issues/5134#issuecomment-3480864832.

It appears that brittle hardcoded list of classes in graalvm/native-build-tools is also only needed for JDK <= 21, see https://github.com/junit-team/junit-framework/issues/5134#issuecomment-3481992112, which we still include in the test matrix though.

In the past coordination between the junit-framework and native-build-tools maintainers worked fine, which is why we had not seen an issue like this before I think. It seems in this case this incompatibility was overlooked by accident before the JUnit release though, see https://github.com/junit-team/junit-framework/issues/5134#issuecomment-3486872002.

But I must admit that I am not that familiar with the details of the native-maven-plugin.

PR #2973, opened 2026-01-22
Add `JsonArray#collector` for `Stream<@Nullable JsonElement>`

MukjepScarlet:
For Java 8.

`JsonObject#collector` will be added later if this is accepted.

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -52,7 +52,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
           });
 
   /**
-   * Returns a collector that accumulates {@link JsonElement}s into a {@code JsonArray}.
+   * Returns a {@link Collector} that accumulates {@link JsonElement}s into a {@code JsonArray}.
    *
    * @since $next-version$
    */
--- a/gson/src/test/java/com/google/gson/JsonArrayTest.java
+++ b/gson/src/test/java/com/google/gson/JsonArrayTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertThrows;
 import com.google.common.testing.EqualsTester;
 import com.google.gson.common.MoreAsserts;
 import java.math.BigInteger;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.junit.Test;
 


gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -354,4 +355,31 @@ public void testToString() {
...
+    assertThat(array.get(3).isJsonObject()).isTrue();
+  }
+
+  @Test
+  public void testCollectorParallel() {
+    JsonArray array =
+        Stream.iterate(0, i -> i + 1)
+            .limit(20)
+            .parallel()
+            .map(JsonPrimitive::new)
```
Marcono1234: Could probably use `IntStream` here:

```suggestion
        IntStream.range(0, 20)
            .parallel()
            .mapToObj(JsonPrimitive::new)
```

gson/src/test/java/com/google/gson/JsonArrayTest.java
```diff
@@ -354,4 +355,31 @@ public void testToString() {
...
+            .collect(JsonArray.collector());
+
+    assertThat(array).hasSize(20);
+    for (int i = 0; i < 20; i++) {
```
Marcono1234: Instead of repeating the value 20 here, check `size()` to be less error-prone in case the test is changed in the future?
```suggestion
    for (int i = 0; i < array.size(); i++) {
```

gson/src/main/java/com/google/gson/JsonArray.java
```diff
@@ -40,6 +42,24 @@
...
+      Collector.of(
+          JsonArray::new,
+          JsonArray::add,
+          (left, right) -> {
+            left.addAll(right);
+            return left;
+          });
+
+  /**
+   * Returns a collector that accumulates {@link JsonElement}s into a {@code JsonArray}.
```
Marcono1234: Could maybe mention `Stream` here to make the purpose of this more obvious for users who might not be that familiar with the `java.util.stream` API?
For example something like this?

```suggestion
   * Returns a {@link Stream} collector that accumulates {@link JsonElement}s into a {@code JsonArray}.
```

What do you think?

MukjepScarlet: I prefer `Returns a {@link Collector} that accumulates {@link JsonElement}s into a {@code JsonArray}.`

MukjepScarlet:
I think if the user is using Java 8+ (Android API 24+), they will definitely import Stream classes somewhere (unless pure Kotlin/Scala), so this could be not a big deal.
By the way in Kotlin I have not found a good way to do similar thing because `JsonArray` doesn't implements `Collection`, so I can't use like `mapTo(JsonArray())`.
I have this:
```kotlin
fun <T> Iterable<T>.mapToJsonArray(transform: (T) -> JsonElement?): JsonArray {
  val arr = if (this is Collection<*>) JsonArray(this.size) else JsonArray()
  for (it in this) arr.add(transform(it))
  return arr
}
```

PR #2982, opened 2026-02-15
Improve `JsonArrayAsListTest#testContains`

Marcono1234:
Follow-up for #2981, which removed `assertThat(list).doesNotContain(null)`

Reverts some previous changes to this method which accidentally converted the assertions to `assertThat(...).contains(...)`.


eamonnmcmanus:
Makes sense. I think in practice `assertThat(list).contains(foo)` _will_ in fact call `list.contains(foo)`, but it's better to be explicit.

eamonnmcmanus (approved):

Issue #2863, opened 2025-06-23
Add Gson.DEFAULT constant


MukjepScarlet:
# Problem solved by the feature

No actual problem

# Feature description

This is in order to provide a DEFAULT **singleton** instance for `Gson`. I saw so many programmers write `new Gson()` anywhere. We might directly provide a default constant to prevent this. Maybe also make `Gson.<init>()` deprecated because we can use `Gson.DEFAULT` instead.

# Alternatives / workarounds

No actual bug

Marcono1234:
You are right that it is preferable to not always create a new `Gson` instance because it internally has a `typeTokenCache` and therefore subsequent usage will be faster.

However, I am not sure if one single `Gson` instance for the complete JVM is ideal: It might lead to higher contention between threads, and possibly completely unrelated dependencies of an application would then share the same `typeTokenCache`. I am afraid this might not only decrease lookup time because the classes used by one dependency (especially for the reflective adapter) might be irrelevant for another dependency, but it can also lead to memory leaks because `typeTokenCache` has 'strong' instead of 'weak' references[^1].

This is only my personal opinion on this though.

[^1]: For separate, per-dependency, `Gson` instances this might not be a big problem because they are then only referenced within the dependency and garbage collected if the objects or classes of that dependency become unreachable.

Marcono1234:
We can definitely improve the Javadoc and User Guide though, mentioning that reusing a `Gson` instance leads to better performance because previously created type adapters are cached.

PR #2983, opened 2026-02-20
Mention in `Gson` Javadoc that reusing instance can improve performance

Marcono1234:
Follow-up for #2863:
Extend the `Gson` Javadoc to mention that type adapters are cached and reusing a `Gson` instance can improve performance.

(Also formats a few "Gson" mentions as `Gson`, where they refer to the class and not the project.)


gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -47,14 +47,16 @@
 import java.util.concurrent.ConcurrentMap;
 
 /**
- * This is the main class for using Gson. Gson is typically used by first constructing a Gson
- * instance and then invoking {@link #toJson(Object)} or {@link #fromJson(String, Class)} methods on
- * it. Gson instances are Thread-safe so you can reuse them freely across multiple threads.
+ * This is the main class for using Gson. Gson is typically used by first constructing a {@code
+ * Gson} instance and then invoking {@link #toJson(Object)} or {@link #fromJson(String, Class)}
+ * methods on it. {@code Gson} instances are Thread-safe so you can reuse them freely across
```
eamonnmcmanus: Could you remove the distracting capitalization of "Thread-safe" while you are here?

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -47,14 +47,16 @@
...
+ * multiple threads.
  *
- * <p>You can create a Gson instance by invoking {@code new Gson()} if the default configuration is
- * all you need. You can also use {@link GsonBuilder} to build a Gson instance with various
- * configuration options such as versioning support, pretty printing, custom newline, custom indent,
- * custom {@link JsonSerializer}s, {@link JsonDeserializer}s, and {@link InstanceCreator}s.
+ * <p>You can create a {@code Gson} instance by invoking {@link #Gson() new Gson()} if the default
+ * configuration is all you need. You can also use {@link GsonBuilder} to build a {@code Gson}
+ * instance with various configuration options such as versioning support, pretty printing, custom
+ * newline, custom indent, custom {@link JsonSerializer}s, {@link JsonDeserializer}s, and {@link
```
eamonnmcmanus: Perhaps rewrite this last bit to talk about `{@link TypeAdapter}s` and `{@link TypeAdapterFactory}s`.

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/Gson.java
Marcono1234: Do you mean mentioning them additionally, or instead of `JsonSerializer` and `JsonDeserializer`?

eamonnmcmanus: My thinking was that we encourage people to use `TypeAdapter[Factory]` rather than `JsonSerializer` and `JsonDeserializer`. (Though I will admit that `TypeAdapter` is _much_ more annoying to use for deserialization than `JsonDeserializer`, even if it _is_ potentially more efficient.) So at least additionally, and perhaps instead of.

Marcono1234: Have reworded and extended that section now. Please let me know in case the new text is too extensive or if you have other suggestions for how to improve it.

eamonnmcmanus (approved):
This is great! Thanks for taking the trouble.

PR #2988, opened 2026-03-05
Add boundary tests for JsonTreeWriter.endArray() error handling

lmj798:
## Summary
Add 3 boundary tests for `JsonTreeWriter.endArray()` to verify proper exception throwing when called in invalid states:
1. `testEndArrayOnEmptyStackThrows` - Verifies IllegalStateException when calling endArray() on empty stack
2. `testEndArrayWithPendingNameThrows` - Verifies IllegalStateException when a pending name exists (object not closed)
3. `testEndArrayWhenStackTopIsNotArrayThrows` - Verifies IllegalStateException when stack top is an object, not an array
## Why
Current tests cover:
- Normal array/object begin/end operations
- Close after write
- Premature close
- Null handling
- NaN/Infinity handling
However, they don't explicitly test `endArray()` error conditions for invalid state detection. These boundary cases ensure proper state validation and prevent regressions from state management changes.
## Verification
mvn -Dtest=JsonTreeWriterTest test
Result: Tests run: 25, Failures: 0, Errors: 0, Skipped: 0 - BUILD SUCCESS

### DIFF
--- a/gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
@@ -100,7 +100,7 @@ public final class JsonTreeWriterTest {
     JsonTreeWriter writer = new JsonTreeWriter();
     writer.setStrictness(Strictness.LENIENT);
     writer.beginArray();
-    IOException e = assertThrows(IOException.class, () -> writer.close());
+    var e = assertThrows(IOException.class, () -> writer.close());
     assertThat(e).hasMessageThat().isEqualTo("Incomplete document");
   }
 


gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
```diff
@@ -100,7 +100,7 @@ public void testPrematureClose() throws Exception {
     JsonTreeWriter writer = new JsonTreeWriter();
     writer.setStrictness(Strictness.LENIENT);
     writer.beginArray();
-    var e = assertThrows(IOException.class, () -> writer.close());
+    IOException e = assertThrows(IOException.class, () -> writer.close());
```
eamonnmcmanus: Is there a reason for this change?

eamonnmcmanus (changes requested):
This seems like a reasonable change. Just one question.

gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
lmj798: I changed it back to explicit typing to keep the coding style consistent with our main source code (which is Java 8). Mixing var in tests while strictly using explicit types in production code can be inconsistent.

Marcono1234: The test code intentionally uses a higher target Java version (11) to make it more convenient and succinct to write. So this inconsistency might not be a problem?

Marcono1234:
Could you please move your last commit (https://github.com/google/gson/pull/2988/changes/073d0f5f2fb364e572e18793c1f2ac1077cec9c0) to a separate PR? It is unrelated to `JsonTreeWriter.endArray()`.

lmj798:
> Could you please move your last commit ([073d0f5](https://github.com/google/gson/commit/073d0f5f2fb364e572e18793c1f2ac1077cec9c0)) to a separate PR? It is unrelated to `JsonTreeWriter.endArray()`.

Sorry about that, and thank you for catching it.
You’re absolutely right that commit 073d0f5 was unrelated. I’ve moved it to a separate PR and force-pushed this branch, so this PR now only contains the JsonTreeWriter.endArray() changes.

PR #2998, opened 2026-03-26
fix: Preserve integral Number type in toJsonTree for byte/short/int

### LINKED ISSUE #2680: TypeAdapterFactory INTEGER_FACTORY will change INTEGER type to LONG type

daguimu:
## Problem

When using `toJsonTree` with `byte`, `short`, or `int` types, the resulting `JsonPrimitive` stores a `Long` instead of the expected `Byte`, `Short`, or `Integer`. This is a regression introduced by commit 3e3266cf.

## Root Cause

The `BYTE`, `SHORT`, and `INTEGER` type adapters call `out.value(value.byteValue())`, `out.value(value.shortValue())`, and `out.value(value.intValue())` respectively. Since `JsonWriter` only has `value(long)` (no `value(int)`), Java auto-widens the primitive to `long`. In `JsonTreeWriter`, `value(long)` creates `new JsonPrimitive(long)`, storing a `Long` regardless of the original type.

## Fix

Cast the narrowed primitive to `Number` before passing to `value()`. This triggers autoboxing to `Byte`/`Short`/`Integer` and causes Java to resolve `value(Number)` instead of `value(long)`. `JsonTreeWriter.value(Number)` preserves the actual Number type in the `JsonPrimitive`. For regular `JsonWriter`, `value(Number)` produces identical text output.

## Tests Added

- `testToJsonTreePreservesIntegralType` — verifies that `toJsonTree` preserves:
  - `Byte` type for `byte`/`Byte.class`
  - `Short` type for `short`/`Short.class`
  - `Integer` type for `int`/`Integer.class`
  - `Long` type for `long`/`Long.class` (unchanged, regression guard)
  - Correct target type after narrowing conversion (e.g., `Double` → `Integer`, `Long` → `Byte`)

## Impact

Only affects code paths using `JsonTreeWriter` (i.e., `toJsonTree`). Regular JSON string serialization via `toJson` is unaffected since `value(Number)` and `value(long)` produce the same text output. All 4587 existing tests pass.

Fixes #2680

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -214,7 +214,7 @@ public final class TypeAdapters {
           if (value == null) {
             out.nullValue();
           } else {
-            out.value(value.byteValue());
+            out.value((Number) value.byteValue());
           }
         }
       };
@@ -249,7 +249,7 @@ public final class TypeAdapters {
           if (value == null) {
             out.nullValue();
           } else {
-            out.value(value.shortValue());
+            out.value((Number) value.shortValue());
           }
         }
       };
@@ -277,7 +277,7 @@ public final class TypeAdapters {
           if (value == null) {
             out.nullValue();
           } else {
-            out.value(value.intValue());
+            out.value((Number) value.intValue());
           }
         }
       };
--- a/gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
+++ b/gson/src/test/java/com/google/gson/functional/PrimitiveTest.java
@@ -161,6 +161,33 @@ public class PrimitiveTest {
     assertThat(gson.toJson(1.5, Integer.class)).isEqualTo("1");
   }
 
+  /** Regression test for https://github.com/google/gson/issues/2680 */
+  @Test
+  public void testToJsonTreePreservesIntegralType() {
+    // toJsonTree should preserve the Number type, not widen byte/short/int to Long
+    assertThat(gson.toJsonTree((byte) 1, byte.class).getAsNumber().getClass())
+        .isEqualTo(Byte.class);
+    assertThat(gson.toJsonTree((byte) 1, Byte.class).getAsNumber().getClass())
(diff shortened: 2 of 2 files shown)


Marcono1234:
Thanks for the PR, for what it's worth, there is already an existing (though by now slightly outdated) PR: #2814

I guess the maintainers will decide which one they prefer. So feel free to keep your PR here open.

eamonnmcmanus:
> Thanks for the PR, for what it's worth, there is already an existing (though by now slightly outdated) PR: #2814
> 
> I guess the maintainers will decide which one they prefer. So feel free to keep your PR here open.

Sorry, I'm not sure why we didn't get around to merging #2814. If that one and this are functionally identical (are they?) then we might go with this one just because #2814 would have to be brought up to date.

I'm testing it against all of Google's internal tests, which will take a little while.

eamonnmcmanus (approved):

Marcono1234:
#2814 also tried to keep reusing the existing `Number` object if it already has the correct type, e.g. use `Integer` as is instead of doing `(Number) integer.intValue()`. Not sure if that provides any performance improvement though. So I guess it can be omitted.

Marcono1234:
The other alternative would be to revert https://github.com/google/gson/commit/3e3266cf48f132928225e1561a6ae4cb5503d08f (as mentioned also in #2814); maybe that change was not worth it, and would have only mattered for rare corner cases.

PR #3000, opened 2026-03-26
fix: Include location info in NumberFormatException from JsonReader

### LINKED ISSUE #1564: JsonReader throws NumberFormatException which does not contain location info

daguimu:
## Problem

`JsonReader.nextDouble()`, `nextInt()`, and `nextLong()` call `Double.parseDouble()` on peeked strings. When parsing fails (e.g., for empty strings like `""`), the `NumberFormatException` thrown by `Double.parseDouble()` contains no JSON location information (line, column, path). This makes it difficult to locate the problematic value in the input JSON.

For example, parsing `{ x: ''}` as an int produces `NumberFormatException: empty String` with no indication of where in the JSON the error occurred.

## Root Cause

The three methods had a comment `// don't catch this NumberFormatException.` and let the exception from `Double.parseDouble()` propagate directly. Other `NumberFormatException` throws in the same methods already include location information via `locationString()`.

## Fix

Catch the `NumberFormatException` from `Double.parseDouble()` and rethrow with a descriptive message that includes the location string, consistent with other error messages in `JsonReader`. The message format follows the existing pattern: `"Expected a <type> but was <value> at line X column Y path $.<path>"`.

## Tests Added

- `testNextDoubleNumberFormatExceptionContainsLocation` — verifies location in path `$[0]`
- `testNextIntNumberFormatExceptionContainsLocation` — verifies location in path `$.x`
- `testNextLongNumberFormatExceptionContainsLocation` — verifies location in path `$.y`

## Impact

Only affects the error path when `Double.parseDouble()` fails on an invalid numeric string. The exception type (`NumberFormatException`) is unchanged. All 4589 existing tests pass.

Fixes #1564

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -1055,8 +1055,7 @@ public class JsonReader implements Closeable {
       result = Double.parseDouble(peekedString);
     } catch (NumberFormatException e) {
       NumberFormatException rethrown =
-          new NumberFormatException(
-              "Expected a double but was " + peekedString + locationString());
+          new NumberFormatException("Expected a double but was " + peekedString + locationString());
       rethrown.initCause(e);
       throw rethrown;
     }
@@ -1118,8 +1117,7 @@ public class JsonReader implements Closeable {
       asDouble = Double.parseDouble(peekedString);
     } catch (NumberFormatException e) {
       NumberFormatException rethrown =
-          new NumberFormatException(
-              "Expected a long but was " + peekedString + locationString());
+          new NumberFormatException("Expected a long but was " + peekedString + locationString());
       rethrown.initCause(e);
       throw rethrown;
     }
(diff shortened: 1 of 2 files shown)


eamonnmcmanus (approved):
Thanks, I think this will be helpful for users!

PR #3011, opened 2026-04-10
Remove `Serializable` from internal `Type` implementation classes.

eamonnmcmanus:
The nested classes `ParameterizedTypeImpl`, `GenericArrayTypeImpl`, and
`WildcardTypeImpl` in `GsonTypes` are implementations of the
corresponding types (without `Impl`) in `java.lang.reflect`. For some
reason, they have always implemented `Serializable`. They are even
documented that way, though the documentation is not published since
`GsonTypes` is an internal implementation class.

The implementations of these interfaces that are returned by
`java.lang.reflect` methods such as `Method.getGenericReturnType()` do
*not* implement `Serializable` so it seems gratuitous for Gson's
implementations to do so. Additionally, `Serializable` classes can
potentially participate in exploits. I think it is highly unlikely that
there is any practical exploit using these classes, but if we can avoid
even having to ask the question then we should.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/GsonTypes.java
+++ b/gson/src/main/java/com/google/gson/internal/GsonTypes.java
@@ -576,12 +576,9 @@ public final class GsonTypes {
       }
       return stringBuilder.append(">").toString();
     }
-
-    private static final long serialVersionUID = 0;
   }
 
   private static final class GenericArrayTypeImpl implements GenericArrayType {
-    @SuppressWarnings("serial")
     private final Type componentType;
 
     GenericArrayTypeImpl(Type componentType) {
@@ -608,8 +605,6 @@ public final class GsonTypes {
     public String toString() {
       return typeToString(componentType) + "[]";
     }
-
-    private static final long serialVersionUID = 0;
   }
 
   /**
@@ -619,10 +614,8 @@ public final class GsonTypes {
    * be Object.class.
    */
   private static final class WildcardTypeImpl implements WildcardType {
-    @SuppressWarnings("serial")
     private final Type upperBound;
 
-    @SuppressWarnings("serial")
     private final Type lowerBound;
 
     WildcardTypeImpl(Type[] upperBounds, Type[] lowerBounds) {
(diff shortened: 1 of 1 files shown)


cpovirk (approved):
I spent a short time but probably still more that was really justified to trace this:

- https://github.com/google/gson/commit/883ce465d5514ee9fd28a73c0891869d0b744d4a
- https://github.com/google/guice/commit/3beaaaff52598e849659281fed35dc29a221fac4

That didn't really answer my question, but it's not like I expect serialization to turn out to be necessary. (I notice that [the JDK implementation](https://github.com/openjdk/jdk/blob/aa5677afcbb4c42248168c364889cbd910327c0c/src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/ParameterizedTypeImpl.java#L38) isn't `Serializable`, either, at least for `ParameterizedType`, which was the only one that I checked.)

eamonnmcmanus:
> (I notice that [the JDK implementation](https://github.com/openjdk/jdk/blob/aa5677afcbb4c42248168c364889cbd910327c0c/src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/ParameterizedTypeImpl.java#L38) isn't Serializable, either, at least for ParameterizedType, which was the only one that I checked.)

Right, I mentioned that in the PR description, which is one reason why I feel pretty confident in making this change. (I also ran this change against all of Googe's internal tests and found no issues.)

cpovirk:
Oops, sorry, I must have clicked the link to jump straight to the diffs :)

gson/src/main/java/com/google/gson/internal/GsonTypes.java
Marcono1234, line 123: Should this comment be removed as well (or adjusted, to mention that type is unsupported)?

eamonnmcmanus: Indeed! Feel free to send a PR. :-)

Marcono1234: Have created #3012; I hope the wording is ok. Please let me know if you want it changed, or want the comment to be removed completely.

PR #3013, opened 2026-04-15
Fix: reject trailing data after top-level JSON null

Senrian:
## Summary

Fixes https://github.com/google/gson/issues/3008

`Gson.fromJson(String/Reader, ...)` and `JsonParser.parseReader(Reader)` documentedly reject trailing data, but both made an exception when the parsed top-level value was JSON `null`.

For non-null values, trailing data already causes a `JsonSyntaxException`. For top-level `null`, the trailing-data check was skipped.

### Gson.java
```java
// Before:
if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {

// After:
if (obj == null || reader.peek() != JsonToken.END_DOCUMENT) {
```

### JsonParser.java
```java
// Before:
if (!element.isJsonNull() && jsonReader.peek() != JsonToken.END_DOCUMENT) {

// After:
if (jsonReader.peek() != JsonToken.END_DOCUMENT) {
```

## Tests

Added `testParseTrailingDataAfterNull()` and `testFromJsonTrailingDataAfterNull()` in `JsonParserTest.java` to verify trailing data is rejected after parsing top-level null.

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1195,7 +1195,7 @@ public final class Gson {
 
   private static void assertFullConsumption(Object obj, JsonReader reader) {
     try {
-      if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {
+      if (obj == null || reader.peek() != JsonToken.END_DOCUMENT) {
         throw new JsonSyntaxException("JSON document was not fully consumed.");
       }
     } catch (MalformedJsonException e) {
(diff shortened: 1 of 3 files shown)


Marcono1234:
Please explain how your PR differs from or is better than PR #3008 (which you claim your PR "fixes").

In particular it seems your PR has two problems:
- your change to use `obj == null` now fails even if JSON `null` has no trailing data
- your test which excepts `JsonParser.parseString("nulltrailing")` to fail most likely does not work because Gson instead parses that as unquoted string (I assume)

PR #3021, opened 2026-04-29
docs: Fix copy-paste @return in JsonArray.getAsCharacter javadoc

daguimu:
## Problem

The `@return` tag on `JsonArray.getAsCharacter()` reads:

```
@return this element as a primitive short if it is single element array.
```

This is a copy-paste from `getAsShort()` and is wrong. The method returns a `char`, not a primitive short, and every other line of the same javadoc block (the description, the `@deprecated` note, and the method signature) talks about a character.

## Root Cause

When the per-method getters were templated, the `getAsCharacter` block reused the `@return` text from `getAsShort` without updating the type description.

## Fix

Replace `primitive short` with `character` in the `@return` tag of `JsonArray.getAsCharacter`. No behavior change; javadoc only.

## Tests

No new tests are required because this is a documentation-only fix. The full test suite still passes:

```
mvn -pl gson test
...
Tests run: 4592, Failures: 0, Errors: 0, Skipped: 19
BUILD SUCCESS
```

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonArray.java
+++ b/gson/src/main/java/com/google/gson/JsonArray.java
@@ -364,7 +364,7 @@ public final class JsonArray extends JsonElement implements Iterable<JsonElement
    * method calls {@link JsonElement#getAsCharacter()} on the element, therefore any of the
    * exceptions declared by that method can occur.
    *
-   * @return this element as a primitive short if it is single element array.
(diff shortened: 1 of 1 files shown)


eamonnmcmanus (approved):
Thanks! This method is deprecated anyway, but we can still fix this incorrect wording.

PR #3026, opened 2026-05-11
Remove obsolete native-maven-plugin JUnit workaround

Marcono1234:
[Version 1.1.0](https://github.com/graalvm/native-build-tools/releases/tag/1.1.0) (which is used since #3032) added the missing built-in configuration for `--initialize-at-build-time=...MethodSegmentResolver` which was added as workaround in #2969; therefore I have removed that workaround again.


eamonnmcmanus (approved):
Thanks!

PR #3027, opened 2026-05-27
fix(proto): cap bitFieldXXX_ index to prevent OOM in LegacyProtoTypeAdapterFactory

adilburaksen:
## Summary

`LegacyProtoTypeAdapterFactory.buildPresenceMaskFromBitFields()` reads `bitField<N>_` entries from the JSON stream to reconstruct a presence bitmask. The index `N` was previously used without bound checks: calling `BigInteger.shiftLeft(N * 32)` with a large attacker-controlled `N` allocates `N / 8` bytes on the heap, making it trivial to OOM the JVM with a small JSON payload (e.g. `{"bitField67108863_": 1}`).

## Fix

Compute `maxGroupIndex = ceil(fieldWithPresenceToBitIndex.size() / 32)`, which is the largest bit-field group index that can influence any presence bit actually checked by this adapter. Any `bitField<N>_` entry with `N > maxGroupIndex` is silently skipped — it cannot affect deserialization correctness and only exists to exhaust memory.

## Test

Added `deserialize_oversizedBitFieldIndex_doesNotCauseOom` to `LegacyProtoTypeAdapterFactoryTest`, which feeds a crafted JSON with `bitField67108863_` and asserts that the valid field `optionalInt32_` is still correctly deserialized without any heap pressure.

## Impact

Any application that deserializes untrusted JSON using `LegacyProtoTypeAdapterFactory` is vulnerable. A single JSON field with a large index is enough to trigger the OOM.

### DIFF
--- a/proto/src/main/java/com/google/gson/protobuf/LegacyProtoTypeAdapterFactory.java
+++ b/proto/src/main/java/com/google/gson/protobuf/LegacyProtoTypeAdapterFactory.java
@@ -217,9 +217,18 @@ public enum LegacyProtoTypeAdapterFactory implements TypeAdapterFactory {
         String fieldName = readFieldName(in);
         Matcher matcher = BIT_FIELD_PATTERN.matcher(fieldName);
         if (matcher.matches()) {
-          int shift = Integer.parseInt(matcher.group(1)) * 32;
-          int mask = in.nextInt();
-          presenceBitmask = presenceBitmask.or(BigInteger.valueOf(mask).shiftLeft(shift));
+          int fieldGroupIndex = Integer.parseInt(matcher.group(1));
+          // Guard against OOM: the only bitField groups that matter are those whose bits are
+          // actually tested in fieldWithPresenceToBitIndex. Groups beyond that bound carry no
+          // semantics and, if allowed, could cause extremely large BigInteger allocations.
+          int maxGroupIndex = (fieldWithPresenceToBitIndex.size() + 31) / 32;
+          if (fieldGroupIndex > maxGroupIndex) {
+            in.skipValue();
+          } else {
+            int shift = fieldGroupIndex * 32;
+            int mask = in.nextInt();
+            presenceBitmask = presenceBitmask.or(BigInteger.valueOf(mask).shiftLeft(shift));
+          }
         } else {
           if (fieldName.endsWith("_")) {
             fieldName = fieldName.substring(0, fieldName.length() - 1);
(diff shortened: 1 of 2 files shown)


adilburaksen:
The CodeQL alert is a valid observation — added a defensive try-catch in the latest commit (a04ecd85). In practice, the `BIT_FIELD_PATTERN` regex constrains the captured group to `\d{1,9}`, so `parseInt` can never throw, but making the safety property explicit is better than relying on the regex invariant.

Marcono1234:
Thanks for highlighting this issue and proposing this fix!

I am wondering if instead of using `skipValue()` for these cases it would be better to throw `JsonParseException`. That would also help troubleshooting in case for whatever reason such a JSON property occurs without any malicious intentions.

However, please keep the code as it is for now and let's wait what the maintainers think about this.

adilburaksen:
@Marcono1234 Thanks for the review!

Both approaches work to prevent the OOM. I went with `skipValue()` because it preserves backward compatibility — existing producers that emit large `bitField` indices (e.g. a newer proto schema with many presence fields) would continue to be parsed rather than throwing. The index beyond `maxGroupIndex` carries no semantic meaning for deserialization anyway, so silently ignoring it felt safe.

That said, `JsonParseException` gives a clearer signal for debugging and is arguably the stricter choice. Happy to update if the maintainers prefer that direction.

proto/src/main/java/com/google/gson/protobuf/LegacyProtoTypeAdapterFactory.java
```diff
@@ -217,9 +217,26 @@ public T read(JsonReader in) throws IOException {
         String fieldName = readFieldName(in);
         Matcher matcher = BIT_FIELD_PATTERN.matcher(fieldName);
         if (matcher.matches()) {
-          int shift = Integer.parseInt(matcher.group(1)) * 32;
-          int mask = in.nextInt();
-          presenceBitmask = presenceBitmask.or(BigInteger.valueOf(mask).shiftLeft(shift));
+          int fieldGroupIndex;
+          try {
+            fieldGroupIndex = Integer.parseInt(matcher.group(1));
+          } catch (NumberFormatException e) {
```
eamonnmcmanus: I think it is annoying to have to pander to static analysis like this, and I would be inclined not to do it.

eamonnmcmanus (changes requested):
Thanks, this is an interesting find! I think, though, that a simpler fix would be to change `{1,9}` to `{1,4}` in the regular expression. It seems very unlikely that anyone has a protobuf message with 10,000 fields, and the allocation of a `BigInteger` to hold 9,999 bits is unlikely to provoke an OOM.

I'm always a bit skeptical about issues arising from deserializing adversarial input, especially for niche cases like this, but I'll note that limiting the size of the `BigInteger` also mitigates a potential performance DoS due to the quadratic `BigInteger.or` in the loop.

adilburaksen:
Good call — the regex cap is much cleaner. I went with the dynamic bound initially because I wasn't sure how large the group index could realistically get, but `{1,4}` is plenty generous for any real proto schema.

Updated: replaced the maxGroupIndex guard with just `{1,4}` in BIT_FIELD_PATTERN, and reverted the surrounding code back to the original three lines. Also updated the test comment to reflect that the protection now happens at the pattern-matching level rather than at runtime.

eamonnmcmanus (approved):
Thanks for following through on this!

PR #3028, opened 2026-05-31
Fix #3008: reject trailing data after top-level null

ErnestHysa:
Fixes google/gson#3008.

Both `Gson.fromJson` and `JsonParser.parseReader` accepted trailing data after a top-level null because the trailing-data check was skipped when the parsed element was null.

## Changes

- **Gson.java** `assertFullConsumption()`: Removed the `obj != null` guard so trailing-data validation runs even when the parsed object is null.
- **JsonParser.java** `parseReader()`: Removed the `!element.isJsonNull()` guard so trailing-data validation runs unconditionally.

Now `null  extra` correctly throws `JsonSyntaxException: Did not consume the entire document` (or the equivalent from `Gson.fromJson`).

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1195,7 +1195,7 @@ public final class Gson {
 
   private static void assertFullConsumption(Object obj, JsonReader reader) {
     try {
-      if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {
+      if (reader.peek() != JsonToken.END_DOCUMENT) {
         throw new JsonSyntaxException("JSON document was not fully consumed.");
       }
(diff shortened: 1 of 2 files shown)


Marcono1234:
Can you please explain what the purpose of this PR here is, or the motivation?

#3008 is a _pull request_ not an issue, and it already tries to fix the issue while also handling corner cases and extending test coverage. It just hasn't been merged yet.

ErnestHysa:
@Marcono1234 Fair point — I missed #3008 before opening this. Looking at it now, the patch is essentially the same change to the two call sites, plus it handles the EOFException corner case (empty/whitespace-only input) and adds test coverage. My #3028 is a strict subset of #3008 with no added value, so I'm closing this PR in favor of #3008. Sorry for the noise.

PR #3029, opened 2026-05-31
Fix #2817: Add precision check to JsonTreeReader.nextInt() and nextLong()

### LINKED ISSUE #2817: Inconsistency JsonReader vs JsonTreeReader about double precision loss
Relates #1737 

# Gson version
2.12.1

# Description
JsonReader and JsonTreeReader do not have the same behavior while reading a long (or an int) when the next token is a double.  
JsonReader will fail.  
JsonTreeReader will cast, it should fail also.

```
String json1 = "{\"value\": 42.123}";
JsonReader reader1 = new JsonReader(new StringReader(json1));
reader1.beginObject();
reader1.nextName();
int result1 = reader1.nextInt(); // fails
System.out.println("Result 1 : " + result1);
```
vs
```
JsonObject json2 = new JsonObject();
json2.addProperty("value", 42.123);
JsonTreeReader reader2 = new JsonTreeReader(json2);
reader2.beginObject();
reader2.nextName();
int result2 = reader2.nextInt(); // cast
System.out.println("Result 2 : " + result2); // -> 42
```

ErnestHysa:
## Description

When a double value is stored in a JsonPrimitive, calling  or  should throw  to be consistent with  behavior, instead of silently truncating the value.

### Issue
Issue #2817 reports that  correctly throws when trying to read a double as an int/long, but  silently casts the value.

**JsonReader behavior (correct):**

**JsonTreeReader behavior (buggy):**

## Fix
This fix adds the same precision validation that  has:
-  now checks if the double value can be exactly represented as an int, and throws if precision would be lost
-  similarly checks for precision loss

## Changes
- **JsonTreeReader.java**: Modified  and  to validate that the stored number can be exactly represented as the requested type without precision loss
- **JsonTreeReaderTest.java**: Added two test cases to verify the fix

## Testing
Added unit tests to verify that  and  throw  when the stored value is a double that would lose precision when cast.

Fixes #2817

### DIFF
--- a/gson/src/main/java/com/google/gson/Gson.java
+++ b/gson/src/main/java/com/google/gson/Gson.java
@@ -1195,7 +1195,7 @@ public final class Gson {
 
   private static void assertFullConsumption(Object obj, JsonReader reader) {
     try {
-      if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {
+      if (reader.peek() != JsonToken.END_DOCUMENT) {
         throw new JsonSyntaxException("JSON document was not fully consumed.");
       }
     } catch (MalformedJsonException e) {
--- a/gson/src/main/java/com/google/gson/JsonParser.java
+++ b/gson/src/main/java/com/google/gson/JsonParser.java
@@ -108,7 +108,7 @@ public final class JsonParser {
     try {
       JsonReader jsonReader = new JsonReader(reader);
       JsonElement element = parseReader(jsonReader);
-      if (!element.isJsonNull() && jsonReader.peek() != JsonToken.END_DOCUMENT) {
+      if (jsonReader.peek() != JsonToken.END_DOCUMENT) {
         throw new JsonSyntaxException("Did not consume the entire document.");
       }
       return element;
--- a/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
@@ -258,7 +258,29 @@ public final class JsonTreeReader extends JsonReader {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    long result = ((JsonPrimitive) peekStack()).getAsLong();
+    Object o = peekStack();
+    long result;
+    if (o instanceof JsonPrimitive) {
+      JsonPrimitive primitive = (JsonPrimitive) o;
+      if (primitive.isNumber()) {
+        Number number = primitive.getAsNumber();
+        if (number instanceof Long || number instanceof Integer) {
+          result = number.longValue();
+        } else {
+          // For other number types (e.g., Double, BigDecimal), check for loss of precision
+          double doubleValue = number.doubleValue();
+          result = (long) doubleValue;
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected a long but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Long.parseLong(primitive.getAsString());
+      }
+    } else {
+      throw new IllegalStateException("Unexpected token: " + o);
+    }
     popStack();
     if (stackSize > 0) {
       pathIndices[stackSize - 1]++;
@@ -273,7 +295,29 @@ public final class JsonTreeReader extends JsonReader {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    int result = ((JsonPrimitive) peekStack()).getAsInt();
+    Object o = peekStack();
+    int result;
+    if (o instanceof JsonPrimitive) {
+      JsonPrimitive primitive = (JsonPrimitive) o;
+      if (primitive.isNumber()) {
+        Number number = primitive.getAsNumber();
+        if (number instanceof Integer) {
+          result = number.intValue();
+        } else {
+          // For other number types (e.g., Long, Double, BigDecimal), check for loss of precision
+          double doubleValue = number.doubleValue();
+          result = (int) doubleValue;
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected an int but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Integer.parseInt(primitive.getAsString());
+      }
+    } else {
+      throw new IllegalStateException("Unexpected token: " + o);
+    }
     popStack();
     if (stackSize > 0) {
       pathIndices[stackSize - 1]++;
--- a/gson/src/test/java/com/google/gson/internal/bind/JsonTreeReaderTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/JsonTreeReaderTest.java
@@ -199,4 +199,27 @@ public class JsonTreeReaderTest {
             "getNestingLimit()");
     MoreAsserts.assertOverridesMethods(JsonReader.class, JsonTreeReader.class, ignoredMethods);
   }
+
+  // Tests for issue #2817: Inconsistency JsonReader vs JsonTreeReader about double precision loss
+  @Test
+  public void testNextIntThrowsForDouble() throws IOException {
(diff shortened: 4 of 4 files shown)


ErnestHysa:
CLA check looks like it didn't pick up my signature — I'm re-checking the linked Google account. The fix itself is ready for review on the code side; will update the CLA status as soon as it's cleared on the Google side.

gson/src/main/java/com/google/gson/Gson.java
```diff
@@ -1195,7 +1195,7 @@ public <T> T fromJson(JsonElement json, TypeToken<T> typeOfT) throws JsonSyntaxE
 
   private static void assertFullConsumption(Object obj, JsonReader reader) {
     try {
-      if (obj != null && reader.peek() != JsonToken.END_DOCUMENT) {
+      if (reader.peek() != JsonToken.END_DOCUMENT) {
```
Marcono1234, line 1198: This accidentally includes changes from your other PR?

gson/src/main/java/com/google/gson/JsonParser.java
```diff
@@ -108,7 +108,7 @@ public static JsonElement parseReader(Reader reader) throws JsonIOException, Jso
     try {
       JsonReader jsonReader = new JsonReader(reader);
       JsonElement element = parseReader(jsonReader);
-      if (!element.isJsonNull() && jsonReader.peek() != JsonToken.END_DOCUMENT) {
+      if (jsonReader.peek() != JsonToken.END_DOCUMENT) {
```
Marcono1234, line 111: This accidentally includes changes from your other PR?

eamonnmcmanus:
Thanks for this! It's a behaviour change, but I think `JsonTreeReader` is relatively little used, and this is something of a corner case anyway, so that's probably OK.

Something seems to have happened to the PR description. There's a lot of missing text, "calling or should throw to be consistent with behavior" and the correct and incorrect examples. Can you fix that? Along with the UnusedVariable finding.

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -258,7 +258,29 @@ public long nextLong() throws IOException {
...
+          // For other number types (e.g., Double, BigDecimal), check for loss of precision
+          double doubleValue = number.doubleValue();
+          result = (long) doubleValue;
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected a long but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Long.parseLong(primitive.getAsString());
```
Marcono1234, line 279: Maybe this could instead still call `getAsLong()`?

```suggestion
        result = primitive.getAsLong();
```

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -273,7 +295,29 @@ public int nextInt() throws IOException {
...
+          // For other number types (e.g., Long, Double, BigDecimal), check for loss of precision
+          double doubleValue = number.doubleValue();
+          result = (int) doubleValue;
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected an int but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Integer.parseInt(primitive.getAsString());
```
Marcono1234, line 316: Maybe this could instead still call `getAsInt()`?

```suggestion
        result = primitive.getAsInt();
```

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -258,7 +258,29 @@ public long nextLong() throws IOException {
...
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected a long but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Long.parseLong(primitive.getAsString());
+      }
+    } else {
+      throw new IllegalStateException("Unexpected token: " + o);
```
Marcono1234, line 282: Might be better to use `AssertionError` to make it clear that this should be unreachable.

```suggestion
      throw new AssertionError("Unexpected token: " + o);
```

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -273,7 +295,29 @@ public int nextInt() throws IOException {
...
+          if (result != doubleValue) {
+            throw new NumberFormatException(
+                "Expected an int but was " + number + locationString());
+          }
+        }
+      } else {
+        result = Integer.parseInt(primitive.getAsString());
+      }
+    } else {
+      throw new IllegalStateException("Unexpected token: " + o);
```
Marcono1234, line 319: Might be better to use `AssertionError` to make it clear that this should be unreachable.

```suggestion
      throw new AssertionError("Unexpected token: " + o);
```

Respectively omit the `o instanceof JsonPrimitive` check? Previously it also performed an unchecked `(JsonPrimitive) peekStack()` (which is guarded by the `peek()` check at the beginning).

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -258,7 +258,29 @@ public long nextLong() throws IOException {
...
-    long result = ((JsonPrimitive) peekStack()).getAsLong();
+    Object o = peekStack();
+    long result;
+    if (o instanceof JsonPrimitive) {
+      JsonPrimitive primitive = (JsonPrimitive) o;
+      if (primitive.isNumber()) {
+        Number number = primitive.getAsNumber();
+        if (number instanceof Long || number instanceof Integer) {
+          result = number.longValue();
+        } else {
```
Marcono1234, line 269: This will probably need explicit handling for `BigInteger` and `BigDecimal` (for `nextLong()` only) because values near `Long.MAX_VALUE` cannot be represented exactly by `double`.
So with your changes this would break behavior which was working previously for such values.

Maybe what you could do is:
- For `BigDecimal`
  1. call `BigDecimal#longValueExact()`
  2. catch the `ArithmeticException` and wrap it in a descriptive `NumberFormatException`
- For `BigInteger`
  1. check `bitLength() >= 64` and throw `NumberFormatException` in that case
     Cannot call `BigInteger#longValueExact()` because it seems that method was only added [recently to Android](https://developer.android.com/reference/java/math/BigInteger#longValueExact()), for an API Level higher than Gson's minimum one. But maybe add a TODO comment to use that method in the future?
    See also https://github.com/google/gson/pull/2031#discussion_r762598044.
  2. call `BigInteger#longValue()`

This is just a suggestion though. Maybe there are other ways to solve this as well.

Marcono1234 (commented):
Feel free to consider my comments only as suggestions, though I hope they are helpful.

----

@eamonnmcmanus, in https://github.com/google/gson/pull/2031#issuecomment-987030702 you had concerns regarding related changes. But that PR changed `JsonPrimitive` behavior while this here only changes `JsonTreeReader`, so maybe this PR here is not as problematic?

gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
Marcono1234, line 269: Similarly, might also need special handling for `LazilyParsedNumber` since that might also be common inside `JsonPrimitive`.

Maybe `LazilyParsedNumber` would need a new `longValueExact()` for that, _maybe_ something like this?
```java
public long longValueExact() {
  try {
    return Long.parseLong(value);
  } catch (NumberFormatException ignored) {
    BigDecimal bigDecimal = asBigDecimal();
    try {
      return bigDecimal.longValueExact();
    } catch (ArithmeticException arithmeticException) {
      NumberFormatException e = new NumberFormatException(arithmeticException.getMessage());
      e.initCause(arithmeticException);
      throw e;
    }
  }
}
```
(but might have to catch the `NumberFormatException` here then, wrap it and include `locationString()`?)

PR #3030, opened 2026-05-31
Bump Error Prone to 2.49.0 and fix new warnings

Marcono1234:
Bumps Error Prone to 2.49.0 and fixes the new warnings it reports.

This should fix the build error of #3022.

The largest changes are:
- Refactoring `LinkedTreeMap` to remove `NATURAL_ORDER` and related reference equality checks (first commit)
- Extracting code out of `assertThrows(...)` calls
Please let me know if you would prefer other local variable names (currently I mostly used "`value`" for the object to be serialized) or if you have any suggestions for changes.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ReflectionAccessFilterHelper.java
+++ b/gson/src/main/java/com/google/gson/internal/ReflectionAccessFilterHelper.java
@@ -73,6 +73,7 @@ public class ReflectionAccessFilterHelper {
   }
 
   /** See {@link AccessibleObject#canAccess(Object)} (Java >= 9) */
+  @SuppressWarnings("InvalidLink") // suppress Error Prone warning about link to Java 9 method
   public static boolean canAccess(AccessibleObject accessibleObject, Object object) {
     return AccessChecker.INSTANCE.canAccess(accessibleObject, object);
   }
--- a/gson/src/test/java/com/google/gson/JsonObjectAsMapTest.java
+++ b/gson/src/test/java/com/google/gson/JsonObjectAsMapTest.java
@@ -103,7 +103,8 @@ public class JsonObjectAsMapTest {
     assertThat(map.put("b", JsonNull.INSTANCE)).isNull();
     assertThat(map.get("b")).isEqualTo(JsonNull.INSTANCE);
 
-    var e = assertThrows(NullPointerException.class, () -> map.put(null, new JsonPrimitive(1)));
+    JsonPrimitive value = new JsonPrimitive(1);
+    var e = assertThrows(NullPointerException.class, () -> map.put(null, value));
     assertThat(e).hasMessageThat().isEqualTo("key == null");
 
     e = assertThrows(NullPointerException.class, () -> map.put("a", null));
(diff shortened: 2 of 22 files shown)


gson/src/main/java/com/google/gson/internal/ReflectionAccessFilterHelper.java
```diff
@@ -73,6 +73,7 @@ public static FilterResult getFilterResult(
   }
 
   /** See {@link AccessibleObject#canAccess(Object)} (Java >= 9) */
+  @SuppressWarnings("InvalidLink") // suppress Error Prone warning about link to Java 9 method
```
Marcono1234, line 76: Refers to the `{@link AccessibleObject#canAccess(Object)}` in the line above

gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
```diff
@@ -233,23 +234,36 @@ public void testStrictNansAndInfinities() throws IOException {
...
 
+  // Suppress Error Prone warning; extracting `(Number) ...` into separate var is redundant
+  @SuppressWarnings("AssertThrowsMinimizer")
   @Test
   public void testStrictBoxedNansAndInfinities() throws IOException {
     JsonTreeWriter writer = new JsonTreeWriter();
     writer.setStrictness(Strictness.LEGACY_STRICT);
     writer.beginArray();
-    assertThrows(IllegalArgumentException.class, () -> writer.value(Float.valueOf(Float.NaN)));
+    assertThrows(IllegalArgumentException.class, () -> writer.value((Number) Float.NaN));
```
Marcono1234, line 244: Used `(Number) ...` casts here to point out that the main purpose is to call the `value(Number)` overload instead of `value(double)` or `value(float)`.

The previous code using `Double.valueOf` and `Float.valueOf` was achieving the same; but Error Prone reported AssertThrowsMinimizer warnings there.
(It still reports those warnings for `(Number) ...` though.)

gson/src/test/java/com/google/gson/internal/bind/JsonTreeWriterTest.java
```diff
@@ -233,23 +234,36 @@ public void testStrictNansAndInfinities() throws IOException {
...
-        () -> writer.value(Double.valueOf(Double.NEGATIVE_INFINITY)));
+        IllegalArgumentException.class, () -> writer.value((Number) Double.NEGATIVE_INFINITY));
     assertThrows(
-        IllegalArgumentException.class,
-        () -> writer.value(Double.valueOf(Double.POSITIVE_INFINITY)));
+        IllegalArgumentException.class, () -> writer.value((Number) Double.POSITIVE_INFINITY));
+  }
+
+  @Test
+  public void testStrictLazyNansAndInfinities() throws IOException {
```
Marcono1234, line 257: Noticed that unlike `JsonWriterTest` the test here did not check `LazilyParsedNumber`, so I added this new method.

gson/src/test/java/com/google/gson/JsonObjectAsMapTest.java
```diff
@@ -224,10 +223,9 @@ public void testEntrySet() {
     // Should contain entries in same order
     assertThat(new ArrayList<>(entrySet)).isEqualTo(expectedEntrySet);
 
+    Map.Entry<String, JsonElement> dummyEntry = new SimpleEntry<>("c", new JsonPrimitive(3));
```
Marcono1234, line 226: Used the name `dummyEntry` instead of `entry` because a few lines below there is already a local variable named `entry`.

test-shrinker/src/test/java/com/google/gson/it/ShrinkingIT.java
```diff
@@ -315,13 +315,9 @@ public void testUnusedClassRemoved() throws Exception {
...
-            () -> {
-              runTest(
-                  className,
-                  c -> {
-                    fail("Class should have been removed during shrinking: " + c);
-                  });
-            });
+            () ->
+                runTest(
+                    className, c -> fail("Class should have been removed during shrinking: " + c)));
```
Marcono1234, line 320: Converted block `-> { ... }` to `-> ...`.

Was reported by Error Prone AssertThrowsBlockToExpression.

eamonnmcmanus (approved):
This is fantastic, thanks for doing it! I wasn't looking forward to fixing all those problems.

PR #3034, opened 2026-06-04
Fix: peekNumber rejects some numerals like 2^64 + "0"

ogolberg:
Issue: the "leading zero" logic in `peekNumber`, which tries to reject octal representations, does not account for long overflows. This results in valid numeric literals like "184467440737095516160" (2^64 + "0") being rejected in `STRICT` parsing mode.

The easiest way to reproduce this is to serialize and deserialize a `BigInteger` with the value above.

This is a minimal fix which adds an overflow check to the leading zero condition.


eamonnmcmanus (approved):
Thanks, that's a great catch!

PR #3038, opened 2026-06-21
Throw JsonSyntaxException instead of NullPointerException for a null AtomicLongArray element

vasiliy-mikhailov:
Deserializing a JSON array that contains a `null` element into an `AtomicLongArray` throws a raw `NullPointerException`:

```java
new Gson().fromJson("[1,null,3]", AtomicLongArray.class); // NullPointerException
```

The array adapter calls `longAdapter.read(in).longValue()`, and the element adapter returns `null` for a JSON `null`, so `longValue()` is invoked on `null`.

This throws a `JsonSyntaxException` for the null element instead, consistent with how the `Long` adapter already reports malformed numbers. Includes a regression test.

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
+++ b/gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
@@ -17,9 +17,11 @@
 package com.google.gson.functional;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonSyntaxException;
 import com.google.gson.LongSerializationPolicy;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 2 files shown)


gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
```diff
@@ -113,4 +115,12 @@ public void testAtomicLongArrayWithStringSerializationPolicy() {
   private static class AtomicLongHolder {
     AtomicLong value;
   }
+
+  @Test
+  public void testAtomicLongArrayWithNullElement() {
+    JsonSyntaxException e =
+        assertThrows(
+            JsonSyntaxException.class, () -> gson.fromJson("[1,null,3]", AtomicLongArray.class));
+    assertThat(e).hasMessageThat().contains("null");
```
Marcono1234: Given that the message is fully controlled by Gson, maybe it would be better to check the exact message with `isEqualTo`?

Marcono1234 (commented):
Thanks a lot!

This looks reasonable; though the direct project members might have additional comments on this.

Are you able to sign the CLA?

(As side note, while the adapter for `AtomicLong` has a similar code pattern, that one is most likely not affected because the `longAdapter` is fully controlled by Gson, so only a JSON `null` could lead to a Java `null` but the adapter is `nullSafe()`.)

gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
vasiliy-mikhailov: Good idea — switched to `isEqualTo("null is not a valid AtomicLongArray element")` to assert the exact message. Thanks!

vasiliy-mikhailov:
CLA is signed now. Also agree on `AtomicLong` — its `longAdapter` is `nullSafe()`, so only `AtomicLongArray` (which reads elements directly) hits the raw NPE, which is why the fix is scoped to just that adapter.

eamonnmcmanus (approved):
Thanks!

PR #3039, opened 2026-06-22
Rethrow Error from @PostConstruct method instead of wrapping it in RuntimeException

vasiliy-mikhailov:
`PostConstructAdapterFactory` rethrows a `RuntimeException` cause as-is, but wraps everything else (`throw new RuntimeException(e.getCause())`). An `Error` thrown by the `@PostConstruct` method — `AssertionError`, `OutOfMemoryError`, `NoClassDefFoundError`, … — is unchecked like `RuntimeException` and should propagate **unchanged**, not be wrapped in a `RuntimeException` (where a `catch (RuntimeException)` could swallow what is meant to be unrecoverable).

This adds the symmetric `else if (e.getCause() instanceof Error) throw (Error) e.getCause();`, matching how the same block already treats `RuntimeException`. The `new RuntimeException(...)` fallback stays for genuinely checked causes. Includes a regression test.

### DIFF
--- a/extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
+++ b/extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
@@ -95,6 +95,37 @@ public class PostConstructAdapterFactoryTest {
     }
   }
 
+  @Test
+  public void testErrorFromPostConstructPropagatesAsError() {
+    Gson gson =
+        new GsonBuilder().registerTypeAdapterFactory(new PostConstructAdapterFactory()).create();
+
+    CustomTestError thrown =
+        assertThrows(
+            CustomTestError.class, () -> gson.fromJson("{\"value\": 1}", ErrorThrowingClass.class));
+
+    assertThat(thrown).hasMessageThat().isEqualTo("error from post-construct");
+  }
+
+  static class ErrorThrowingClass {
+    public int value;
+
+    @PostConstruct
+    private void init() {
+      throw new CustomTestError("error from post-construct");
+    }
+  }
+
+  // A dedicated Error subclass so the test does not rely on a JDK Error type (e.g. AssertionError,
+  // which Gson re-tags with a version prefix) and is not confused with a meaningful JVM error.
+  private static final class CustomTestError extends Error {
+    private static final long serialVersionUID = 1L;
+
+    CustomTestError(String message) {
+      super(message);
+    }
+  }
+
   @SuppressWarnings({"overrides", "EqualsHashCode"}) // for missing hashCode() override
   static class MultipleSandwiches {
     public List<Sandwich> sandwiches;
--- a/extras/src/main/java/com/google/gson/typeadapters/PostConstructAdapterFactory.java
+++ b/extras/src/main/java/com/google/gson/typeadapters/PostConstructAdapterFactory.java
@@ -65,6 +65,8 @@ public class PostConstructAdapterFactory implements TypeAdapterFactory {
         } catch (InvocationTargetException e) {
           if (e.getCause() instanceof RuntimeException) {
             throw (RuntimeException) e.getCause();
+          } else if (e.getCause() instanceof Error) {
+            throw (Error) e.getCause();
           }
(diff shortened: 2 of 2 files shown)


extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
```diff
@@ -95,6 +95,29 @@ public boolean equals(Object o) {
     }
   }
 
+  @Test
+  public void testErrorFromPostConstructPropagatesAsError() {
+    Gson gson =
+        new GsonBuilder().registerTypeAdapterFactory(new PostConstructAdapterFactory()).create();
+
+    Throwable thrown =
+        org.junit.Assert.assertThrows(
```
Marcono1234: Test compilation fails because this should use an import for `org.junit.Assert` instead of the qualified name, respectively a static import.

extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
```diff
@@ -95,6 +95,29 @@ public boolean equals(Object o) {
...
+            Throwable.class, () -> gson.fromJson("{\"value\": 1}", ErrorThrowingClass.class));
+
+    // The Error thrown from @PostConstruct should propagate as-is, not wrapped in RuntimeException
+    assertThat(thrown).isInstanceOf(Error.class);
+    assertThat(thrown).isNotInstanceOf(RuntimeException.class);
```
Marcono1234: Could make this more specific to check for the exact custom error:

```suggestion
    assertThat(thrown).isInstanceOf(AssertionError.class);
    assertThat(thrown).hasMessageThat().isEqualTo("error from post-construct");
```

(Or maybe even change `assertThrows(Throwable.class, ...)` to `AssertionError error = assertThrows(AssertionError.class, ...)`?)

Marcono1234 (commented):
Thanks! I have added two suggestions.

The direct project members might have additional comments though.

extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
vasiliy-mikhailov: Good catch — switched to the static import so `assertThrows` is used directly. The qualified `org.junit.Assert.assertThrows(...)` left the `import static` unused, which the build rejected. Thanks!

extras/src/test/java/com/google/gson/typeadapters/PostConstructAdapterFactoryTest.java
vasiliy-mikhailov: Done — switched to `assertThrows(AssertionError.class, ...)`, so it now also asserts the propagated throwable is an `AssertionError` rather than a wrapping `RuntimeException`.

One wrinkle on the exact message: `Gson.fromJson` re-tags `AssertionError`s with an `"AssertionError (GSON <version>): "` prefix (keeping the original as the cause), so `isEqualTo("error from post-construct")` does not hold on the top-level message. I used `hasMessageThat().contains("error from post-construct")` to keep the specificity while staying robust to the version tag.

Marcono1234: > One wrinkle on the exact message: `Gson.fromJson` re-tags `AssertionError`s with an `"AssertionError (GSON <version>): "` prefix

Ah, maybe it would be better to use a different `Error` subclass then (or directly `Error`), possibly even a custom subclass of `Error` (to no cause confusion when using any of the existing `Error` subclasses, e.g. throwing an error such as `StackOverflowError` might be confusing).

vasiliy-mikhailov: Good call — switched to a dedicated `CustomTestError extends Error` thrown from `@PostConstruct`, so there is no `AssertionError` re-tagging and no overlap with a meaningful JVM error such as `StackOverflowError`. The message assertion is now exact (`isEqualTo("error from post-construct")`).

Issue #2737, opened 2024-09-07
Switch from Gson `LinkedTreeMap` to JDK `LinkedHashMap` as Map implementation


Marcono1234:
# Problem solved by the feature
Gson uses its internal `LinkedTreeMap` as `Map` implementation when a `Map<String, ...>` (or a raw `Map`) should be created. The reasons for this `LinkedTreeMap` seem to be that old JDK versions were vulnerable to denial-of-service attacks due to hash code collision. See previous discussions in https://github.com/google/gson/pull/1992#issuecomment-941732915 (that also covers another Gson Map implementation which we had removed in the meantime) and https://github.com/google/gson/pull/2152#discussion_r957618584.

The problem is that Gson's `LinkedTreeMap`:
- Does not permit `null` keys
- Requires that keys must be `Comparable`

This can cause issues such as #1247, but also for less contrived cases where the user tries to add `null` or non-`Comparable` to the deserialized map, such as:
```java
@SuppressWarnings("unchecked")
Map<Object, Object> map = new Gson().fromJson("{}", Map.class);
map.put(null, 1);

var map2 = new Gson().fromJson("{}", new TypeToken<Map<String, Integer>>() {});
map2.put(null, 1);
```

Note: I have added the https://github.com/google/gson/labels/java8 label because for Java >= 8 we can probably be relatively sure that `LinkedTreeMap` is not needed as denial-of-service protection, see [JDK-8046170](https://bugs.openjdk.org/browse/JDK-8046170).
But we could also already make this change while Gson is still targeting Java 7 as minimum.

# Feature description
We should consider not creating Gson's `LinkedTreeMap` in `ConstructorConstructor` anymore, but only JDK `LinkedHashMap`.
(And adjust unit tests which assert `isInstanceof(LinkedTreeMap.class)` and `isNotInstanceof(LinkedTreeMap.class)`.)

Note: We can probably not completely remove `LinkedTreeMap` because:
- External users rely on it (even though they shouldn't because the class is internal); we could mark it as `@Deprecated` though
- `JsonObject` uses it, and especially its `asMap` method relies on `LinkedTreeMap` not permitting `null` keys and values
Otherwise we would have to write a custom wrapper class which disallows `null` keys and values (similar to what we have for `JsonArray#asList`)

This would then:
- Resolve #1247
- Obsolete #2152

# Alternatives / workarounds
- Try to fix at least some cases of `ClassCastException` due to `LinkedTreeMap` usage, see #2152

[...]

PR #3042, opened 2026-06-25
Preserve string-keyed map hardening for ConcurrentMap

M0nd0R:
## Summary

This change preserves Gson's existing string-keyed map hardening behavior for
`ConcurrentMap<String, ...>` interface targets.

`Map<String, ...>` currently uses `LinkedTreeMap` to avoid hash-table collision
behavior on older runtimes. `ConcurrentMap<String, ...>` was instead resolved to
`ConcurrentHashMap`; this change resolves string-keyed concurrent map interface
targets to `ConcurrentSkipListMap` while preserving `ConcurrentHashMap` for
non-string keys and explicit `ConcurrentHashMap` targets.

## Tests

- `mise x java@temurin-17 maven@3.9.11 -- mvn -pl gson -Dtest=MapTest test`
- `mise x java@temurin-17 maven@3.9.11 -- mvn -pl gson spotless:check`

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
+++ b/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java
@@ -368,6 +368,10 @@ public final class ConstructorConstructor {
     else if (rawType.isAssignableFrom(TreeMap.class)) {
       return TreeMap::new;
     }
+    // Preserve string-key collision hardening for ConcurrentMap interfaces.
+    else if (rawType.isAssignableFrom(ConcurrentSkipListMap.class) && hasStringKeyType(type)) {
+      return ConcurrentSkipListMap::new;
+    }
     // Then try ConcurrentMap implementation
     else if (rawType.isAssignableFrom(ConcurrentHashMap.class)) {
       return ConcurrentHashMap::new;
--- a/gson/src/test/java/com/google/gson/functional/MapTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapTest.java
@@ -279,11 +279,23 @@ public class MapTest {
   public void testConcurrentMap() {
     Type typeOfMap = new TypeToken<ConcurrentMap<Integer, String>>() {}.getType();
     ConcurrentMap<Integer, String> map = gson.fromJson("{\"123\":\"456\"}", typeOfMap);
+    assertThat(map).isInstanceOf(ConcurrentHashMap.class);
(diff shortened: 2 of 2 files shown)


Marcono1234:
Thanks!

I am a bit skeptical how useful this is: Based on [JDK-8046170](https://bugs.openjdk.org/browse/JDK-8046170) and the Git history of the OpenJDK and Android source code the solution for frequent hash collisions existed in `ConcurrentHashMap` before `HashMap` (but also only since JDK 8).
And Gson already requires Java >= 8 for the latest versions.

I suspect `ConcurrentSkipListMap` is not as efficient as `ConcurrentHashMap` due to its sorting and 'skip list' feature, unless the users needs exactly that, but then they should deserialize as that specific class and not as `ConcurrentMap`.

Or do you have any insights that the mitigations for frequent hash collisions in `ConcurrentHashMap` and `HashMap` in JDK 8 are not enough?

Even for `Map<String, ...>` it might make sense if Gson migrated away from its own `LinkedTreeMap`, see #2737 and the issues linked there.

Note that I am not a direct member of this project; maybe they think differently about this.

M0nd0R:
Thanks for the detailed context.

My rationale here was not that the JDK 8+ `ConcurrentHashMap` tree-bin mitigation is known to be insufficient. In my local validation on a modern JDK, the colliding keys also ended up in a `ConcurrentHashMap$TreeBin`, not in a reproducer that times out.

The motivation was consistency with Gson's existing string-keyed map policy in `ConstructorConstructor`: `Map<String, ...>` is still special-cased to avoid hash-table collision behavior on older runtimes, but `ConcurrentMap<String, ...>` currently bypasses that policy and routes promoted JSON object names to `ConcurrentHashMap`.

That said, your performance/semantics point is fair. `ConcurrentSkipListMap` is not a drop-in performance equivalent for `ConcurrentHashMap`, and if Gson maintainers consider the supported-runtime `ConcurrentHashMap` hardening sufficient, this PR may not be worth that tradeoff. I am happy to rework or close this depending on which direction the project prefers.

Marcono1234:
#3057 has been merged and `LinkedHashMap` will now be used as `Map` implementation.
Following that rationale, I think this PR here is not needed (but feel free to keep it open for an official confirmation by a maintainer).

Thanks anyway for your effort and the constructive discussion!

M0nd0R:
Closing.

PR #3045, opened 2026-06-26
Read an empty Locale string back as Locale.ROOT

vasiliy-mikhailov:
Gson serializes `Locale.ROOT` to an empty string (`Locale.toString()` returns `""`), but deserializing `""` back to a `Locale` threw a `NullPointerException`: the tokenizer produced no tokens, so `language` stayed `null` and was passed to `new Locale(null)`.

Defaulting `language` to the empty string makes the empty form round-trip to `Locale.ROOT`. Added a regression test.

## AI assistance disclosure

This contribution was produced with the help of an AI pipeline. The pipeline processed a large amount of source code to surface suspected bugs, reproduced a subset of them with failing unit tests and generated candidate fixes, and prepared pull requests from the ones that held up. Each PR was then reviewed and verified by a human before being opened: the fix and test were checked by hand and the test was confirmed to fail before the change and pass after.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -909,7 +909,7 @@ public final class TypeAdapters {
           }
           String locale = in.nextString();
           StringTokenizer tokenizer = new StringTokenizer(locale, "_");
-          String language = null;
+          String language = "";
           String country = null;
           String variant = null;
           if (tokenizer.hasMoreElements()) {
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -909,7 +909,7 @@ public final class TypeAdapters {
           }
           String locale = in.nextString();
           StringTokenizer tokenizer = new StringTokenizer(locale, "_");
-          String language = "";
+          String language = null;
           String country = null;
           String variant = null;
           if (tokenizer.hasMoreElements()) {
@@ -921,7 +921,10 @@ public final class TypeAdapters {
           if (tokenizer.hasMoreElements()) {
             variant = tokenizer.nextToken();
           }
-          if (country == null && variant == null) {
+          if (language == null) {
+            return Locale.ROOT;
+          }
+          if (country == null) {
             return new Locale(language);
           } else if (variant == null) {
             return new Locale(language, country);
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -909,7 +909,7 @@ public Locale read(JsonReader in) throws IOException {
           }
           String locale = in.nextString();
           StringTokenizer tokenizer = new StringTokenizer(locale, "_");
-          String language = null;
+          String language = "";
```
Marcono1234: This change seems to work (`new Locale("")` creates something which acts like `Locale.ROOT`), but maybe it would be better to explicitly handle this case.
That is, keep `= null` here and then below add a `if (language == null) { return Locale.ROOT; }`?

What do you think?

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
Marcono1234: Not directly related to these changes, but maybe could omit the `variant == null` here?
```diff
- if (country == null && variant == null) {
+ if (country == null) {
```

If I understand the code above correctly, due to the `hasMoreElements()` checks `country == null` implies `variant == null`. The current (redundant?) `variant == null` check irritates null analysis (e.g. in IntelliJ IDEA), making it think for `new Locale(language, country, variant)` `country` can actually be null.

(Side note: IntelliJ did also identify the `new Locale(language)` as potentially causing a `NullPointerException`, which you proved in your PR here.)

This is only a suggestion though; feel free to ignore this and keep the code as it is.

gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -263,6 +263,14 @@ public void testLocaleDeserializationWithLanguage() {
     assertThat(locale.getLanguage()).isEqualTo("en");
   }
 
+  @Test
+  public void testLocaleDeserializationRoot() {
+    // Gson serializes Locale.ROOT to "", so it must read "" back to Locale.ROOT.
```
Marcono1234, line 268: For completeness it would probably be good to have also a serialization test which shows that `Locale.ROOT` is actually serialized as `"\"\""`.

Marcono1234 (commented):
Good catch!

I have added a few suggestion comments, hopefully they are useful. The direct project members might have additional comments.

vasiliy-mikhailov:
Thanks @Marcono1234! Adopted all three:

1. Switched to the explicit `if (language == null) { return Locale.ROOT; }` guard — agreed it reads more clearly than relying on `new Locale("")`.
2. Dropped the redundant `variant == null`, so it's now just `if (country == null)`. As you noted, the `hasMoreElements()` order means `country == null` already implies `variant == null`, and it keeps null analysis happy.
3. Added `testLocaleSerializationRoot` asserting `Locale.ROOT` serializes to `""`, so the round-trip is covered from both sides.

(And credit to IntelliJ for flagging the `new Locale(language)` NPE.)

PR #3046, opened 2026-06-26
Use Double and Long hashCode functions

lllincoln:
### Purpose

`JsonPrimitive.hashCode` currently contains a comment and the hashCode bitwise computation itself. This can be simplified using the hashCode functions that JDK 1.8+ provides through Long and Double. This PR aims to improve the readability of the function by using those Long and Double hashCode functions.

### Description

Java 1.8 introduced [Double.hashCode](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#hashCode-double-) and [Long.hashCode](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#hashCode-long-).

[`Double.hashCode`](https://github.com/keerath/openjdk-8-source/blob/master/jdk/src/share/classes/java/lang/Double.java):

```java
/**
 * Returns a hash code for a {@code double} value; compatible with
 * {@code Double.hashCode()}.
 *
 * @param value the value to hash
 * @return a hash code value for a {@code double} value.
 * @since 1.8
 */
public static int hashCode(double value) {
    long bits = doubleToLongBits(value);
    return (int)(bits ^ (bits >>> 32));
}
```

[`Long.hashCode`](https://github.com/keerath/openjdk-8-source/blob/master/jdk/src/share/classes/java/lang/Long.java):

```java
/**
 * Returns a hash code for a {@code long} value; compatible with
 * {@code Long.hashCode()}.
 *
 * @param value the value to hash
 * @return a hash code value for a {@code long} value.
 * @since 1.8
 */
public static int hashCode(long value) {
    return (int)(value ^ (value >>> 32));
}
```

Currently, GSON performs the same operations by essentially inlining these functions. By using the JDK provided hashCode functions and not inlining them, this PR improves the readability of the `JsonPrimitive.hashCode` function. Thanks for considering!

### Checklist

- [X] New code follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html)\
- [X] `mvn clean verify javadoc:jar` passes without errors

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -262,14 +262,11 @@ public final class JsonPrimitive extends JsonElement {
     if (value == null) {
       return 31;
     }
-    // Using recommended hashing algorithm from Effective Java for longs and doubles
     if (isIntegral(this)) {
-      long value = getAsNumber().longValue();
-      return (int) (value ^ (value >>> 32));
+      return Long.hashCode(getAsNumber().longValue());
     }
     if (value instanceof Number) {
-      long value = Double.doubleToLongBits(getAsNumber().doubleValue());
-      return (int) (value ^ (value >>> 32));
+      return Double.hashCode(getAsNumber().doubleValue());
     }
     return value.hashCode();
   }


Marcono1234:
Thanks!

The "check-android-compatibility" CI workflow currently fails because those methods were only added in Android API level 24 ([`Double#hashCode(double)`](https://developer.android.com/reference/java/lang/Double?hl=en#hashCode(double)), [`Long#hashCode(long)`](https://developer.android.com/reference/java/lang/Long?hl=en#hashCode(long))), but Gson's minimum supported API level is 23 (the README is outdated, saying still 21):
https://github.com/google/gson/blob/c9f3fd55854a743b66f857ace3c7b268ea3e2ef7/pom.xml#L526-L528

@eamonnmcmanus, would it make sense to raise the minimum API level to 24? There are a few other TODOs in the code which could be addressed then as well (in subsequent PRs).

cpovirk:
- Google has done its periodic bump of minSdkVersion in foundational libraries, this time moving to 24 (e.g., [AndroidX](https://issuetracker.google.com/issues/474169350), [Guava](https://github.com/google/guava/pull/8573)). I'd imagine that GSON can follow.
- Even if not for that: These methods are available under all versions of Android because they're "backported" ([`Double`](https://r8.googlesource.com/r8/+/cc98a00bcf9288bc1069f55dee03f7acda4121f8/src/main/java/com/android/tools/r8/ir/desugar/BackportedMethodRewriter.java#802), [`Long`](https://r8.googlesource.com/r8/+/cc98a00bcf9288bc1069f55dee03f7acda4121f8/src/main/java/com/android/tools/r8/ir/desugar/BackportedMethodRewriter.java#895)), making them available even to users who don't opt in to library desugaring. As in https://github.com/google/gson/pull/2949#issuecomment-3577649591, I endorse using the Animal Sniffer scents from open-toast/gummy-bears to recognize backporting.

eamonnmcmanus:
> @eamonnmcmanus, would it make sense to raise the minimum API level to 24? There are a few other TODOs in the code which could be addressed then as well (in subsequent PRs).

Done in #3082.

@lllincoln this PR conflicted with changes from #3059. I've adapted your idea to those changes. Does it look reasonable?

lllincoln:
> @lllincoln this PR conflicted with changes from #3059. I've adapted your idea to those changes. Does it look reasonable?

Looks great! Thanks for following up on this and adapting it!

eamonnmcmanus (approved):
Thanks!

Issue #3047, opened 2026-06-26
AtomicIntegerArray deserializer crashes on JSON null (IllegalStateException) — symmetric fix to #3038


mahirhir:
## Bug

Deserializing a JSON array that contains a `null` element into an `AtomicIntegerArray` crashes with an uncaught `IllegalStateException`:

```java
new Gson().fromJson("[1,null,3]", AtomicIntegerArray.class);
// throws IllegalStateException: Expected an int but was NULL at line 1 column 4 path $[1]
```

## Root cause

In `TypeAdapters.java`, the `ATOMIC_INTEGER_ARRAY` adapter's `read()` method calls `in.nextInt()` inside a loop that only catches `NumberFormatException`:

```java
public AtomicIntegerArray read(JsonReader in) throws IOException {
  List<Integer> list = new ArrayList<>();
  in.beginArray();
  while (in.hasNext()) {
    try {
      int integer = in.nextInt();
      list.add(integer);
    } catch (NumberFormatException e) {
      throw new JsonSyntaxException(e);
    }
  }
  in.endArray();
  ...
}
```

When the `JsonReader` is positioned at a `NULL` token, `nextInt()` throws `IllegalStateException` — which is **not** caught here, so it propagates raw instead of being converted to a `JsonSyntaxException`.

## Relationship to #3038

PR #3038 (merged 2026-06-22) fixed the symmetric issue in `AtomicLongArray`. That adapter delegates to `longAdapter.read(in)`, which returns `null` for a JSON null; the fix added an explicit null-check:

```java
Number value = longAdapter.read(in);
if (value == null) {
  throw new JsonSyntaxException("null is not a valid AtomicLongArray element");
}
list.add(value.longValue());
```

`AtomicIntegerArray` was not updated with the same treatment.

## Expected behaviour

A JSON null element inside an `AtomicIntegerArray` should throw a `JsonSyntaxException` (same as `AtomicLongArray` after #3038), not a raw `IllegalStateException`.

## Suggested fix

Mirror the approach used in the `AtomicLongArray` adapter — either:

**Option A** — peek before reading:
```java
while (in.hasNext()) {
  try {
    if (in.peek() == com.google.gson.stream.JsonToken.NULL) {
      throw new JsonSyntaxException("null is not a valid AtomicIntegerArray element");
    }
    int integer = in.nextInt();
    list.add(integer);
  } catch (NumberFormatException e) {
    throw new JsonSyntaxException(e);
  }
}
```

[...]

PR #3048, opened 2026-06-28
Wrap IllegalStateException at the AtomicIntegerArray adapter level, consistent with AtomicLongArray

amitmishra11:
This PR was written primarily by Claude Code; I reviewed the change and ran the test suite before submitting.

## Background

I originally opened this PR believing it fixed a crash described in #3047 (a non-numeric `AtomicIntegerArray` element causing a raw, uncaught `IllegalStateException`). On closer testing, I confirmed that crash does not actually reproduce via the public `Gson.fromJson()` API, since `Gson.fromJson(Reader, TypeToken)` already catches `IllegalStateException` generically and wraps it as `JsonSyntaxException` before it reaches the caller (Gson.java around line 1123). See the discussion on the issue and on this PR for the full back-and-forth - my apologies for initially missing that contradiction.

## What this PR actually does

- `ATOMIC_INTEGER_ARRAY.read()` now catches `IllegalStateException` (in addition to the existing `NumberFormatException`) and wraps it at the adapter level, the same way `AtomicLongArray`'s adapter already handles its null check (see #3038) rather than relying on the caller (`Gson.fromJson`) to do the wrapping.
- This only matters if the adapter's `read(JsonReader)` is invoked directly, bypassing `Gson.fromJson`. Through the normal `fromJson` path, the only visible effect is a more specific exception message/cause at the point of failure instead of the generic message from the outer catch.

This is a minor consistency/message-quality change, not a crash fix. I'm leaving it open in case the consistency improvement is still wanted, but I understand if maintainers would rather close it given the original premise doesn't hold.

## Testing

Added `testAtomicIntegerArrayWithNullElement` and `testAtomicIntegerArrayWithNonNumericElement`. `mvn spotless:check` and the full `gson` module test suite pass.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -344,7 +344,7 @@ public final class TypeAdapters {
             try {
               int integer = in.nextInt();
               list.add(integer);
-            } catch (NumberFormatException e) {
+            } catch (NumberFormatException | IllegalStateException e) {
               throw new JsonSyntaxException(e);
             }
           }
--- a/gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
+++ b/gson/src/test/java/com/google/gson/functional/JavaUtilConcurrentAtomicTest.java
@@ -89,6 +89,22 @@ public class JavaUtilConcurrentAtomicTest {
   }
 
   @Test
+  public void testAtomicIntegerArrayWithNullElement() {
+    JsonSyntaxException e =
+        assertThrows(
+            JsonSyntaxException.class, () -> gson.fromJson("[1,null,3]", AtomicIntegerArray.class));
+    assertThat(e).hasCauseThat().isInstanceOf(IllegalStateException.class);
+  }
+
+  @Test
+  public void testAtomicIntegerArrayWithNonNumericElement() {
+    JsonSyntaxException e =
+        assertThrows(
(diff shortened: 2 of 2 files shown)


amitmishra11:
I have signed the Google CLA.

amitmishra11:
I signed it!

Marcono1234:
Can you please explain the purpose of this PR? In https://github.com/google/gson/issues/3047#issuecomment-4825384716 you confirmed that the issue reported there is non-existent, and now _you_ propose a PR here which attempts to fix that non-existent issue.

amitmishra11:
Fair callout, and apologies for the confusion. You're right that I confirmed on the issue that the originally reported crash doesn't reproduce via the public \`fromJson\` API, since \`Gson.fromJson(Reader, TypeToken)\` already catches \`IllegalStateException\` generically (Gson.java:1123) and wraps it as \`JsonSyntaxException\` before it reaches the caller. So this PR does not fix a crash, and I shouldn't have written "Fixes #3047" - that's not accurate, and I should have caught the contradiction before pushing this rather than after.

What this PR actually changes, now described accurately:
- If \`ATOMIC_INTEGER_ARRAY.read()\` is invoked directly (e.g. via \`TypeAdapter.read(JsonReader)\` rather than through \`Gson.fromJson\`), the exception is now wrapped at the adapter level instead of relying on the caller. This is the same pattern \`AtomicLongArray\`'s adapter already uses for its null check.
- Via \`Gson.fromJson\`, the only effect is a more specific exception message/cause at the point of failure instead of the generic message from the outer catch.

That's a minor consistency/message-quality change, not a bug fix. I'll leave it to you/the maintainers whether that's worth keeping - happy to close this PR if it's not, since the issue itself doesn't actually need it.

PR #3049, opened 2026-06-29
refactor: improve code quality by enforcing final on utility classes and extracting complex methods

nafisa1602:
## What issues were addressed
- Class With Only Private Constructors Should Be Final: Made NumberLimits, 
  PreJava9DateFormatProvider, ReflectionAccessFilterHelper, TroubleshootingGuide, 
  FormattingStyle, ISO8601Utils, and SqlTimestampTypeAdapter final
- Loose Coupling: Changed ArrayDeque to Deque interface in GsonBuilder
- Cognitive Complexity/ Cyclomatic Complexity: Extracted excludeField() checks 
  into private named methods in Excluder

## Why the changes improve the system
- final on utility classes prevents unintended subclassing and communicates 
  design intent clearly to future maintainers
- Interface types over implementation types improves flexibility
- Extracted methods reduce cognitive complexity of excludeField() from 18 to 
  near 0 and make each exclusion rule independently readable and testable

## Refactoring approach
Extract Method and programming to interfaces patterns, guided by PMD static analysis.

## Functionality unchanged
All 4603 existing tests pass with 0 failures after every change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,3 @@ local.properties
 build
 
 .DS_Store
-apache-maven-3.9.6-bin.tar.gz
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ local.properties
 build
 
 .DS_Store
+apache-maven-3.9.6-bin.tar.gz
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -926,7 +926,7 @@ public final class GsonBuilder {
   List<TypeAdapterFactory> createFactories(
       ConstructorConstructor constructorConstructor,
       JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory) {
-    ArrayList<TypeAdapterFactory> factories = new ArrayList<>();
+      List<TypeAdapterFactory> factories = new ArrayList<>();
 
     // built-in type adapters that cannot be overridden
     factories.add(TypeAdapters.JSON_ELEMENT_FACTORY);
@@ -997,8 +997,7 @@ public final class GsonBuilder {
             excluder,
             jsonAdapterFactory,
             newImmutableList(reflectionFilters)));
-
-    factories.trimToSize();
+    
     return Collections.unmodifiableList(factories);
   }
 
--- a/gson/src/main/java/com/google/gson/GsonBuilder.java
+++ b/gson/src/main/java/com/google/gson/GsonBuilder.java
@@ -926,7 +926,7 @@ public final class GsonBuilder {
   List<TypeAdapterFactory> createFactories(
       ConstructorConstructor constructorConstructor,
       JsonAdapterAnnotationTypeAdapterFactory jsonAdapterFactory) {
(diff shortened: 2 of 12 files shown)


gson/src/main/java/com/google/gson/GsonBuilder.java
```diff
@@ -997,7 +998,6 @@ List<TypeAdapterFactory> createFactories(
             jsonAdapterFactory,
             newImmutableList(reflectionFilters)));
 
-    factories.trimToSize();
```
Marcono1234, line 1000: Maybe it would be worth discussing whether this `trimToSize()` here is really needed or is premature optimization, but since the purpose of your PR is only to use "interface types over implementation types", I think this functionality change should be reverted for this PR. That is, please change the local variable back to `ArrayList` and add back the `trimToSize()` call.

Generally "interface types over implementation types" is a good principle, but for internal code it is bit less relevant, even more so if this only applies to a local variable and the specific implementation type is needed because the needed functionality does not exist for the interface type (as it is the case here).

.gitignore
```diff
@@ -18,3 +18,4 @@ local.properties
 build
 
 .DS_Store
+apache-maven-3.9.6-bin.tar.gz
```
Marcono1234: Can you please revert this? It seems to be specific to your development setup and is probably not useful for most other contributors.

Marcono1234 (commented):
Thanks!

The changes look mostly good to me. I have added two comments, feel free to consider them only as suggestions.
The direct project members might have additional comments.

nafisa1602:
Thanks for the review! I've reverted the createFactories change back to
ArrayList with trimToSize() and removed the maven tarball entry from
.gitignore. Let me know if anything else needs adjusting.

gson/src/main/java/com/google/gson/internal/Excluder.java
```diff
@@ -156,40 +156,52 @@ private TypeAdapter<T> delegate() {
   }
 
   public boolean excludeField(Field field, boolean serialize) {
-    if ((modifiers & field.getModifiers()) != 0) {
+    if (isExcludedByModifier(field)) {
```
eamonnmcmanus: This is fine, but I think then I would go a step further and just have a single return statement, like this:

```java
return isExcludedByModifier(field)
    || isExcludedByVersion(field)
    || ...
    || isExcludedByStrategy(field, serialize);
```

In particular, the current `return isExcludedByStrategy(field, serialize);` reads a little strangely because it makes it look as if that condition is somehow different from the preceding ones.

eamonnmcmanus (changes requested):
Thanks for doing this! One small suggestion, and also it looks as if you have a conflict to resolve.

gson/src/main/java/com/google/gson/internal/Excluder.java
nafisa1602: Done. Consolidated all the conditions into a single return statement with ||, so isExcludedByStrategy no longer looks disconnected from the rest. Pushed in d694216.

eamonnmcmanus (approved):
Thanks again!

Issue #1443, opened 2018-12-11
JSON Pointer spec (RFC 6901) support


markkolich:
Consider the following JSON object:

```json
{
  "foo": {
    "bar": {
       "baz": "bat"
    }
  }
}
```

Using GSON, without registering a custom type adapter, if I want to safely access member `foo.bar.baz` I end up with something like this:

```java
final JsonObject object = new GsonBuilder().create().fromJson("{...}", JsonObject.class);

final JsonObject foo = object.getAsJsonObject("foo");
if (foo != null) {
  final JsonObject bar = foo.getAsJsonObject("bar");
  if (bar != null) {
    final JsonElement baz = bar.get("baz");
    if (baz != null && baz.isJsonPrimitive()) {
      final String bat = baz.getAsString();
      // ...
    }
  }
}
```

Instead, I'd love to be able to call a single method on the initial `JsonObject` itself, something like `getStringAtPath()`:

```java
final JsonObject object = new GsonBuilder().create().fromJson("{...}", JsonObject.class);

final String baz = object.getStringAtPath("foo.bar.baz"); // <<-- !!
if (baz == null) {
  // ... didn't exist.
}
```

In lieu of this missing API, I cooked up a simple helper method for String primitives:

```java
import com.google.common.base.Preconditions;
import com.google.common.base.Splitter;

@Nullable
public static final String getStringAtPath(
    final JsonObject object,
    final String path) {
  Preconditions.checkNotNull(object, "JSON object cannot be null.");
  Preconditions.checkNotNull(path, "JSON object path cannot be null.");

  JsonObject base = object;
  final Iterable<String> segments = Splitter.on(".").split(path);
  for (final String segment : segments) {
    final JsonElement element = base.get(segment);
    if (element == null) {
      return null;
    }

    if (element.isJsonPrimitive()) {
      final JsonPrimitive primitive = element.getAsJsonPrimitive();
      return primitive.getAsString();
    } else if (element.isJsonObject()) {
      base = element.getAsJsonObject();
    } else {
      return null;
    }
  }

  return null;
}
```

It would be tremendously helpful if similar methods were conveniently folded into the `JsonObject` API itself.  It could easily be extended for the other types as well: int, boolean, char, double, etc.  Admittedly, things get a bit hairy with arrays but I'd imagine support for paths like `foo.bar[8].baz` could work.

Apologies if this is a dupe, I did briefly search the existing set of issues but didn't immediately find anything related.

[...]

PR #3050, opened 2026-06-30
json pointers as local-only references

scottslewis:
The purpose of this pull request is to add support for the enhancement #1443.

The pr implements the [JRef 'local-only' specification](https://github.com/hyperjump-io/json-reference/blob/main/spec.md). Local-pnly means that only json pointers starting with '#' (local to the document) are supported.

The main two classes for the implementation are JRefTypeAdapter and JRefTypeAdapterFactory.  The JRefTypeAdapterFactory implements the Gson TypeAdapterFactory and creates instances of JRefTypeAdapter.

JRefTypeAdapter implements Gson TypeAdapter and so implements the write (serialization) and read (deserialization) with JRef support.

Two support classes from the [Jackson 3.0 databind project](https://github.com/FasterXML/jackson-databind) have been included as part of this contribution, with permission from the Jackson maintainer: @cowtowncoder.  [Please see this dialog](https://github.com/FasterXML/jackson-databind/pull/6027#issuecomment-4763834920) about these two classes.  The two classes from Jackson are:
JsonPointer (implementation of the rfc-6901 json pointer specification for local-only json pointers).  Also includes parent/child ptr creation/validation/compare code.

TokenStreamContext - Allows the creation of a context for parsers (like Jackson and Gson) that treat the wirting and reading of json as a sequentially processed stream of tokens.  This class is used within JsonPointer to go back and forth between JsonPointers as strings (e.g. /foo/bar/0) and JsonPointers as parent and child contexts (e.g. /foo parent for bar parent for /0 container element 0).

The JRefTypeAdapter. uses the JsonReader.getPath() (json path syntax...e.g. $foo, $bar[0]) in order to  get a contextual name from the token stream, and contruct JsonPointer instances during deserialization.  See JRefTypeAdapter.read(JsonReader) and JRefTypeAdapter.getJsonPointerFromJsonPath methods.

The only modification to an existing Gson class, other than new classes described above is the addition of a single method to JsonWriter class

```java
  public String getDeferredName() {
	  return this.deferredName;
  }
 ```
This method is used in the JRefTypeAdapter.write method to get the deferred name in serialization context where required (e.g. objects, and containers).

Also included in this pr are test classes, in the test src tree in new package named com.google.gson.jref package.

All tests are passing using junit 4.

I'm happy to break this up into multiple prs (e.g. gson changes, jref impl, tests) if desired.  Or to refactor.

### DIFF
--- a/gson/src/main/java/com/google/gson/JRefTypeAdapter.java
+++ b/gson/src/main/java/com/google/gson/JRefTypeAdapter.java
@@ -1,0 +1,296 @@
+package com.google.gson;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.gson.internal.Streams;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import com.google.gson.stream.MalformedJsonException;
+
+public class JRefTypeAdapter extends TypeAdapter<Object> {
+
+	public static final String JREF_NAME = "$ref";
+	public static final String HASH = "#";
+
+	class JRefTokenStreamContext extends TokenStreamContext {
+
+		private JRefTokenStreamContext parent;
+		private String currentName;
+
+		// Arrays for serialization (known length)
+		JRefTokenStreamContext(JRefTokenStreamContext parent, int type, int index) {
+			this.parent = parent;
+			this._type = type;
+			this._index = index;
+		}
+
+		// Arrays
+		JRefTokenStreamContext(JRefTokenStreamContext parent, int index) {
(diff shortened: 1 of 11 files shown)


scottslewis:
The Jackson databind impl of this approach (local only reference support through json pointers), is [moving foward](https://github.com/FasterXML/jackson-databind/pull/6045#issuecomment-5028302857).  I would appreciate a review of this contribution.

scottslewis:
For all information:

The json ser/des behavior of this pr (using json pointers + jref for efficient local-only json) has now been contributed to Jackson

https://github.com/FasterXML/jackson-databind/pull/6045

And made available via Javascript standard 'JSON' api:

https://www.npmjs.com/package/@openmcptools/jref

This provides interoperability across languages/processes for size efficient json ser/des of complex data structures (trees, networks, graphs).

PR #3053, opened 2026-07-01
Bump Error Prone to 2.50.0

Marcono1234:
### Purpose
Bumps Error Prone to 2.50.0 and fixes new warnings
Unblocks #3051

### Description
The change to the Error Prone check `ReferenceEquality` and the new check `ExposedPrivateType` required some code changes.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/GsonTypes.java
+++ b/gson/src/main/java/com/google/gson/internal/GsonTypes.java
@@ -169,7 +169,9 @@ public final class GsonTypes {
 
   /** Returns true if {@code a} and {@code b} are equal. */
   public static boolean equals(Type a, Type b) {
-    if (a == b) {
+    @SuppressWarnings("ReferenceEquality")
+    boolean areSame = a == b;
+    if (areSame) {
       // also handles (a == null && b == null)
       return true;
 
@@ -366,7 +368,9 @@ public final class GsonTypes {
         }
 
         toResolve = resolveTypeVariable(context, contextRawType, typeVariable);
-        if (toResolve == typeVariable) {
+        @SuppressWarnings("ReferenceEquality")
+        boolean areSame = toResolve == typeVariable;
+        if (areSame) {
           break;
         }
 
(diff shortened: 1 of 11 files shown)


gson/src/main/java/com/google/gson/internal/bind/JsonAdapterAnnotationTypeAdapterFactory.java
```diff
@@ -35,19 +35,19 @@
  * @since 2.3
  */
 public final class JsonAdapterAnnotationTypeAdapterFactory implements TypeAdapterFactory {
-  private static class DummyTypeAdapterFactory implements TypeAdapterFactory {
+  private static final class DummyTypeAdapterFactory implements TypeAdapterFactory {
```
Marcono1234, line 38: Needed to prevent a `ReferenceEquality` warning when checking for the dummy instance using `==`.

gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -474,7 +477,7 @@ abstract void readIntoField(JsonReader reader, Object target)
   public abstract static class Adapter<T, A> extends TypeAdapter<T> {
     private final FieldsData fieldsData;
 
-    Adapter(FieldsData fieldsData) {
+    private Adapter(FieldsData fieldsData) {
```
Marcono1234, line 480: Was flagged by `ExposedPrivateType`; maybe that check is actually a bit strict in that it flags even package-private vs. `private`, instead of just checking `protected` and `public`.

Not sure if this will cause a synthetic bridge method now, since Gson targets Java 8 which does not have "JEP 181: Nest-Based Access Control".
But maybe that is negligible.

gson/src/main/java/com/google/gson/internal/GsonTypes.java
```diff
@@ -421,14 +425,14 @@ private static Type resolve(
...
-          if (lowerBound != originalLowerBound[0]) {
+          if (!equal(lowerBound, originalLowerBound[0])) {
             toResolve = supertypeOf(lowerBound);
             break;
           }
         } else if (originalUpperBound.length == 1) {
           Type upperBound =
               resolve(context, contextRawType, originalUpperBound[0], visitedTypeVariables);
-          if (upperBound != originalUpperBound[0]) {
+          if (!equal(upperBound, originalUpperBound[0])) {
```
Marcono1234, line 435: This could technically be a behavior change, but it is now consistent with the rest of the `resolve` method which already uses `equal` for other kinds of nested types, e.g. the type arguments of parameterized types or array components.

eamonnmcmanus (approved):

PR #3054, opened 2026-07-05
Cache reflective accessibility check result per BoundField

### LINKED ISSUE #2202: `ReflectiveTypeAdapterFactory.BoundField` should not call `checkAccessible` for every read and write
Currently `ReflectiveTypeAdapterFactory.BoundField` might end up calling `checkAccessible` for every field read and write.

# Feature description
It would probably make more sense to only call this method once when the `BoundField` is created and then store the result in a field and check that when reading or writing.

The underlying issue is that `java.lang.reflect.AccessibleObject.canAccess(Object)` requires an instance as argument. It appears the internal JDK implementation only needs its class to check if the caller can access a `protected` member. So most likely any instance would be suitable to determine this once, because Gson's `ReflectionAccessFilterHelper$AccessChecker` does not have a super class so the `canAccess` result should always be the same.

nileshpatil6:
Closes #2202

`checkAccessible` calls `ReflectionAccessFilterHelper.canAccess` reflectively on every single `write()` and `readIntoField()` call when `blockInaccessible` is true, even though the result only depends on the field/accessor and the `ReflectionAccessFilter`s configured on the `Gson` instance, not on the specific object being serialized or deserialized. So for repeated (de)serialization of many instances of the same type, the exact same reflective check gets redone for every single object.

This adds a small `AccessibleCache` per `BoundField` that runs the check lazily on the first `write()`/`readIntoField()` call and caches the result (accessible, or the inaccessible message) for reuse afterwards, so the reflective check effectively runs once per field/accessor instead of once per object.

The constructor accessibility check for records (`RecordAdapter`) already only runs once at adapter creation time, so it did not need this change.

### Testing

I don't have a maven setup in the environment I worked in, so I could not run `mvn clean verify`. To still verify correctness I compiled the `gson` module's main sources directly with `javac` (stubbing out the generated `GsonBuildConfig` and pulling `error_prone_annotations` from Maven Central), then compiled and ran the existing `ReflectionAccessFilterTest` and `ReflectionAccessTest` against it with JUnit + Truth pulled from Maven Central:

- `ReflectionAccessFilterTest`: all 15 tests pass.
- `ReflectionAccessTest`: 2/3 pass; the third (`testRestrictiveSecurityManager`) fails both with and without my change, with a `BootstrapMethodError` coming from `Truth`'s lambda metafactory colliding with a custom `SecurityManager` under JDK 17, so it's a pre-existing environment issue unrelated to this change, not a regression.

I also wrote a small standalone repro exercising a `Gson` configured with a `BLOCK_INACCESSIBLE` filter: a public field serializes/deserializes correctly across repeated calls (the cached "accessible" result gets reused), and a private field consistently throws `JsonIOException` on every call (the cached "inaccessible" result doesn't get lost or silently become a false positive after the first check).

Would appreciate someone running the full `mvn clean verify` on CI to double check, since I couldn't do that locally.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -183,11 +183,11 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
 
   /**
    * Caches the result of {@link ReflectionAccessFilterHelper#canAccess} for a single (non-static)
-   * member, so the reflective check only has to run once per {@code BoundField} instead of on
-   * every {@link BoundField#write} or {@link BoundField#readIntoField} call. The result only
-   * depends on the member itself and the {@code ReflectionAccessFilter}s configured for the
-   * enclosing {@code Gson} instance, not on the specific object instance being serialized or
-   * deserialized, so it is safe to compute it lazily on the first check and reuse it afterwards.
+   * member, so the reflective check only has to run once per {@code BoundField} instead of on every
+   * {@link BoundField#write} or {@link BoundField#readIntoField} call. The result only depends on
+   * the member itself and the {@code ReflectionAccessFilter}s configured for the enclosing {@code
+   * Gson} instance, not on the specific object instance being serialized or deserialized, so it is
+   * safe to compute it lazily on the first check and reuse it afterwards.
    */
   private static final class AccessibleCache<M extends AccessibleObject & Member> {
     private static final String ACCESSIBLE = "";
--- a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
@@ -169,12 +169,49 @@ public final class ReflectiveTypeAdapterFactory implements TypeAdapterFactory {
       Object object, M member) {
     if (!ReflectionAccessFilterHelper.canAccess(
         member, Modifier.isStatic(member.getModifiers()) ? null : object)) {
-      String memberDescription = ReflectionHelper.getAccessibleObjectDescription(member, true);
-      throw new JsonIOException(
-          memberDescription
-              + " is not accessible and ReflectionAccessFilter does not permit making it"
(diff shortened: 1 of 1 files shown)


gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java
```diff
@@ -169,12 +169,49 @@ private static <M extends AccessibleObject & Member> void checkAccessible(
...
+
+  /**
+   * Caches the result of {@link ReflectionAccessFilterHelper#canAccess} for a single (non-static)
+   * member, so the reflective check only has to run once per {@code BoundField} instead of on every
+   * {@link BoundField#write} or {@link BoundField#readIntoField} call. The result only depends on
+   * the member itself and the {@code ReflectionAccessFilter}s configured for the enclosing {@code
+   * Gson} instance, not on the specific object instance being serialized or deserialized, so it is
+   * safe to compute it lazily on the first check and reuse it afterwards.
+   */
+  private static final class AccessibleCache<M extends AccessibleObject & Member> {
```
eamonnmcmanus, line 192: Caching the result of the accessibility check seems like a good idea, but I think we could do something a bit simpler. The usual case is that the member _is_ accessible, and we would like to remember when we have already determined that. But in the unusual case where it is _not_ accessible, I think it is fine to make the redundant accessibility check again and construct the exception message again. That makes me think that we could have a smaller change where `checkAccessible` acquires a parameter `AtomicBoolean knownAccessible`. If it's already true, we don't need to check again, and otherwise we check and update.

eamonnmcmanus (changes requested):

Marcono1234:
@nileshpatil6, did you intentionally close this PR? What do you think about the suggestion by @eamonnmcmanus above?

nileshpatil6:
@Marcono1234 not intentional, no. I deleted my fork while cleaning up old repos and only realized afterwards that it kills the PR permanently, GitHub won't reopen once the head repo is gone.

Eamonn's suggestion is better than what I had, the AccessibleCache class was more machinery than the problem needed. I've reopened this as #3086 implementing exactly that: an AtomicBoolean knownAccessible parameter on checkAccessible, skip when already true, and the inaccessible case just redoes the check each time.

Also managed to get maven working locally this time, so unlike this PR the new one has a real mvn -pl gson verify run behind it: 4619 tests, 0 failures.

PR #3056, opened 2026-07-13
Throw JsonSyntaxException for null map keys in array form

### LINKED ISSUE #3037: Uncaught NullPointerException ("key == null") when deserializing a JSON object with a null key into a Map
## Summary

Deserializing certain crafted JSON into a `Map` via `Gson.fromJson(...)` reaches `com.google.gson.internal.LinkedTreeMap.put` with a `null` key and throws a raw `java.lang.NullPointerException` with message `key == null`, instead of one of Gson's documented parse exceptions (`JsonSyntaxException` / `JsonParseException`). The map type adapter accepts the JSON array-encoded entry form (`[[key, value], ...]`) on read regardless of configuration, and when the encoded key is the JSON literal `null` it is handed directly to the backing map.

This matters because the exception escapes Gson's documented error-handling contract. `Gson.fromJson` is documented to throw `JsonSyntaxException` (a `JsonParseException`) for malformed input, so a caller deserializing untrusted JSON typically guards with `catch (JsonParseException ...)`. A `NullPointerException` is unchecked and not part of that family, so it slips past such handlers and propagates up the stack, allowing crafted input to crash a caller that otherwise correctly handles malformed JSON (an availability / robustness issue). The JVM remains memory-safe; this is not memory corruption.

## Root Cause

For a target `Map` type, `MapTypeAdapterFactory$Adapter.read` supports two on-the-wire encodings: a JSON object (`{"k": v}`) and a JSON array of two-element arrays (`[[k, v], ...]`, the "complex map key" encoding). In the array branch, the key is decoded with the key type adapter and passed straight to `map.put(key, value)`. When the encoded key is the JSON literal `null`, the key adapter returns `null`, and the default map implementation, `LinkedTreeMap`, rejects a `null` key with a bare `NullPointerException`:

```java
// com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter#read (array branch)
in.beginArray();
while (in.hasNext()) {
  in.beginArray();                 // one [key, value] entry
  K key = keyTypeAdapter.read(in); // resolves to null for JSON `null`
  V replaced = map.put(key, value); // <-- NPE thrown here for null key
  if (replaced != null) {
    throw new JsonSyntaxException("duplicate key: " + key);
  }
  in.endArray();
}
in.endArray();
```

```java
// com.google.gson.internal.LinkedTreeMap#put
@Override
public V put(K key, V value) {
  if (key == null) {
    throw new NullPointerException("key == null");
  }
  ...
}
```

Because the `null`-key rejection happens deep inside the internal map rather than at the adapter boundary, malformed input surfaces as an unchecked `NullPointerException` instead of a typed `JsonSyntaxException`.

## PoC

The following JSON, deserialized into `Map<String, Object>`, triggers the bug. It uses the array-encoded map-entry form with a single entry whose key is the JSON literal `null`:

```
[[null,1]]
```

## Reproduction

Build/obtain gson 2.11.0 and run the public API on the input:

```java
import com.google.gson.Gson;
import java.util.Map;

public class Repro {
    String json = "[[null,1]]";
    Type mapType = new TypeToken<Map<String, Object>>() {}.getType();
    Object result = new Gson().fromJson(json, mapType); // throws NPE
    System.out.println("No exception, result = " + result);
  }
}
```

Compile and run:

```
javac -cp gson-2.11.0.jar Repro.java
java -cp .:gson-2.11.0.jar Repro
```

Output:

```
Exception in thread "main" java.lang.NullPointerException: key == null
	at com.google.gson.internal.LinkedTreeMap.put(LinkedTreeMap.java:118)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:188)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:154)
	at com.google.gson.Gson.fromJson(Gson.java:1361)
	at com.google.gson.Gson.fromJson(Gson.java:1262)
	at com.google.gson.Gson.fromJson(Gson.java:1171)
	at com.google.gson.Gson.fromJson(Gson.java:1137)
	at Repro.main(Repro.java:11)
```

## Suggested Fix

In `MapTypeAdapterFactory$Adapter.read`, validate the decoded key before insertion and translate a `null` key into a typed parse error rather than letting the backing map throw an unchecked `NullPointerException`, e.g.:

```java
if (key == null) {
  throw new JsonSyntaxException("Map key is null");
}
```

This keeps all malformed-input outcomes inside the documented `JsonParseException` contract. (The same guard applies to both the array-encoded and object-encoded branches.)

## PoC bytes (self-contained)

The trigger JSON is `[[null,1]]` (10 bytes, plain ASCII). No binary decoding is required; pass the literal string above to `Gson.fromJson` with target type `Map<String, Object>` to reproduce.

## Credit

Aisle Research (Ze Sheng (O2Lab & TAMU), Dmitrijs Trizna, Luigino Camastra, Guido Vranken).

ATKasem:
Fixes #3037

## Summary
- Reject `null` map keys in `MapTypeAdapterFactory.Adapter.read` before insertion
- Throw `JsonSyntaxException` instead of letting `LinkedTreeMap.put` surface an unchecked `NullPointerException`

## Root cause
Array-encoded map entries (`[[key, value], ...]`) decode a JSON `null` key to Java `null` and pass it straight to the backing map. `LinkedTreeMap` rejects that with `NullPointerException: key == null`, which escapes Gson's documented `JsonParseException` error contract.

## Test plan
- [x] `mvn -pl gson -Dtest=MapAsArrayTypeAdapterTest test`
- [x] `mvn -pl gson -Dtest=MapTest,MapAsArrayTypeAdapterTest test`

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
@@ -187,6 +187,9 @@ public final class MapTypeAdapterFactory implements TypeAdapterFactory {
         while (in.hasNext()) {
           in.beginArray(); // entry array
           K key = keyTypeAdapter.read(in);
+          if (key == null) {
+            throw new JsonSyntaxException("Map key is null");
+          }
           V value = valueTypeAdapter.read(in);
           if (map.containsKey(key)) {
             throw new JsonSyntaxException("duplicate key: " + key);
@@ -200,6 +203,9 @@ public final class MapTypeAdapterFactory implements TypeAdapterFactory {
         while (in.hasNext()) {
           JsonReaderInternalAccess.INSTANCE.promoteNameToValue(in);
           K key = keyTypeAdapter.read(in);
+          if (key == null) {
+            throw new JsonSyntaxException("Map key is null");
+          }
           V value = valueTypeAdapter.read(in);
           if (map.containsKey(key)) {
             throw new JsonSyntaxException("duplicate key: " + key);
--- a/gson/src/test/java/com/google/gson/functional/MapAsArrayTypeAdapterTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapAsArrayTypeAdapterTest.java
@@ -91,6 +91,15 @@ public class MapAsArrayTypeAdapterTest {
   }
 
   @Test
+  public void testNullKeyArrayForm() {
+    Gson gson = new Gson();
+    Type type = new TypeToken<Map<String, Object>>() {}.getType();
+    JsonSyntaxException e =
+        assertThrows(JsonSyntaxException.class, () -> gson.fromJson("[[null,1]]", type));
+    assertThat(e).hasMessageThat().isEqualTo("Map key is null");
+  }
+
+  @Test
   public void testMultipleEnableComplexKeyRegistrationHasNoEffect() {
     Type type = new TypeToken<Map<Point, String>>() {}.getType();
     Gson gson =


Marcono1234:
The `NullPointerException` has been fixed by #3057.

@eamonnmcmanus, should `null` keys be rejected nonetheless (as proposed by this PR here) or should they be allowed now?

eamonnmcmanus:
> @eamonnmcmanus, should `null` keys be rejected nonetheless (as proposed by this PR here) or should they be allowed now?

It's an interesting question. I don't see a strong reason why we should reject null keys. Users might then find that they get an NPE when they try to convert the `Map` we give them into a null-hostile map (like `ConcurrentHashMap`), but I think that's probably the right behaviour.

It's not clear to me if the null case can arise with object-encoded maps, like `{null: 17}`. That's not valid JSON. Maybe if a `TypeAdapter` returns null? But again I don't think we should then reject the null.

If @ATKasem wants to convert this PR into tests that verify these various cases then I think that would be useful.

Marcono1234:
> If @ATKasem wants to convert this PR into tests that verify these various cases then I think that would be useful.

Have created #3068 to extend the tests (had these changes already in an older PR). But in case there is still room for improvement, please either of you let me know.

I hope that is ok; thanks nonetheless @ATKasem for your effort!

ATKasem:
Agreed — with #3057 fixing the NPE and the consensus that `null` keys should be allowed rather than rejected, the premise of this PR no longer holds. #3068 covers the test cases @eamonnmcmanus suggested, so there's nothing left here worth keeping.

Closing. Thanks @Marcono1234 and @eamonnmcmanus for working through it.

PR #3057, opened 2026-07-13
Use LinkedHashMap as the default Map.

### LINKED ISSUE #1247: Confusing Exceptions

### LINKED ISSUE #2737: Switch from Gson `LinkedTreeMap` to JDK `LinkedHashMap` as Map implementation

### LINKED ISSUE #3037: Uncaught NullPointerException ("key == null") when deserializing a JSON object with a null key into a Map

eamonnmcmanus:
The JDK and Android have had DoS protection at least as good as LinkedTreeMap for over a decade now.

Fixes #3037. Fixes #2737. Fixes #1247.

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/MapTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapTest.java
@@ -166,6 +166,14 @@ public class MapTest {
   }
 
   @Test
+  public void testMapDeserializationWithNullKeyAndListOfEntries() {
+    Type typeOfMap = new TypeToken<Map<String, Integer>>() {}.getType();
+    Map<String, Integer> map = gson.fromJson("[[null,123]]", typeOfMap);
+    assertThat(map).hasSize(1);
+    assertThat(map.get(null)).isEqualTo(123);
+  }
+
+  @Test
   public void testMapSerializationWithIntegerKeys() {
     Map<Integer, String> map = new LinkedHashMap<>();
     map.put(123, "456");
--- a/gson/src/test/java/com/google/gson/functional/MapTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapTest.java
@@ -233,33 +233,6 @@ public class MapTest {
   }
 
   @Test
-  public void testMapStringSupertypeKeyDeserialization() {
-    // Should only use Gson's LinkedTreeMap for String as key, but not for supertypes (e.g. Object)
-    Type typeOfMap = new TypeToken<Map<Object, Integer>>() {}.getType();
-    Map<?, ?> map = gson.fromJson("{\"a\":1}", typeOfMap);
-
-    assertWithMessage("Map<Object, ...> should not use Gson Map implementation")
-        .that(map)
-        .isNotInstanceOf(LinkedTreeMap.class);
-
-    Map<?, ?> expectedMap = Collections.singletonMap("a", 1);
-    assertThat(map).isEqualTo(expectedMap);
-  }
-
-  @Test
-  public void testMapNonStringKeyDeserialization() {
-    Type typeOfMap = new TypeToken<Map<Integer, Integer>>() {}.getType();
-    Map<?, ?> map = gson.fromJson("{\"1\":1}", typeOfMap);
-
-    assertWithMessage("Map<Integer, ...> should not use Gson Map implementation")
-        .that(map)
-        .isNotInstanceOf(LinkedTreeMap.class);
-
-    Map<?, ?> expectedMap = Collections.singletonMap(1, 1);
-    assertThat(map).isEqualTo(expectedMap);
-  }
-
-  @Test
   public void testHashMapDeserialization() {
     Type typeOfMap = new TypeToken<HashMap<Integer, String>>() {}.getType();
     HashMap<Integer, String> map = gson.fromJson("{\"123\":\"456\"}", typeOfMap);
--- a/gson/src/test/java/com/google/gson/functional/MapTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapTest.java
@@ -220,12 +220,6 @@ public class MapTest {
     Type typeOfMap = new TypeToken<Map<String, Integer>>() {}.getType();
     Map<?, ?> map = gson.fromJson("{\"a\":1}", typeOfMap);
 
-    assertWithMessage(
-            "Map<String, ...> should use LinkedTreeMap to protect against DoS in older JDK"
-                + " versions")
-        .that(map)
-        .isInstanceOf(LinkedTreeMap.class);
-
     Map<?, ?> expectedMap = Collections.singletonMap("a", 1);
     assertThat(map).isEqualTo(expectedMap);
   }
--- a/gson/src/test/java/com/google/gson/functional/MapTest.java
+++ b/gson/src/test/java/com/google/gson/functional/MapTest.java
@@ -17,7 +17,6 @@
 package com.google.gson.functional;
 
 import static com.google.common.truth.Truth.assertThat;
-import static com.google.common.truth.Truth.assertWithMessage;
 import static org.junit.Assert.assertThrows;
 
 import com.google.gson.Gson;
@@ -33,7 +32,6 @@ import com.google.gson.JsonSerializer;
 import com.google.gson.JsonSyntaxException;
 import com.google.gson.common.TestTypes;
 import com.google.gson.internal.GsonTypes;
-import com.google.gson.internal.LinkedTreeMap;
 import com.google.gson.reflect.TypeToken;
 import java.lang.reflect.Type;
 import java.util.AbstractMap;
--- a/gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
+++ b/gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
@@ -158,15 +158,12 @@ public class ConstructorConstructorTest {
 
   @Test
   public void testStringMapCreation() {
-    // When creating raw Map should use Gson's LinkedTreeMap, assuming keys could be String
     Object actual = constructorConstructor.get(TypeToken.get(Map.class)).construct();
-    assertThat(actual).isInstanceOf(LinkedTreeMap.class);
+    assertThat(actual).isInstanceOf(LinkedHashMap.class);
 
-    // When creating a `Map<String, ...>` should use Gson's LinkedTreeMap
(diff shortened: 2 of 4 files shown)


gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
Marcono1234: Maybe this comment is overly specific or a bit misleading now, given that `LinkedHashMap` is always used?

gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
```diff
@@ -158,15 +158,12 @@ public void testCustomCollectionInterfaceCreation() {
 
   @Test
   public void testStringMapCreation() {
```
Marcono1234: Maybe rename this test method as well to make it more generic?

```suggestion
  public void testMapCreation() {
```

Previously the test was specific to how `Map<String, ...>` was handled, but now with all the comments removed the test method name might be a bit misleading.

gson/src/test/java/com/google/gson/functional/MapTest.java
Marcono1234: Might be good to remove or adjust these test methods (`testMapStringSupertypeKeyDeserialization` and `testMapNonStringKeyDeserialization`) as well, since they might be a bit misleading now?

Marcono1234 (commented):
Thanks a lot for addressing this!

Maybe some of the tests need some further adjustments to avoid confusion, see my review comments. What do you think?

gson/src/test/java/com/google/gson/functional/MapTest.java
eamonnmcmanus: Thanks, yes.

gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
eamonnmcmanus: Yes, good point. Probably the test could be simplified further, but I've just renamed it and removed a no-longer-relevant comment.

gson/src/test/java/com/google/gson/internal/ConstructorConstructorTest.java
eamonnmcmanus: Done.

cpovirk (approved):
> The JDK and Android have had DoS protection at least as good as LinkedTreeMap for over a decade now.

As https://github.com/google/gson/issues/2737 says, the Android protection dates back to (though Code Search / Git makes it harder than it should be to trace)...

https://cs.android.com/android/_/android/platform/libcore/+/9efb6d12ce4d2ffedb73d6e9887ea2c89f8ec129

It landed in time for 8.0.0...

https://cs.android.com/android/platform/superproject/+/android-8.0.0_r1:libcore/ojluni/src/main/java/java/util/HashMap.java;l=143;drc=15a062b16485b913fbca72ce5ebafb9e16aab284

...but not 7.1.2...

https://cs.android.com/android/platform/superproject/+/android-7.1.2_r39:libcore/ojluni/src/main/java/java/util/HashMap.java;l=134;drc=d2449bb576ad1e3a3877364e5e1ae28625f69e35

So, you know, Android has had the protection for only _almost nine_ years :)

https://en.wikipedia.org/wiki/Android_version_history#Overview

Google's standard minSdkVersion will soon be Nougat (7.0), so we still won't be quite there. But generally, the ability to DoS an Android app by sending it hostile data is less of a concern than the server side is, and devices on Nougat-like versions of Android quite probably have bigger security concerns than GSON maps :)

PR #3059, opened 2026-07-17
Fix JsonPrimitive.hashCode to honor the equals/hashCode contract (#992)

### LINKED ISSUE #992: JsonPrimitive with LazilyParsedNumber integers breaks contract for hashCode()

earfman:
Fixes #992.

### Problem

`JsonPrimitive.equals` considers numbers with the same value but different `Number` backing types equal — e.g. `new JsonPrimitive(42)`, a lazily-parsed `"42"` (the type produced by round-tripping a number through JSON), and `new JsonPrimitive(42.0)` are all equal to one another. But `hashCode` branched on `isIntegral(this)`:

```java
if (isIntegral(this)) {
  long value = getAsNumber().longValue();
  return (int) (value ^ (value >>> 32));
}
if (value instanceof Number) {
  long value = Double.doubleToLongBits(getAsNumber().doubleValue());
  return (int) (value ^ (value >>> 32));
}
```

An integral primitive (`42`, hashed as a `long`) and an equal non-integral one (lazily-parsed `"42"`, hashed via `doubleToLongBits`) therefore get different hash codes, violating the `Object.hashCode` contract. In practice this breaks `HashMap`/`HashSet`: a value stored under `new JsonPrimitive(42)` can't be found using the equal primitive you get back after parsing the same number from JSON.

### Fix

Derive the hash code solely from the `double` value that every branch of `equals` ultimately compares:

- Mathematically-integral values within `long` range are hashed as the `long` (this keeps the hash unchanged for the common integral types, and normalizes `-0.0` to `0`, which `equals` treats as equal to `0`).
- Everything else is hashed via `doubleToLongBits(doubleValue)` (which canonicalizes NaN, matching `equals`).

`equals` is left untouched.

### Scope / disclosure

- **Some hash codes change — by design.** The old `hashCode` used two different algorithms (integral → hash the `long`; otherwise → hash `doubleToLongBits`), which is the root cause of the bug. Now anything integral-valued and in `long` range hashes as the `long`. Concretely: an integral-valued floating primitive like `42.0` now hashes like `42`; `-0.0` now hashes like `0`; and an integral value with magnitude `> 2^53` (a `Long`/`BigInteger` a `double` can't represent exactly) now hashes by its `double` value, so `9007199254740993L` matches `9007199254740992.0`. In every one of those cases the new hash agrees with another primitive that `equals` already considers equal. Purely fractional `Double`/`Float`/`BigDecimal` values are unchanged. Hash codes aren't part of gson's serialized contract, but I'm flagging the change explicitly.
- **Pre-existing, out of scope:** `equals` itself is non-transitive around the `2^53` boundary (two distinct `long`s that both round to the same `double` are each "equal" to that `double`, but not to each other). This PR does not touch `equals` and does not try to fix that; it only makes `hashCode` consistent with `equals` as currently defined. Happy to file a separate issue if that's useful.

### Testing

- New `testEqualsAndHashCodeAcrossNumberTypes` asserts equal-implies-equal-hashcode across `Integer`/`LazilyParsedNumber`/`Double`/`BigDecimal`/`BigInteger`, the JSON round-trip case from the issue, `-0.0` vs `0`, and the `> 2^53` long/double case.
- Full `gson` module test suite passes (`mvn test`); Spotless clean.

<sub>Authored with AI assistance; the change was independently reviewed and verified before submission (including a fuzz check confirming no `equals`-equal pair produces differing hash codes). I'll sign the Google CLA.</sub>

### DIFF
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -262,14 +262,24 @@ public final class JsonPrimitive extends JsonElement {
     if (value == null) {
       return 31;
     }
-    // Using recommended hashing algorithm from Effective Java for longs and doubles
-    if (isIntegral(this)) {
-      long value = getAsNumber().longValue();
-      return (int) (value ^ (value >>> 32));
-    }
     if (value instanceof Number) {
-      long value = Double.doubleToLongBits(getAsNumber().doubleValue());
-      return (int) (value ^ (value >>> 32));
+      // equals(Object) can consider primitives backed by different Number classes equal (for
+      // example Integer 42, a lazily parsed number 42 and Double 42.0), and every case in which
+      // equals(Object) considers two numbers equal implies that they have the same `double` value.
+      // Therefore the hash code must be derived only from the `double` value, otherwise equal
+      // primitives could have different hash codes, see https://github.com/google/gson/issues/992
+      double doubleValue = getAsNumber().doubleValue();
+      long longValue = (long) doubleValue;
+      if (doubleValue == longValue) {
+        // Mathematically integral values within `long` range use the hashing algorithm recommended
+        // by Effective Java for longs; this keeps the hash code unchanged for the common integral
+        // types and also normalizes -0.0 to 0.0 (which equals(Object) considers equal)
+        return (int) (longValue ^ (longValue >>> 32));
+      }
+      // doubleToLongBits (unlike doubleToRawLongBits) collapses all NaN bit patterns to a single
+      // canonical value, matching equals(Object) which considers all NaN values equal
+      long bits = Double.doubleToLongBits(doubleValue);
+      return (int) (bits ^ (bits >>> 32));
     }
     return value.hashCode();
   }
--- a/gson/src/main/java/com/google/gson/JsonPrimitive.java
+++ b/gson/src/main/java/com/google/gson/JsonPrimitive.java
@@ -262,26 +262,45 @@ public final class JsonPrimitive extends JsonElement {
     if (value == null) {
       return 31;
     }
+    // The conditions below parallel the structure of equals(Object). Unlike in equals, every
+    // numeric branch must delegate to the same double-based hash: equals can consider primitives
+    // from *different* branches equal (for example Integer 42 equals Double 42.0 and a lazily
+    // parsed number 42 through the floating-point comparison), so a branch hashing anything other
+    // than the double value would give equal primitives different hash codes, see
+    // https://github.com/google/gson/issues/992
+    if (isIntegral(this)) {
+      // equals compares two integral primitives by their BigInteger values or long values;
+      // primitives equal under either comparison also have the same double value, so the
+      // double-based hash is consistent with this branch of equals. It is also consistent with
+      // the mixed comparisons handled below, such as Integer 42 vs Double 42.0.
+      return hashOfDoubleValue(getAsNumber().doubleValue());
+    }
     if (value instanceof Number) {
-      // equals(Object) can consider primitives backed by different Number classes equal (for
-      // example Integer 42, a lazily parsed number 42 and Double 42.0), and every case in which
-      // equals(Object) considers two numbers equal implies that they have the same `double` value.
-      // Therefore the hash code must be derived only from the `double` value, otherwise equal
-      // primitives could have different hash codes, see https://github.com/google/gson/issues/992
-      double doubleValue = getAsNumber().doubleValue();
-      long longValue = (long) doubleValue;
-      if (doubleValue == longValue) {
-        // Mathematically integral values within `long` range use the hashing algorithm recommended
-        // by Effective Java for longs; this keeps the hash code unchanged for the common integral
-        // types and also normalizes -0.0 to 0.0 (which equals(Object) considers equal)
-        return (int) (longValue ^ (longValue >>> 32));
-      }
-      // doubleToLongBits (unlike doubleToRawLongBits) collapses all NaN bit patterns to a single
-      // canonical value, matching equals(Object) which considers all NaN values equal
-      long bits = Double.doubleToLongBits(doubleValue);
-      return (int) (bits ^ (bits >>> 32));
+      // equals compares the remaining numbers by their double values. (For two BigDecimals it
+      // uses compareTo, which only considers values equal that also have the same double value.)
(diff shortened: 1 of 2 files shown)


eamonnmcmanus (changes requested):
Thanks for looking into this! I agree that it is a problem that the current `JsonPrimitive.hashCode()` violates the contract for `hashCode()`.

Would it be possible to structure the code of the rewritten `hashCode` so it is more similar to the code of `equals`? That way it would be easier to see that the rewritten code does indeed respect the contract. So in other words it would have basically the same conditions in the same order (`isIntegral { instanceof BigInteger / getAsNumber().longValue() }` etc).

gson/src/main/java/com/google/gson/JsonPrimitive.java
```diff
@@ -262,28 +262,47 @@ public int hashCode() {
...
-        // types and also normalizes -0.0 to 0.0 (which equals(Object) considers equal)
-        return (int) (longValue ^ (longValue >>> 32));
-      }
-      // doubleToLongBits (unlike doubleToRawLongBits) collapses all NaN bit patterns to a single
-      // canonical value, matching equals(Object) which considers all NaN values equal
-      long bits = Double.doubleToLongBits(doubleValue);
-      return (int) (bits ^ (bits >>> 32));
+      // equals compares the remaining numbers by their double values. (For two BigDecimals it
+      // uses compareTo, which only considers values equal that also have the same double value.)
+      return hashOfDoubleValue(getAsNumber().doubleValue());
```
eamonnmcmanus, line 281: I don't think this approach is going to work. Not all `BigDecimal` or `BigInteger` values will fit in a `double`. I suggested logic parallel to the `equals` logic because I think that is necessary.

For `BigDecimal`, I think canonicalizing via `stripTrailingZeros()` should be enough to ensure that two values for which `equals` returns true will have the same `hashCode`.

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/JsonPrimitive.java
earfman: Thanks — you're right that not every `BigDecimal`/`BigInteger` fits in a `double`, and I think that's exactly the crux, so I want to lay out carefully why the double-based hash is forced by the current `equals`, and where that leaves your `stripTrailingZeros()` suggestion.

**Why the hash must be a function of the double value (given `equals` as it stands):** every numeric `JsonPrimitive` is `equals()` to some primitive whose comparison runs through the double path. Cross-type pairs (`BigDecimal` vs `Double`, `BigInteger` vs `Double`, etc.) skip both the integral-integral branch and the both-`BigDecimal` branch, and compare `getAsDouble()`. So, for example, on current `main`:

- `new JsonPrimitive(new BigDecimal("42"))` equals `new JsonPrimitive(42)` — via the double path. With `stripTrailingZeros().hashCode()` for the `BigDecimal` branch these would hash `1302` vs `42`, a contract violation on the most ordinary input.
- `new JsonPrimitive(new BigDecimal("0.1"))` equals `new JsonPrimitive(0.1)` (same double), but `stripTrailingZeros` hashes the exact decimal `0.1`, not the double — again unequal hashes for equal values.

I did check the suggestion empirically before replying: adding `stripTrailingZeros().hashCode()` for `BigDecimal` on top of this PR's structure produces 39 equal-but-different-hash ordered pairs on a ~26-value all-pairs matrix plus 300k randomized cross-type pairs (current `main` scores 75 on the same harness; this PR scores 0). Happy to share the harness or fold its cases into the test.

**Beyond-double values can't hash distinctly under the current `equals` either:** `BigDecimal("1e400")` and `BigDecimal("2e400")` are each `equals()` to `JsonParser.parseString("1e400")` (a `LazilyParsedNumber`, whose `doubleValue()` is `+Infinity`) — so by the contract they're transitively forced to share one hash even though `compareTo` distinguishes them. That's parser-produced, so it can't be avoided by rejecting infinite `Double`s at construction. Under this PR they simply collide (legal for `hashCode`, and only reachable for magnitudes beyond ~1.8e308).

**One behavior note I should make explicit:** for the same reason, this PR changes the hashes of `Long`/`BigInteger` values above 2^53 (e.g. `Long(2^53+1)`). That's forced: on `main`, `Long(2^53+1)` equals `Double(2^53)`, so they must share a hash.

**The fork in the road:** if you'd like values beyond double precision to hash distinctly, I believe `equals` itself has to be refined first (e.g. exact numeric comparison for all `Number` pairs) — at which point `stripTrailingZeros`-style hashing becomes exactly right. That would be a visible behavior change (`Double(0.1)` would stop equaling `BigDecimal("0.1")`), but it would also fix a pre-existing non-transitivity: today `Long(2^53+1)` ~ `Double(2^53)` ~ `Long(2^53)` while `Long(2^53+1)` ≠ `Long(2^53)`. I deliberately kept `equals` untouched in this PR, but I'm happy to take that on here or in a follow-up if you'd prefer that direction — your call on scope.

eamonnmcmanus: Oh wow, what a mess!

The existing `equals` seems rather poorly conceived, not least because of the intransitivity that you mentioned. Conceptually, I think two `JsonPrimitive` instances should be equal iff their `toString()` values are equal. That would be very expensive if implemented naïvely, but we could imagine various optimizations to avoid actually having to do the `toString()`. Then the `hashCode()` could also be defined in terms of `toString()`, again with potential optimizations. It's possible that the change in semantics would break a lot of existing code, though.

With this PR, it's a bit concerning that large `long` values will have fewer distinct hashes, meaning potentially more collisions if they are used as keys in a `HashMap`. But I think that's still better than having a `hashCode()` that violates its contract.

I'm going to try and run this change past all of Google's internal tests. If that doesn't show up any problems, the change is at least an improvement.

eamonnmcmanus: The test run passed, so I think this is good to go.

eamonnmcmanus (approved):

PR #3061, opened 2026-07-20
Add StructTypeAdapter for direct Struct ↔ JsonObject conversion

### LINKED ISSUE #2945: Get JsonObject from google.protobuf.Struct
Currently the easiest way to convert a protobuf Struct to a JsonObject is to first print it to String using JsonFormat.printer().print() and the read the String using gson.fromJson method. Something like this:
```
String jsonString = JsonFormat.printer().omittingInsignificantWhitespace().print(struct);
JsonObject jsonObject = gson.FromJson(jsonString, JsonObject.class);
```

This is slightly inefficient in my opinion since there is a two step conversion and can be optimized to have a single conversion from Struct to JsonObject

# Feature description
Please, is it possible to implement a method fromProtobufStruct (or fromStruct) that converts a Struct to a JsonObject?

mayokunyusuf:
Purpose

Closes #2945.

Description

Adds StructTypeAdapter, a utility for converting directly between protobuf's Struct well-known type and Gson's JsonObject without round-tripping through an intermediate JSON string, as would be required with JsonFormat.printer() and JsonFormat.parser(). This avoids the additional parsing and serialization overhead during conversion.

fromStruct(Struct) recursively converts nested Struct, ListValue, and Value instances into the corresponding JsonObject, JsonArray, JsonPrimitive, and JsonNull. Its output matches what JsonFormat.printer().print(struct) produces when parsed back into JSON. This behavior is verified by the matchesExistingJsonFormatRoundTripForVariousStructs test.

NaN and infinite numeric values throw IllegalArgumentException because they do not have valid JSON representations.

Checklist

New code follows the Google Java Style Guide.
New public API validates arguments where necessary, including rejecting null values.
New public API includes Javadoc.
Javadoc includes the @since $next-version$ tag.
Unit tests have been added where necessary.
Unit tests use Truth assertions.
No JUnit 3 features are used.
A regression test has been added for the fixed behavior, where applicable.
mvn clean verify javadoc:jar passes without errors.

### DIFF
--- a/proto/src/test/java/com/google/gson/protobuf/functional/StructTypeAdapterTest.java
+++ b/proto/src/test/java/com/google/gson/protobuf/functional/StructTypeAdapterTest.java
@@ -22,138 +22,138 @@ import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.gson.protobuf.StructTypeAdapter;
 import com.google.protobuf.ListValue;
+import com.google.protobuf.NullValue;
 import com.google.protobuf.Struct;
 import com.google.protobuf.Value;
 import com.google.protobuf.util.JsonFormat;
 import java.util.Arrays;
 import java.util.List;
 import org.junit.Test;
-import com.google.protobuf.NullValue;
 
 public class StructTypeAdapterTest {
-    private final Gson gson = new Gson();
+  private final Gson gson = new Gson();
 
-    @Test
-    public void emptyStructConvertsToEmptyJsonObject() {
-        Struct struct = Struct.newBuilder().build();
+  @Test
+  public void emptyStructConvertsToEmptyJsonObject() {
+    Struct struct = Struct.newBuilder().build();
 
-        JsonObject json = StructTypeAdapter.fromStruct(struct);
+    JsonObject json = StructTypeAdapter.fromStruct(struct);
 
-        assertThat(json.size()).isEqualTo(0);
+    assertThat(json.size()).isEqualTo(0);
+  }
+
+  @Test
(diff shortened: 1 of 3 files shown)


proto/src/test/java/com/google/gson/protobuf/functional/StructTypeAdapterTest.java
```diff
@@ -0,0 +1,164 @@
...
+                .build());
+
+    for (Struct struct : samples) {
+      String jsonFormatString =
+          JsonFormat.printer().omittingInsignificantWhitespace().print(struct);
+      JsonObject expected = gson.fromJson(jsonFormatString, JsonObject.class);
+
+      JsonObject actual = StructTypeAdapter.fromStruct(struct);
+
+      assertThat(actual).isEqualTo(expected);
```
eamonnmcmanus: Would you mind using [`Expect`](https://truth.dev/api/latest/com/google/common/truth/Expect.html) here? That way, if one of the cases fails, it doesn't prevent seeing failures from later cases. You could also potentially do this in the other methods that have more than one `assertThat`.

eamonnmcmanus (changes requested):
Thanks for doing this! It looks great. Just one small thing.

proto/src/test/java/com/google/gson/protobuf/functional/StructTypeAdapterTest.java
mayokunyusuf: Alright, will do.

eamonnmcmanus (approved):
Thanks again!

PR #3062, opened 2026-07-21
Include location in NumberFormatException thrown by JsonTreeReader

dong0713:
## Problem

`JsonReader.nextDouble()`, `nextLong()` and `nextInt()` already include the JSON path in the `NumberFormatException` thrown when a value cannot be parsed. This was added in #3000 as a fix for #1564.

However, the corresponding methods in `JsonTreeReader` were missing this handling. As a result, the two public deserialization entry points behave inconsistently:

```java
class Model { int x; }

// Streaming path — already includes location (since #3000)
new Gson().fromJson("{\"x\":\"\"}", Model.class);
// -> JsonSyntaxException: ... Expected an int but was  at line 1 column 8 path $.x

// Tree path — no location information
JsonObject obj = new JsonObject();
obj.add("x", new JsonPrimitive(""));
new Gson().fromJson(obj, Model.class);
// -> JsonSyntaxException: ... NumberFormatException: For input string: ""
```

Both paths consume untrusted JSON and report parse failures to the caller, so they should report the same location information.

## Fix

Mirror the existing `JsonReader` handling: wrap the `JsonPrimitive` conversion in `nextDouble()`, `nextLong()` and `nextInt()` with a try-catch, and rethrow a `NumberFormatException` containing the offending value and `locationString()`, preserving the original exception as the cause (via `initCause(e)`), exactly as the streaming `JsonReader` does.

## Tests

Added three regression tests in `JsonTreeReaderTest` (`testNextDouble/NextInt/NextLongNumberFormatExceptionContainsLocation`), asserting the message contains the JSON path (`path $.x` / `path $.y`), paralleling the existing tests added for `JsonReader` in `JsonReaderTest`.

Full test suite: `Tests run: 4607, Failures: 0, Errors: 0, Skipped: 20` — no regressions.

## Notes

- This is the tree-path counterpart of #3000; `JsonTreeReader` overrides the same `JsonReader` methods, so the omission is a direct symmetric gap.
- The outer `Gson.fromJson` only catches `EOFException`/`IllegalStateException`/`IOException`, so the bare `NumberFormatException` from the tree path reaches the caller unwrapped, making the missing location directly user-visible.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
@@ -434,8 +434,7 @@ public final class JsonTreeReader extends JsonReader {
   }
 
   /** Creates a {@link NumberFormatException} whose message includes the current path. */
-  private NumberFormatException numberFormatException(
-      String message, NumberFormatException cause) {
+  private NumberFormatException numberFormatException(String message, NumberFormatException cause) {
     NumberFormatException exception = new NumberFormatException(message + locationString());
     exception.initCause(cause);
     return exception;
--- a/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
@@ -243,7 +243,13 @@ public final class JsonTreeReader extends JsonReader {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    double result = ((JsonPrimitive) peekStack()).getAsDouble();
+    JsonPrimitive primitive = (JsonPrimitive) peekStack();
+    double result;
+    try {
+      result = primitive.getAsDouble();
+    } catch (NumberFormatException e) {
+      throw numberFormatException("Expected a double but was " + primitive.getAsString(), e);
+    }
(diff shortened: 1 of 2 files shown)


gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -240,7 +240,17 @@ public double nextDouble() throws IOException {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    double result = ((JsonPrimitive) peekStack()).getAsDouble();
+    JsonPrimitive primitive = (JsonPrimitive) peekStack();
+    double result;
+    try {
+      result = primitive.getAsDouble();
+    } catch (NumberFormatException e) {
+      NumberFormatException rethrown =
```
eamonnmcmanus: It's annoying that `NumberFormatException` doesn't have a constructor with a `cause` parameter!

Given that the code pattern here occurs three times, I think it would make sense to introduce a helper method, so you could write this:

```java
throw numberFormatException("Expected a double " ..., e);
```

eamonnmcmanus (changes requested):
Thanks for helping improve consistency here!

dong0713:
Done — introduced a `numberFormatException` helper as suggested. Thanks!

eamonnmcmanus (approved):
Thanks again!

PR #3063, opened 2026-07-21
Reject non-ASCII digits in JsonTreeReader.nextInt/nextLong

dong0713:
## Summary

`JsonTreeReader.nextInt()` and `nextLong()` silently accept non-ASCII digit variants (e.g. full-width digits U+FF10–FF19, Arabic-Indic, Bengali) when a JSON **string** token is parsed as an integer/long. This is the tree-based counterpart of the streaming-reader gap fixed in #2995 (issue #2994), and was missed when that fix landed.

## Problem

Since #2995, the streaming `JsonReader.nextInt()` / `nextLong()` reject non-ASCII characters via `validateAscii()` before calling `Integer.parseInt()` / `Long.parseLong()`. But the tree-based `JsonTreeReader` — used by `Gson.fromJson(JsonElement, Type)`, `fromJsonTree(...)`, and any path that deserializes from an already-parsed `JsonElement` — has no such check. It goes straight to `JsonPrimitive.getAsInt()` → `Integer.parseInt(getAsString())`, and `Integer.parseInt` uses `Character.digit(...)`, which recognizes *any* Unicode decimal digit, not just ASCII `0-9`.

The result is a behavioral divergence for identical input:

```java
// streaming path (fixed by #2995): throws MalformedJsonException
new Gson().fromJson("{\"x\":\"１２３\"}", Payload.class);   // JsonReader

// tree-based path (this PR's bug): silently returns 123
new Gson().fromJson(JsonParser.parseString("{\"x\":\"１２３\"}"), Payload.class);
```

This is also an incomplete fix for the security concern in #2994 (validation bypass via silent Unicode-digit coercion): any caller that deserializes from a `JsonElement` rather than a raw string still gets the old, vulnerable behavior.

Confirmed on `main` (JDK 21): `Integer.parseInt("１２３")` returns `123`; `JsonTreeReader.nextInt()` on `"１２３"` returns `123` without throwing. (`Double.parseDouble` rejects non-ASCII digits, so `nextDouble()` is unaffected and intentionally not touched here — matching the scope of #2995.)

## Fix

Mirror the approach taken by #2995 in the streaming reader: on the **string** token path of `JsonTreeReader.nextInt()` / `nextLong()`, call a new `validateAscii(primitive.getAsString())` that throws `MalformedJsonException` (including `locationString()`, consistent with the rest of `JsonTreeReader`) if any character exceeds U+007F. The `NUMBER` token path is left untouched — a `JsonPrimitive` that is a `Number` already holds a legitimate numeric value, exactly as the streaming reader only validates the string-to-number path.

```java
JsonPrimitive primitive = (JsonPrimitive) peekStack();
if (token == JsonToken.STRING) {
  validateAscii(primitive.getAsString());
}
int result = primitive.getAsInt();
```

This is a strictly-tightening change with no effect on ASCII input, and keeps the tree reader consistent with the streaming reader.

## Tests

Added to `JsonTreeReaderTest`:

- `testNextIntRejectsNonAsciiDigits` — `"１２３"` via `nextInt()` throws `MalformedJsonException` whose message contains `"non-ASCII"` and `"path $.x"`.
- `testNextLongRejectsNonAsciiDigits` — same for `nextLong()` with `"path $.y"`.
- `testNextIntAndLongAcceptAsciiDigitsInString` — confirms plain ASCII `"123"` / `"456"` in string tokens still parse normally.

All 4607 existing tests pass (3 new ones added), 0 failures, 20 skipped.

## Scope

- Touches only `JsonTreeReader.java` and its test.
- Same one-method `validateAscii` shape as the streaming reader, adapted to throw `MalformedJsonException` + `locationString()` (the streaming reader uses `syntaxError()`).
- No behavioral change for ASCII input; only the tree-based string-to-int/long path is tightened to match the streaming path.

This completes the tree-based counterpart of the #2994 / #2995 fix.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
@@ -258,7 +258,11 @@ public final class JsonTreeReader extends JsonReader {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    long result = ((JsonPrimitive) peekStack()).getAsLong();
+    JsonPrimitive primitive = (JsonPrimitive) peekStack();
+    if (token == JsonToken.STRING) {
+      validateAscii(primitive.getAsString());
+    }
+    long result = primitive.getAsLong();
     popStack();
     if (stackSize > 0) {
       pathIndices[stackSize - 1]++;
@@ -273,7 +277,11 @@ public final class JsonTreeReader extends JsonReader {
       throw new IllegalStateException(
           "Expected " + JsonToken.NUMBER + " but was " + token + locationString());
     }
-    int result = ((JsonPrimitive) peekStack()).getAsInt();
+    JsonPrimitive primitive = (JsonPrimitive) peekStack();
+    if (token == JsonToken.STRING) {
+      validateAscii(primitive.getAsString());
+    }
+    int result = primitive.getAsInt();
     popStack();
     if (stackSize > 0) {
       pathIndices[stackSize - 1]++;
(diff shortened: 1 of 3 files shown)


gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -391,4 +399,18 @@ public String getPreviousPath() {
   private String locationString() {
     return " at path " + getPath();
   }
+
+  // Mirror the ASCII validation added to the streaming JsonReader in #2995:
```
eamonnmcmanus: I think I would omit this information. It won't be all that interesting a few months from now to know what happened when.

The code here duplicates a method that was introduced in #2995. Perhaps we could reduce the duplication by having `public boolean isAllAscii(String s)` in this class and calling it from the other class. Since this is an `.internal.` package, that doesn't change the API shown to users.

eamonnmcmanus (changes requested):
Thanks, this is a great idea! Just one small thing.

dong0713:
Done — extracted `public static boolean isAllAscii(String)` in `JsonTreeReader` and made `JsonReader.validateAscii` call it. Also removed the history comment. Thanks!

eamonnmcmanus (approved):
Thanks!

PR #3064, opened 2026-07-21
Don't suggest Moshi for Kotlin anymore

Marcono1234:
See https://redirect.github.com/square/moshi/issues/2079#issuecomment-4051193660
> I also don't recommend Moshi to others writing Kotlin. I recommend kotlinx.serialization because while it has many things I think are worse it has enough that's better to justify its use.

However that comment was written in March 2026 and might be outdated, and the maintainer says they are only speaking for themselves and not the other maintainers.


eamonnmcmanus (approved):
Thanks, I think this is a good recommendation.

PR #3067, opened 2026-07-22
Fix URI type adapter to preserve identity for non-ASCII URIs

### LINKED ISSUE #1587: URI type adapter does not preserve object identity
The type adapter included by default for `java.net.URI` does not preserve the identity of all URI values. In other words:

```java
URI uri = new URI("s3://bucket/path/Kankyō.png");
URI roundTripUri = gson.fromJson(gson.toJson(uri), URI.class);
assert roundTripUri.equals(uri); // AssertionError
```

The problem is that the implementation [here](https://github.com/google/gson/blob/d1828736eb425a575667e0c86117ced787d1cf5b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java#L502-L520) uses `URI#toASCIIString`, which normalizes the URI contents instead of perserving them like this:

```java
Normalizer.normalize(s, Normalizer.Form.NFC)
```

This is very bad, because it is changing the meaning of the URI on serialization. This issue has caused numerous bugs at our business dealing with international filenames being uploaded to our SAAS products and then being changed after being passed between multiple microservices.

Here is an alternate implementation that we have started to use in our code:

```java
public class NonNormalizingUriTypeAdapter extends TypeAdapter<URI> {
    @Override
    public URI read(JsonReader in) throws IOException {
        if (in.peek() == JsonToken.NULL) {
            in.nextNull();
            return null;
        }
        try {
            String nextString = in.nextString();
            return "null".equals(nextString) ? null : new URI(nextString);
        }
        catch (URISyntaxException e) {
            throw new JsonIOException(e);
        }
    }

    @Override
    public void write(JsonWriter out, URI value) throws IOException {
        out.value(value == null ? null : value.toString());
    }
}
```

Really the only difference is using `toString()` instead of `toASCIIString()`, which URL-encodes the result without changing the character byte representations.

anishesg:
The URI type adapter was calling `URI.toASCIIString()` in its `write()` method, which percent-encodes non-ASCII characters and applies Unicode normalization. The `read()` method does not reverse this percent-encoding, causing URIs with international characters to lose their original identity during round-trip serialization. For example, `s3://bucket/path/Kankyō.png` would fail an equality check after being serialized and deserialized.

The fix replaces `toASCIIString()` with `toString()` in line 733 of TypeAdapters.java. This preserves the original URI representation without normalization while remaining JSON-spec compliant, as JSON allows UTF-8 non-ASCII characters in strings. The change is backward compatible because pure-ASCII URIs produce identical output with both methods.

A regression test was added to verify that URIs with non-ASCII characters maintain equality after round-trip serialization.

Fixes #1587

### DIFF
--- a/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
+++ b/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
@@ -188,6 +188,15 @@ public class DefaultTypeAdaptersTest {
   }
 
   @Test
+  public void testUriRoundTripWithNonAscii() throws Exception {
+    URI uri = new URI("s3://bucket/path/Kankyō.png");
+    String json = gson.toJson(uri);
+    URI roundTripUri = gson.fromJson(json, URI.class);
+    assertThat(roundTripUri).isEqualTo(uri);
+    assertThat(roundTripUri.toString()).isEqualTo(uri.toString());
+  }
+
+  @Test
   public void testNullSerialization() {
     testNullSerializationAndDeserialization(Boolean.class);
     testNullSerializationAndDeserialization(Byte.class);
--- a/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
+++ b/gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
@@ -184,7 +184,7 @@ public class DefaultTypeAdaptersTest {
     String uriValue = "http://google.com/";
     String json = '"' + uriValue + '"';
     URI target = gson.fromJson(json, URI.class);
-    assertThat(target.toASCIIString()).isEqualTo(uriValue);
+    assertThat(target.toString()).isEqualTo(uriValue);
   }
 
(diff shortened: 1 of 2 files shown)


gson/src/test/java/com/google/gson/functional/DefaultTypeAdaptersTest.java
```diff
@@ -187,6 +187,15 @@ public void testUriDeserialization() {
     assertThat(target.toASCIIString()).isEqualTo(uriValue);
```
Marcono1234: For consistency use `toString()` here now as well? (for that specific URI it does not make a difference though)

```suggestion
    assertThat(target.toString()).isEqualTo(uriValue);
```

anishesg: good catch, updated

eamonnmcmanus (approved):
I ran this against all of Google's internal tests and did not find any issues. That suggests that the change in behaviour is unlikely to be a problem in practice.

PR #3075, opened 2026-07-28
Only deserialize `InetAddress` from numeric IP addresses.

eamonnmcmanus:
When *serializing* `InetAddress`, we serialize the numeric form, like `"127.0.0.1"` or `"0:0:0:0:0:0:0:1"`. Because of historic deficiencies in the Java API, Gson uses a method that deserializes these forms, but also host names
like `"localhost"`. Resolving names like that can be expensive, and given that the canonical serial form is numeric, Gson should not do it by default.

Google has many thousands of internal tests that rely directly or indirectly on Gson. We ran them all with a changed version of Gson that produced an exception on deserializing *any* non-null string to an `InetAddress`, even a numeric one. No tests failed. So apparently Google's giant source base has zero occurrences of deserializing an `InetAddress` with Gson. If, however, you do have code that does that, and if it does rely on being able to accept host names, you can set the new system property `gson.allowDnsInetAdress` to `true` to restore the previous behaviour.

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
@@ -737,16 +737,16 @@ public final class TypeAdapters {
 
   public static final TypeAdapterFactory URI_FACTORY = newFactory(URI.class, URI);
 
-  /**
-   * A pattern that matches every IP address and no DNS address. It matches plenty of things that
-   * aren't either of those, which is fine. An IPv4 address is n.n.n.n where each n is a
-   * non-negative integer. An IPv6 address contains at least one colon. (There are further
-   * constraints in both cases, but they don't matter here.)
-   */
-  private static final Pattern IP_ADDRESS_PATTERN = Pattern.compile(".*:.*|[0-9]+(\\.[0-9]+){3}");
-
   public static final TypeAdapter<InetAddress> INET_ADDRESS =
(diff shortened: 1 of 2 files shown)


kluever (approved):
LGTM

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -738,17 +739,32 @@ public void write(JsonWriter out, URI value) throws IOException {
 
   public static final TypeAdapter<InetAddress> INET_ADDRESS =
       new TypeAdapter<InetAddress>() {
+
+        // A pattern that matches every IP address and no DNS address. It matches plenty of things
+        // that aren't either of those, which is fine. An IPv4 address is n.n.n.n where each n is a
+        // non-negative integer. An IPv6 address contains at least one colon. (There are further
+        // constraints in both cases, but they don't matter here.)
+        private final Pattern ipAddressPattern = Pattern.compile(".*:.*|[0-9]+(\\.[0-9]+){3}");
```
Marcono1234, line 747: ~~Should this be potentially more lenient for user-provided IP address strings (and not obtained from `InetAddress#getHostAddress`)? Should it at least allow IPv6 enclosed by`[` and `]`?~~

Edit: Nevermind, sorry. The `.*` do match the `[` and `]`, so that should work.

PR #3076, opened 2026-07-29
Make `LazilyParsedNumber` implement `Comparable`.

eamonnmcmanus:
This means that it can be put in a `TreeSet` or used as a key in a `TreeMap`, though only if the other elements or keys are also instances of `LazilyParsedNumber`. Typically that will be the case if they all came from the same place, for example by parsing incoming JSON in the same way.


kluever (approved):
LGTM

PR #3081, opened 2026-08-03
Address https://docs.zizmor.sh/audits/#artipacked findings.

cpovirk:
Google has been rolling out `zizmor` to more of its repos, and `zizmor` wants `persist-credentials: false`.

(Gemini leads me to believe that the risk without `persist-credentials: false` has been minimal since [version 6](https://github.com/actions/checkout/releases/tag/v6.0.0) (https://github.com/actions/checkout/pull/2286?). I haven't tried to verify that. It [doesn't seem like the default `persist-credentials` value is likely to change](https://github.com/actions/checkout/issues/485).)

`zizmor` also wants us to pin `google/oss-fuzz`. I posted https://github.com/google/oss-fuzz/issues/6836#issuecomment-5169599063.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
       - name: "Set up JDK ${{ matrix.java }}"
         uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95  # v5.6.0
         with:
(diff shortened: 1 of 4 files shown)


eamonnmcmanus (approved):

Marcono1234:
> `zizmor` also wants us to pin `google/oss-fuzz`. I posted [google/oss-fuzz#6836 (comment)](https://github.com/google/oss-fuzz/issues/6836#issuecomment-5169599063)

It doesn't look like the situation with oss-fuzz will change any time soon, and pinning to an arbitrary commit might prevent Dependabot from updating the commit ref (see comments on the oss-fuzz issue).
The workflow using it is dedicated to just running oss-fuzz and has quite minimal permissions, so maybe adding a suppression comment is fine?

The current approach of just not updating the `cifuzz.yml` workflow at all (see also commits of #3107) seems problematic in the long run.

PR #3083, opened 2026-08-03
Fix LegacyProtoTypeAdapterFactoryTest bitfield regex pattern.

eamonnmcmanus:
This will allow the test to pass even after some upcoming changes to the internal representation of Java protos.

This is internal cl/958647397.

### DIFF
--- a/proto/src/test/java/com/google/gson/protobuf/functional/LegacyProtoTypeAdapterFactoryTest.java
+++ b/proto/src/test/java/com/google/gson/protobuf/functional/LegacyProtoTypeAdapterFactoryTest.java
@@ -199,6 +199,7 @@ public final class LegacyProtoTypeAdapterFactoryTest {
     Pattern pattern =
         Pattern.compile(
             "^(memoized.?(Hash.?Code|Is.?Initialized|Size))|unknown.?Fields"
+                + "|bit.?[fF]ield.*"
                 + "|.*Memoized.?Serialized.?Size$",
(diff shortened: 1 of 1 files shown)


proto/src/test/java/com/google/gson/protobuf/functional/LegacyProtoTypeAdapterFactoryTest.java
Marcono1234, line 209: Not directly related, but instead of collecting the keys in `keysToRemove` and then doing `keySet().removeAll(keysToRemove)`, would it be possible to directly do `keySet().removeIf(...)`?

eamonnmcmanus: Yes, I think so. Obviously if we can do streams we can do `removeIf`. Feel free to add that change in here, or we can make a separate PR.

eamonnmcmanus: Oh I got confused, I'm the one who could make this change. I think I'll defer it, though.

Marcono1234: Have created #3124 for this now.

cpovirk:
(apologies for never getting around to providing the requested review on this, especially given how trivial it was)

PR #3084, opened 2026-08-05
Delete Maven binary

Marcono1234:
Was accidentally added by #3049

Alternatively, is force pushing an option to remove that commit? It would be really unfortunate if all future contributors have to pay this redundant ~ 9MB.

(Side note: This was partially my fault because I suggested that the contributor should remove their specific gitignore entry again (https://github.com/google/gson/pull/3049#discussion_r3491623382). I still think that comment was appropriate; `.gitignore` should ideally not contain user-specific ignores, but I had not considered that the user might then accidentally commit the file, and that it would accidentally be overlooked during review.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,13 @@ updates:
       # Name is used for branch name and pull request title
       maven:
         patterns:
-        # Create a single pull request for all dependencies and plugins
-        - "*"
+          # Create a single pull request for all dependencies and plugins
+          - "*"
+        exclude-patterns:
+          # Exclude JUnit 6 from the group (that is, create a separate PR for it)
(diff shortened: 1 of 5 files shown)


Marcono1234:
@eamonnmcmanus, would it be an option to rewrite (and force push) the currently last 5 commits on `main` to remove `apache-maven-3.9.6-bin.tar.gz` from `main` completely?

eamonnmcmanus:
Well this is a bit unfortunate. I should have noticed that in review. But 9MB is not all that much in the global scheme of things, and the [long list of warnings](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository#side-effects-of-rewriting-history) about removing files from a repo makes me think it is not worth the potential upheaval.

eamonnmcmanus (approved):

PR #3085, opened 2026-08-05
Update code and docs for minimum Android API level 24

Marcono1234:
### Purpose
Follow-up for #3082

### Description
- Updates the README with the minimum Android API level
- Updates code to use API available since API level 24 (does **not** include #3046)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ see [`GsonBuilder.disableJdkUnsafe()`](https://javadoc.io/doc/com.google.code.gs
 #### Minimum Android API level
 
 - Gson 2.15.0 and newer: API level 24
-- Gson 2.14.0 and newer: API level 23
 - Gson 2.11.0 and newer: API level 21
 - Gson 2.10.1 and older: API level 19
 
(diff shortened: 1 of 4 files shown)


gson/src/main/java/com/google/gson/internal/NonNullElementWrapperList.java
```diff
@@ -117,6 +119,16 @@ public <T> T[] toArray(T[] a) {
...
 
+  @Override
+  public void sort(Comparator<? super E> c) {
+    delegate.sort(c);
+  }
+
+  @Override
+  public Spliterator<E> spliterator() {
+    return delegate.spliterator();
+  }
```
Marcono1234, line 130: These are already covered by the tests in `JsonArrayAsListTest`.

README.md
```diff
@@ -69,6 +69,7 @@ see [`GsonBuilder.disableJdkUnsafe()`](https://javadoc.io/doc/com.google.code.gs
 
 #### Minimum Android API level
 
+- Gson 2.15.0 and newer: API level 24
```
eamonnmcmanus, line 72: I had been planning on releasing 2.14.1 next, but I think you're right that the change in Android SDK level implies upping the minor level.

eamonnmcmanus (approved):
Thanks for doing this! I had meant to update the documentation, but I hadn't thought of all those other changes.

omusab476-web (approved):

PR #3094, opened 2026-08-14
fix: JsonTreeReader nextInt/nextLong silently truncates fractional numbers

### LINKED ISSUE #2817: Inconsistency JsonReader vs JsonTreeReader about double precision loss

Jerryyy985:
## Summary
`JsonTreeReader.nextInt()/nextLong()` call `Number.intValue()/longValue()` directly, which silently truncates fractional values (`42.123` → `42`), while the streaming `JsonReader` throws `NumberFormatException` for the same input. Behavior between the two readers is inconsistent.

## Fix
Mirror `JsonReader`'s precision check: for NUMBER tokens, reject values whose double representation is not exactly equal to the requested integer type (`result != getAsDouble()` guard, same as `JsonReader.nextInt/long`).

Verified against a 12-case boundary matrix (fractional / integral / overflow / NaN / BigInteger / BigDecimal paths) — TreeReader now matches JsonReader behavior with no regressions on valid inputs.

Fixes #2817


gson/src/main/java/com/google/gson/internal/bind/JsonTreeReader.java
```diff
@@ -274,6 +274,10 @@ public long nextLong() throws IOException {
     long result;
     try {
       result = primitive.getAsLong();
+      if (token == JsonToken.NUMBER && result != primitive.getAsDouble()) {
+        // Mirror JsonReader: reject numbers that cannot be exactly represented as a long
+        throw new NumberFormatException("Expected a long but was " + primitive.getAsString());
```
Marcono1234, line 279: Probably have to move this out of the `try-catch(NumberFormatException)`, otherwise this exception is redundantly wrapped.

(same for `nextInt` below as well)

Marcono1234 (commented):
Thanks!

Could you please also add tests (probably in `JsonTreeReaderTest`)?
And do you think you will be able to sign the Contributor License Agreement?

PR #3095, opened 2026-08-14
fix: AtomicIntegerArray null element produces misleading error

### LINKED ISSUE #3047: AtomicIntegerArray deserializer crashes on JSON null (IllegalStateException) — symmetric fix to #3038

Jerryyy985:
## Summary
`ATOMIC_INTEGER_ARRAY` reads elements via raw `in.nextInt()` and only catches `NumberFormatException`; a JSON null element escapes as `IllegalStateException` wrapped in a misleading `adapter-not-null-safe` JsonSyntaxException (with a wrong troubleshooting hint).

## Fix
Mirror `atomicLongArrayAdapter` (from #3038): factory-based adapter reading elements through a Number TypeAdapter, throwing a clean `JsonSyntaxException("null is not a valid AtomicIntegerArray element")` on null. Registered symmetrically in GsonBuilder.

Verified: `[1,null,3]` now produces the same clean error as AtomicLongArray; normal/empty/top-level-null paths unaffected.

Fixes #3047


Marcono1234:
Thanks!

Instead of delegating to the `TypeAdapter` for `Integer`, an alternative would be to add a `in.peek() == JsonToken.NULL` check before calling `in.nextInt()`. That would most likely be more efficient because it avoids boxing.

----

> wrapped in a misleading `adapter-not-null-safe` JsonSyntaxException (with a wrong troubleshooting hint)

The [hint](https://github.com/google/gson/blob/main/Troubleshooting.md#adapter-not-null-safe) says though "a built-in adapter does not support JSON null values", so it might not be completely misleading. But any PR for improving the Troubleshooting Guide is welcome!

----

My main concern here is if this change is really needed. #3038 was definitely an issue because it fixed a `NullPointerException`, which would have had no detail information for the user.
But here for `AtomicIntegerArray` this is only about a slight error message improvement.

I suspect only few users actually deserialize `AtomicIntegerArray`, and even fewer would encounter this error. And the current exception message is probably still somewhat useful.

~~Additionally, this code pattern of directly calling a `JsonReader` method (`nextInt()`, `nextBoolean()`, ...) without explicit check for JSON null and custom error reporting first is pretty common in the Gson code base. Adjusting just one specific case seems a bit inconsistent; and trying to adjust all seems unnecessary.~~
Edit: Maybe I am wrong with this, and it only affects a few adapters which read arrays and objects; specifically `TypeAdapters#ATOMIC_INTEGER_ARRAY` (which you are changing in this PR), `TypeAdapters$IntegerFieldsTypeAdapter`, `JavaTimeTypeAdapters`. But still not completely sure if this is worth it.

(Note that I am not a direct member of this project; this is my personal opinion on this.)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -337,40 +337,40 @@ public void write(JsonWriter out, AtomicBoolean value) throws IOException {
...
+    Objects.requireNonNull(intAdapter);
+    return new TypeAdapter<AtomicIntegerArray>() {
+      @Override
+      public AtomicIntegerArray read(JsonReader in) throws IOException {
+        List<Integer> list = new ArrayList<>();
+        in.beginArray();
+        while (in.hasNext()) {
+          Number value = intAdapter.read(in);
+          if (value == null) {
+            throw new JsonSyntaxException("null is not a valid AtomicIntegerArray element");
```
Marcono1234, line 351: Would probably be good to include `in.getPreviousPath()` here in the exception messsage, see also #3096

PR #3096, opened 2026-08-14
Fix some adapter exceptions not including JSON document location

Marcono1234:
### Purpose
Improve exceptions & exception messages

### Description
Some of the exceptions thrown by built-in adapters did not include the JSON document location, making it difficult for the user to troubleshoot.


gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
```diff
@@ -187,10 +187,11 @@ public Map<K, V> read(JsonReader in) throws IOException {
         while (in.hasNext()) {
           in.beginArray(); // entry array
           K key = keyTypeAdapter.read(in);
-          V value = valueTypeAdapter.read(in);
           if (map.containsKey(key)) {
-            throw new JsonSyntaxException("duplicate key: " + key);
+            throw new JsonSyntaxException(
+                "Duplicate key '" + key + "'; at path " + in.getPreviousPath());
```
Marcono1234, line 192: Also changed message to start with capital letter to be consistent with the other exception messages

gson/src/main/java/com/google/gson/internal/bind/MapTypeAdapterFactory.java
```diff
@@ -187,10 +187,11 @@ public Map<K, V> read(JsonReader in) throws IOException {
...
           }
+          V value = valueTypeAdapter.read(in);
```
Marcono1234, line 194: Moved reading of the value down so that `getPreviousPath()` is more accurate (pointing to the key instead of the value)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -723,11 +729,15 @@ public URI read(JsonReader in) throws IOException {
             in.nextNull();
             return null;
           }
+          String nextString = in.nextString();
+          if (nextString.equals("null")) {
+            return null;
+          }
```
Marcono1234, line 744: Moved this out of the `try-catch (URISyntaxException)`

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -723,11 +729,15 @@ public URI read(JsonReader in) throws IOException {
...
           try {
-            String nextString = in.nextString();
-            return nextString.equals("null") ? null : new URI(nextString);
+            return new URI(nextString);
           } catch (URISyntaxException e) {
-            throw new JsonIOException(e);
+            throw new JsonSyntaxException(
+                "Failed parsing '" + nextString + "' as URI; at path " + in.getPreviousPath(), e);
           }
```
Marcono1234, line 750: Changed this from throwing `JsonIOException` to `JsonSyntaxException` for consistency; this was the only case where `JsonIOException` was thrown in this class.

(Technically this is a behavior change, though it is probably unlikely that any user relies on this exact exception.)

gson/src/main/java/com/google/gson/internal/bind/TypeAdapters.java
```diff
@@ -704,7 +711,15 @@ public URL read(JsonReader in) throws IOException {
             return null;
           }
           String nextString = in.nextString();
-          return nextString.equals("null") ? null : new URL(nextString);
+          if (nextString.equals("null")) {
+            return null;
+          }
+          try {
+            return new URL(nextString);
+          } catch (MalformedURLException e) {
```
Marcono1234, line 719: Added handling for (~~unchecked~~[^1]) `MalformedURLException`, for better exception message and for consistency with `URI` handling (though there the exception is a checked one)

[^1]: Edit: Actually a checked exception but a subclass of `IOException` and therefore needed no `try-catch` because the enclosing `read` method already has `throws IOException`.

eamonnmcmanus (approved):
Thanks, this looks great!

I'm guessing that some people will have tests that need to be updated because of these changes, but those tests are fragile anyway. I'm guessing that that includes some of Google's internal tests, which I can easily update.

PR #3101, opened 2026-08-25
Use `switch` statements instead of if-chains.

eamonnmcmanus:
This is mostly cosmetic. There is unlikely to be a significant performance difference either way, but `switch` statements can be a bit easier to read because you don't need to check that every case is comparing the same thing.


gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -2217,24 +2216,38 @@ private static void assertDocument(String document, Object... expectations) thro
...
+          case STRING:
+            assertThat(reader.nextString()).isEqualTo("string");
+            break;
+          case NUMBER:
+            assertThat(reader.nextInt()).isEqualTo(123);
+            break;
+          case NULL:
+            reader.nextNull();
+            break;
+          default:
```
cpovirk, line 2248:
```suggestion
          case END_DOCUMENT:
```

Is it worth having exhaustive cases here instead of a `default`, since it's easy? This seems like a case in which it's academic because JSON is not going to change?

cpovirk (approved):

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
eamonnmcmanus: Yeah, not sure it really matters all that much given that this is just a test helper method.

PR #3102, opened 2026-08-26
Add JsonReader.getCharacterOffset to track stream position

barakhub:
### Purpose

Adds `getCharacterOffset()` to `JsonReader` to allow retrieving the current absolute character position in the JSON stream. 

This enables stream-reading large (e.g. ~2GB) JSON files to collect index positions of specific elements, and later efficiently direct-read specific parts of the JSON file based on those collected positions.

### Description

* Adds `getCharacterOffset()` to `JsonReader` which tracks and returns the 0-based character offset from the beginning of the stream.
* Implements the method in `JsonTreeReader` to return `-1L` (since it does not read from a character stream).
* Additional overhead is minimal (one 64-bit integer addition for every ~1K of characters).
* This does implicitly "limit" the size of the stream to `2^63-1` characters, which is highly reasonable for current JSON use cases.

### Checklist


eamonnmcmanus (approved):
This looks great. (I already reviewed it internally at Google.)

PR #3104, opened 2026-08-31
Support building with JDK 25

Marcono1234:
### Purpose
Support building with JDK 25

Follow-up for #2847 and #2921

### Description
Previously building with JDK 25 was possible (and was done by CI), but was not officially supported.

The underlying issue is that recent JDK builds (at least those from Temurin) do not include `jmod` files anymore but ProGuard [still depends on them](https://redirect.github.com/Guardsquare/proguard/issues/473).

This PR here now adjusts the Maven build config so that it gracefully skips ProGuard and related tests when building with JDK 25 but `jmod` files are missing.
Additionally it uses `actions/setup-java`'s new [`jmod` file support](https://redirect.github.com/actions/setup-java/pull/1149) to actually test JDK 25 with `jmod` files being present.

This means it runs now intentionally two tests for JDK 25 on CI:
- without `jmod` files: this matches the behavior if a user would use Temurin JDK 25, and ensures that the build works
- with `jmod` files: this verifies that a full build including ProGuard-related tests works for JDK 25, and also matches the behavior if a user would use a JDK distribution which includes `jmod` files

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,19 +17,26 @@ env:
 
 jobs:
   build:
-    name: "Build on JDK ${{ matrix.java }}"
+    name: "Build on JDK ${{ matrix.java }} (${{ matrix.java-package }})"
     strategy:
       matrix:
-        java: [ 17, 21 ]
+        java: [ 17, 21, 25 ]
+        java-package: [ jdk ]
         include:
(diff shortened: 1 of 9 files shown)


eamonnmcmanus (approved):
Thanks! That jmod business seems not to have a trivial!

PR #3109, opened 2026-09-04
Fix date-only parsing failure when local midnight is skipped by DST

### LINKED ISSUE #2539: Bug Parse Date 1966-11-01
# Gson version

Version: 2.10.1

# Java / Android version

Android target sdk 33

# Used tools

- [ ] Maven; version: 
- [X] Gradle; version: 
- [ ] ...

# Description

Failure in parse date '1966-11-01'

## Expected behavior

Failed parsing '1966-11-01' as Date; at path $.birthday
image

If it's a day's difference from success

image

zgj-ssslab:
## Description
Parsing a date-only string (e.g. `"1966-11-01"`) fails with `IllegalArgumentException: HOUR_OF_DAY: 0 -> 1` when the default time zone skips midnight on that date due to a DST transition (e.g. `America/Sao_Paulo` in 1966).

## Root cause
For date-only input, `ISO8601Utils.parse` constructs a `GregorianCalendar` for local midnight in the default time zone with `setLenient(false)`. When a DST transition shifts clocks from 0:00 to 1:00, that wall time does not exist and the non-lenient calendar rejects it — even though the calendar date itself is valid.

## Fix
When the strict calendar fails, resolve leniently **only if** the date fields (year/month/day) are unchanged by the lenient resolution — i.e. the date exists but its midnight was skipped; the result is the first existing instant of that day, matching the semantics of `java.time`'s `LocalDate.atStartOfDay(ZoneId)`. Genuinely invalid dates (e.g. `2021-02-30`) shift to a different day under lenient resolution and keep failing, preserving the strict behavior.

## How I checked
- Reproduced the failure with `America/Sao_Paulo` + `1966-11-01`; verified all other time zones/dates unchanged
- Added regression tests: DST-gap date parses to the first existing instant; `2021-02-30`, `2021-13-01`, `1966-00-01` still fail
- Full module test suite passes (4635 tests)

Fixes #2539

### DIFF
--- a/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
+++ b/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
@@ -178,9 +178,26 @@ public final class ISO8601Utils {
       if (!hasT && (date.length() <= offset)) {
         Calendar calendar = new GregorianCalendar(year, month - 1, day);
         calendar.setLenient(false);
-
-        pos.setIndex(offset);
-        return calendar.getTime();
+        try {
+          pos.setIndex(offset);
+          return calendar.getTime();
+        } catch (IllegalArgumentException e) {
+          // The calendar date can be valid while local midnight does not exist in the default
+          // time zone, for example '1966-11-01' in America/Sao_Paulo where clocks were shifted
+          // forward from 0:00 to 1:00. In that case resolve the date like java.time's
+          // LocalDate.atStartOfDay(ZoneId) by using the first existing instant of that day.
+          // Truly invalid dates (such as '2021-02-30') also fail above, but lenient resolution
+          // would shift them to a different day, so they are re-thrown below.
+          Calendar lenientCalendar = new GregorianCalendar(year, month - 1, day);
+          Date resolved = lenientCalendar.getTime();
+          if (lenientCalendar.get(Calendar.YEAR) != year
+              || lenientCalendar.get(Calendar.MONTH) != month - 1
+              || lenientCalendar.get(Calendar.DAY_OF_MONTH) != day) {
+            throw e;
+          }
+          pos.setIndex(offset);
+          return resolved;
+        }
       }
 
       if (hasT) {
--- a/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
+++ b/gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
@@ -44,6 +44,43 @@ public class ISO8601UtilsTest {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


Marcono1234:
Could you please run `mvn spotless:apply` and commit the changes? Your PR currently changes the line endings of the modified files from LF (`\n`) to CRLF (`\r\n`); this is also the reason for the CI failure.

gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
```diff
@@ -43,6 +43,50 @@ private static GregorianCalendar createUtcCalendar() {
     return calendar;
   }
 
+  @Test
+  public void testParseDateOnlyDSTGap() throws ParseException {
+    TimeZone defaultTimeZone = TimeZone.getDefault();
+    try {
+      // 1966-11-01 00:00 does not exist in America/Sao_Paulo because clocks were shifted
```
eamonnmcmanus, line 50: This example makes it look as if this is a problem that almost nobody will have. Googling suggests that there are present-day timezones that switch to and from DST at midnight. I believe you can use America/Santiago and 2026-09-06 for a more realistic test case.

gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java
```diff
@@ -178,9 +178,26 @@ public static Date parse(String date, ParsePosition pos) throws ParseException {
       if (!hasT && (date.length() <= offset)) {
         Calendar calendar = new GregorianCalendar(year, month - 1, day);
         calendar.setLenient(false);
-
-        pos.setIndex(offset);
-        return calendar.getTime();
+        try {
+          pos.setIndex(offset);
+          return calendar.getTime();
+        } catch (IllegalArgumentException e) {
```
eamonnmcmanus, line 184: This makes me a bit uneasy. I'm sure there's more than one way we could get this exception.

Could we just construct a `GregorianCalendar` instance that specifies noon instead of midnight? Then we should be able to avoid the issue.

I also think ultimately we should rewrite this method so it uses `java.time` APIs, but that's a bigger project, and might have implications for Android (with old target SDKs and no desugaring).

eamonnmcmanus (changes requested):

gson/src/test/java/com/google/gson/internal/bind/util/ISO8601UtilsTest.java
Marcono1234: > that almost nobody will have

It is the exact reproducer from #2539 though; and based on that issue it is parsed as date of birth, so not such an unreasonable / uncommon case?

(Not saying though that a reproducer with a more recent date wouldn't be useful.)

PR #3111, opened 2026-09-05
Fix parse position for hour-only timezone offsets

mehuljariwala:
### Purpose

Keep `ParsePosition` within the consumed input when parsing an hour-only ISO 8601 timezone offset.

### Description

The parser pads `+01` to `+0100` for timezone lookup, then counts the padded length as consumed input. As a result, successfully parsing `2018-06-25T00:00:00+01` advances the position two characters past the end of the string.

Count the original timezone text before padding it. Tests cover positive, negative, and zero hour-only offsets, full offsets with and without a colon, unchanged parsed instants, and a nonzero starting parse position. Both new tests fail before the fix. This changes position accounting without changing the accepted timezone formats or resulting dates.

### Checklist

- [x] Google Java formatting checked with Spotless.
- [x] Added regression tests using Truth and JUnit 4.
- [x] `mvn clean verify javadoc:jar` passes on JDK 17, including all eight reactor modules.
- [x] No new public API or argument-validation contract.

The Surefire reports contain 4,854 tests, no failures or errors, and 19 skips. The new legacy-Date test uses a method-scoped `JavaUtilDate` lint suppression, following the neighboring tests, after JDK 21 CI flagged the Date API comparisons. Prepared with OpenAI Codex assistance; tests and build verification were executed locally.


Marcono1234:
Did you intentionally close this PR, or was it accidental when you deleted your fork?

Issue #1741, opened 2020-07-21
Unsatisfactory handling of type variables


TaylanUB:
GSON Version: 2.8.6

When `$Gson$Types.getRawType(Type)` is passed a `TypeVariable` instance, it simply gives up and returns `Object.class`:

```java
    ...
    } else if (type instanceof TypeVariable) {
      // we could use the variable's bounds, but that won't work if there are multiple.
      // having a raw type that's more general than necessary is okay
      return Object.class;
    } ...
```

That's not very good, and giving up just because there may be several bounds seems unnecessary. Since the behavior causes issues with my custom type adapter, I've had to resort to the following hack. (The `TypedField` class is a component of my custom type adapter. If you want to see the full code, [here's the whole shebang](https://gist.github.com/TaylanUB/c58447c87f63c8923cce747160a0cdf6).)

```java
    private static class TypedField {
        private final Field field;
        private final Type type;

        private TypedField(Type staticContext, Class<?> actualClass, Field field) {
            this.field = field;
            Type type = $Gson$Types.resolve(staticContext, actualClass, field.getGenericType());
            // As of Gson version 2.8.6, there is absolutely no way of making it use something
            // other than ObjectTypeAdapter for a Type that is a TypeVariable.  For that reason
            // we change the type here to the upper bound of the type variable.
            if (type instanceof TypeVariable) {
                this.type = Util.ifNull(getUpperBound(type), Object.class);
            } else {
                this.type = type;
            }
        }

        ...

        private static Type getUpperBound(Type type) {
            if (isNormalClass(type)) {
                return type;
            } else if (type instanceof ParameterizedType) {
                return ((ParameterizedType) type).getRawType();
            } else if (type instanceof TypeVariable) {
                for (Type bound : ((TypeVariable<?>) type).getBounds()) {
                    Type upperBound = getUpperBound(bound);
                    if (upperBound != null) {
                        return upperBound;
                    }
                }
            } else if (type instanceof WildcardType) {
                for (Type bound : ((WildcardType) type).getUpperBounds()) {
                    Type upperBound = getUpperBound(bound);
                    if (upperBound != null) {
                        return upperBound;
                    }
                }
            }
            return null;
        }

        private static boolean isNormalClass(Type type) {
            return type instanceof Class && ((Class<?>) type).isAssignableFrom(Object.class);
        }
    }
```

[...]

PR #3112, opened 2026-09-05
Use the runtime type for a wildcard element type

### LINKED ISSUE #1870: [bug] sub class field missing during serialization if collection's element type is a WildcardType
I am serializing a list like this:
```
List<? extends TestBeanBase> list1;
list1.add(new TestBeanSub());    // TestBeanSub is a child class of TestBeanBase
```
I find that the fields of TestBeanSub is missed in the serialized json.

You can try my example code：
```java
    @Test
    public void testList() {
        List<TestBeanBase> list = new ArrayList<>();

        //base class
        TestBeanBase baseBean = new TestBeanBase();
        baseBean.a = "1";
        list.add(baseBean);

        //sub class
        TestBeanSub subBean = new TestBeanSub();
        subBean.a = "1";
        subBean.b = "2";//b is a field declared in sub class, this field value is missed during serialization
        list.add(subBean);

        TestBeanList beanList = new TestBeanList();
        beanList.list1 = list;

        Gson gson = new GsonBuilder().setPrettyPrinting().create();
        System.out.println(gson.toJson(beanList));
    }

    public static class TestBeanList {
        public List<? extends TestBeanBase> list1; //element type is a WildcardType
    }

    public static class TestBeanBase {
        public String a;
    }

    public static class TestBeanSub extends TestBeanBase {
        public String b;
    }
```

Output json :
```
{
  "list1": [
    {
      "a": "1"
    },
    {
      "a": "1"   //field b is missed
    }
  ]
}
```

HuzaifaChaudary:
Fixes #1870

`getRuntimeTypeIfMoreSpecific` only looks at the runtime type for a `Class` or a
`TypeVariable` . `? extends Base` is a `WildcardType` , so it is neither , and the adapter for
the declared bound is used instead :

```
List<? extends Base>   ->  {"list":[{"a":"1"}]}            b is missing
List<Base>             ->  {"list":[{"b":"2","a":"1"}]}
```

same value , same field , different json .

a wildcard says nothing more than its upper bound , so this reads the bound and then decides
the same way it already does for a plain type . the bound is only used to make the decision ,
the declared type is still what gets kept when it is parameterized , so `? extends List<String>`
behaves like `List<String>` and does not lose its type arguments . that is why this reads the
bound rather than just adding `|| type instanceof WildcardType` , which would swap in the raw
runtime class and drop them .

while testing i found a worse case of the same cause that is not in the report .
`List<? extends CharSequence>` holding a `String` and a `StringBuilder` serialized as
`[{},{}]` , because gson fell back to a reflective adapter for the interface . the strings were
simply gone . `List<CharSequence>` gives `["s","t"]` . that is fixed by the same change .

four tests added , covering a wildcard list , a wildcard map value , and a wildcard reached
through a type variable . the fourth goes the other way and pins that
`Collection<? extends ParameterizedBase<String>>` still keeps the declared type , so the change
is not too loose . it passes with and without .

4637 tests pass in the gson module , 0 failures , and the full reactor is 4856 across all eight
modules . `spotless:check` is clean .

one related thing i deliberately left out . `registerTypeAdapter(Base.class, adapter)` is still
ignored when the element type is `? extends Base` , because the lookup token never matches one
registered for `Base.class` . that needs the wildcard stripped at adapter lookup time , where
there is already a `// TODO: strip wildcards?` in `GsonTypes` , and it is a wider change to
your internals , so it seemed better to leave that call to you . this change still improves that
case , the subclass fields stop disappearing .


HuzaifaChaudary:
@googlebot I signed it!

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
```diff
@@ -100,8 +101,21 @@ private static boolean isReflective(TypeAdapter<?> typeAdapter) {
...
+    Type declaredType = type;
+    // A wildcard says nothing more than its upper bound, so `? extends Base` is decided the same
+    // way as `Base`. Note that the bound is only used to make this decision; when it is a
+    // parameterized type the declared type is kept, the same as for a plain parameterized type,
+    // because the runtime class would lose the type arguments.
+    if (declaredType instanceof WildcardType) {
+      declaredType = ((WildcardType) declaredType).getUpperBounds()[0];
+    }
+
+    if (declaredType instanceof Class<?> || declaredType instanceof TypeVariable<?>) {
```
eamonnmcmanus, line 117: Should we do the same thing for `TypeVariable`? If you have `<T extends Number>` then we could treat `T` as `Number`. Unlike wildcards, that might require looping: if you have `<U extends T>` then we could also treat `U` as `Number`.

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
```diff
@@ -100,8 +101,21 @@ private static boolean isReflective(TypeAdapter<?> typeAdapter) {
...
+    }
+
+    Type declaredType = type;
+    // A wildcard says nothing more than its upper bound, so `? extends Base` is decided the same
+    // way as `Base`. Note that the bound is only used to make this decision; when it is a
+    // parameterized type the declared type is kept, the same as for a plain parameterized type,
+    // because the runtime class would lose the type arguments.
+    if (declaredType instanceof WildcardType) {
+      declaredType = ((WildcardType) declaredType).getUpperBounds()[0];
+    }
```
Marcono1234, line 115: Have you also considered what effect your changes will have for wildcards with lower bound, e.g. `List<? super String>`?

The PR description and the comment here only covers upper bounds. With your changes, `getUpperBounds()[0]` returns for a wildcard with lower bounds just `Object`.

I think this is fine though because it is effectively the same behavior as the previous code:
- previously: wildcard with lower bound was kept as is, but `TypeToken#getRawType()` [treats that as `Object`](https://github.com/google/gson/blob/gson-parent-2.14.0/gson/src/main/java/com/google/gson/internal/GsonTypes.java#L150-L154), so eventually it uses the built-in `ObjectTypeAdapter` which uses the runtime type during serialization
- with your changes: it directly chooses the runtime type adapter here within `TypeAdapterRuntimeTypeWrapper`

But could you please double-check to be safe?

Marcono1234 (commented):
Thanks for the changes! I have one question regarding wildcards with lower bounds (see comment).

Note that I am not a direct project member; feel free to consider my comment only as suggestion.

eamonnmcmanus (changes requested):

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
Marcono1234: Maybe it would be better to handle that in a separate PR though?

There are some other closely related issues with handling type variables (#1741, #2563). Just changing this here for `TypeAdapterRuntimeTypeWrapper` might be inconsistent, and the open question is also how to handle type variables with multiple bounds.
And also, because Gson does not permit capturing type variables with `TypeToken` anymore by default, it might be less common to encounter this (only for raw types now?).

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
HuzaifaChaudary: checked it , and it comes out the way you describe .

a field typed `Collection<? super Sub>` holding a `Sub` gives `{"collection":[{"s":3,"b":2}]}` , and that is the same string before and after this change . the subclass field is there in both , so nothing moves for lower bounds .

the two routes end in the same place , as you said . `getUpperBounds()[0]` is `Object` for `? super Sub` , so this picks the runtime type . before , the wildcard was kept and `getRawType()` reads a lower bounded wildcard as `Object` , so it ended up in `ObjectTypeAdapter` , which also writes the runtime type . i printed the type this method is handed to be sure the wildcard branch is the one that runs , and it is , `kind=WildcardTypeImpl type=? super ...Sub` .

gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
HuzaifaChaudary: i measured this before answering , and one part of the question above turns out to have a different answer than expected .

what a type variable does today . the existing branch already returns `value.getClass()` for a `TypeVariable` , so the runtime type is used , and this change does not touch that . measured on this branch and on main with the change reverted , identical both ways :

`<T extends Number>` holding 7 gives `{"t":7}` . `<T extends ParameterizedBase<String>>` holding a `ParameterizedSub<String>` gives `{"t":{"s":"three","t":"two"}}` . `<T extends Base & Marker>` holding a marked `Sub` gives `{"t":{"s":3,"b":2}}` .

on whether it is only raw types now . it is not . i printed the type this method is handed , and a type variable arrives for an ordinary parameterized owner too : `Container<Base>` holding a `Sub` comes in as `kind=TypeVariableImpl type=T` , same as the raw `Container` . so the path is reachable in normal code , not only through raw types .

what treating `T` as `Number` would actually change . it is the opposite direction from the current code , which goes straight to the runtime class . reading the bound instead only changes the answer where the bound is parameterized , which is the same case this change keeps the declared type for on the wildcard side . so it would make the two consistent . multiple bounds are never consulted today , and #1741 names that as the reason `getRawType()` gives up on type variables at all , so the looping you mention would have to decide that question too .

that is why i have not put it in this change : it alters a path this one leaves alone , and it lands in the middle of #1741 and #2563 . happy either way , say which you prefer and i will do it . if you want it here i will add the bound walk with a rule for multiple bounds , and if you would rather keep this to wildcards i will open it separately and link both issues .

Marcono1234: Ah sorry, I forget to consider that it is already handling `TypeVariable` so this would indeed about making it stricter; the referenced type variable issues are then probably not so relevant, but the question how to handle multiple bounds might remain?

-----

> on whether it is only raw types now . it is not . i printed the type this method is handed , and a type variable arrives for an ordinary parameterized owner too : `Container<Base>` holding a `Sub` comes in as `kind=TypeVariableImpl type=T` , same as the raw `Container` . so the path is reachable in normal code , not only through raw types .

How did you serialize the `Container<Base>`? Though I just noticed that my assumption regarding raw types was wrong; they are much more common during serialization because simply doing `gson.toJson(obj)` uses the raw (runtime) type for `obj`, so a `List<String>` would for example be serialized as `List.class`.

----

Sorry for the disruption; please only consider what @eamonnmcmanus says about this.

eamonnmcmanus: OK, so I think we're saying that, even if there is something to be done for type variables, it's not all that clear exactly what that should be. So I agree that we should not try to include that here.

eamonnmcmanus (approved):
Thanks! This looks fine, and it passed all of Google's internal tests.

PR #3114, opened 2026-09-06
Fix the Multiset code sample in TypeAdapterFactory's JavaDoc

### LINKED ISSUE #1335: The code sample in TypeAdapterFactory's JavaDoc does not work

alorentiar:
The Multiset sample in TypeAdapterFactory's javadoc compared the raw type against Multiset.class, so it rejected ImmutableMultiset and HashMultiset even though those are what most people actually use. The example then fell through to Gson's default collection handling and duplicates were not aggregated.

The sample now uses isAssignableFrom, the same pattern as CollectionTypeAdapterFactory, and falls back to Object.class when the token carries no type information.

Fixes #1335


alorentiar:
Closing this, sorry for the noise.

PR #3116, opened 2026-09-08
fix: reject unpaired surrogates in strict JSON

### LINKED ISSUE #3113: JsonReader silently accepts \uXXXX escapes containing lone surrogates in string values and field names
**The problem**

`JsonReader.readEscapeCharacter()` ([`JsonReader.java:1855`](https://github.com/google/gson/blob/main/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1855)) reads 4 hex digits, casts to `(char)`, and returns unchecked at [line 1882](https://github.com/google/gson/blob/main/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1882). No `Character.isSurrogate` / `0xD800..0xDFFF` gate anywhere. Callers at [line 1244](https://github.com/google/gson/blob/main/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1244) (string-value) and [line 1340](https://github.com/google/gson/blob/main/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1340) (skip-string) consume the raw char with no validation. Both `setLenient(false)` and `Strictness.STRICT` accept the malformed input.

**The specs**

- [RFC 8259 §8.2](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.2): "behavior of software that receives JSON texts containing such values is unpredictable; for example, implementations might return different values for the length of a string value or even suffer fatal runtime exceptions."
- [RFC 7493 §2.1](https://www.rfc-editor.org/rfc/rfc7493.html#section-2.1) (I-JSON): "Object member names, and string values in arrays and object members, MUST NOT include code points that identify Surrogates or Noncharacters as defined by [UNICODE]... `\uDEAD` is invalid because it is an unpaired surrogate."
- [RFC 8785 §3.2.2.2](https://www.rfc-editor.org/rfc/rfc8785.html#section-3.2.2.2) (JCS): "occurrences of such data MUST cause a compliant JCS implementation to terminate with an appropriate error."

**Reproducer (gson 2.14.0, JDK 21, `Strictness.STRICT`)**

```java
import com.google.gson.*;
import com.google.gson.stream.*;
import java.io.StringReader;

public class Repro3113 {
    public static void main(String[] args) throws Exception {
        String[] docs = {"{\"k\":\"\\uD800\"}", "{\"k\":\"\\uDC00\"}", "{\"\\uDC00\":1}"};
        for (String doc : docs) {
            JsonReader r = new JsonReader(new StringReader(doc));
            r.setStrictness(Strictness.STRICT);
            r.beginObject();
            String name = r.nextName();
            String val = r.peek() == JsonToken.NUMBER
                    ? String.valueOf(r.nextInt()) : r.nextString();
            System.out.println("input=" + doc
                + " ACCEPT name.char0=0x" + Integer.toHexString(name.charAt(0))
                + " val.char0=0x" + Integer.toHexString(val.charAt(0)));
        }
    }
}
```

Output:
```
input={"k":"\uD800"} ACCEPT name.char0=0x6b val.char0=0xd800
input={"k":"\uDC00"} ACCEPT name.char0=0x6b val.char0=0xdc00
input={"\uDC00":1}  ACCEPT name.char0=0xdc00 val.char0=0x31
```

**How to read the output**

- `input=` shows the JSON document fed in.
- `ACCEPT` means the parser returned successfully (no exception).
- `name.char0` is the hex value of the first char of the field name.
- `val.char0` is the hex value of the first char of the value.
- Row 1: field name is `"k"` (`0x6b`), value's first char is `0xd800` — a lone high surrogate returned as-is in the Java `String`.
- Row 2: same field name, value's first char is `0xdc00` — a lone low surrogate returned as-is.
- Row 3: field NAME itself starts with `0xdc00` (a lone low surrogate); the value is just the digit `1` (`0x31`).

All three should have rejected under `Strictness.STRICT` per the RFC 7493 / RFC 8785 requirements above.

**Suggested fix**

The return at [`JsonReader.java:1882`](https://github.com/google/gson/blob/main/gson/src/main/java/com/google/gson/stream/JsonReader.java#L1882) is the site to add the guard. Rejecting under `Strictness.STRICT` preserves current behavior in `Strictness.LEGACY_STRICT` / `Strictness.LENIENT` — no Android-compat break.

rksharma-owg:
Fixes #3113

`JsonReader` currently accepts `\uD800` / `\uDC00` escapes as lone UTF-16 surrogate code units even when `Strictness.STRICT` is selected. This can return malformed Java strings for both values and object member names. Validate the fully decoded quoted string at the boundary where it is returned, rejecting unpaired surrogates while allowing valid surrogate pairs and leaving legacy/lenient modes unchanged.

## Validation

- `mvn -pl gson -Denforcer.skip=true -Dtest=com.google.gson.stream.JsonReaderTest test` (164 tests, 3 skipped; JDK 26 requires the repository enforcer override)
- `git diff --check`

The regression test covers escaped lone high/low surrogates in values and names, plus a valid escaped pair.

### DIFF
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -1228,10 +1228,10 @@ public class JsonReader implements Closeable {
           pos = p;
           int len = p - start - 1;
           if (builder == null) {
-            return new String(buffer, start, len);
+            return validateString(new String(buffer, start, len));
           } else {
             builder.append(buffer, start, len);
-            return builder.toString();
+            return validateString(builder.toString());
           }
         } else if (c == '\\') {
           pos = p;
@@ -1261,6 +1261,24 @@ public class JsonReader implements Closeable {
         throw syntaxError("Unterminated string");
       }
     }
+  }
+
+  /** Validates that a string does not contain unpaired UTF-16 surrogate characters. */
+  private String validateString(String value) throws IOException {
+    if (strictness != Strictness.STRICT) {
+      return value;
+    }
+    for (int i = 0; i < value.length(); i++) {
+      char c = value.charAt(i);
+      if (Character.isHighSurrogate(c)) {
+        if (i + 1 >= value.length() || !Character.isLowSurrogate(value.charAt(++i))) {
+          throw syntaxError("Unpaired surrogate characters are not allowed in strict mode");
+        }
+      } else if (Character.isLowSurrogate(c)) {
+        throw syntaxError("Unpaired surrogate characters are not allowed in strict mode");
+      }
+    }
+    return value;
   }
 
   /** Returns an unquoted value as a string. */
--- a/gson/src/main/java/com/google/gson/stream/JsonReader.java
+++ b/gson/src/main/java/com/google/gson/stream/JsonReader.java
@@ -1270,11 +1270,12 @@ public class JsonReader implements Closeable {
     }
     for (int i = 0; i < value.length(); i++) {
       char c = value.charAt(i);
-      if (Character.isHighSurrogate(c)) {
-        if (i + 1 >= value.length() || !Character.isLowSurrogate(value.charAt(++i))) {
-          throw syntaxError("Unpaired surrogate characters are not allowed in strict mode");
+      if (Character.isSurrogate(c)) {
+        if (Character.isHighSurrogate(c)
+            && i + 1 < value.length()
+            && Character.isLowSurrogate(value.charAt(++i))) {
+          continue;
         }
-      } else if (Character.isLowSurrogate(c)) {
         throw syntaxError("Unpaired surrogate characters are not allowed in strict mode");
       }
     }
--- a/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
+++ b/gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
@@ -138,6 +138,33 @@ public final class JsonReaderTest {
   }
 
   @Test
+  public void testStrictModeRejectsUnpairedSurrogates() throws IOException {
+    for (String json : new String[] {"\"\\uD800\"", "\"\\uDC00\"", "{\"\\uD800\":1}"}) {
+      JsonReader reader = new JsonReader(reader(json));
+      reader.setStrictness(Strictness.STRICT);
+      if (json.startsWith("{")) {
+        reader.beginObject();
+        IOException expected = assertThrows(IOException.class, reader::nextName);
+        assertThat(expected)
+            .hasMessageThat()
+            .startsWith("Unpaired surrogate characters are not allowed in strict mode");
+      } else {
+        IOException expected = assertThrows(IOException.class, reader::nextString);
+        assertThat(expected)
+            .hasMessageThat()
+            .startsWith("Unpaired surrogate characters are not allowed in strict mode");
+      }
+    }
+  }
+
+  @Test
+  public void testStrictModeAllowsPairedSurrogates() throws IOException {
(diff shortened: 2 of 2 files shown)


gson/src/main/java/com/google/gson/stream/JsonReader.java
```diff
@@ -1263,6 +1263,24 @@ private String nextQuotedValue(char quote) throws IOException {
...
   }
 
+  /** Validates that a string does not contain unpaired UTF-16 surrogate characters. */
+  private String validateString(String value) throws IOException {
+    if (strictness != Strictness.STRICT) {
+      return value;
+    }
+    for (int i = 0; i < value.length(); i++) {
+      char c = value.charAt(i);
+      if (Character.isHighSurrogate(c)) {
```
eamonnmcmanus: As a minor optimization, you could do this:

```java
if (Character.isSurrogate(c)) {
  if (Character.isHighSurrogate(c)
      && i + 1 < value.length()
      && Character.isLowSurrogate(value.charAt(++i))) {
    continue;
  }
  throw syntaxError(...);
}
```

That way, in the common case where there are no surrogate characters, there is only one `if` per iteration rather than two.

I thought there might be some clever way to use `CharsetEncoder.canEncode(CharSequence)` here, but it doesn't look as if that would be very efficient.

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
```diff
@@ -137,6 +137,33 @@ public void testStrictModeAllowsOtherControlCharacters() throws IOException {
     assertThat(reader.nextString()).isEqualTo("\u007F\u009F");
   }
 
+  @Test
+  public void testStrictModeRejectsUnpairedSurrogates() throws IOException {
+    for (String json : new String[] {"\"\\uD800\"", "\"\\uDC00\"", "{\"\\uD800\":1}"}) {
```
eamonnmcmanus: I believe the new code will also reject inputs that contain unpaired surrogates directly as characters, not expressed via `\u` escapes in the JSON. That makes sense to me, but can you test it too? I mean `"\"\uD800\""` for example.

eamonnmcmanus (changes requested):
Thanks for putting this together! Just a couple of small things.

rksharma-owg:
Addressed the review feedback in commit `1b4fc9b`:\n\n- Added coverage for an unpaired surrogate present directly in the JSON input, alongside escaped values and names.\n- Simplified the strict-mode validation loop to check `Character.isSurrogate` once per character and skip valid pairs.\n\nThe focused `JsonReaderTest` suite passes: 164 tests, 3 skipped.

eamonnmcmanus (approved):
Thanks!

gson/src/test/java/com/google/gson/stream/JsonReaderTest.java
Marcono1234: I am wondering if raw unpaired surrogates are rather the responsibility of the user? The user is in control of how the underlying `Reader` is created, and could probably have prevented the unpaired surrogates if they wanted to. (But if the JSON is present as `String` already, then it would be cumbersome for the user to validate it; effectively duplicating the code of the PR here.)

I am mentioning this because I am wondering whether it would suffice to implement this validation only in the context of `readEscapeCharacter()` (maybe with an additional local boolean to store whether a low surrogate is expected).

The current implementation of this PR also has the side-effect that it allows creating surrogate pairs with mixed raw and escaped surrogates, as long as they form a valid pair in the end. Maybe a bit weird, but not necessarily bad?

But maybe this is not as important; the current implementation of this PR is fine too.

(Sorry that this comment is a bit late now.)

PR #3117, opened 2026-09-11
Remove redundant `recordPromotedName` calls

Marcono1234:
### Purpose
Follow-up for #3098

### Description
These `recordPromotedName` calls are guarded by a `case PEEKED_LONG`, but it seems `PEEKED_LONG` is only used if an actual number occurred in the JSON document. Whereas when promoting a name to a value, it will always be a JSON string value.

These redundant calls were also the "expensive" ones, converting a long to a String.


gson/src/test/java/com/google/gson/stream/JsonReaderPathTest.java
```diff
@@ -423,6 +423,38 @@ public void promoteNameToValueUpdatesPath() throws IOException {
     reader.endObject();
   }
 
+  @Test
+  public void promoteNameToValueUpdatesPath_Skip() throws IOException {
+    JsonReader reader = factory.create("{\"name\":\"value\"}");
+    reader.beginObject();
+    JsonReaderInternalAccess.INSTANCE.promoteNameToValue(reader);
+    reader.skipValue();
+    String expectedPath = factory == Factory.STRING_READER ? "$.<skipped>" : "$.name";
```
Marcono1234, line 434: This also highlights a difference between `JsonReader` and `JsonTreeReader`. But I guess this is fine: `JsonReader` uses the dummy name `<skipped>` for efficiency to avoid having to obtain the name as String. Whereas `JsonTreeReader` has the name available as String anyway.

For `JsonTreeReader` the behavior for `skipValue()` now differs between a regular name and a promoted one, but that is probably still fine.
Trying to fix that would make the `JsonTreeReader` implementation more complex because it would now need a dedicated field as well to remember whether a name had been promoted.

eamonnmcmanus (approved):
Thanks for noticing this!

PR #3118, opened 2026-09-11
Throw JsonSyntaxException for null element in primitive array

jdymitarai:
### Purpose
Throw `JsonSyntaxException` instead of a raw `IllegalArgumentException` when a JSON array with a `null` element is deserialized into a primitive array.

### Description
Deserializing a JSON array that contains a `null` element into any primitive array (e.g. `int[]`, `boolean[]`, `double[]`, `long[]`, `char[]`, `byte[]`, `short[]`, `float[]`) currently causes reflection (`Array.set(array, i, null)`) to throw an uncaught `java.lang.IllegalArgumentException`:

```java
new Gson().fromJson("[1, null, 3]", int[].class); // java.lang.IllegalArgumentException
```

Because `IllegalArgumentException` is neither wrapped nor caught, it terminates execution with a raw runtime exception and lacks any JSON document path context.

This PR adds a null check in `ArrayTypeAdapter.read` when `componentType.isPrimitive()`:
1. Rejects null elements early and fails fast with `JsonSyntaxException` (`"null is not a valid " + componentType.getName() + "[] element; at path " + in.getPreviousPath()`), directly mirroring the precedent established for `AtomicLongArray` in #3038 and #3096.
2. Accurately reports the JSON document path (e.g. `$[1]` or `$[0][1]`).
3. Leaves Object arrays (e.g. `Integer[]`, `String[]`, and outer dimensions of multidimensional arrays like `int[][]`) unaffected so they can continue to contain `null` elements as valid Java objects.
4. Adds comprehensive unit tests in `ArrayTest` covering all primitive types, boxed arrays, and multidimensional arrays.

### Checklist
- [x] New code follows the Google Java Style Guide
- [x] New unit tests have been added
  - [x] Assertions in unit tests use Truth
  - [x] No JUnit 3 features are used
  - [x] New tests added in `ArrayTest` for situations which failed previously and are now fixed


Marcono1234:
Thanks! There is already an existing recent PR which fixes that issue (as well as a few related ones): #3108

Would it be ok for you if that existing PR was preferred over your one? Maybe it would be helpful if you had a look at that PR and verify that it covers everything your PR covers. The tests there are not as extensive, but maybe they suffice?

jdymitarai:
Hi @Marcono1234,

Thanks for pointing that out! I took a close look at #3108 — it looks great and addresses the primitive array null issue cleanly along with the related collection adapters (TreeSet, ArrayDeque, PriorityQueue) and float/double number parsing.

I'm completely fine with preferring #3108 over this PR! The implementation approach in ArrayTypeAdapter is essentially identical. The only minor difference was that our tests in this PR explicitly asserted against all 8 Java primitive array types (int[], boolean[], double[], long[], char[], byte[], short[], float[]), but since componentType.isPrimitive() handles all 8 uniformly in ArrayTypeAdapter, the coverage in #3108 is more than sufficient.

Feel free to close this PR in favor of #3108. Thanks again for your review and stewardship of Gson!

jdymitarai:
Closing in favor of #3108 as discussed above. Thank you!

PR #3121, opened 2026-09-15
Reject unpaired surrogates in JsonWriter STRICT mode

sushant-me:
`JsonReader` rejects unpaired UTF-16 surrogate characters in strict mode (that was #3113 / #3116), but `JsonWriter` has no equivalent validation, so the two sides of the API disagree about what a strict JSON string is.

An unpaired surrogate in a Java `String` has no faithful JSON encoding: a JSON string is a sequence of Unicode characters (RFC 8259 section 7). Two consequences today, both silent:

- `JsonWriter` with `Strictness.STRICT` writes the value anyway, producing a document that its own Javadoc promises conforms to RFC 8259 but which `JsonReader` in the same strictness mode then refuses to read back. Gson's own default output is not readable by Gson.
- On the documented write-to-UTF-8 path the character becomes `?`, so data is silently lost.

### What this changes

`JsonWriter.string(...)` is the single sink for both names and values, so validating there covers `value(String)`, `name(String)` and deferred names alike:

```java
private void string(String value) throws IOException {
  validateString(value);
  ...
```

`validateString` mirrors `JsonReader.validateString` exactly, including the message, so a caller that catches one path understands the other.

### Scope

The check is limited to `Strictness.STRICT`. `LEGACY_STRICT` and `LENIENT` are unchanged, so no existing caller sees different output. That mirrors how the read side was introduced, and is why this is not a behaviour change for anyone who did not already opt into strict mode.

### Tests

Adds the writer-side coverage that never existed:

- `testStrictModeRejectsUnpairedSurrogates` — lone high, lone low, embedded, and reversed pair.
- `testStrictModeRejectsUnpairedSurrogateInName` — the name path (rejection surfaces when the entry is completed, because names are written lazily).
- `testStrictModeAllowsPairedSurrogates` — a valid pair is written and then read back in strict mode, which is the round-trip property the bug broke.
- `testNonStrictModesStillAllowUnpairedSurrogates` — pins the legacy behaviour so it cannot regress accidentally.

`mvn -pl gson test` passes: 4671 tests, 0 failures, 0 errors.


sushant-me:
@googlebot I signed it!

PR #3123, opened 2026-09-16
Fix eager runtime TypeAdapter lookup discarding usable declared adapters

### LINKED ISSUE #3122: `TypeAdapterRuntimeTypeWrapper` eagerly resolves the runtime-type adapter, so a failure building an adapter it would have discarded escapes to the caller
## Gson version

2.14.0 (also reproduced on 2.12.1, 2.9.0 and 2.8.9)

## Java version

Reproduced on OpenJDK 21 and OpenJDK 26. Any JDK with the module system should be affected.

## Description

When a value is serialized as an element of an array or a collection, Gson wraps the
component's `TypeAdapter` in `TypeAdapterRuntimeTypeWrapper`. On write, that wrapper resolves an
adapter for the value's **runtime** type and only afterwards decides which of the two adapters to
use:

```java
// com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java
TypeAdapter<T> chosen = delegate;
Type runtimeType = getRuntimeTypeIfMoreSpecific(type, value);
if (runtimeType != type) {
  TypeAdapter<T> runtimeTypeAdapter = (TypeAdapter<T>) context.getAdapter(TypeToken.get(runtimeType));
  if (!(runtimeTypeAdapter instanceof ReflectiveTypeAdapterFactory.Adapter)) {
    chosen = runtimeTypeAdapter;
  } else if (!isReflective(delegate)) {
    chosen = delegate;              // <-- runtimeTypeAdapter is discarded here
  } else {
    chosen = runtimeTypeAdapter;
  }
}
chosen.write(out, value);
```

The `context.getAdapter(...)` call is unconditional. When no adapter is registered for the runtime
type, Gson falls back to `ReflectiveTypeAdapterFactory`, whose construction throws `JsonIOException`
if a field of that class cannot be made accessible — for instance any class in a package that
`java.base` does not open.

In the `else if (!isReflective(delegate))` branch the result of that call is thrown away: the
user-registered adapter for the declared type is used instead. The exception is therefore raised
while computing a value the method has already decided not to use.

The practical consequence is that **registering an adapter for an interface is enough to serialize a
value directly, but not to serialize the same value inside an array or a collection**, as soon as the
implementation class is one Gson cannot reflect over. This makes `java.lang.reflect.Type` graphs
(whose implementations live in `sun.reflect.generics.reflectiveObjects`) impossible to serialize
without registering an adapter for every internal JDK class, even though a perfectly good adapter for
the declared interface is registered.

## Reproduction

```java
import com.google.gson.Gson;
import com.google.gson.GsonBuilder;
import com.google.gson.TypeAdapter;
import com.google.gson.stream.JsonReader;

import java.io.IOException;
import java.lang.reflect.ParameterizedType;
import java.lang.reflect.WildcardType;
import java.util.Arrays;
import java.util.List;

public class GsonRuntimeTypeLookupBug {

    static class Holder {
        List<? extends Number> field;
    }

    /** A trivial adapter for WildcardType. It never uses reflection. */
    static final TypeAdapter<WildcardType> WILDCARD_ADAPTER = new TypeAdapter<WildcardType>() {
        @Override
        public void write(JsonWriter out, WildcardType value) throws IOException {
            out.value(value == null ? null : value.getTypeName());
        }

        @Override
        public WildcardType read(JsonReader in) throws IOException {
            in.nextString();
            return null;
        }
    };

    public static void main(String[] args) throws Exception {
        ParameterizedType listOfWildcard =
                (ParameterizedType) Holder.class.getDeclaredField("field").getGenericType();
        WildcardType wildcard = (WildcardType) listOfWildcard.getActualTypeArguments()[0];
        // wildcard.getClass() is sun.reflect.generics.reflectiveObjects.WildcardTypeImpl

        Gson gson = new GsonBuilder()
                .registerTypeAdapter(WildcardType.class, WILDCARD_ADAPTER)
                .create();

        // works
        System.out.println(gson.toJson(wildcard, WildcardType.class));

        // throws
        System.out.println(gson.toJson(new WildcardType[]{wildcard}, WildcardType[].class));

        // throws as well
        System.out.println(gson.toJson(Arrays.asList(wildcard),
                new TypeToken<List<WildcardType>>() {}.getType()));
    }
}
```

## Expected behaviour

All three calls serialize the wildcard with `WILDCARD_ADAPTER`, since it is registered for the
declared type and is not reflective — which is exactly what `TypeAdapterRuntimeTypeWrapper` decides
in its `else if (!isReflective(delegate))` branch.

## Actual behaviour

The first call succeeds. The other two fail:

```
"? extends java.lang.Number"

Exception in thread "main" com.google.gson.JsonIOException: Failed making field
'sun.reflect.generics.reflectiveObjects.WildcardTypeImpl#upperBounds' accessible;
either increase its visibility or write a custom TypeAdapter for its declaring type.
See https://github.com/google/gson/blob/main/Troubleshooting.md#reflection-inaccessible
	at com.google.gson.internal.reflect.ReflectionHelper.makeAccessible(ReflectionHelper.java:52)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(...)
	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(...)
	at com.google.gson.Gson.getAdapter(Gson.java:557)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(...)
```

The error message is also misleading here: it advises writing a custom `TypeAdapter` for the
declaring type, but one *is* registered for the declared interface, and the internal JDK class the
message names is not something a user can reasonably register an adapter for.

On Gson 2.8.9 the same reproduction fails with the underlying
`java.lang.reflect.InaccessibleObjectException` instead, so the behaviour predates the
`JsonIOException` wrapping added in 2.9.0.

## Suggested fix

Resolve the runtime-type adapter only when its outcome can actually be used, or tolerate its failure
when the declared-type delegate is going to be preferred anyway. For example:

```java
TypeAdapter<T> chosen = delegate;
Type runtimeType = getRuntimeTypeIfMoreSpecific(type, value);
if (runtimeType != type) {
  TypeAdapter<T> runtimeTypeAdapter = null;
  try {
    runtimeTypeAdapter = (TypeAdapter<T>) context.getAdapter(TypeToken.get(runtimeType));
  } catch (JsonIOException e) {
    // No usable adapter for the runtime type. Only rethrow if the delegate cannot serve either,
    // i.e. if we would have chosen the runtime type adapter.
    if (isReflective(delegate)) {
      throw e;
    }
  }
  if (runtimeTypeAdapter != null) {
    ...unchanged selection...
  }
}
chosen.write(out, value);
```

This keeps the documented order of preference intact and only changes what happens when the third
preference — the reflective adapter for the runtime type — cannot be built at all.

## Workaround

Registering the adapter with `registerTypeHierarchyAdapter` instead of `registerTypeAdapter` avoids
the problem, because the lookup for the runtime type then returns the user adapter rather than a
reflective one:

```java
Gson gson = new GsonBuilder()
        .registerTypeHierarchyAdapter(WildcardType.class, WILDCARD_ADAPTER)
        .create();
```

Both the array and the collection then serialize as `["? extends java.lang.Number"]`.

The equivalent workaround for a `TypeAdapterFactory` is to match assignable types rather than the
exact `TypeToken`, so that the factory also answers for the runtime implementation class.

jaideeppyne:
### Purpose
Fixes #3122 — `TypeAdapterRuntimeTypeWrapper` eagerly calls `context.getAdapter(...)` for the value's runtime type even when that adapter would be discarded in favor of a non-reflective declared-type adapter. If reflective construction of the runtime-type adapter fails (e.g. inaccessible JDK-internal implementation classes under the module system), the exception escapes and serialization of arrays/collections fails even though a registered interface adapter would have worked.

### Description
Wrap the runtime-type `getAdapter` call and only rethrow `JsonIOException` when the declared-type delegate is reflective (i.e. when the runtime-type adapter would actually have been chosen). Selection preference order is unchanged.

Adds a regression test using a registered non-reflective `WildcardType` adapter with array/collection serialization — the scenario from the issue report.

Locally verified: `mvn -pl gson test -Dtest=TypeAdapterRuntimeTypeWrapperTest` (8 tests, including the new one). The new test fails on unfixed code with the same `JsonIOException` described in #3122.

### Checklist
- [x] New code follows the Google Java Style Guide
  - [x] Assertions in unit tests use Truth
  - [x] No JUnit 3 features are used
- [ ] `mvn clean verify javadoc:jar` passes without errors (ran targeted unit test suite locally)

Note: Google CLA will be required for this contribution (https://cla.developers.google.com/).


Marcono1234:
Can you please run `mvn spotless:apply` and commit the changes to fix the formatting issues (and the CI failures)?

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
```diff
@@ -202,4 +208,64 @@ public void testGsonFutureAdapter() {
...
+    WildcardType wildcard = sampleWildcard();
+    // Runtime class is typically sun.reflect.generics.reflectiveObjects.WildcardTypeImpl, which is
+    // not reflectively accessible under the module system.
+    try {
+      var unused = new Gson().getAdapter(TypeToken.get(wildcard.getClass()));
+      // If reflective access happens to succeed on this JDK, the bug cannot be reproduced here.
+      return;
+    } catch (JsonIOException expected) {
+      // Continue with the regression assertions.
+    }
```
Marcono1234, line 274: Instead of relying on the JDK to disallow access to internal fields, an alternative would be to register a Gson `ReflectionAccessFilter` which returns `BLOCK_ALL`. That should also allow reproducing this issue ([related internal code](https://github.com/google/gson/blob/698ba9ebfdbf629ca4ccc393d9cecd69c1e7115f/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L142-L149)).

Though that triggers a different `JsonIOException` than the one this test currently covers. So it would also be an option to add such a test using `ReflectionAccessFilter` additionally.

jaideeppyne: Added a deterministic `ReflectionAccessFilter`/`BLOCK_ALL` regression test in 066a2df. It verifies that resolving the runtime `Subclass` adapter raises `JsonIOException`, while a non-reflective declared `Base` serializer still writes the container successfully. This avoids relying on JDK-internal access rules.

jaideeppyne:
Ran `mvn spotless:apply` and committed its formatting changes in 066a2df. I also added the suggested deterministic `ReflectionAccessFilter.BLOCK_ALL` regression test. With JDK 25 (the local JDK 26 is outside the repository's enforced `[17,26)` range), `mvn -pl gson verify -q` and `mvn spotless:check -q` pass.

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
```diff
@@ -202,4 +225,63 @@ public void testGsonFutureAdapter() {
...
+      new TypeAdapter<WildcardType>() {
+        @Override
+        public void write(JsonWriter out, WildcardType value) throws IOException {
+          out.value(value == null ? null : value.getTypeName());
+        }
+
+        @Override
+        public WildcardType read(JsonReader in) throws IOException {
+          var unused = in.nextString();
+          return null;
```
Marcono1234: Minor: You could replace this to always throw an exception instead, since this method should not be called within the test. For example `throw new UnsupportedOperationException()` similar to the other test in this file.

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
```diff
@@ -202,4 +225,63 @@ public void testGsonFutureAdapter() {
...
+        }
+      };
+
+  private static WildcardType sampleWildcard() throws Exception {
+    ParameterizedType listOfWildcard =
+        (ParameterizedType) Holder.class.getDeclaredField("field").getGenericType();
+    return (WildcardType) listOfWildcard.getActualTypeArguments()[0];
+  }
```
Marcono1234, line 260: Would it be possible to make `Holder` a local class of this `sampleWildcard()` method? Because only its wildcard type is needed, but nothing else.

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
```diff
@@ -202,4 +225,63 @@ public void testGsonFutureAdapter() {
...
+  @Test
+  public void testNonReflectiveInterfaceAdapter_PrefersDelegateWhenRuntimeReflectiveFails()
+      throws Exception {
+    WildcardType wildcard = sampleWildcard();
+    // Runtime class is typically sun.reflect.generics.reflectiveObjects.WildcardTypeImpl, which is
+    // not reflectively accessible under the module system.
+    try {
+      var unused = new Gson().getAdapter(TypeToken.get(wildcard.getClass()));
+      // If reflective access happens to succeed on this JDK, the bug cannot be reproduced here.
+      return;
```
Marcono1234: Could maybe add a JUnit `assumeTrue("message", false)` here, or replace the `return` with `throw new AssumptionViolatedException(...)` to mark the test as skipped?

Marcono1234 (commented):
Thanks for the changes!

I have added a few comments with additional suggestions; hopefully they are useful. But the project members might add additional review comments.

There has been further discussion in #3122 after you had created the PR here, but the proposed changes there will probably have to be discussed with a project member first.
If you want feel free to include those changes, or leave them out for now and then in the future they can be included in a follow-up PR (by you or anyone else).

Marcono1234:
There seems to be a general issue with this exception catching approach (see https://github.com/google/gson/issues/3122#issuecomment-5718476335 and the previous comment): `TypeAdapterRuntimeTypeWrapper` cannot easily know if the exception was thrown because the reflective factory was used directly for the requested type (in which case the exception can be ignored) or a custom user-defined factory delegated handling (e.g. for a nested field) and the reflective factory failed for a nested type (in which case the exception must not be ignored).

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
jaideeppyne: Updated in 3f16bcfc: the adapter read method now always throws UnsupportedOperationException, since this serialization-only test must never call it.

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
jaideeppyne: Moved Holder into sampleWildcard in 3f16bcfc so it is scoped to its only use.

gson/src/test/java/com/google/gson/functional/TypeAdapterRuntimeTypeWrapperTest.java
jaideeppyne: Updated in 3f16bcfc: the reflectively-accessible JDK case now throws AssumptionViolatedException and is reported as skipped rather than silently returning. Focused tests and Spotless pass.

jaideeppyne:
Thanks for flagging this. I agree the current broad JsonIOException catch cannot distinguish a direct ReflectiveTypeAdapterFactory failure for the requested runtime type from a nested reflective failure reached through a user factory; silently falling back in the latter case would be wrong.

The issue discussion now has two materially different designs (a typed creation failure carrying its TypeToken, or a dedicated adapter lookup that stops before the reflective factory), and either expands this PR beyond the original fix. I am leaving that design choice to a Gson project member rather than selecting a new internal/public contract unilaterally. The current tests and cleanup suggestions are addressed; I can revise the implementation once maintainers choose the intended lookup contract.

PR #3124, opened 2026-09-16
Simplify key removal in `LegacyProtoTypeAdapterFactoryTest`

Marcono1234:
### Purpose
Simplify test code

### Description
See https://github.com/google/gson/pull/3083#discussion_r3730020998

The code currently does `var keysToRemove = json.keySet().stream()...; json.keySet().removeAll(keysToRemove);`. That should be identical to `json.keySet().removeIf(...)` (unless I am overlooking something).


eamonnmcmanus (approved):
Indeed, I think I had forgotten about `removeIf`.

PR #3126, opened 2026-09-21
fix: throw JsonSyntaxException for null elements in AtomicIntegerArray

### LINKED ISSUE #3047: AtomicIntegerArray deserializer crashes on JSON null (IllegalStateException) — symmetric fix to #3038

priyadarshnisundararajan:
## What
`new Gson().fromJson("[1,null,3]", AtomicIntegerArray.class)` threw a raw `IllegalStateException` ("Expected an int but was NULL") instead of `JsonSyntaxException`.

## Why
In `TypeAdapters.ATOMIC_INTEGER_ARRAY.read()`, `in.nextInt()` was called directly inside a try block that only caught `NumberFormatException`. A JSON null token makes `nextInt()` throw `IllegalStateException`, which escaped uncaught.

## Fix
Mirrors the already-merged `AtomicLongArray` fix from PR #3038: peek for a `NULL` token before reading each element and throw `JsonSyntaxException("null is not a valid AtomicIntegerArray element; at path " + in.getPath())`. `in.getPath()` (not `getPreviousPath()`) is used because the null token has not been consumed yet, which yields the identical `$[1]` path format as the `AtomicLongArray` fix.

Fixes #3047

## Tests
Added `testAtomicIntegerArrayWithNullElement` to `JavaUtilConcurrentAtomicTest`, asserting `JsonSyntaxException` with message `null is not a valid AtomicIntegerArray element; at path $[1]` (mirrors `testAtomicLongArrayWithNullElement`). Ran the full `JavaUtilConcurrentAtomicTest` class: 9 tests, all passing (8 existing + 1 new). Maven could not resolve build extensions over this network (TLS interception on repo.maven.apache.org), so tests were compiled and run directly with javac/java 21 using jars fetched from Maven Central (junit 4.13.2, truth 1.4.5, guava 33.7.0-jre + transitive deps).


Marcono1234:
Thanks for the pull request! However, have you verified that the issue you are trying to fix here even existed in the first place, and have you read the discussion in #3047?

Even the author of that issue confirmed in the end that the claimed issue does not actually exist: https://github.com/google/gson/issues/3047#issuecomment-5646892313

priyadarshnisundararajan:
You're right, thank you for checking. I verified against current `main` and the reported crash does not reproduce: `Gson.fromJson(JsonReader, TypeToken)` already catches the `IllegalStateException` and wraps it as `JsonSyntaxException`, so there is no uncaught exception here. Closing this as invalid, apologies for the noise.

